PR #7, opened 2024-04-24
[llama] Enable flash attention path

rsuderman:
We can use `torch.nn.functional.scaled_dot_product_attention` to target hte `iree` flash attention path. Substitution is a numerically identical behavior that hits `linalg_ext.attention`.


rsuderman:
Kunwar is going to work on causal / attention mask for `linalg_ext`. Once that lands we can land the flash attention change.

stellaraccident (approved):

stellaraccident:
We can land this, but I'll probably also want to make attention be an inference op so we can interface it with the sharding stuff. Fine to sequence that after this, since I'll be working on sharding the regular matmuls in the model for the rest of the week.

PR #9, opened 2024-04-25
Support moving theta and models to a specific device.

stellaraccident:
* Threads explicit device through models.
* Implements functional InferenceTensor, Theta and Dataset transformations and uses it to implement `to(device=)`.
* Adds `--device foo` to example runner.
* With https://github.com/iree-org/iree-turbine/pull/3 and supporting patches, this allows custom ops and kernels to transparently be used on CUDA/ROCM devices (instead of just CPU).

### DIFF
--- a/sharktank/sharktank/layers/rotary_embedding.py
+++ b/sharktank/sharktank/layers/rotary_embedding.py
@@ -4,6 +4,8 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+from typing import Optional
+
 import torch
 
 from .base import BaseLayer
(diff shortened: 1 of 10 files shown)


sharktank/sharktank/layers/rotary_embedding.py
```diff
@@ -81,12 +91,15 @@ def apply_batched_mask(
...
-        max_seqlen: int, dim: int, theta_value: float = 10000.0
+        self,
+        max_seqlen: int,
+        dim: int,
+        theta_value: float = 10000.0,
     ):
         freqs = 1.0 / (
-            theta_value ** (torch.arange(0, dim, 2)[: (dim // 2)].float() / dim)
+            theta_value
+            ** (torch.arange(0, dim, 2, device=self.device)[: (dim // 2)].float() / dim)
```
rsuderman, line 102: Gross line break but I assume this is an autoformat choice. In brackets so it should be fine.

rsuderman (approved):

sharktank/sharktank/layers/rotary_embedding.py
stellaraccident: 'tis

PR #15, opened 2024-04-30
Allow files with multiple suffixes in dataset loading helper.

ScottTodd:
Some GGUF files have names like [`mistral-7b-v0.1.Q8_0.gguf`](https://huggingface.co/TheBloke/Mistral-7B-v0.1-GGUF/tree/main). for which `Path.suffixes` returns a list like `['.1', '.Q8_0', '.gguf']`. We can check the last element of the suffix list to support both `foo.gguf` and `foo.bar.baz.gguf`.


ScottTodd:
Superseded by https://github.com/nod-ai/sharktank/pull/4

PR #17, opened 2024-05-03
Try de-flaking api_server_test.

ScottTodd:
The test was failing with `ERROR:    [Errno 98] error while attempting to bind on address ('0.0.0.0', 8000): address already in use` (https://github.com/nod-ai/sharktank/actions/runs/8944166670/job/24570503674?pr=16#step:7:1330). Some other process on the test runner may be using port 8000, so this attempts to find a free port before starting the server.


ScottTodd:
Ping?

stellaraccident (approved):

PR #20, opened 2024-05-06
Add `--output` argument to `export_paged_llm_v1.py`.

ScottTodd:
This helps when changing the script, exporting multiple versions, or using different inputs (via the `--hf-dataset` or `--gguf-file` input flags).


stellaraccident (approved):
Thanks. You can kind of tell where I stopped typing on that :)

PR #21, opened 2024-05-07
Add support for tensor parallel sharding of datasets.

stellaraccident:
* Adds a new ShardedTensor and ShardedPrimitiveTensor (ShardedQuantizedTensor is TBI).
* Adds an example `shard_llm_dataset` tool which will shard most of the LLM MLP layers for standard models.
* Extends dataset saving and loading to save 'global' IRPA files by default, but if shards are present in the dataset, they will be written/read from dedicated `*.rank{n}.irpa` files. For serving, this will allow each rank to have a subset of only the parameters it needs in certain advanced topologies.
  * May make this optional in the future, also allowing to save a flat IRPA.
* Threads through some additional CLI handling niceties for dataset input/output tools.
* Next steps will add op support for these ShardedTensor cases by emitting appropriate device-affinity and reduction ops.

### DIFF
--- a/sharktank/sharktank/utils/cli.py
+++ b/sharktank/sharktank/utils/cli.py
@@ -37,7 +37,7 @@ def parse(parser: argparse.ArgumentParser):
 def add_input_dataset_options(parser: argparse.ArgumentParser):
     """Adds options to load a GGUF dataset.
 
-    Either the `--hf-dataset` or `--gguf-file` argument can be present.
+    Either the `--hf-dataset`, `--gguf-file`, or `--irpa-file` argument can be present.
     """
     group = parser.add_mutually_exclusive_group(required=True)
     group.add_argument(
--- a/sharktank/sharktank/utils/cli.py
+++ b/sharktank/sharktank/utils/cli.py
@@ -9,7 +9,10 @@
 from typing import Dict, Optional
 
 import argparse
+import logging
 from pathlib import Path
+
+from ..types import Dataset
 
 from . import hf_datasets
 from . import tokenizer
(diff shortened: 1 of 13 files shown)


rsuderman (approved):

sharktank/sharktank/utils/cli.py
```diff
@@ -22,16 +25,16 @@ def create_parser(
...
-    return parser.parse_args()
+    args = parser.parse_args()
+    return args
 
 
-def add_gguf_dataset_options(parser: argparse.ArgumentParser):
+def add_input_dataset_options(parser: argparse.ArgumentParser):
     """Adds options to load a GGUF dataset.
 
     Either the `--hf-dataset` or `--gguf-file` argument can be present.
```
ScottTodd, line 40: nit: docstring is out of date (thanks for adding the IRPA argument!)
```suggestion
    Either the `--hf-dataset`, `--gguf-file`, or `--irpa-file` argument can be present.
```

sharktank/sharktank/utils/io.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+
+    def __init__(self, save_path: Path):
+        super().__init__()
+        self.save_path = save_path
+        self._rank_builders: dict[int, ParameterArchiveBuilder] = {}
+
+    def for_rank(self, rank: int) -> ParameterArchiveBuilder:
+        """Returns a ParameterArchiveBuilder for"""
+        if rank in self._rank_builders:
```
ScottTodd, line 29: nit: docstring sentence fragment - "for" what?

ScottTodd (approved):
Some tests need updating? https://github.com/nod-ai/sharktank/actions/runs/8979537381/job/24661703170?pr=21

PR #27, opened 2024-05-14
Revert "Remove InferenceOps from theta. (#23)"

rsuderman:
This reverts commit 6431a892da826639c138958740ad535fe87487d0.

Export was failing due to symbolic tracing.


ScottTodd:
Can you elaborate on what is failing? Should we go with a partial revert instead of a full revert?

rsuderman:
> Can you elaborate on what is failing? Should we go with a partial revert instead of a full revert?

Sounds like there is an upstream failure for `torch`. Corresponding bug logged here:
https://github.com/pytorch/pytorch/issues/126242

ScottTodd (approved):

PR #31, opened 2024-05-16
Implement explicit sharded arithmetic operations using duck typing.

stellaraccident:
This level of math is aimed at the inner-most level of the device hierarchy, where it is assumed that the devices connected to a process are connected via high-speed links. By coding the model with explicit sharded arithmetic, we naturally get multi-device distribution by having a few points where we insert communication primitives. In this approach, the same math expressions run on either unary or nary-sharded tensor-parallel weight partitions. Using duck typing, the nary-sharded partitions are seen as a `ShardedTensor` and corresponding duplication of the math is undertaken. Unlike fully automatic solutions, this approach *requires* that the model author does some minimal coding for placement and distribution. In practice, this is unavoidable anyway, so by taking the layers of automation out of it, it is at least transparent, can be inspected eagerly, etc.

New ops:

* `elementwise`: Apply a sharded or unsharded unary/binary elementwise operator.
* `sharded_cat`: Concatenate across the sharding dimension or return an unsharded tensor as is.
* `sharded_sum`: Sum all shards or return unsharded tensor as is.

Adds implementation variants for:

* `matmul_sharded_rhs`: Matmul operating on a global lhs and a sharded rhs. This is typically the first layer in a multi step sequence. Returns a sharded tensor.
* `matmul_sharded`: Matmul operating on a sharded lhs and rhs. Performs partial matmuls on each shard, returning a sharded result tensor.

It is left to the user, based on the arrangement of the math to choose how to reduce a final result from a ShardedTensor to a unary one.

For now, not implemented on QuantizedTensors because additional infra will be needed for row-wise sharding.

### DIFF
--- a/sharktank/sharktank/ops/sharded_impls.py
+++ b/sharktank/sharktank/ops/sharded_impls.py
@@ -1,0 +1,94 @@
+# Copyright 2024 Advanced Micro Devices, Inc
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import torch
+from torch import Tensor, dtype
+import torch.nn.functional as F
+
+from ..types import InferenceTensor, ShardedPrimitiveTensor
+from ._registry import unbox_tensor
+from .signatures import *
+
+# Sharded elementwise.
+
+
+@elementwise.override(ShardedPrimitiveTensor)
+def sharded_elementwise_unary(operator, x: ShardedPrimitiveTensor):
+    partials = [operator(unbox_tensor(pt)) for pt in x.shards]
+    return ShardedPrimitiveTensor(shard_dim=x.shard_dim, shape=x.shape, ts=partials)
+
+
+@elementwise.override(ShardedPrimitiveTensor, ShardedPrimitiveTensor)
+def sharded_elementwise_binary(
+    operator, x: ShardedPrimitiveTensor, y: ShardedPrimitiveTensor
+):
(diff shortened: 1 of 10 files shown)


sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+@sharded_sum.override(ShardedPrimitiveTensor)
+def sharded_sum_sharded(maybe_sharded: ShardedPrimitiveTensor):
+    # TODO: Should implement as an all reduce.
+    shards = maybe_sharded.shards
+    accum = shards[0].as_torch()
+    for shard in shards[1:]:
+        accum = torch.add(accum, shard.as_torch())
+    return accum
```
rsuderman, line 94: Missing newline at end of file

sharktank/sharktank/ops/signatures.py
```diff
@@ -91,3 +111,39 @@ def _rms_norm_trampoline(
...
+
+@sharded_sum.trampoline
+def _sharded_sum_trampoline(d: SignatureDispatcher, maybe_sharded: AnyTensor):
+    tensors = (maybe_sharded,)
+    for override in d.find_overrides(tensors):
+        result = override(maybe_sharded)
+        if result is not NotImplemented:
+            return override, result
+    else:
+        d.fail(tensors)
```
rsuderman, line 149: Same on newline

rsuderman (approved):

sharktank/sharktank/ops/sharded_impls.py
stellaraccident: weird. maybe a stale push.

PR #36, opened 2024-05-21
Enable Mixtral LLM model

archana-ramalingam:
Add Mixture of experts, router logic for Mixtral LLM


sharktank/sharktank/layers/ffn_moe.py
```diff
@@ -0,0 +1,213 @@
...
+
+from .base import Theta, ThetaLayer, LinearLayer, RMSNormLayer
+
+__all__ = [
+    "MixtralBlockSparseTop2MLP",
+    "MixtralSparseMoeBlock",
+]
+
+
+class MixtralBlockSparseTop2MLP(ThetaLayer):
```
stellaraccident, line 21: Naming nit: This looks just like a normal FFN network. I probably should have made this for the base model. We don't need to call this MixtralBlockSparseTop2MLP. It could just be LlmFfnMLP (or something like that). I'll need to do work on this module to shard/distribute it, and would like to make it common.

stellaraccident (commented):
Thanks for doing this. Left one small note. There are some things we can commonize as a followon, but best to get something working first.

sharktank/sharktank/layers/ffn_moe.py
archana-ramalingam: Thanks for taking a look Stella. Yes, it's a common FFN and I will extract it into it's own file. Once I wrap Mixtral, will have a PR to add the FFN to existing Llama model too.

rsuderman (approved):

archana-ramalingam:
MoE model changes from this PR have been cherry picked in #169, for Grok model bringup.

PR #39, opened 2024-05-21
Export using "direct" kv_cache_type with batch size 1.

ScottTodd:
It sounds like we'll end up with a bs1-only variant for client use and a bsN variant for serving with a couple of batch sizes compiled in that can be switched between dynamically to optimize for throughput.

* A new `--bs` argument can be used to specify batch sizes:

    flag | result
    -- | --
    --bs=1 | direct cache, batch size 1
    --bs=4 | paged cache, batch size 4
    --bs=1,4 | paged cache, batch sizes 1 and 4
    --bs=2,4,8 | paged cache, batch sizes 2, 4, and 8

* `prefill_bs1` exports successfully, with `generate_batch_decode(bs)` commented out and a command like

    ```
    python -m sharktank.examples.export_paged_llm_v1 \
        --hf-dataset=open_llama_3b_v2_f16_gguf \
        --output_mlir=/tmp/llama_prefill_bs1.mlir \
        --output_config=/tmp/llama_prefill_bs1.json \
        --bs=1
    ```

* `decode_bs1` fails during fx tracing, see the TODO in `transact_cache_direct()` to use index ops

---

TODO: 

- [ ] Resolve TODOs in `validate_direct_llama_model.py`, unless we don't want that script in the first place
- [ ] Update decode so it exports, by switching to using index ops

### DIFF
--- a/sharktank/sharktank/examples/export_paged_llm_v1.py
+++ b/sharktank/sharktank/examples/export_paged_llm_v1.py
@@ -4,7 +4,7 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-"""Inference support for the PagedLLMV1 protocol of models."""
+"""Export support for the PagedLLMV1 protocol of models."""
 
 import json
 import torch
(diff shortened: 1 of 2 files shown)


stellaraccident (approved):

ScottTodd:
Going to merge with the TODOs unresolved so PRs don't start stacking up.

PR #50, opened 2024-06-13
Add sharding of conv2d, group norm and layer norm

sogartar:
* Sharding of output channel dimension of conv2d with a single group.
* Sharding by groups in group normalization.
* Sharding of non-normalization dimension in layer norm.


stellaraccident (approved):
Nice! thank you

sogartar:
Rebased before merging.

PR #54, opened 2024-06-14
Add replicated tensor type and handle conv2d with sharded input channels

sogartar:


sharktank/tests/ops/sharded_test.py
```diff
@@ -13,6 +13,74 @@
...
+        ]
+        expected_result = torch.cat(shards, dim=shard_dim)
+
+        sharded = ShardedPrimitiveTensor(shard_dim=shard_dim, ts=shards)
+        actual_result = ops.all_gather(sharded)
+
+        for shard in actual_result.shards:
+            torch.testing.assert_close(shard.as_torch(), expected_result)
+
+    def testCov2dShardedInputAndOutputChannelsOneGroup(self):
```
stellaraccident, line 31: Nit: misspelled "testCov"

stellaraccident (approved):
Nice, thanks!

stellaraccident:
Discussed offline about possibly changing `__eq__` to something like `is_deep_equivalent` and documenting specifically what that means.

sogartar:
@stellaraccident I renamed `__eq__` to `is_deep_equal`.

sogartar:
I rebased and resolved conflicts.

sogartar:
@stellaraccident, I ran into this problem that when loading the dataset if we mem map, then even when deleting the Python objects the underlying files are still open. I see that the files are memory mapped in IREE. I assume we are missing somewhere an ownership relation.

PR #58, opened 2024-06-18
Integer kernels: Matmul_per_axis_q8, attention_per_channel_q8, batch_matmul, convolution

KyleHerndon:
This patch adds a bunch of integer and quantized kernels.

I have some edits I want to do before we look to land this, but opening a PR to get some feedback on a bug in torch-mlir through the conv_2d_nchw_fchw kernel path.


sharktank/tests/kernels/conv_2d_nchw_fchw_test.py
```diff
@@ -0,0 +1,70 @@
...
+                return kernels.conv_2d_nchw_fchw(a, b, c, d, e)
+
+        mod = MyModule()
+        dtype = torch.int32
+        ep = torch.export.export(
+            mod,
+            args=(
+                (torch.rand([8,8,8,8]) * 64).to(dtype),
+                (torch.rand([8,8,1,1]) * 64).to(dtype),
+                "1, 1", "0, 0", "1, 1"
```
stellaraccident, line 60: This is what is triggering your torch-mlir issue. Could definitely be a better error message, but the gist of it is that you are declaring that your conv op requires an attribute (effectively a compile time constant) for stride/padding/dilation but here you are asking torch to create a program where you pass that as inputs (not a compile time constant).

So I think this is just an artifact of how you are testing it: you need to inline the stride/padding/dilation, not pass as an arg like this.

stellaraccident:
I'm going to lift the stuff we need out of this into a new branch/PR. Will leave the q8 variants here as we are not using them yet.

PR #60, opened 2024-06-19
Make matmul's transpose_rhs default to False and use linear instead

sogartar:
Add the `permute` function that is equivalent to `torch.permute`. Add the `T` property to `InferenceTensor`.
Implement `linear` for some sharded tensor cases.

### DIFF
--- a/sharktank/sharktank/types/tensors.py
+++ b/sharktank/sharktank/types/tensors.py
@@ -10,7 +10,6 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass
 
 import torch
-import shark_turbine.ops as ops
 from ..utils.math import ceildiv
 from shark_turbine.aot import (
     ExternalTensorTrait,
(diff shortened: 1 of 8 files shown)


sharktank/sharktank/types/tensors.py
```diff
@@ -240,6 +239,14 @@ def _clone_with_globals(
             f"InferenceTensor {type(self)} does not implement _clone_with_globals"
         )
 
+    @property
+    def T(self) -> "InferenceTensor":
+        from ..ops import permute
+
+        return permute(
+            self, dims=torch.flip(torch.arange(len(self.shape)), dims=[0]).tolist()
```
sogartar, line 247: @stellaraccident, I am not sure how `torch.flip(torch.arange(len(self.shape)), dims=[0]).tolist()` would be handled by model tracing. `torch.arange` returns a `torch.Tensor`. At the end we conver to a `list`. This must not end up as part of an exported model.

sogartar: I fixed this by using more pythonic constructs to generate the reverse dimensions list. This seems to make the torch export tracer happy.

stellaraccident (approved):

PR #62, opened 2024-06-20
Add a few cases of sharded matmul

sogartar:
* Sharded parallel dimensions for RHS and LHS.
* Sharded LHS and replicated RHS.
* Sharded LHS.


stellaraccident (approved):
Nice!

PR #63, opened 2024-06-22
Expand sharded element-wise binary ops support

sogartar:
Add polymorphic `replicate`, `shard` and `shard_like` ops.
Add operator `+` for InferenceTensor.
Add `ops.equal`, analogous to `torch.equal` that operates on sharktank tensor types.

### DIFF
--- a/sharktank/sharktank/ops/sharded_impls.py
+++ b/sharktank/sharktank/ops/sharded_impls.py
@@ -14,7 +14,7 @@ from ..types import (
     UnreducedTensor,
     ShardedTensor,
 )
-from ._registry import unbox_tensor
+from ._registry import unbox_tensor, AnyTensor
 from .signatures import *
 
 
@@ -224,6 +224,16 @@ def elementwise_binary_sharded_lhs_replicated_rhs(
     return elementwise(operator, x, y_sharded)
 
 
+@equal.override(ReplicatedTensor)
+def equal_replicated(a: ReplicatedTensor, b: AnyTensor) -> bool:
+    return a.is_deep_equal(b)
+
+
+@equal.override(ShardedPrimitiveTensor)
+def equal_sharded(a: ShardedPrimitiveTensor, b: AnyTensor) -> bool:
+    return a.is_deep_equal(b)
+
+
 @group_norm_affine.override(
     ShardedPrimitiveTensor, ShardedPrimitiveTensor, ShardedPrimitiveTensor
 )
@@ -379,20 +389,20 @@ def permute_sharded(tensor: ShardedPrimitiveTensor, dims: List[int]):
     return ShardedPrimitiveTensor(ts=permuted_shards, shard_dim=permuted_shard_dim)
 
 
-@replicate.override(Tensor)
-def replicate_tensor(input, *, count: int) -> ReplicatedTensor:
-    torch_input = unbox_tensor(input)
-    return ReplicatedTensor(ts=torch_input, shard_count=count)
-
-
 @replicate.override(ReplicatedTensor)
-def replicate_tensor(input: ReplicatedTensor, *, count: int) -> ReplicatedTensor:
+def replicate_replicated(input: ReplicatedTensor, *, count: int) -> ReplicatedTensor:
     assert input.shard_count == count
     return input
 
 
+@replicate.override(Tensor)
+def replicate_unsharded(input, *, count: int) -> ReplicatedTensor:
+    torch_input = unbox_tensor(input)
+    return ReplicatedTensor(ts=torch_input, shard_count=count)
+
+
 @shard.override(Tensor)
-def shard_tensor(input, *, dim: int, count: int) -> ShardedPrimitiveTensor:
+def shard_unsharded(input, *, dim: int, count: int) -> ShardedPrimitiveTensor:
     torch_input = unbox_tensor(input)
     return ShardedPrimitiveTensor(ts=torch_input, shard_dim=dim, shard_count=count)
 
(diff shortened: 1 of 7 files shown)


sogartar:
I would actually like to add a few more tests.

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -204,6 +204,36 @@ def sharded_elementwise_binary(
     return ShardedPrimitiveTensor(shard_dim=x.shard_dim, shape=x.shape, ts=partials)
 
 
+@elementwise.override(ReplicatedTensor, ShardedPrimitiveTensor)
+def elementwise_binary_replicated_lhs_sharder_rhs(
+    operator, x: ReplicatedTensor, y: ShardedPrimitiveTensor
+):
+    assert x.shard_count == y.shard_count
```
stellaraccident, line 211: This will be a user error message: Fine to still assert, but include a more detailed message. For something like this, better to do it in an if/raise.

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -359,6 +389,109 @@ def permute_sharded(tensor: ShardedPrimitiveTensor, dims: List[int]):
     return ShardedPrimitiveTensor(ts=permuted_shards, shard_dim=permuted_shard_dim)
 
 
+@replicate.override(ReplicatedTensor)
+def replicate_replicated(input: ReplicatedTensor, *, count: int) -> ReplicatedTensor:
+    assert input.shard_count == count
```
stellaraccident, line 394: Ditto on error message.

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -359,6 +389,109 @@ def permute_sharded(tensor: ShardedPrimitiveTensor, dims: List[int]):
...
+def shard_unsharded(input, *, dim: int, count: int) -> ShardedPrimitiveTensor:
+    torch_input = unbox_tensor(input)
+    return ShardedPrimitiveTensor(ts=torch_input, shard_dim=dim, shard_count=count)
+
+
+@shard.override(ShardedPrimitiveTensor)
+def shard_sharded(
+    input: ShardedPrimitiveTensor, *, dim: int, count: int
+) -> ShardedPrimitiveTensor:
+    assert input.shard_count == count
```
stellaraccident, line 414: Error message. Probably collapse these two asserts into one check/raise.

sharktank/sharktank/ops/signatures.py
```diff
@@ -145,6 +150,38 @@ def _embedding_lookup_trampoline(
         d.fail(tensors)
 
 
+@overridable
+def equal(a: AnyTensor, b: AnyTensor) -> bool:
+    """Compares 2 tensors for equality, such that if one is substituted with the other
+    in sharktank polymorphic calls, the results will be essentially the same.
+    Meaning, they would also compare equal.
+
+    torch.Tensor and DefaultPrimitiveTensor with the same contents would compare equal."""
```
stellaraccident, line 159: Based on your trampoline, I think this needs a note like "overrides are matched first against both tensor types and failing that, then on just the first. Therefore, each override must internally decide whether it can handle an equality check with an arbitrary `b` tensor."

sharktank/sharktank/ops/signatures.py
```diff
@@ -145,6 +150,38 @@ def _embedding_lookup_trampoline(
...
+    raise NotImplementedError
```
stellaraccident, line 160: Just `...` like the rest

sharktank/sharktank/ops/signatures.py
```diff
@@ -315,6 +352,75 @@ def _rms_norm_trampoline(
...
+    else:
+        d.fail(tensors)
+
+
+@overridable
+def shard(input: AnyTensor, *, dim: int, count: int) -> ShardedTensor:
+    """Shard `input` along `dim`.
+    This does not mean that a sharded tensor is further sharded.
+    It is not composition of sharding operations.
+    TODO: Maybe we can rename to `reshard` if it turns out to be confusing.
```
stellaraccident, line 381: +1 to rename

sharktank/sharktank/types/tensors.py
```diff
@@ -250,6 +250,13 @@ def T(self) -> "InferenceTensor":
 
         return permute(self, dims=dims)
 
+    def __add__(
+        self, rhs: "sharktank.ops._registry.AnyTensor"
```
stellaraccident, line 254: Just leave off the type annotation. We're at the limit of what a useful annotation can do there.

sharktank/tests/ops/ops_test.py
```diff
@@ -11,6 +11,35 @@
 
 from sharktank import ops
 from sharktank.types import *
+from copy import deepcopy
+
+
+class EqualTest(unittest.TestCase):
+    def testEqualTorchTensors(self):
+        a = torch.rand(2, 3, dtype=torch.float32)
+        b = deepcopy(a)
```
stellaraccident, line 20: I think just use torch.clone vs deepcopy (here and below).

sharktank/tests/ops/sharded_test.py
```diff
@@ -10,6 +10,7 @@
 
 from sharktank import ops
 from sharktank.types import *
+from copy import deepcopy
```
stellaraccident, line 13: Switch to torch.clone

sharktank/tests/ops/sharded_test.py
```diff
@@ -134,6 +135,108 @@ def testCov2dShardedOutputChannelsOneGroup(self):
...
+        shard_count = 3
+        sharded_a = ops.shard(a, dim=shard_dim, count=shard_count)
+        sharded_b = ops.shard(b, dim=shard_dim, count=shard_count)
+        sharded_result = sharded_a + sharded_b
+        actual_result = ops.shard_like(sharded_result, expected_result)
+
+        torch.testing.assert_close(actual_result, expected_result)
+
+    def testBinaryOperators(self):
+        operators = [torch.add, torch.div, torch.fmin, torch.fmax, torch.sub]
```
stellaraccident, line 155: Prefer using `parameterized` (like this: https://github.com/nod-ai/sharktank/blob/ba89559f1c7e8661af977c7764c1d063acb8b401/sharktank/tests/kernels/mmt_block_scaled_q8_test.py#L24)

stellaraccident (approved):
Thanks. A number of small comments. Please have a look at them before landing.

PR #64, opened 2024-06-24
Fix quantizers_test on Windows.

ScottTodd:
Each call to `_roundtrip` was creating a file with the same name in the temp directory and the second write would fail. This gives each temp file a unique name manually.

We could also
* Close the file to allow it to be reopened
* Use other functions in tempfile to create uniquely named _files_, not just a directory

Sample error: `FAILED sharktank\tests\types\quantizers_test.py::StaticScaledQuantizerTest::testPerTensorQuantDequant - RuntimeError: Error building parameter archive: D:\a\iree\iree\c\runtime\src\iree\base\internal\file_io.c:381: UNKNOWN; failed to open file 'C:\Users\RUNNER~1\AppData\Local\Temp\tmpejgrnsnpStaticScaledQuantizerTest\poodoo.irpa'`. Full logs: https://github.com/nod-ai/sharktank/actions/runs/9509442435/job/26212392662.

References:
* https://github.com/iree-org/iree/blob/fe571e4d5efde141ec437cb4699307df67a38b9c/runtime/bindings/python/io.cc#L424-L491
* https://github.com/iree-org/iree/blob/fe571e4d5efde141ec437cb4699307df67a38b9c/runtime/src/iree/base/internal/file_io.c#L372-L382
* https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea

### DIFF
--- a/sharktank/tests/types/quantizers_test.py
+++ b/sharktank/tests/types/quantizers_test.py
@@ -17,8 +17,8 @@ class StaticScaledQuantizerTest(TempDirTestBase):
         super().setUp()
         torch.manual_seed(0)
 
-    def _roundtrip(self, it):
-        dataset_path = self._temp_dir / "poodoo.irpa"
+    def _roundtrip(self, it, suffix=""):
+        dataset_path = self._temp_dir / f"poodoo{suffix}.irpa"
         theta = Theta([it])
         Dataset({}, theta).save(dataset_path)
         ds = Dataset.load(dataset_path)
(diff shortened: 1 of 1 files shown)


rsuderman (approved):

ScottTodd:
Still seeing another flake in `ThetaTest`, reported here: https://github.com/nod-ai/sharktank/pull/43#issuecomment-2187545440

PR #65, opened 2024-06-25
Disentangle the sharded and split notions

sogartar:
We are using the term sharded as a general term to mean a tensor that is distributed across a number of logical devices.
On the other hand we are also using it to mean a tensor split along a dimension and distributed across the logical devices. To avoid confusion in the latter context we will talk about a sharded split tensor or simply a split tensor.


stellaraccident (approved):
Good improvement. Thank you

PR #67, opened 2024-06-25
Add sharding specs

sogartar:
Sharding specs are concise description of tensor and layer sharding. They can be applied to tensors and thetas to get the sharded counterpart.

This is mostly the pluming around specific cases. Only sharing of the channel dimension in group normalization is provided.

### DIFF
--- a/sharktank/sharktank/types/sharding.py
+++ b/sharktank/sharktank/types/sharding.py
@@ -1,0 +1,83 @@
+# Copyright 2024 Advanced Micro Devices, Inc
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+"""Specifications describing how a tensor, ops, layers and blocks are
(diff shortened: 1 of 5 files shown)


sharktank/sharktank/types/sharding.py
```diff
@@ -0,0 +1,83 @@
...
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+"""Specifications describing how a tensor, ops, layers and blocks are
+sharded."""
+
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from . import Theta
```
stellaraccident, line 12: I think this needs to be `from .theta import Theta` or something like that. Within a package, you always want to import from the leaf vs a partially initalized parent.

stellaraccident (approved):

sharktank/sharktank/types/sharding.py
sogartar: I removed as it was actually unused, but will look out for this in the future.

PR #70, opened 2024-06-25
Add test script and workflow for llama export, compile, serve.

ScottTodd:
Progress on https://github.com/nod-ai/sharktank/issues/22

Sample runs on my fork:
* https://github.com/ScottTodd/sharktank/actions/runs/9670685134
* https://github.com/ScottTodd/sharktank/actions/runs/9715408887

I decided to run this on a nightly `schedule` and on `workflow_dispatch`. It takes around 10 minutes so it _could_ run on `pull_request` if we want too.

As these components stabilize and we spend less time hacking on individual steps using the full toolkit (python -> manual `iree-compile` vs. using the in-process compiler API) we can switch the test from a bash script to a pytest file. Need to start somewhere :)


tests/llama_export_compile_serve.sh
ScottTodd, line 1: Is https://github.com/nod-ai/sharktank/tree/main/sharktank/sharktank/models/punet running through IREE yet? We could set up a similar set of tests for that model too.

ScottTodd, line 1: Opinions on this top level `tests/` folder? This exercises both `sharktank/` and `shortfin/`

tests/llama_export_compile_serve.sh
```diff
@@ -0,0 +1,47 @@
...
+LOCAL_VMFB_FILE="${TEMP_DIR}/model.vmfb"
+
+mkdir -p ${TEMP_DIR}
+
+huggingface-cli download --local-dir ${TEMP_DIR} ${HUGGING_FACE_MODEL_NAME} ${HUGGING_FACE_MODEL_FILE}
+
+python -m sharktank.examples.export_paged_llm_v1 \
+  --gguf-file="${LOCAL_GGUF_FILE}" \
+  --output-mlir="${LOCAL_MLIR_FILE}" \
+  --output-config="${LOCAL_CONFIG_FILE}"
```
ScottTodd, line 36: > I decided to run this on a nightly `schedule` and on `workflow_dispatch`. It takes around 10 minutes so it _could_ run on `pull_request` if we want too.

Ehhh, maybe on `pull_request` makes more sense... that will get us coverage for when the python code changes.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,69 @@
+name: Integration Tests
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Weekdays at 13:00 UTC = 05:00 PST / 06:00 PDT.
+    - cron: "5 4 * * 1-5"
```
ScottTodd, line 7: > I decided to run this on a nightly `schedule` and on `workflow_dispatch`. It takes around 10 minutes so it _could_ run on `pull_request` if we want too.

Ehhh, maybe on `pull_request` makes more sense... that will get us coverage for when the python code changes.

stellaraccident: Let's start scheduled/on-demand. Gotta start somewhere.

tests/llama_export_compile_serve.sh
stellaraccident: Yes, but fragile. Can start with just eager and import/export. Need to fetch from an HF repo to do it.

stellaraccident: Let's put these scripts in build_tools/integration_tests. Python projects end up using the tests/ folder for more unit-testy things.

stellaraccident (approved):
Thanks. I suspect we'll want to run a variant of this on a real runner with a GPU, and that should cache pytorch/rocm and HF download locally. Will need that for punet since that really needs to run on a GPU.

ScottTodd:
> Thanks. I suspect we'll want to run a variant of this on a real runner with a GPU, and that should cache pytorch/rocm and HF download locally. Will need that for punet since that really needs to run on a GPU.

Good to know for punet. In other repos (IREE, SHARK-TestSuite) we have a few persistent runners and test scripts are already set up to use a local cache.

This is currently using a 7GB .gguf file (3b parameters in f16) from an open model on huggingface. We can test with other quantized data types and larger parameter sizes too. That may work well enough on nightly runs using standard runners but yeah - should use larger runners with persistent caches for best results.

PR #81, opened 2024-06-28
Add sharding of ResnetBlock2D

sogartar:
Add one case of resnet block sharding
`ResnetBlock2DSplitOutputChannelsSharding`.

Move AnyTensor to avoid cyclic dependencies.

Add broadcast_dim(s) functions and handle broadcasting for element-wise binary with sharded split tensors.

Make sharded linear override for a cartesian product of type sets. Let the constituent ops fail if they can't handle the input.

Add matmul for replicated LHS and split RHS.

Fix resharding of thetas.

Add unshard op.

Fix slicing of split tensor.

Refactor Theta to reflect that it is actually a tree of tensors and expose the tree.
Fix some asserts during construction.

### DIFF
--- a/sharktank/sharktank/ops/shape.py
+++ b/sharktank/sharktank/ops/shape.py
@@ -4,15 +4,15 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-from typing import Collection
+from typing import Sequence
 from ..types.tensors import AnyTensor
 
 
 def broadcast_dim(
-    dim: int, shaped_or_shape: Collection[Collection[int] | AnyTensor]
+    dim: int, shaped_or_shape: Sequence[Sequence[int] | AnyTensor]
 ) -> int:
-    """Returns the dimension corresponding to `args[0]`'s dimension `dim` after
-    broadcasting `args`.
+    """Returns the dimension corresponding to `shaped_or_shape[0]`'s dimension `dim` after
+    broadcasting `shaped_or_shape`.
 
     Parameters
     ----------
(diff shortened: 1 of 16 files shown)


sharktank/sharktank/ops/shape.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Parameters
+    ----------
+    args: is a collection of shapes or tensors.
+
+    ```python
+    shape1 = [2, 3, 1]
+    shape2 = [4, 2, 3, 5]
+    dims = [2, 2]
+    res = broadcast_dims(2, [shape1, shape2])
```
stellaraccident, line 52: Nit: I think the first arg is a list, right?

sharktank/sharktank/ops/shape.py
```diff
@@ -0,0 +1,62 @@
...
+    #     # Tensors case.
+    #     return broadcast_dim(dim, *[tensor.shape for tensor in args])
+    # ranks = [len(shape) for shape in args]
+    # broadcast_rank = max(ranks)
+    # return dim + max(0, broadcast_rank - len(args[0]))
+    return broadcast_dims([dim], shaped_or_shape)[0]
+
+
+def broadcast_dims(
+    dims: Collection[int], shaped_or_shape: Collection[Collection[int] | AnyTensor]
```
stellaraccident, line 39: I think you want the Sequence type here, not Collection.

sharktank/sharktank/types/sharding.py
```diff
@@ -93,3 +93,58 @@ def theta_sharding(self) -> ThetaSharding:
...
+                ),
+                "bias": Split(
+                    shard_count=self.shard_count,
+                    shard_dim=self.weight_and_bias_spit_dim,
+                ),
+            }
+        )
+
+
+class ResnetBlock2DSplitOutputChannelsSharding(ThetaLayerSharding):
```
stellaraccident, line 120: Ok to have this here for now, but it should probably be with the resnet layer (which is not a "general" resnet but specific to unet). Fine for now.

sharktank/tests/layers/resnet_test.py
```diff
@@ -0,0 +1,134 @@
...
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import unittest
+
+import torch
+
+from sharktank.models.punet.layers import ResnetBlock2D
```
stellaraccident, line 11: Yeah, we'll want to move this test and the sharding into punet, since this resnet block is just named "resnet" but is more resnet-like for that specific purpose.

stellaraccident (approved):
Nice. A couple of small nits and a request to move some stuff. I'm fine landing with the current structure and we can move later.

sharktank/sharktank/ops/shape.py
sogartar: done

sharktank/sharktank/ops/shape.py
sogartar: I change it.

sharktank/sharktank/types/sharding.py
sogartar: Moved them to punet.

PR #83, opened 2024-06-28
Add sharded interpolate

sogartar:
Add fully replicated case.
Add split batch or channel dimensions.


sharktank/sharktank/ops/__init__.py
```diff
@@ -17,6 +17,7 @@
 """
 
 from . import _registry
+from ._registry import unbox_tensor
```
stellaraccident, line 20: For later, was thinking of adding these utilities to types so everyone can use

stellaraccident (approved):

Issue #76, opened 2024-06-26
[punet] CI for quantization import/compilation/golden check


stellaraccident:

ScottTodd:
Could you take a look at https://github.com/nod-ai/sharktank/pull/70 and https://github.com/nod-ai/SHARK-TestSuite/pull/272 ? Those are for llama, and it would be nice to have similar testing for punet and llama. For the first, I'm not particularly attached to the bash script but just wanted something running in this repo somehow. The second distills the program down to a common test format we can use from IREE on presubmit.

stellaraccident:
Oh thanks, will have a look.

PR #84, opened 2024-06-29
[punet] Add integration tests.

stellaraccident:
* Imports/validates the FP16 model running eagerly.
* Imports/validates the int8 model running eagerly.
* Exports the models.

Progress on #76

### DIFF
--- a/sharktank/integration/models/punet/integration_test.py
+++ b/sharktank/integration/models/punet/integration_test.py
@@ -1,0 +1,248 @@
+# Copyright 2024 Advanced Micro Devices, Inc
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+from pathlib import Path
+import pytest
+
+from sharktank.utils import testing
+
+
+@pytest.fixture
+def temp_dir():
+    with testing.temporary_directory(__name__) as td:
+        yield Path(td)
+
+
+@pytest.fixture
+def punet_goldens():
+    from huggingface_hub import hf_hub_download
+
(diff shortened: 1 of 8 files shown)


sharktank/sharktank/utils/testing.py
```diff
@@ -42,3 +43,84 @@ def run_main(self, main_func, *args):
...
+            print(f"Restoring debug flag {k} = {v}")
+            setattr(flags, k, v)
+
+
+def get_best_torch_device() -> str:
+    import torch
+
+    if torch.cuda.is_available() and torch.cuda.device_count() > 0:
+        return "cuda:0"
+    return "cpu"
```
ScottTodd, line 90: May want a way to set this device index via environment variable or flag.

On some of our CI runners we try to only use a specific GPU (like 6), since we and possibly other groups are running multiple jobs on the same "machine" / node.

sharktank/integration/models/punet/integration_test.py
ScottTodd, line 1: Until this is running on CI, it would be useful to see the logs produced (e.g. in a gist).

Things I'd look for:

* Time taken for each step / test case
* Format of logs on success
* Format of logs on failure
* How easy the flow is to understand and reproduce outside of pytest
* What artifacts are downloaded
* What artifacts are produced

sharktank/sharktank/utils/testing.py
```diff
@@ -42,3 +43,84 @@ def run_main(self, main_func, *args):
...
+
+
+def assert_golden_safetensors(actual_path, ref_path):
```
ScottTodd, line 93: This function is nice, giving a good reason to stay in Python instead of using bash or just the native tools like `iree-run-module`.

(I'm still slowly coming to terms with moving infrastructure from C/C++ to python :P)

ScottTodd (commented):
Nice!

PR #96, opened 2024-07-09
Quark dataset importer for fp8

dan-garvey:
:edited:

Provides a entry point for quark models that are stored as a safetensor file + config.json. With a little effort this could also be adapted to become the equivalent of the hf importer from gguf without the intermediate step. 

I removed a lot of the useful debugging tooling I developed for this because it doesn't play nicely with torch.export, but I imagine a lot of it could be re-added if I guarded it in some way, so I'm keeping a branch with those changes for reference.


stellaraccident (changes requested):
Please create an import script specific to llama. Probably in models/llama/tools. Fine to start from here but these will diverge a lot

sharktank/sharktank/models/llama/tools/import_quark_dataset.py
```diff
@@ -0,0 +1,366 @@
...
+                quant_theta, layer_name, updated_tensors, split_sizes=split_sizes
+            )
+
+    # Update the non quantized weights (norm layers)
+    for layer_idx in model_layers:
+        update_norm_layer(
+            quant_theta,
+            layer_idx,
+            updated_tensors,
+            head_count=updated_properties["llama.attention.head_count"],
```
IanNod, line 345: head_count and split_sizes don't exist in update_norm_layer

sharktank/sharktank/models/llama/tools/import_quark_dataset.py
```diff
@@ -0,0 +1,366 @@
...
+        "mlp.down_proj",
+        "mlp.up_proj",
+        "self_attn.o_proj",
+        "self_attn.qkv",
+    ]
+    for layer in model_layers:
+        for sub in sub_layers:
+            layer_name = layer + "." + sub
+            apply_per_layer_quant(
+                quant_theta, layer_name, updated_tensors, split_sizes=split_sizes
```
IanNod, line 336: missing n_head arg

IanNod (changes requested):

sharktank/sharktank/models/llama/tools/import_quark_dataset.py
dan-garvey: oops, I think I cleaned it up from the function decl but not the call, good catch

sharktank/sharktank/models/llama/tools/import_quark_dataset.py
dan-garvey: nice catch, the default is correct for 7b but will be a problem for 70

sharktank/sharktank/models/llama/tools/import_quark_dataset.py
```diff
@@ -0,0 +1,366 @@
+# Copyright 2024 Advanced Micro Devices, Inc
```
IanNod, line 1: Would be great to get some unit tests for this

sharktank/sharktank/layers/linear.py
```diff
@@ -54,19 +52,23 @@ def __init__(
...
+        qdq_output = self.qdq_output
+        original_input = x.clone().detach()
         if self.premul_input is not None:
             x = ops.elementwise(torch.mul, x, self.premul_input)
 
         if q_input is not None:
             x = q_input.quantize(x)
         elif qdq_input is not None:
+            # TODO: probably need a way to only do q_input if exporting.
+            print("qdq input")
```
IanNod, line 71: extraneous print

sharktank/sharktank/models/llama/tools/import_quark_dataset.py
dan-garvey: hmm

I could generate a safetensors file with one attn block of random numbers,  upstream a config json, and then save to irpa and load and compare against the initial random values. I'm not sure how useful that is as a test though. Do you have something else in mind?

IanNod (approved):

stellaraccident (approved):

PR #101, opened 2024-07-10
Fixed bug in cli.py

Eliasj42:
in cli.py, `data_files` is initialized as a set, but its use makes it look like its intended to be a dict.  Also I edited the model_cookbook.md as convert.py was replaced by convert_hf_to_gguf.py


stellaraccident (approved):
Oof. Thanks for the catch. That syntax is a footgun.

PR #102, opened 2024-07-11
Allow sharktank tensors to be part of export function signatures

sogartar:
Adds flattening and unflattening of our tensor types according to torch requirements.
This allows us to pass and return such tensors when exporting functions.

Add function to populate names of tensors in a theta with the natural name corresponding to its path in the theta.
When exporting a dataset each tensor must have a unique name.

Make name a property in InferenceTensor.
Renaming of sharded tensors needs to rename the shards as well.

### DIFF
--- a/sharktank/sharktank/types/tensors.py
+++ b/sharktank/sharktank/types/tensors.py
@@ -4,7 +4,18 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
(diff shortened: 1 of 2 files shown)


sogartar:
@stellaraccident, will you be able to review this PR?

stellaraccident (approved):
Nice. I think you have it coming next but would be good to have an example and/or test that verifies that this usage stays working. (doesn't have to be in this PR if it is in your stack somewhere)

sogartar:
@stellaraccident, thank you for your review. I have more PRs coming.
I rebased to verify the CI is OK as this PR has been opened for a while.

PR #106, opened 2024-07-12
[punet] Implement QKV quantization fusion and adapt to Brevitas mechanism.

stellaraccident:
This patch enables quantization activation fusion of either qkv (for MHA) or kv (for MCHA) attention layers.

In the model, this is selected based on the presence of either a `to_kv` or `to_qkv` theta on each attention layer. Normally, a LinearLayer handles all quantization inside of itself, but in this case, we decompose what the LinearLayer usually does by explicitly performing activation quantization using either qkv or kv shared input conditioning, then proceed with the normal computation and dequant.

The Brevitas import is a bit finnicky in this case because the current version of it concats the kv or qkv weights/scales/bias/zp in an attempt to only have one set of each. However, it is a more natural representation (and leaves less room for optimization failures) to simply populate the dataset with the data at rest how we intend to use it. So we reorganize into this structure:

```
attn0.to_qkv
  premul_input
  qinput
  to_q.weight
  to_q.bias
  to_k.weight
  to_k.bias
  to_v.weight
  to_v.bias
```

(to_kv is similar but does not include the q within it, leaving it at the top, unfused level)

We may want to switch the quantizer to a more natural (for us) representation.

### DIFF
--- a/sharktank/sharktank/models/punet/tools/import_brevitas_dataset.py
+++ b/sharktank/sharktank/models/punet/tools/import_brevitas_dataset.py
@@ -169,7 +169,7 @@ def apply_per_layer_quant(
         updated_tensors[bias_quant.name] = bias_quant
 
     # If dealing with a fused QKV layer, then we need to split the weight as
-    # Brevitas concats it along the [n * input_dim, output_dim] axis, where
+    # Brevitas concats it along the [n * output_dim, input_dim] axis, where
     # n is 2 (to_kv) or 3 (to_qkv).
     # We do this here vs in the model since it is much more efficient to
     # operate on the weights contiguously at rest.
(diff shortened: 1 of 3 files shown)


sharktank/sharktank/models/punet/tools/import_brevitas_dataset.py
```diff
@@ -162,11 +165,83 @@ def _get_json_tensor(
...
         )
-        bias_quant = bias_quantizer.quantize(bias, name=bias.name)
+        bias_quant = bias_quantizer.quantize(bias, name=bias_name)
         updated_tensors[bias_quant.name] = bias_quant
-        # Spot check that things look sane.
-        bias_dequant = bias_quant.unpack().dequant()
-        bias_diff = bias.as_torch() - bias_dequant
+
+    # If dealing with a fused QKV layer, then we need to split the weight as
+    # Brevitas concats it along the [n * input_dim, output_dim] axis, where
```
nickfraser, line 172: Typo in comment, should read: `Brevitas concats it along the [n * output_dim, input_dim] axis, where`

rsuderman (approved):

PR #108, opened 2024-07-15
[punet] Generalize the brevitas importer slightly to handle symmetric quant.

stellaraccident:
These were just some bug fixes from never getting this combination before.


IanNod (commented):
Looks good to me

PR #120, opened 2024-08-05
Llama Testing

rohan-tan-bhowmik:
Created an FP32 test to verify and compare the numerical outputs of sharktank's PagedLlamaAttentionBlock and HuggingFace's equivalent LlamaDecoderLayer. Expects, given random weights and inputs, exact equality.

Enjoy!

### DIFF
--- a/sharktank/sharktank/layers/rotary_embedding.py
+++ b/sharktank/sharktank/layers/rotary_embedding.py
@@ -20,17 +20,62 @@ class RotaryEmbeddingLayer(BaseLayer):
         rope_dimension_count: int,
         max_seqlen: int,
         device: Optional[torch.device] = None,
+        hf: bool = False,
     ):
         super().__init__()
         self.device = device
         self._table = self._create_rotary_embed_table(
             max_seqlen=max_seqlen,
             dim=rope_dimension_count,
+            hf=hf,
         )
+        self.hf = hf
 
     def forward(self, *, xq: torch.Tensor, xk: torch.Tensor, start_index: int):
         # xq_, xk_ shape: bs, sl, _, dim
         # freqs_cis shape: max_sl, dim
+        
+        def create_interleaved_tensor(dim):
+            """Creates a tensor which indexes an tensor such that
+            it alternates between elements of its first and second
+            half. Intended for use for HuggingFace's rotation
+            implementation.
+
+            Args:
+              dim: Size of tensor
+
+            Returns:
+              Interleaved indexing tensor
+            """
+            first_half = torch.arange(dim//2)
+            second_half = torch.arange(dim//2, dim)
+            
+            interleaved_tensor = torch.empty(dim, dtype=torch.long)
+            interleaved_tensor[0::2] = first_half
+            interleaved_tensor[1::2] = second_half
+            
+            return interleaved_tensor
+        
+        def create_ordering_tensor(dim):
+            """Creates a tensor which indexes an tensor such that
+            it reverses the alternation induced by create_interleaved_tesnor.
+            Intended for use for HuggingFace's rotation implementation.
+
+            Args:
+              dim: Size of tensor
+
+            Returns:
+              Ordering indexing tensor
+            """
+            order_tensor = torch.empty(dim, dtype=torch.long)
+            order_tensor[:dim // 2] = torch.arange(0, dim, 2)
+            order_tensor[dim // 2:] = torch.arange(1, dim, 2)
+            return order_tensor
+        
+        if self.hf:
+            xq = xq[..., create_interleaved_tensor(xq.shape[-1])]
+            xk = xk[..., create_interleaved_tensor(xq.shape[-1])]
+
         xq_ = torch.view_as_complex(xq.reshape(*xq.shape[:-1], -1, 2))
         xk_ = torch.view_as_complex(xk.reshape(*xk.shape[:-1], -1, 2))
         _, sl, _, dim = xq_.shape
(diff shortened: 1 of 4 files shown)


sharktank/sharktank/models/llama/llama.py
```diff
@@ -275,27 +284,46 @@ def __init__(
...
-            "ffn_norm", RMSNormLayer(theta("ffn_norm"), epsilon=rms_epsilon)
-        )
-        self.add_module("ffn_gate", LinearLayer(theta("ffn_gate")))
-        self.add_module("ffn_up", LinearLayer(theta("ffn_up")))
-        self.add_module("ffn_down", LinearLayer(theta("ffn_down")))
+        if hf:
+            self.add_module(
+                "attn_norm", RMSNormLayer(theta("input_layernorm"), epsilon=rms_epsilon)
+            )
+            self.add_module("attn_q", LinearLayer(theta("self_attn.q_proj")))
```
raikonenfnu, line 294: Do we need this block of code to rename the parameters? Since in the test we are specifying the paramter names anyways, couldn't we just use the old names?

sharktank/sharktank/models/llama/llama.py
```diff
@@ -46,6 +47,9 @@ class LlamaModelConfig:
     # Dtype to use for attention.
     attention_dtype: torch.dtype = torch.float16
 
+    # Indicates if running with HuggingFace implementation.
```
raikonenfnu, line 50: Maybe add a little comment on how HF is different, also maybe call the flag `use_hf`

sharktank/sharktank/models/llama/testing.py
```diff
@@ -0,0 +1,54 @@
...
+
+from typing import List
+
+import torch
+
+from ...types.tensors import *
+from ...types.theta import Theta
+
+
+def make_rand_torch(shape, dtype):
```
raikonenfnu, line 15: Add small comment that you are shifting the distribution

raikonenfnu (commented):
Thanks for the PR, overall looks good, just some NITs.

sharktank/tests/models/llama/attention_test.py
```diff
@@ -0,0 +1,151 @@
...
+        attention_embedding = RotaryEmbeddingLayer(
+            rope_dimension_count=rope_dimension_count,
+            max_seqlen=max_seq_len,
+            device="cpu",
+            hf=True,
+        )
+
+        input_tensor = torch.rand(1, seq_len, head_count * head_dim) * 2 - 1
+
+        h = attention_block(
```
aviator19941, line 71: NIT: don't use single letter variable names

sharktank/tests/models/llama/attention_test.py
```diff
@@ -0,0 +1,151 @@
...
+
+        llama_decoder_layer = LlamaDecoderLayer(
+            config=llama_config, layer_idx=block_index
+        )
+        llama_decoder_layer.self_attn = llama_attention_block
+        llama_decoder_layer.mlp = llama_mlp
+        llama_decoder_layer.input_layernorm = llama_input_layernorm
+        llama_decoder_layer.post_attention_layernorm = llama_post_attention_layernorm
+
+        g = llama_decoder_layer(
```
aviator19941, line 141: NIT: don't use single letter variable names

sharktank/sharktank/models/llama/llama.py
aviator19941: I think this is because sharktank is based on gguf instead of huggingface, so the name mappings are different and need to be specified when using gguf weights vs. huggingface weights.

sharktank/tests/models/llama/attention_test.py
```diff
@@ -0,0 +1,151 @@
...
+        )
+        attention_block = PagedLlamaAttentionBlock(
+            theta=attention_block_theta,
+            block_index=block_index,
+            cache=paged_kv_cache,
+            head_count=head_count,
+            head_dim=head_dim,
+            head_count_kv=head_count_kv,
+            rms_epsilon=rms_epsilon,
+            hf=True,
```
aviator19941, line 60: So this isn't a direct call to huggingface, it's simulating the same math that huggingface uses?

sharktank/sharktank/models/llama/llama.py
raikonenfnu: But looks like the huggingface "weight" is just being loaded here, where we specify through string what the weights are, but technically, we can load it using the old string/params name no? https://github.com/nod-ai/sharktank/pull/120/files#diff-3605352cc960ff4b5e59079e25b9f8db20a8bd83646b1788d73e6d2d14bd4ec3R108-R131

rohan-tan-bhowmik: Decided to use original string/params names

raikonenfnu: awesome!

sharktank/tests/models/llama/attention_test.py
raikonenfnu: It's using all of HF llama layers but we are integrating/building ourselves I believe

raikonenfnu: actually not sure anymore what you were pointing at haha, but yes this PR should introduce the math that HF uses for rope embedding

rohan-tan-bhowmik: Yep, this attention block is our own sharktank implementation. We are comparing its outputs to its corresponding equivalent in huggingface (decoder layer)

sharktank/sharktank/layers/rotary_embedding.py
```diff
@@ -31,6 +33,48 @@ def __init__(
...
+              Ordering indexing tensor
+            """
+            order_tensor = torch.empty(dim, dtype=torch.long)
+            order_tensor[: dim // 2] = torch.arange(0, dim, 2)
+            order_tensor[dim // 2 :] = torch.arange(1, dim, 2)
+            return order_tensor
+
+        if self.use_hf:
+            xq = xq[..., create_interleaved_tensor(xq.shape[-1])]
+            xk = xk[..., create_interleaved_tensor(xq.shape[-1])]
```
raikonenfnu, line 76: Not too much context here, but can we (trivially) move this xq/xk into the other `if self.use_hf:`  region?

sharktank/tests/models/llama/attention_test.py
rohan-tan-bhowmik: Addressed, greatly appreciated

sharktank/tests/models/llama/attention_test.py
rohan-tan-bhowmik: Addressed, much appreciated

sharktank/sharktank/models/llama/testing.py
rohan-tan-bhowmik: Good suggestion, have done

sharktank/sharktank/models/llama/llama.py
rohan-tan-bhowmik: I agree this makes more sense. Changed

PR #123, opened 2024-08-08
Add test that shards, exports and runs with IREE a Unet Resnet block

sogartar:
The test targets llvm-cpu with the local-task IREE driver.

### DIFF
--- a/sharktank/tests/models/punet/sharded_resnet_block_with_iree_test.py
+++ b/sharktank/tests/models/punet/sharded_resnet_block_with_iree_test.py
@@ -228,7 +228,13 @@ def test_sharded_resnet_block_with_iree(
     The model is tensor sharded across 2 devices.
     """
 
-    with tempfile.TemporaryDirectory() as tmp_dir:
+    with tempfile.TemporaryDirectory(
+        # TODO: verify hypothesis and remove ignore_cleanup_errors=True
+        # torch.export.export is spawning some processes that don't exit when the
+        # function returns, this causes some objects to not get destroyed, which
+        # in turn holds files params.rank0.irpa and params.rank1.irpa open.
+        ignore_cleanup_errors=True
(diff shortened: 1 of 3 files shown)


sogartar:
This PR depends on
https://github.com/iree-org/iree-turbine/pull/73
https://github.com/iree-org/iree-turbine/pull/20

sharktank/tests/models/punet/sharded_resnet_block_with_iree_test.py
```diff
@@ -0,0 +1,247 @@
...
+    assert sharded_input_image.shard_count == sharded_input_time_emb.shard_count
+    shard_count = sharded_input_image.shard_count
+    logger.info(f"Running module.")
+    hal_driver = iree.runtime.get_driver("local-task")
+    vm_instance = iree.runtime.VmInstance()
+    available_devices = hal_driver.query_available_devices()
+    # Use the same actual device for all devices.
+    devices = [
+        hal_driver.create_device(available_devices[0]) for _ in range(shard_count)
+    ]
```
sogartar, line 61: @stellaraccident, you mentioned that we can target multiple IREE CPU devices.
What I have done here is effectively equivalent to
```
iree-run-module \
  --device=local-task \
  --device=local-task \
  ....
```
This sets the assigns `local-task` to the 2 device promises in the model.
Is this what you had in mind?

sogartar:
The test in this PR is failing with large numerical imprecisions compared to the expectation.

stellaraccident (approved):
Thanks. I may refactor/break this up into a better example at some point, but good to at least have the test for now.

stellaraccident:
> The test in this PR is failing with large numerical imprecisions compared to the expectation.

If you can't find the imprecision quickly, get it submitted as a WIP with the precision check commented. Then will need to triage.

PR #125, opened 2024-08-15
[libshortfin] Wire Python asyncio to IREE concurrency.

stellaraccident:
* Add user APIs to start workers on the System.
* Rework System destruction to have an explicit Shutdown phase. Use this from Python in an atexit handler to explicitly shutdown any system instances live at exit (needed to avoid various deadlocks and unsavory things).
* Adds a specialized PyWorker with overrides to do Python accounting (set up event loop on the thread, GIL maintenance).
* Add Worker call and call_threadsafe APIs.
* Create an AbstractEventLoop that is sufficient to launch coroutines on a worker.
* Clean up various init and shutdown order issues to make use resilient.
* Skating towards having lightweight Process and queues to drive device work, letting them oversubscribe worker threads as needed.

The same underlying APIs should be able to be made to work for C++ coroutines but that will be a separate effort.


libshortfin/src/shortfin/local/system.h
```diff
@@ -61,7 +75,20 @@ class SHORTFIN_API System : public std::enable_shared_from_this<System> {
   // Creates a new Scope bound to this System (it will internally
   // hold a reference to this instance). All devices in system order will be
   // added to the scope.
-  std::unique_ptr<Scope> CreateScope();
+  std::shared_ptr<Scope> CreateScope(Worker &worker);
+
+  // Creates a scope bound to the init worker.
+  std::shared_ptr<Scope> CreateScope();
+
+  // Workers.
```
benvanik, line 83:
```suggestion
```

benvanik (approved):

PR #128, opened 2024-08-17
[shortfin] Plumb through await handling to Python asyncio.

stellaraccident:
* Stubs are in place that can be adapted for C++ coroutines but left for later.
* Makes `Process` awaitable (for termination).
* Makes `ScopedDevice` awaitable (for sync).
* There's substantial room to profile/optimize this on the Python side, but focusing on getting everything to connect first.

Handful of other changes:

* Adds CTest tests for some of the native support library.
* Adds a thread local to track the current native worker (saves a lot of Python jumping).
* Introduces a CompletionEvent C++ class and associated Python binding. This supports basic signaling and will be enhanced later with value producing async objects.
* Namespaces the `iree_` support stuff under `iree::`
* Adds a BlockingExecutor to the system which can run arbitrary blocking thunks. Uses this to workaround an issue with exporting hal semaphores to wait_sources (this will get attention this week to remove the hack).
* Fixes a bad copy vs ref of slim_mutex, rendering mutexes all unique (caught by unit tests). Deleted copy constructors from all classes not meant to be copyable.
* Added a shared_event class which heap allocated and reference counts an event. This is useful for various situations where an event producer and consumer are separated in terms of object ownership.
* Added a device_sync example which demonstrates interleaved fills/syncs.

### DIFF
--- a/libshortfin/src/shortfin/local/async.cc
+++ b/libshortfin/src/shortfin/local/async.cc
@@ -1,0 +1,49 @@
+// Copyright 2024 Advanced Micro Devices, Inc
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "shortfin/local/async.h"
+
+namespace shortfin::local {
+
+SingleWaitFuture::SingleWaitFuture()
+    : wait_source_(iree_wait_source_immediate()) {}
+
+SingleWaitFuture::SingleWaitFuture(iree_shared_event::ref event)
+    : wait_source_(event->await()) {
+  // It is sufficient to simply capture the event ref in the lambda. It
+  // will be lifetime extended for as long as the resource control
+  // or its copies are alive.
+  resource_control_ = [event](ResourceCommand) {};
+}
+
+SingleWaitFuture::SingleWaitFuture(iree_hal_semaphore_ptr sem, uint64_t payload)
+    : wait_source_(iree_hal_semaphore_await(sem, payload)) {
+  resource_control_ = [sem](ResourceCommand) {};
+}
+
+SingleWaitFuture::~SingleWaitFuture() {
+  if (resource_control_) {
+    resource_control_(ResourceCommand::RELEASE);
+  }
+}
+
+bool SingleWaitFuture::is_ready() {
+  iree_status_code_t status_code;
+  SHORTFIN_THROW_IF_ERROR(iree_wait_source_query(wait_source_, &status_code));
+  return status_code == IREE_STATUS_OK;
+}
+
+bool SingleWaitFuture::BlockingWait(iree_timeout_t timeout) {
+  auto status = iree_wait_source_wait_one(wait_source_, timeout);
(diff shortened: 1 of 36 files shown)


libshortfin/src/shortfin/local/async.cc
```diff
@@ -0,0 +1,45 @@
...
+#include "shortfin/local/async.h"
+
+namespace shortfin::local {
+
+CompletionEvent::CompletionEvent()
+    : wait_source_(iree_wait_source_immediate()) {}
+
+CompletionEvent::CompletionEvent(iree::shared_event::ref event)
+    : wait_source_(event->await()) {
+  // // It is sufficient to simply capture the event ref in the lambda. It
```
benvanik, line 16: TODO?

libshortfin/src/shortfin/local/scheduler.cc
```diff
@@ -36,6 +39,29 @@ void Account::active_deps_extend(iree_hal_semaphore_list_t sem_list) {
   }
 }
 
+CompletionEvent Account::OnSync() {
+  bool hack_has_working_wait_source = false;
+  if (hack_has_working_wait_source) {
+    return CompletionEvent(sem_, idle_timepoint_);
+  } else {
+    // TODO: Burn this path with fire! No attempt has been made to make this
```
benvanik, line 47: 200w 1

libshortfin/src/shortfin/support/blocking_executor.cc
```diff
@@ -0,0 +1,162 @@
...
+    if (iree_time_now() > report_deadline) {
+      logging::warn(
+          "Still waiting for blocking executor threads to terminate (live "
+          "threads = {}, total created = {}).",
+          last_live_thread_count, total_thread_count);
+      report_deadline = iree_timeout_as_deadline_ns(warn_timeout);
+    }
+
+    // Short spin delay to signal/wait.
+    iree_wait_until(iree_timeout_as_deadline_ns(iree_make_timeout_ms(250)));
```
benvanik, line 78: 250ms is quite awhile - shouldn't be needed? you can use `iree_thread_join` to join with the thread - if the issue is that you don't want an infinite wait there we could add a iree_thread_try_join that takes a timeout

benvanik (approved):

libshortfin/src/shortfin/local/async.cc
stellaraccident: Oh, thanks for the catch. Refactor that I didn't finish. Meant to initialize the resource_baton (was using a naked lambda for this, which is where the comment came from).

libshortfin/src/shortfin/local/scheduler.cc
stellaraccident: lol, 'tis ok. Forced me to wire in a blocking executor, which is needed for other things.

libshortfin/src/shortfin/support/blocking_executor.cc
stellaraccident: Yeah, isn't great but I really do want the deadline (have debugged too many shutdown issues to not have hang reporting). Dropped the timeout to 50ms and added a TODO: `    // TODO: Introduce iree_thread_try_join and refactor the shutdown sequence
    // to be based on thread joining with a warn deadline.`

PR #129, opened 2024-08-20
[shortfin] Start building out serving demo spike.

stellaraccident:
This is the first step in a spike to add minimal features and API for standing up a mobilenet server from an ONNX model in the ONNX model zoo, and it is being used to build out basic features:

* Make Worker monomorphic and add a mechanism so that the Python bindings can stash needed state and get lifecycle events that make interop possible (without subclassing). This allows us to move the current worker TLS lookup to the C++ side, where the rest of the framework can use it.
* Adds basic bindings for `ProgramModule` and `Program`, allowing loading from VMFB and setting up with all devices in a scope.
* Adds a `Message` base class that will be used for signaling between processes. A little tricky because of the vague ownership and binding/extension needs.
* Adds a `Queue` class with very simple support for readers and writers. (This will be extended a lot later to support fancy features like back-pressure shutdown sequencing, etc)
* Adds a `Future` class that can be resolved with a concrete type or exception. `VoidFuture` is hand-coded. Others are via `TypedFuture` template. Implement `MessageFuture`.
* Tie `Future` into asyncio so it is awaitable.
* Add script to build demo model and scaffolding for a server.

There is some torturous code in some of this because there are only so many ways to make it idiomatic C++ and bindable to Python. There are some shutdown and exception reporting issues still to work out (which is typical of this state of implementation).


libshortfin/src/shortfin/local/async.h
```diff
@@ -64,6 +64,105 @@ class SHORTFIN_API CompletionEvent {
...
+  // Access the raw done wait source.
+  operator const iree_wait_source_t() { return state_->done_event_.await(); }
+
+ protected:
+  struct BaseState {
+    BaseState() : done_event_(false) {}
+    virtual ~BaseState() = default;
+    int ref_count_ = 1;
+    iree::event done_event_;
+    iree_status_t failure_status_ = iree_ok_status();
```
benvanik, line 102: it may be worth it to make a C++ wrapper for status - I found it useful - I'm not sure how many statuses (if any) are being leaked here. e.g. it may be a contractual requirement that ThrowFailure is always called before a Future ever goes out of scope, but I don't think that's guaranteed here and if failure_status_ points to a non-ok status then it'll leak.

all status instances must have iree_status_ignore or iree_status_free called on them - they both do the same thing but ignore improves readability when the user is explicitly ignoring an error (and we may have a debug flag at some point that prints out ignored errors to help track down missing ones). `runtime/src/iree/base/status_cc.h` is the one we provide in IREE that works like abseil Status (mostly), supports move semantics for transferring statuses, etc, and also has StatusOr if you need it.

libshortfin/src/shortfin/local/program.cc
```diff
@@ -0,0 +1,55 @@
...
+                                  bool mmap) {
+  iree::file_contents_ptr contents;
+  iree_file_read_flags_t flags =
+      mmap ? IREE_FILE_READ_FLAG_MMAP : IREE_FILE_READ_FLAG_PRELOAD;
+  SHORTFIN_THROW_IF_ERROR(iree_file_read_contents(
+      path.c_str(), flags, system.host_allocator(), contents.for_output()));
+
+  // Ownership hazard: iree_vm_bytecode_module_create only assumes ownership
+  // when it returns *sucessfully*. In the exceptional case, ownership
+  // remains with
```
benvanik, line 27: (truncated comment)

libshortfin/src/shortfin/local/scope.cc
```diff
@@ -82,6 +88,40 @@ std::vector<std::string_view> Scope::device_names() const {
...
+  std::vector<iree_hal_device_t *> raw_devices;
+
+  // By default, bind all devices in the scope in order to the program.
+  for (Device *d : devices_) {
+    raw_devices.push_back(d->hal_device());
+  }
+
+  // Add a HAL module.
+  iree::vm_module_ptr hal_module;
+  SHORTFIN_THROW_IF_ERROR(iree_hal_module_create(
```
benvanik, line 103: fyi: at some point you may want to change this to something similar to what the tooling does in `iree_tooling_resolve_modules` - it uses `iree_vm_module_enumerate_dependencies` to walk the dependencies and add the required modules only as needed. to start you could use it just to see if the hal is used, but as you add other module types for exposing sharkfin functionality (or module versions; `iree_vm_module_dependency_t` has the minimum version required so you can switch between them, and whether they are optional/required) you'll probably want to build around that

libshortfin/src/shortfin/local/scope.cc
```diff
@@ -82,6 +88,40 @@ std::vector<std::string_view> Scope::device_names() const {
...
+  all_modules.push_back(hal_module);
+
+  // Add explicit modules.
+  for (auto &pm : modules) {
+    all_modules.push_back(pm.vm_module());
+  }
+
+  // Create the context.
+  iree::vm_context_ptr context;
+  iree_vm_context_flags_t flags = IREE_VM_CONTEXT_FLAG_CONCURRENT;
```
benvanik, line 116: heads up that today this mostly just changes the way tracy reports things such that each context gets a fiber track - not sure if you've got tracy running yet but if you see it looking different than normal this is the reason

libshortfin/src/shortfin/support/iree_helpers.h
```diff
@@ -11,6 +11,7 @@
 #include <stdexcept>
 
 #include "iree/base/api.h"
+#include "iree/base/internal/file_io.h"
```
benvanik, line 14: fyi: file_io.h sucks and will be removed at some point (why it's in internal/) - fine for now and when it does get axed or changed you can copy it local if you still want it

benvanik (approved):

libshortfin/src/shortfin/local/async.h
stellaraccident: Thanks, did this. The use cases for wanting an ignorable status are pretty small, mainly in this kind of Future object where it is legal to completely ignore it from the user side.

libshortfin/src/shortfin/local/scope.cc
stellaraccident: Thanks. Converted to a TODO.

libshortfin/src/shortfin/support/iree_helpers.h
stellaraccident: ack. It is better than what was done in the prior bindings. Will take the guts off your hands when needed.

libshortfin/src/shortfin/local/scope.cc
stellaraccident: ack

benvanik (approved):

PR #131, opened 2024-08-20
[shortfin] Introduce system options

marbre:
These options allow to deselect specific runtime drivers. With this libshortfin can be built, even if a driver is not available in the IREE build directory, avoiding a linker failure. The components are still built, but do not get linked.


marbre:
Superseded. Will be addressed in a new PR.

PR #132, opened 2024-08-20
[build] Improve dependency management for builds

Groverkss:
Add logic to cmake configuration to automatically download dependencies if they cannot be found. Also adds a flag to specify SHORTFIN_IREE_SOURCE_DIR to build iree out-of-tree.

### DIFF
--- a/libshortfin/CMakeLists.txt
+++ b/libshortfin/CMakeLists.txt
@@ -23,21 +23,14 @@ set(CMAKE_CXX_STANDARD 20)
 # https://discourse.cmake.org/t/cmake-3-28-cmake-cxx-compiler-clang-scan-deps-notfound-not-found/9244/3
 set(CMAKE_CXX_SCAN_FOR_MODULES 0)
 
+# Problems with libfmt without fPIC
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
+
 option(SHORTFIN_BUILD_PYTHON_BINDINGS "Builds Python Bindings" OFF)
 option(SHORTFIN_BUILD_TESTS "Builds C++ tests" ON)
 
-if(SHORTFIN_BUILD_TESTS)
-  include(FetchContent)
-  FetchContent_Declare(
-    googletest
-    URL https://github.com/google/googletest/archive/03597a01ee50ed33e9dfd640b249b4be3799d395.zip
-  )
-  # For Windows: Prevent overriding the parent project's compiler/linker settings
-  set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
-  FetchContent_MakeAvailable(googletest)
-  include(GoogleTest)
-  enable_testing()
-endif()
+include(FetchContent)
 
 # Includes.
 list(APPEND CMAKE_MODULE_PATH
@@ -46,9 +39,77 @@ list(APPEND CMAKE_MODULE_PATH
 include(shortfin_library)
 
 # Dependencies.
-find_package(spdlog)
-find_package(xtensor)
-find_package(IREERuntime)
+
+## fmt
+FetchContent_Declare(
+  fmt
+  GIT_REPOSITORY https://github.com/fmtlib/fmt.git
+  GIT_TAG        e69e5f977d458f2650bb346dadf2ad30c5320281 # 10.2.1 (sync with spdlog)
+  # Do not download if package is already available
+  FIND_PACKAGE_ARGS NAMES fmt
+)
+
+## spdlog
+set(SPDLOG_FMT_EXTERNAL ON)
+FetchContent_Declare(
+  spdlog
+  GIT_REPOSITORY https://github.com/gabime/spdlog.git
+  GIT_TAG        2d4acf8cc321d7783d8f2e22e17a794c6d0e9450 # v1.14.1
+  # Do not download if package is already available
+  FIND_PACKAGE_ARGS NAMES spdlog
+)
+
+## xtl: required for xtensor
+FetchContent_Declare(
+  xtl
+  GIT_REPOSITORY https://github.com/xtensor-stack/xtl.git
+  GIT_TAG        a7c1c5444dfc57f76620391af4c94785ff82c8d6 # v0.7.7
+  # Do not download if package is already available
+  FIND_PACKAGE_ARGS NAMES xtl
+)
+
+## xtensor
+FetchContent_Declare(
+  xtensor
+  GIT_REPOSITORY https://github.com/xtensor-stack/xtensor.git
+  GIT_TAG        3634f2ded19e0cf38208c8b86cea9e1d7c8e397d # v0.25.0
+  # Do not download if package is already available
+  FIND_PACKAGE_ARGS NAMES xtensor
+)
+
+FetchContent_MakeAvailable(fmt spdlog xtl xtensor)
+
+## iree runtime
+
+set(SHORTFIN_IREE_SOURCE_DIR "" CACHE FILEPATH "Path to IREE source")
+if(SHORTFIN_IREE_SOURCE_DIR)
+  set(IREE_BUILD_COMPILER OFF)
+  set(IREE_BUILD_TESTS OFF)
+  add_subdirectory(${SHORTFIN_IREE_SOURCE_DIR} iree SYSTEM)
+else()
+  # Try to find iree using find_package
+  # TODO: Instead fetch bundled iree from releases and build using that
+  find_package(IREERuntime)
+endif()
+
+# tests
+
+if(SHORTFIN_BUILD_TESTS)
+  if (SHORTFIN_IREE_SOURCE_DIR)
+    # Kind of a bug with iree? It pulls googletest even when tests are disabled.
+    # IREE already builds googletest from source
+  else()
+    FetchContent_Declare(
+      googletest
+      URL https://github.com/google/googletest/archive/03597a01ee50ed33e9dfd640b249b4be3799d395.zip
+    )
+    # For Windows: Prevent overriding the parent project's compiler/linker settings
+    set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
+    FetchContent_MakeAvailable(googletest)
+  endif()
+  include(GoogleTest)
+  enable_testing()
+endif()
 
 add_subdirectory(src)
 


libshortfin/CMakeLists.txt
```diff
@@ -23,21 +23,14 @@ set(CMAKE_CXX_STANDARD 20)
 # https://discourse.cmake.org/t/cmake-3-28-cmake-cxx-compiler-clang-scan-deps-notfound-not-found/9244/3
 set(CMAKE_CXX_SCAN_FOR_MODULES 0)
 
+# Problems with libfmt without fPIC
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
```
marbre, line 28: Instead of modifying these CMAKE flags you probably want 
```suggestion
set (CMAKE_POSITION_INDEPENDENT_CODE ON)
```

libshortfin/CMakeLists.txt
```diff
@@ -23,21 +23,14 @@ set(CMAKE_CXX_STANDARD 20)
 # https://discourse.cmake.org/t/cmake-3-28-cmake-cxx-compiler-clang-scan-deps-notfound-not-found/9244/3
 set(CMAKE_CXX_SCAN_FOR_MODULES 0)
 
+# Problems with libfmt without fPIC
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
```
stellaraccident, line 27: Just do (vs munging c flags): set(CMAKE_POSITION_INDEPENDENT_CODE ON)

We will probably want to not do this on Windows.

libshortfin/CMakeLists.txt
```diff
@@ -46,9 +39,77 @@ list(APPEND CMAKE_MODULE_PATH
 include(shortfin_library)
 
 # Dependencies.
-find_package(spdlog)
-find_package(xtensor)
-find_package(IREERuntime)
+
+## fmt
```
stellaraccident, line 43: We need to condition this on some kind of SHORTFIN_BUNDLE_DEPS option and if it is false, use find_package. Not bundling the deps is important for certain scenarios and package managers.

libshortfin/CMakeLists.txt
```diff
@@ -46,9 +39,77 @@ list(APPEND CMAKE_MODULE_PATH
...
+  GIT_TAG        3634f2ded19e0cf38208c8b86cea9e1d7c8e397d # v0.25.0
+  # Do not download if package is already available
+  FIND_PACKAGE_ARGS NAMES xtensor
+)
+
+FetchContent_MakeAvailable(fmt spdlog xtl xtensor)
+
+## iree runtime
+
+set(SHORTFIN_IREE_SOURCE_DIR "" CACHE FILEPATH "Path to IREE source")
```
stellaraccident, line 84: Can you move options to the top of the file?

libshortfin/CMakeLists.txt
Groverkss: oh that's cool! I'm not sure why this is happening though. My guess is it's some weird dependency web between spdlog and fmt.

libshortfin/CMakeLists.txt
```diff
@@ -46,9 +39,77 @@ list(APPEND CMAKE_MODULE_PATH
...
+else()
+  # Try to find iree using find_package
+  # TODO: Instead fetch bundled iree from releases and build using that
+  find_package(IREERuntime)
+endif()
+
+# tests
+
+if(SHORTFIN_BUILD_TESTS)
+  if (SHORTFIN_IREE_SOURCE_DIR)
```
stellaraccident, line 98: We should modernize how IREE does this so it plays nicely with others. The right way to do it would be to teach IREE how to fetch googletest dynamically and then make ours available before including IREE (these work as first one wins).

libshortfin/CMakeLists.txt
```diff
@@ -46,9 +39,77 @@ list(APPEND CMAKE_MODULE_PATH
...
+
+FetchContent_MakeAvailable(fmt spdlog xtl xtensor)
+
+## iree runtime
+
+set(SHORTFIN_IREE_SOURCE_DIR "" CACHE FILEPATH "Path to IREE source")
+if(SHORTFIN_IREE_SOURCE_DIR)
+  set(IREE_BUILD_COMPILER OFF)
+  set(IREE_BUILD_TESTS OFF)
+  add_subdirectory(${SHORTFIN_IREE_SOURCE_DIR} iree SYSTEM)
```
stellaraccident, line 88: Let's also only enable backends that we care about: right now just local and amdgpu.

Also, for my own edification, why do you need SYSTEM here?

stellaraccident (changes requested):

libshortfin/CMakeLists.txt
Groverkss: done

libshortfin/CMakeLists.txt
Groverkss: done

libshortfin/CMakeLists.txt
marbre: Following the discussion on discord, seems like there is a `unused function` error propagating from IREE and due to `-Wno-used-function` set. However, just building with `IREE_BUILD_COMPILER_OFF` shouldn't trigger that error (we don't see it in iree-bare-metal-arm) but I can dig into with @Groverkss tomorrow.

Groverkss: The error comes because IREE uses -Wno-unused-function and other compilation options, while our CMakeLists.txt doesn't (for now at least). IREE will fail to compile without these flags. I used SYSTEM to isolate IREE's compilation flags to itself. I'm guessing this could also bite us in future if we wanted to use some of our own flags. Not sure whats preferred here.

libshortfin/CMakeLists.txt
Groverkss: I'm not very sure how to do that on the IREE side. Can we namespace the iree googletest dependency?

libshortfin/CMakeLists.txt
```diff
@@ -46,9 +44,98 @@ list(APPEND CMAKE_MODULE_PATH
...
+    FetchContent_Declare(
+      googletest
+      URL https://github.com/google/googletest/archive/03597a01ee50ed33e9dfd640b249b4be3799d395.zip
+    )
+    # For Windows: Prevent overriding the parent project's compiler/linker settings
+    set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
+    FetchContent_MakeAvailable(googletest)
+  endif()
+    include(GoogleTest)
+    enable_testing()
```
marbre, line 136:
```suggestion
  include(GoogleTest)
  enable_testing()
```

libshortfin/CMakeLists.txt
```diff
@@ -46,9 +44,98 @@ list(APPEND CMAKE_MODULE_PATH
...
+  # Try to find iree using find_package
+  find_package(IREERuntime)
+endif()
+
+# tests
+
+if(SHORTFIN_BUILD_TESTS)
+  if (NOT SHORTFIN_IREE_SOURCE_DIR)
+    # Due to the above mentioned bugged with IREE CMakeLists.txt, we reuse
+    # IREE's googletest.
```
marbre, line 126:
```suggestion
    # For now we use gtest shipped alongside with IREE.
```

libshortfin/CMakeLists.txt
```diff
@@ -46,9 +44,98 @@ list(APPEND CMAKE_MODULE_PATH
...
+    FetchContent_Populate(iree)
+  endif()
+  set(SHORTFIN_IREE_SOURCE_DIR ${iree_SOURCE_DIR})
+endif()
+
+if(SHORTFIN_IREE_SOURCE_DIR)
+  set(IREE_BUILD_COMPILER OFF)
+  set(IREE_BUILD_TESTS OFF)
+  set(IREE_BUILD_SAMPLES OFF)
+  # Disable missing submodules error because we are only building the runtime
```
marbre, line 108:
```suggestion
  # Disable missing submodules error because we are only building the runtime.
```

libshortfin/CMakeLists.txt
```diff
@@ -46,9 +44,98 @@ list(APPEND CMAKE_MODULE_PATH
...
+  set(IREE_HAL_DRIVER_LOCAL_SYNC ON)
+  set(IREE_HAL_DRIVER_LOCAL_TASK ON)
+  set(IREE_HAL_DRIVER_HIP ON)
+  add_subdirectory(${SHORTFIN_IREE_SOURCE_DIR} shortfin_iree SYSTEM)
+else()
+  # Try to find iree using find_package
+  find_package(IREERuntime)
+endif()
+
+# tests
```
marbre, line 121: If you want to highlight that this a section, maybe change to
```suggestion
#-------------------------------------------------------------------------------
# Tests
#-------------------------------------------------------------------------------
```
? But I think we can also introduce sections at a later time when the CMake config has grown a little more.

libshortfin/CMakeLists.txt
```diff
@@ -46,9 +44,98 @@ list(APPEND CMAKE_MODULE_PATH
...
+
+if (NOT SHORTFIN_IREE_SOURCE_DIR AND SHORTFIN_BUNDLE_DEPS)
+  FetchContent_Declare(
+    iree
+    GIT_REPOSITORY https://github.com/iree-org/iree.git
+    GIT_TAG candidate-20240821.992
+    # TODO: We shouldn't have to pull googletest when we are not building tests.
+    #       This needs to be fixed with IREE.
+    GIT_SUBMODULES "third_party/benchmark third_party/cpuinfo third_party/flatcc third_party/hip-build-deps third_party/googletest"
+    GIT_SHALLOW ON
```
marbre, line 95: As `GIT_SHALLOW` is only documented for https://cmake.org/cmake/help/latest/module/ExternalProject.html and as I couldn't find it for `FetchContent`, does this has an effect?

libshortfin/CMakeLists.txt
```diff
@@ -46,9 +44,98 @@ list(APPEND CMAKE_MODULE_PATH
...
+  set(IREE_BUILD_TESTS OFF)
+  set(IREE_BUILD_SAMPLES OFF)
+  # Disable missing submodules error because we are only building the runtime
+  set(IREE_ERROR_ON_MISSING_SUBMODULES OFF)
+  # Only enable local_sync/local_task/hip drivers for now.
+  set(IREE_HAL_DRIVER_DEFAULTS OFF)
+  set(IREE_HAL_DRIVER_LOCAL_SYNC ON)
+  set(IREE_HAL_DRIVER_LOCAL_TASK ON)
+  set(IREE_HAL_DRIVER_HIP ON)
+  add_subdirectory(${SHORTFIN_IREE_SOURCE_DIR} shortfin_iree SYSTEM EXCLUDE_FROM_ALL)
```
marbre, line 115: An alternative to `SYSTEM` could be to explicitly adjust the flags via
```cmake
set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-unused-variable")
```

marbre (commented):
Some first thoughts. Will test locally before hitting the approve button.

libshortfin/CMakeLists.txt
Groverkss: seems a bit unrelated to the PR

libshortfin/CMakeLists.txt
Groverkss: FetchContent internally uses ExternalProject afaiu. It does the same thing as documented for ExternalProject, i.e. --depth 1.

libshortfin/CMakeLists.txt
Groverkss: We could yeah, I'm fine with either.

libshortfin/CMakeLists.txt
```diff
@@ -46,9 +44,98 @@ list(APPEND CMAKE_MODULE_PATH
...
+## iree runtime
+
+if (NOT SHORTFIN_IREE_SOURCE_DIR AND SHORTFIN_BUNDLE_DEPS)
+  FetchContent_Declare(
+    iree
+    GIT_REPOSITORY https://github.com/iree-org/iree.git
+    GIT_TAG candidate-20240821.992
+    # TODO: We shouldn't have to pull googletest when we are not building tests.
+    #       This needs to be fixed with IREE.
+    GIT_SUBMODULES "third_party/benchmark third_party/cpuinfo third_party/flatcc third_party/hip-build-deps third_party/googletest"
```
stellaraccident, line 94: Interesting. I would have expected that this was semicolon delimitted or not quoted. I expect this is working like this as a quirk of passthrough to the underlying git command.

stellaraccident (approved):

libshortfin/CMakeLists.txt
marbre: Great, thanks for the clarification!

libshortfin/CMakeLists.txt
marbre: No strong preference on my side. Let's go with what you have for now :)

libshortfin/CMakeLists.txt
marbre: I agree we should take a look into modernizing IREE. For now I am fine with this but we should add it to the list for future improvements.

libshortfin/CMakeLists.txt
marbre: Let's go with what you have (at least) for now :)

libshortfin/CMakeLists.txt
```diff
@@ -46,9 +44,98 @@ list(APPEND CMAKE_MODULE_PATH
...
+    GIT_SHALLOW ON
+  )
+  FetchContent_GetProperties(iree)
+  if(NOT iree_POPULATED)
+    FetchContent_Populate(iree)
```
marbre, line 99: This gives me
> CMake Warning (dev) at /usr/share/cmake-3.30/Modules/FetchContent.cmake:1953 (message):
  Calling FetchContent_Populate(iree) is deprecated, call
  FetchContent_MakeAvailable(iree) instead.  Policy CMP0169 can be set to OLD
  to allow FetchContent_Populate(iree) to be called directly for now, but the
  ability to call it with declared details will be removed completely in a
  future version.
Call Stack (most recent call first):
  CMakeLists.txt:99 (FetchContent_Populate)
This warning is for project developers.  Use -Wno-dev to suppress it.

Fine if addresses in a follow-up though.

libshortfin/CMakeLists.txt
marbre: FYI, looking into what's actually cloned, the submodules are fully fetched and `GIT_SHALLOW ON` does not seem to have an effect here. Anyway, this is probably worth a look if there someday is a need to build the compiler and fetch LLVM.

marbre (approved):

libshortfin/CMakeLists.txt
Groverkss: Yeah, that's a weird thing about FetchContent. They recommend using FetchContent_MakeAvailable, which already does a add_subdirectory. I like fetching and setting the directory directly. Can change it later if needed.

libshortfin/CMakeLists.txt
Groverkss: Interesting, I think some options don't work as expected. I'll remove GIT_SHALLOW if it's a no-op anyway.

marbre: It works for IREE but not for its submodules. So no need to remove it.

PR #133, opened 2024-08-21
Add GH workflow to test libshortfin

marbre:
This adds a very basic GitHub action to

 * Build the IREE runtime from source (no caching, not using any of
   IREE's `build_tools/cmake/*.sh` scripts yet)
 * Build libshortfin
 * Run libshortfin tests via CMake

### DIFF
--- a/.github/workflows/ci_linux_x64-libshortfin.yml
+++ b/.github/workflows/ci_linux_x64-libshortfin.yml
@@ -1,0 +1,105 @@
+#!/bin/bash
+# Copyright 2024 Advanced Micro Devices, Inc
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+name: CI - libshorthin
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_linux_x64-libshortfin.yml
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
+# Copyright 2024 Advanced Micro Devices, Inc
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+name: CI - libshorthin
```
Groverkss, line 8: nit: libshortfin

stellaraccident (approved):

.github/workflows/ci_linux_x64-libshortfin.yml
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
```
ScottTodd, line 1: I suspect this shebang is breaking syntax highlighting in this file by having GitHub treat the file as a bash script instead of a yaml file.

marbre: Thanks, accidentally copied that with the header. Will push the fix with one of the next refactorings.

PR #134, opened 2024-08-21
[shortfin][build] Fetch iree if SHORTFIN_BUNDLE_DEPS is specified

Groverkss:


Groverkss:
Depends on https://github.com/nod-ai/sharktank/pull/132

PR #137, opened 2024-08-22
[libshortfin] Introduce `SHORTFIN_SYSTEMS_AMDGPU`

marbre:
Introduces an `SHORTFIN_SYSTEMS_AMDGPU` option to allow building
libshortfin for host-only systems. The option defaults to ON.

### DIFF
--- a/libshortfin/src/CMakeLists.txt
+++ b/libshortfin/src/CMakeLists.txt
@@ -13,6 +13,11 @@ target_include_directories(
                            $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>)
 
 
+set(_INIT_INTERNAL_DEPS)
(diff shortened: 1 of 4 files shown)


libshortfin/src/CMakeLists.txt
```diff
@@ -13,13 +13,19 @@ target_include_directories(
...
+
 shortfin_public_library(
   NAME
     shortfin
   COMPONENTS
     shortfin_array
     shortfin_local
     shortfin_support
-    shortfin_systems_amdgpu
     shortfin_systems_host
```
Groverkss, line 29: This is part of _INIT_INTERNAL_DEPS now, so this should be removed from here

Groverkss (approved):

PR #139, opened 2024-08-22
[shortfin] Teach setup.py to build using cmake

Groverkss:


Groverkss:
Looking for an initial review. This setup works, the only thing I don't like is hardcoding clang/clang++, because libshortfin doesn't compile with GCC (right now, due to some comparisons between data types)

libshortfin/setup.py
```diff
@@ -55,8 +70,127 @@ def copy_extensions_to_source(self, *args, **kwargs):
...
+            else:
+                print(f"Not re-configing (already configured)", file=sys.stderr)
+
+            # Build.
+            subprocess.check_call(["cmake", "--build", "."], cwd=CMAKE_BUILD_DIR)
+            print("Build complete.", file=sys.stderr)
+
+            # We only take _shortfin_default from the build.
+            target_dir = os.path.join(
+                os.path.abspath(self.build_lib), "_shortfin_default"
```
stellaraccident, line 158: Fine to leave for now but we're going to end up with multiple cmake invocations producing different variants of this directory (ie for tracing builds, etc). For context, see how iree runtime/setup.py does it.

libshortfin/setup.py
```diff
@@ -18,35 +23,45 @@
...
 # paths relative to the directory containing setup.py. Why? No one knows.
-current_dir = Path(__file__).resolve().parent
-rel_source_dir = source_dir.relative_to(current_dir, walk_up=True)
-rel_binary_dir = binary_dir.relative_to(current_dir, walk_up=True)
+REL_SOURCE_DIR = SOURCE_DIR.relative_to(SETUPPY_DIR, walk_up=True)
+REL_BINARY_DIR = BINARY_DIR.relative_to(SETUPPY_DIR, walk_up=True)
 
 
-class BuiltExtension(Extension):
+class CMakeExtension(Extension):
```
stellaraccident, line 52: I think that if is_cpp_prebuilt you still want to no-op this.

libshortfin/setup.py
```diff
@@ -71,16 +205,18 @@ def copy_extensions_to_source(self, *args, **kwargs):
...
-        "shortfin": str(python_src_dir / "shortfin"),
+        "shortfin": str(PYTHON_SOURCE_DIR / "shortfin"),
     },
     ext_modules=[
-        BuiltExtension("_shortfin_default.lib"),
+        CMakeExtension("_shortfin_default.lib")
         # TODO: Conditionally map additional native library variants.
     ],
     cmdclass={
+        "build": CustomBuild,
```
stellaraccident, line 218: I think you may need to condition this somehow on is_cpp_prebuilt

stellaraccident (approved):
Looks reasonable as a starting point. I would like to make sure the current way where when invoked from the build dir still works.

libshortfin/setup.py
Groverkss: Yeah i used runtime/setup.py as reference. I think it should be easy to do that, I would add a suffix here for the specific build, something like _shortfin_default_tracy.

libshortfin/setup.py
Groverkss: With is_cpp_prebuilt, the logic in CMakeBuildPy makes sure that we don't try to run cmake if we have prebuilt things.

PR #141, opened 2024-08-22
[tuner] Add candidate generator script

RattataKing:
The plan is to move the tuner from the [sdxl-scripts repo](https://github.com/nod-ai/sdxl-scripts.git) to the sharktank repo.
This PR is the first step, moving `candidate_gen.py` (the candidate generator for tuning) along with its CI and unit tests. 
More changes will come to fully decouple it.

New changes in `candidate_gen.py`:
- Edited comments
- Added a registry class for dispatch kind
- Moved dispatch kind specific functions from global to their belonged classes
- Changed how it walks op and when it decides which functions to use for candidate generation.
- Checked if input mlir translation info contains `LLVMGPUVectorDistribute`

`candidate_gen_test.py`:
- No new tests added, updated the existing function to match changes in `candidate_gen.py`


tuner-requirements.txt
ScottTodd, line 1: I'd recommend putting this file in the `sharktank/sharktank/tools/tuner/` folder to make it easier to lift from sharktank to a new repo later on.

I've also been starting to name requirement files `requirements-[description].txt` instead of `[description]-requirements.txt` so that alphabetical sorting puts all requirements files next to each other. I looked around a bit to see if there is a standard convention across repos but wasn't sure.

.github/workflows/ci.yaml
```diff
@@ -53,9 +53,11 @@ jobs:
           # from non default locations first. Installing the PyTorch CPU
           # wheels saves multiple minutes and a lot of bandwidth on runner setup.
           pip install --no-compile -r pytorch-cpu-requirements.txt
+          pip install --no-compile -r tuner-requirements.txt
```
ScottTodd, line 56: I think we don't (currently?) want the tuner to spread its dependencies to sharktank, so this probably should be kept separate. If the tuner becomes an integral part of sharktank then the deps should move into the base `requirements.txt` IMO.

Because of that, I suggest creating a separate job in this workflow file or creating a separate workflow file entirely, rather than add new dependencies here and try to run the tests in the same step sequence.

The main functional difference between the two options in this case is that separate workflow files can have different triggers, but you probably just want what is in this file:
```yml
on:
  workflow_dispatch:
  pull_request:
  push:
    branches:
      - main
```

## Option 1: stay in this file

```diff
jobs:
  test:
    name: "Unit Tests and Type Checking"
    strategy:
      matrix:
    ...

+  test_tuner:
+    name: "Unit Tests and Type Checking for Tuner"
+    ...  (copy/fork other steps as needed, just doing what is needed to run the tuner tests/lint)
```

## Option 2: create a new file

Like how https://github.com/nod-ai/sharktank/blob/main/.github/workflows/ci_linux_x64-libshortfin.yml is its own file, create a new `ci-tuner.yml`:

```yml
name: CI - Tuner
on:
  workflow_dispatch:
  pull_request:
  push:
    branches:
      - main
concurrency:
  # A PR number if a pull request and otherwise the commit hash. This cancels
  # queued and in-progress runs for the same PR (presubmit) or commit
  # (postsubmit). The workflow name is prepended to avoid conflicts between
  # different workflows.
  group: ${{ github.workflow }}-${{ github.event.number || github.sha }}
  cancel-in-progress: true
jobs:
  test:
    name: "Unit Tests and Type Checking"
    ...
```

sharktank/sharktank/tools/tuner/candidate_gen.py
```diff
@@ -0,0 +1,1312 @@
...
+from typing import Callable
+from textwrap import indent
+
+import iree.compiler as ireec
+from iree.compiler import ir
+from iree.compiler.dialects import _linalg_ops_gen, _util_ops_gen
+
+"""
+Usage: ./candidate_gen.py 121.mlir -o "tuning/candidates" -l 1024 --lhs-dims=mk --rhs-dims=nk --tile-dims=mnk
+"""
```
ScottTodd, line 28: I'd merge this documentation with the comment above, rather than separate the two with all the `import` statements.
```python
# Given an input dispatch, this code modifies the hyperparameters
# in the code and runs it.

import foo  # <-- these are right between the docs
import bar

"""
Usage: ./candidate_gen.py 121.mlir -o "tuning/candidates" -l 1024 --lhs-dims=mk --rhs-dims=nk --tile-dims=mnk
"""
```

Maybe following this format, all above the `import`s:
```python
"""(One line description of what purpose the script has)

(Other context, information about prerequisites, links to documentation, etc.)
(More context...)

Usage: ./candidate_gen.py 121.mlir -o "tuning/candidates" -l 1024 --lhs-dims=mk --rhs-dims=nk --tile-dims=mnk
"""
```

sharktank/sharktank/tools/tuner/candidate_gen.py
```diff
@@ -0,0 +1,1312 @@
...
+    apply_params_fn: (
+        Callable[[ProblemSize, list[str], Configuration], tuple[str, str]] | None
+    ) = None
+    if walk_result.dispatch_kind == DispatchKind.conv:
+        get_shapes_fn = get_shapes_conv
+        apply_params_fn = apply_params_conv
+    elif walk_result.dispatch_kind == DispatchKind.mmt:
+        get_shapes_fn = get_shapes_mmt
+        apply_params_fn = apply_params_mmt
+    elif walk_result.dispatch_kind == DispatchKind.contraction:
```
ScottTodd, line 1203: This is a place that could be refactored to make the tuner more modular / easier to extend without needing to keep adding code to this single 1300+ line file.

There are all sorts of ways this could be set up / design patterns this could follow. The first that comes to mind for me right now would be to
1. have a registry of "dispatch kinds"
2. have this code loop over each item in the registry
3. if the item matches the walk result, use the `get_shapes_fn` and `apply_params_fn` from that item

Or... maybe that sort of already exists in the `walk_mlir_op` function - maybe that could set `get_shapes_fn` and `apply_params_fn`?

I'm basically wondering how many places in the script need to be changed to support a new operation. That code would ideally be grouped somehow (e.g. in a function or a subclass of some sort)

sharktank/sharktank/tools/tuner/candidate_gen.py
```diff
@@ -0,0 +1,1312 @@
...
+            MfmaIntrinsic.mfma_i8_16x16x32_i32(),
+            MfmaIntrinsic.mfma_i8_32x32x16_i32(),
+        ]
+
+
+@dataclass
+class Configuration:
+    subgroup_size: int
+    workgroup_size: list[int]
+    intrinsic: MfmaIntrinsic
```
ScottTodd, line 152: This looks specific to one particular hardware target. Would be good to generalize _or_ rename this `Configuration` dataclass to more accurately describe the included data.

If I wanted to tune for Vulkan, how much of this script would I need to fork? Could I just implement some interface that ROCm is already implementing?

ScottTodd (commented):
A few comments from an initial scan. Definitely want to resolve the github workflow / requirement file topics before merge here. Could defer the script refactoring.

.github/workflows/ci.yaml
RattataKing: thanks, took the option 2

sharktank/sharktank/tools/tuner/candidate_gen.py
RattataKing: I added a registry for dispatch kind and did some workaround with op_walk function

sharktank/sharktank/tools/tuner/candidate_gen.py
RattataKing: Can we leave this generalizing task until we add the first client that needs it?

sharktank/sharktank/tools/tuner/requirements-dev.txt
```diff
@@ -0,0 +1,2 @@
+pre-commit==3.8.0
```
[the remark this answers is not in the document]
RattataKing: decide to keep it for now until some conflicts arises

.github/workflows/ci-tuner.yml
```diff
@@ -0,0 +1,49 @@
...
+        with:
+          python-version: '3.10.12'
+
+      - name: Install dev dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r sharktank/sharktank/tools/tuner/requirements-dev.txt
+
+      - name: Run pre-commit test
+        run: pre-commit run --all-files --show-diff-on-failure --color=always
```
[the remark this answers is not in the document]
RattataKing: thanks, removed it

sharktank/sharktank/tools/tuner/candidate_gen.py
```diff
@@ -0,0 +1,1386 @@
...
+
+
+class DispatchTuner(ABC):
+    @abstractmethod
+    def supports(self, mlir: str) -> bool:
+        pass
+
+    @abstractmethod
+    def get_candidate_gen_fn(self) -> CandidateGenFn:
+        pass
```
kuhar, line 1139: Nice! I think we could go one step further and instead of `DispatchTuner` returning `CandidateGenFn`, we can make these two functions be methods in `DispatchTuner`, and convert all these helper functions `get_shapes_*`, `apply_params_*` , etc. to methods in each tuner subclass.

kuhar, line 1139: Could you also add documentation comments explaining what these two abstract methods do?

sharktank/sharktank/tools/tuner/candidate_gen.py
```diff
@@ -0,0 +1,1386 @@
...
+    intrinsic: MfmaIntrinsic
+    tile_sizes: list[int]
+    subgroup_m_count: int
+    subgroup_n_count: int
+    waves_per_eu: int
+
+
+class MlirRegex(str, Enum):
+    ssa_value = r"%[a-zA-Z0-9-_]+"
+    tensor_type = r"tensor<(([0-9]+x)+((f|i)[0-9]+))>"
```
ScottTodd, line 164: This looks tricky to reason about and might be missing coverage for some types like bf16 and more exotic types like `"f8e4m3fnuz"`. Is a regex matcher really needed? How is this used? If it's being used to parse through the input program, there should be some MLIR Python APIs you can use for programmatic manipulation instead of needing to parse through the format.

sharktank/sharktank/tools/tuner/candidate_gen.py
```diff
@@ -0,0 +1,1386 @@
...
+        return rf"ins\({MlirRegex.ssa_value}, {MlirRegex.ssa_value} : (?P<LHS>{MlirRegex.tensor_type}), (?P<RHS>{MlirRegex.tensor_type})\)"
+
+    @staticmethod
+    def dps_outs_one_arg() -> str:
+        return rf"outs\({MlirRegex.ssa_value} : (?P<RES>{MlirRegex.tensor_type})\)"
+
+
+def read_input_mlir(filename: str) -> list[str]:
+    with open(filename, "r") as f:
+        return f.readlines()
```
ScottTodd, line 177: (Future refactoring, no action needed here) Good candidate for moving into a utils file that can be tested and included from multiple scripts. This is king of buried in the middle of this file.

Actually, this also depends on the file being MLIR text, right? This relates to my comment about the regex too... ideally this tool would accept MLIR bytecode as well and would use APIs to manipulate the IR instead of regex / strings.

sharktank/sharktank/tools/tuner/candidate_gen.py
```diff
@@ -0,0 +1,1386 @@
...
+        lhs_shaped_type = parse_tensor_type(lhs_tensor_type)
+        assert lhs_shaped_type.rank() == 2
+        lhs_M, lhs_K = lhs_shaped_type.shape
+
+        rhs_shaped_type = parse_tensor_type(rhs_tensor_type)
+        assert rhs_shaped_type.rank() == 2
+        rhs_N, rhs_K = rhs_shaped_type.shape
+
+        assert lhs_shaped_type.element_type == rhs_shaped_type.element_type
+        assert lhs_K == rhs_K
```
ScottTodd, line 624: Are these asserts possible to hit on valid programs? If so, this should probably be wrapped in some error handling code that prints a more detailed error message ("unimplemented feature", "unexpected input program structure", etc.)

sharktank/sharktank/tools/tuner/candidate_gen.py
```diff
@@ -0,0 +1,1386 @@
...
+
+        return ProblemSize(
+            MatmulSize(M, N, K0),
+            lhs_type=lhs_shaped_type,
+            rhs_type=rhs_shaped_type,
+            res_type=res_shaped_type,
+            dispatch_kind=DispatchKind.contraction,
+        )
+
+    assert False, "Shape not found"
```
ScottTodd, line 750: Could include more information about what was being searched for in this message

sharktank/sharktank/tools/tuner/candidate_gen.py
```diff
@@ -0,0 +1,1386 @@
...
+
+def tune(
+    input: str,
+    output: str = "",
+    limit: int = 4096,
+    num_subgroups: int = 4,
+    lhs_dims: str = "mk",
+    rhs_dims: str = "nk",
+    tile_dims: str = "mnk",
+):
```
ScottTodd, line 1265: As this is the main entrypoint function to this script, can you add some docstring comments for each argument?

sharktank/sharktank/tools/tuner/candidate_gen.py
kuhar: The parsing / mutation is still text-based. Note that this parsing is pipeline-specific -- it only supports dispatches covered by `LLVMGPUVectorDistribution`. IMO this should be fine as long as we check that this is indeed the pipeline used, and pick a less general name like `VectorDistributionMlirRegex`. Longer term, we should definitely move all of the parsing to the mlir/iree bindings.

Would you be opposed to landing the first version that still depends on text processing for the input mlir?

sharktank/tests/tuner/candidate_gen_test.py
```diff
@@ -0,0 +1,784 @@
...
+
+
+def test_get_shaped_type_element_bitwidth():
+    assert (
+        candidate_gen.ShapedType([1024, 2048], candidate_gen.ElementType.i8).bitwidth
+        == 8
+    )
+    assert (
+        candidate_gen.ShapedType([2048], candidate_gen.ElementType.i32).bitwidth == 32
+    )
```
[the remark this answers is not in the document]
kuhar: I think this doesn't stop all tests from executing, only the subsequent checks within the current function? IE I though that the difference is similar to `checkXYZ` and `assertXYZ` in GoogleTest, with `assert condition` matching the latter.

If yes, I don't think it's worth switching to unittest -- it's a lot of boilerplate for marginal gains in this context

sharktank/sharktank/tools/tuner/candidate_gen.py
kuhar: No, they can't hit on valid mlir but can triger on invalid mlir. Invalid mlir should fail to parse earlier on and this won't fire there either.

sharktank/tests/tuner/candidate_gen_test.py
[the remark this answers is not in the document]
ScottTodd: Correct - only subsequent checks within the current function will not run.

I don't think it should be a lot of boilerplate, and it's all useful boilerplate too.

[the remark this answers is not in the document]
RattataKing: ok so stay with what it is

sharktank/sharktank/tools/tuner/candidate_gen.py
RattataKing: thanks! A very good point, added it to the TODO list for future

sharktank/sharktank/tools/tuner/candidate_gen.py
RattataKing: leave it unchanged for now

sharktank/sharktank/tools/tuner/candidate_gen.py
```diff
@@ -0,0 +1,1408 @@
...
+    return mlir_module
+
+
+@dataclass
+class TFMLIR:
+    """Transformation of MLIR context"""
+
+    template: str
+    modified: str
+    embeddable: str
```
ScottTodd, line 493: I would use a name other than `TFMLIR` here. I still (regrettably) associate `TF` with `TensorFlow`, not `Transformation` :P

Maybe
* `TransformedMLIR`
* `MlirTransformation`
* `MLIRTransformation`

ScottTodd (approved):
A `README.md` in `sharktank/tools/tuner/` would be helpful too

sharktank/sharktank/tools/tuner/candidate_gen.py
RattataKing: Added `LLVMGPUVectorDistribution` checks for now

kuhar (approved):
LGTM, thanks for all the cleanups

ScottTodd (approved):

kuhar (approved):
The new location looks good to me

PR #144, opened 2024-08-22
[quant] Prevent faketensoring shape values

rsuderman:
All numpy arrays are automatically converted to fake tensors. To support reshape constants we need to ensure that they are cast to integers first.


stellaraccident (approved):
Oh ffs. I'm sorry. That must have really hurt to find. Thank you.

PR #146, opened 2024-08-23
Execute Python tests in CI

marbre:
With the Python tests are executed in the CI via pytest. Tests that are
not capable to run due to missing hardware are not executed whereas
failing test that are expected to pass in the future are currently
marked xfail.

### DIFF
--- a/libshortfin/tests/amdgpu_system_test.py
+++ b/libshortfin/tests/amdgpu_system_test.py
@@ -4,7 +4,10 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+import pytest
 
+
+@pytest.mark.requires_amd_gpu
 def test_create_host_cpu_system():
     from _shortfin import lib as sfl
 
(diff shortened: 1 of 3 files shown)


libshortfin/tests/amdgpu_system_test.py
```diff
@@ -4,7 +4,12 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+import os
+import pytest
 
+GITHUB_ACTIONS = os.getenv("GITHUB_ACTIONS") == "true"
+
+@pytest.mark.skipif(GITHUB_ACTIONS, reason="No AMD GPU in GitHub action.")
```
Groverkss, line 12: I'm not sure if i like this. Can we instead label these tests as "requires-amdgpu" and in github actions filter these tests out?

marbre: I don't know if we can use a label here, but let me rename the environment variable and set it explicitly.

Groverkss (approved):

PR #149, opened 2024-08-23
[libshortfin] Build debug/asan

marbre:
This adds a GitHub action to build a debug/asan version of libshortfin. Shallow cloning of dependencies obtained via FetchContent is reverted as the dependencies cannot be fetched in the CI.

Depends on #148

### DIFF
--- a/.github/workflows/ci_linux_x64_asan-libshortfin.yml
+++ b/.github/workflows/ci_linux_x64_asan-libshortfin.yml
@@ -1,0 +1,160 @@
+# Copyright 2024 Advanced Micro Devices, Inc
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+name: CI - libshortfin - ASan
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+      - main
+    paths:
+      - '.github/workflows/ci_linux_x64_asan-libshortfin.yml'
+      - 'libshortfin/**'
+
+permissions:
+  contents: read
+
+env:
+  PYENV_ROOT: ${{ github.workspace }}/pyenv
+  PYENV_REF: 9ecd803bffaffb949fbdd8c70cb086227f6a3202 # v2.4.10
+  PYTHON_VER: 3.12.3
+  CACHE_ASAN_VER: 1
+  CACHE_DEPS_VER: 1
+  IREE_SOURCE_DIR: ${{ github.workspace }}/iree
+  LIBSHORTFIN_DIR: ${{ github.workspace }}/libshortfin/
+
+
+jobs:
+  setup-python-asan:
+    name: Setup Python ASan
+    runs-on: ubuntu-24.04
+
+    steps:
+    - name: Cache Python ASan
+      id: cache-python-asan
+      uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
+      with:
+        path: ${{ env.PYENV_ROOT }}
+        key: ${{ runner.os }}-python-asan-${{ env.PYENV_REF }}-${{ env.PYTHON_VER }}-v${{ env.CACHE_ASAN_VER }}
+        lookup-only: 'true'
+
+    - name: Install dependencies
+      if: steps.cache-python-asan.outputs.cache-hit != 'true'
+      run: |
+        sudo apt update
+        sudo apt install clang lld cmake ninja-build
+        sudo apt install build-essential libssl-dev zlib1g-dev libbz2-dev libreadline-dev libsqlite3-dev curl git libncursesw5-dev xz-utils tk-dev libxml2-dev libxmlsec1-dev libffi-dev liblzma-dev
+
+    - name: Checkout pyenv
+      if: steps.cache-python-asan.outputs.cache-hit != 'true'
+      uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      with:
+        repository: pyenv/pyenv
+        ref: ${{ env.PYENV_REF }}
+        path: ${{ env.PYENV_ROOT }}
+
+    - name: Install pyenv & Python
+      if: steps.cache-python-asan.outputs.cache-hit != 'true'
+      run: |
+        cd ${{ env.PYENV_ROOT }}
+        src/configure && make -C src
+        export PATH=${{ env.PYENV_ROOT }}/bin:$PATH && eval "$(pyenv init -)"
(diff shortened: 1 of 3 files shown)


.github/workflows/ci_linux_x64_asan-libshortfin.yml
```diff
@@ -0,0 +1,151 @@
...
+        pyenv versions
+        pyenv global ${{ env.PYTHON_VER }}-debug
+        pyenv local ${{ env.PYTHON_VER }}-debug
+        pyenv shell ${{ env.PYTHON_VER }}-debug
+        pip install nanobind
+        pip install -r ${{ env.LIBSHORTFIN_DIR }}/requirements-tests.txt
+
+    - name: Build libshortfin
+      run: |
+        export PYENV_ROOT=${{ env.PYENV_DIR }}
```
stellaraccident, line 121: Could just set this in the job env and use it exclusively (vs also having PYENV_DIR).

.github/workflows/ci_linux_x64_asan-libshortfin.yml
```diff
@@ -0,0 +1,151 @@
...
+        export PATH=${{ env.PYENV_DIR }}/bin:$PATH && eval "$(pyenv init -)"
+        pyenv global ${{ env.PYTHON_VER }}-debug
```
stellaraccident, line 123: I don't think this is necessary to do again as it is changing symlinks and such, not mucking with the environment.

.github/workflows/ci_linux_x64_asan-libshortfin.yml
```diff
@@ -0,0 +1,151 @@
...
+        path: ${{ env.PYENV_DIR }}
+        key: ${{ runner.os }}-python-asan-${{ env.PYENV_REF }}-${{ env.PYTHON_VER }}-${{ env.CACHE_VER }}
+    - name: Install Python packages
+      # TODO: Switch to `pip install -r requirements.txt -e libshortfin/`.
+      run: |
+        export PYENV_ROOT=${{ env.PYENV_DIR }}
+        export PATH=${{ env.PYENV_DIR }}/bin:$PATH && eval "$(pyenv init -)"
+        pyenv versions
+        pyenv global ${{ env.PYTHON_VER }}-debug
+        pyenv local ${{ env.PYTHON_VER }}-debug
```
stellaraccident, line 114: I think you just need the pyenv global, not local/shell.

.github/workflows/ci_linux_x64_asan-libshortfin.yml
```diff
@@ -0,0 +1,151 @@
...
+        pyenv shell ${{ env.PYTHON_VER }}-debug
+        pip install nanobind
```
stellaraccident, line 116: You should add this to requirements.txt and pin it (to 2.0). This can/has broken before and will poison the cache.

(should also add the requirements.txt and requirements-tests.txt hashes to the pyenv hash if installing these packages as part of the pyenv build)

.github/workflows/ci_linux_x64_asan-libshortfin.yml
```diff
@@ -0,0 +1,151 @@
...
+      id: cache-python-asan
+      uses: actions/cache/restore@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
+      with:
+        path: ${{ env.PYENV_DIR }}
+        key: ${{ runner.os }}-python-asan-${{ env.PYENV_REF }}-${{ env.PYTHON_VER }}-${{ env.CACHE_VER }}
+    - name: Install Python packages
+      # TODO: Switch to `pip install -r requirements.txt -e libshortfin/`.
+      run: |
+        export PYENV_ROOT=${{ env.PYENV_DIR }}
+        export PATH=${{ env.PYENV_DIR }}/bin:$PATH && eval "$(pyenv init -)"
```
stellaraccident, line 111: There's a way to have env vars survive to subsequent steps. Would be good to do that vs needing to repeat at every step.

libshortfin/CMakeLists.txt
```diff
@@ -127,6 +123,9 @@ if (NOT SHORTFIN_IREE_SOURCE_DIR AND SHORTFIN_BUNDLE_DEPS)
 endif()
 
 if(SHORTFIN_IREE_SOURCE_DIR)
+  if(SHORTFIN_ENABLE_ASAN)
```
stellaraccident, line 126: Necessary? I would have expected that the -fsanitize=address from the parent project applied to the sub project.

stellaraccident (approved):
Thanks. Left a number of comments but if it passes like this fine to land and iterate.

.github/workflows/ci_linux_x64_asan-libshortfin.yml
marbre: Done.

.github/workflows/ci_linux_x64_asan-libshortfin.yml
marbre: Thanks for the pointer. `pyenv global` is indeed sufficient.

.github/workflows/ci_linux_x64_asan-libshortfin.yml
marbre: I have cut it down to what is strictly needed.

.github/workflows/ci_linux_x64_asan-libshortfin.yml
marbre: I added the the dependency to the `requirements-tests.txt` and pinned the version to `2.0.0`. As it also specified as a build-system dep with line
https://github.com/nod-ai/sharktank/blob/b1d09dbf1496949929c9aaf11f3cd9c551975958/libshortfin/pyproject.toml#L6
`nanobind` version `2.1.0` is installed when executing `pip install -v -e.` later. Should this be pinned to `2.0.0` as well?

Based on the work you've reviewed so far, I further improved caching. Instead of adding hashes for the requirements.txt files to the pyenv/python build step, I added an additional cache on top to cache the dependencies. Assuming that the dependencies are changing more rapidly, this will reduce the build time of the cache. The step to install and cache the dependencies takes about 60s whereas building pyenv/python takes ~ 5m 45s.

.github/workflows/ci_linux_x64_asan-libshortfin.yml
marbre: It cut it down as far as it was possible. Still `eval "$(pyenv init -)"` is needed but setting the path and `PYENV_ROOT` is no longer done in each step.

libshortfin/CMakeLists.txt
marbre: It might be necessary if building the compiler, but it is not for the runtime. Removed it for now.

marbre:
Thanks for the review and the suggestions @stellaraccident. With commit 7e7b77ed60ce6de5d1d6a2384ccc8ae21714ea59, the ASan build fails but it succeeds without  the commit, see for example run https://github.com/marbre/sharktank/actions/runs/10560961944/job/29255953907.

PR #153, opened 2024-08-26
Split running ctest and pytest

marbre:


.github/workflows/ci_linux_x64_asan-libshortfin.yml
```diff
@@ -154,12 +154,17 @@ jobs:
...
-    - name: Test libshortfin
+    - name: Run ctest
       env:
         CTEST_OUTPUT_ON_FAILURE: 1
       run: |
-        eval "$(pyenv init -)"
         cd ${{ env.LIBSHORTFIN_DIR }}/build
         cmake --build . --target test
+      # TOOD: Stop continuing on error after tests got fixed
+      continue-on-error: true
```
ScottTodd, line 164: Another pattern for this is here: https://github.com/nod-ai/sharktank/blob/46b34c9df417c034ff64a598e7968ec420eb2175/.github/workflows/ci.yaml#L60-L68

That's also a bit buggy though - if the setup steps fail, the test steps will try to run.

You can be explicit and give jobs names, then allow running a step based on the conclusion of that step. See https://stackoverflow.com/a/75123818

marbre: Thanks for the pointer, that was what I was looking for! For simplicity using `!cancelled()` for now and improve robustness in a follow up.

ScottTodd (approved):

PR #155, opened 2024-08-26
[llama] Explicit option to use static tables.

stellaraccident:
When exporting, it is better to leave table construction to be dynamic and let the compiler move things to initialization time (versus materializing large tables, which can be max_context_length^2). We set static_tables=False unconditionally on export while leaving it True for eager use.

Contains a workaround for #156.


saienduri (approved):
Thanks

IanNod (approved):

PR #157, opened 2024-08-27
[shortfin] Make TimelineResource hold scopes alive.

stellaraccident:
Prior to this, TimelineResource held a raw C++ reference to the scope but did not hold a shared_ptr ref to it. The storage class, which holds buffers alive, was keeping the raw IREE buffer and device alive but not the shortfin object hierarchy. On the happy path, this does not cause issues, but during abnormal termination and other race conditions, it can cause lifetime problems.

Here we:

* Make TimelineResource hold a std::shared_ptr<Scope>, which transitively keeps everything alive that can be accessed.
* Clean up incorrect ordering of fields in key classes, which could cause dependent objects to be destroyed out of order.
* Remove cases in iree::object_ptr which could have resulted in incorrect accounting in certain scenarios.
* Adds compile-time flags to perform verbose shortfin lifetime logging and application-side IREE reference checks.
* Change System::Shutdown() to not clear object references that still may be live (leave that for the destructor).
* Correct an issue where drivers could be retained forever, which was then masking another lifetime issue during abnormal termination.
* Find a low rate shutdown race triggering use-after-free in the BlockingExecutor and fix it (found by lifetime logging).
* Ensure that all tests are ASAN and LSAN clean, even with a new abnormal termination case added (see updates to include buffer copying, which is currently triggering a buffer permission exception, which set this whole triage session in motion).

### DIFF
--- a/libshortfin/src/shortfin/support/iree_helpers.h
+++ b/libshortfin/src/shortfin/support/iree_helpers.h
@@ -17,6 +17,10 @@
 #include "iree/vm/api.h"
 #include "shortfin/support/api.h"
 
+#if !defined(SHORTFIN_IREE_LOG_RC)
+#define SHORTFIN_IREE_LOG_RC 0
+#endif
+
 namespace shortfin {
 
 // -------------------------------------------------------------------------- //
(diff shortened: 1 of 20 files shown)


libshortfin/src/shortfin/support/iree_helpers.h
```diff
@@ -105,41 +192,60 @@ class object_ptr {
     }
   }
   object_ptr(object_ptr &&other) : ptr(other.ptr) { other.ptr = nullptr; }
+  object_ptr &operator=(const object_ptr &other) = delete;
   object_ptr &operator=(object_ptr &&other) {
+    reset();
```
stellaraccident, line 197: Note that this was causing a reference leak, which just happened to be perfectly offsetting another ownership issue in the happy path.

libshortfin/src/shortfin/support/iree_helpers.h
```diff
@@ -105,41 +192,60 @@ class object_ptr {
...
+  constexpr Assignment for_output() noexcept {
     reset();
-    return &ptr;
+    return Assignment(this);
   }
 
   operator bool() const { return ptr != nullptr; }
   T *get() const { return ptr; }
-  void reset(T *other = nullptr) {
+  void reset() {
```
stellaraccident, line 244: This was a footgun since it allowed an unintended steal. Removed the capability as we have "steal_reference" and "borrow_reference" that do exactly what they say on the label.

PR #158, opened 2024-08-28
[tuner] Add tuner files

RattataKing:
Move remaining tuner files from SDXL repo to sharktank


tuner/examples/punet/README.md
```diff
@@ -0,0 +1,33 @@
...
+## Shell Scripts
+
+The required shell scripts can be downloaded from: [sdxl-scripts](https://github.com/nod-ai/sdxl-scripts)
+
+These scripts include:
+1. `compile-punet-base.sh` - Used for compiling model candidates.
+2. `compile_candidate.sh` - Used for compiling dispatch candidates.
+3. `punet.sh` - Invoked by `compile_candidate.sh`.
+
+Please configure the file paths and update commands in `PunetClient`.
```
kuhar, line 15: Alternatively, we could require these scripts to be in PATH

tuner/examples/punet/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Running the Tuner
+
+### Recommended Trial Run
+For an initial trial to test the tuning loop, use:
+```
+python punet_autotune.py 1286.mlir --num-candidates=1
+```
+### Dry Run Test
```
kuhar, line 24:
```suggestion

### Dry Run Test
```

tuner/generate_tuning_config.py
```diff
@@ -0,0 +1,85 @@
...
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# Given an input tuning spec, generates the config files used by the tuner.
+
+import argparse
+
+
+def main() -> int:
```
kuhar, line 12: We should also drop this file until it's used by libtuner.py

RattataKing: removed

tuner/examples/punet/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+The required shell scripts can be downloaded from: [sdxl-scripts](https://github.com/nod-ai/sdxl-scripts)
+
+These scripts include:
+1. `compile-punet-base.sh` - Used for compiling model candidates.
+2. `compile_candidate.sh` - Used for compiling dispatch candidates.
+3. `punet.sh` - Invoked by `compile_candidate.sh`.
+
+Please configure the file paths and update commands in `PunetClient`.
+**Note:** For convenience, you can alternatively add these scripts to your `PATH` environment variable.
```
kuhar, line 16: We should update the python code to look use commands that assume these scripts will be in PATH, and update the suggestion to modify the paths and commands in `PunetClient`.

tuner/examples/punet/punet_autotune.py
```diff
@@ -69,7 +69,7 @@ def get_model_compile_command(
         command = [
             "timeout",
             "300s",
-            (script_dir / "compile-punet-base.sh").as_posix(),
+            "./compile-punet-base.sh",
             "./tools/iree-compile",
```
kuhar, line 73: Also here -- we can use the iree binaries from path

tuner/examples/punet/punet_autotune.py
```diff
@@ -0,0 +1,181 @@
...
+    def get_dispatch_benchmark_command(
+        self,
+        candidate_tracker: libtuner.CandidateTracker,
+    ) -> list[str]:
+        compiled_vmfb_path = candidate_tracker.compiled_dispatch_path
+        assert compiled_vmfb_path is not None
+
+        command = [
+            "timeout",
+            "16s",
```
ScottTodd, line 49: `16s` seems arbitrary, and the `timeout` command is not portable to Windows (https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/timeout is very different from https://man7.org/linux/man-pages/man1/timeout.1.html)

How about using the `timeout` parameter on https://docs.python.org/3/library/subprocess.html#subprocess.run instead, and then setting this one to 15s or 30s instead of 16s? You'd need to change this function to return something other than just a list of arguments .

tuner/libtuner.py
```diff
@@ -0,0 +1,1302 @@
...
+# Default values for max number of workers
+DEFAULT_MAX_CPU_WORKERS = (
+    multiprocessing.cpu_count() // 2
+)  # the actual amount of worker that will be generated = max(min(max_cpu_workers//2, len(task_list)), 1)
+"""note: Do not use all CPU cores"""
+
+# Declare global variables at the module level for multiprocessing
+worker_id = None
+device_id = None
+"""Do not need to change"""
```
ScottTodd, line 55: These `"""` comments look out of place. Use `#` for comments or remove them?

tuner/libtuner.py
```diff
@@ -0,0 +1,1302 @@
...
+        if not self.result_str:
+            return None
+        pattern = r"process_time/real_time_mean\s+([\d.]+)\s\w{2}"
+        match = re.search(pattern, self.result_str)
+        if not match:
+            return None
+        try:
+            return float(match.group(1))
+        except ValueError:
+            return None
```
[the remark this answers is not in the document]
kuhar: I didn't know about this, thanks @ScottTodd!

tuner/libtuner.py
```diff
@@ -0,0 +1,1302 @@
...
+
+"""
+Provides fundamental functions for tuning:
+    - generate_candidates()
+    - compile_dispatches()
+    - benchmark_dispatches()
+    - compile_models()
+    - benchmark_models()
+
+Requires a wrapper Python script to import `libtuner`, use the `TuningClient` API, customize compilation and benchmarking commands, and implement a complete tuning loop for a specific model.
```
ScottTodd, line 15: nit: format / wrap this at 80 or 100 characters

tuner/libtuner.py
```diff
@@ -0,0 +1,1302 @@
...
+            return None
+        try:
+            return float(match.group(1))
+        except ValueError:
+            return None
+
+
+def generate_display_DBR(
+    candidate_id: int = 0, mean_time: float = random.uniform(100.0, 500.0)
+) -> str:
```
ScottTodd, line 229: Why does `mean_time` have a random default value? That seems really confusing.

tuner/libtuner.py
```diff
@@ -0,0 +1,1302 @@
...
+    """Generate dispatch_benchmark_result string for displaying"""
+    return f"{candidate_id}\tMean Time: {mean_time:.1f}"
+
+
+def generate_display_MBR(
+    candidate_vmfb_path_str: str = "baseline.vmfb",
+    device_id: str = "0",
+    t1: float = random.uniform(100.0, 500.0),
+    calibrated_diff: Optional[float] = None,
+) -> str:
```
ScottTodd, line 239: Same comment about defaulting to a random value.

tuner/libtuner.py
```diff
@@ -0,0 +1,1302 @@
...
+        default=50,
+    )
+    parser.add_argument(
+        "--dry-run",
+        action="store_true",
+        help="Do not attempt to run any modules or initialize the IREE runtime",
+    )
+
+    # candidate_gen.tune() options
+    parser.add_argument(
```
ScottTodd, line 365: You can use argument groups for these: https://docs.python.org/3/library/argparse.html#argument-groups

tuner/libtuner.py
```diff
@@ -0,0 +1,1302 @@
...
+    task_results = [
+        generate_sample_task_result(
+            f"process_time/real_time_mean    {random.uniform(100.0, 500.0):.3g} ms",
+            i,
+            str(0),
+        )
+        for i in compiled_candidates
+    ]
+
+    return task_results
```
ScottTodd, line 905: More wondering about use of `random` here. I'd rather show some obvious placeholder than show truly random data.

tuner/libtuner.py
```diff
@@ -0,0 +1,1302 @@
...
+    append_to_file(
+        dispatch_benchmark_dump_list,
+        filepath=path_config.output_unilog,
+        title="All Dispatch Benchmark Results",
+    )
+
+    benchmarking_rate = (len(parsed_benchmark_results) / len(benchmark_results)) * 100
+    logging.critical(
+        f"Total: {len(benchmark_results)} | Benchmarked: {len(parsed_benchmark_results)} | Failed: {len(benchmark_results) - len(parsed_benchmark_results)} | Benchmarking Rate: {benchmarking_rate:.1f}%"
+    )
```
ScottTodd, line 979: Are you sure you want `critical` ("A serious error, indicating that the program itself may be unable to continue running.") here? That's a step _above_ error: https://docs.python.org/3/library/logging.html#logging-levels

How about `info`? https://docs.python.org/3/library/logging.html#logging.Logger.info

tuner/libtuner.py
```diff
@@ -0,0 +1,1302 @@
...
+
+    # Check if model candidate produces tbe same .vmfb
+    collision_detected, unique_model_candidates_indexes = collision_handler(
+        model_candidates_hash_list
+    )
+
+    if collision_detected:
+        logging.critical(
+            f"Remains [{len(unique_model_candidates_indexes)}] unique candidate indexes"
+        )
```
ScottTodd, line 1061: Similar comment about logging levels. I'd use `warning` or `error` before `critical`.

tuner/examples/punet/punet_autotune.py
```diff
@@ -0,0 +1,181 @@
...
+
+    print("Validating devices")
+    libtuner.validate_devices(args.devices)
+    print("Validation successful!\n")
+
+    print("Generating candidates...")
+    candidates = libtuner.generate_candidates(args, path_config, candidate_trackers)
+    print(f"Stored candidates in {path_config.candidates_dir}\n")
+    if stop_after_phase == libtuner.ExecutionPhases.generate_candidates:
+        return
```
ScottTodd, line 131: Feature request: track the time spent in each phase. At least when running in a console with timestamps (like GitHub Actions), that can be reconstructed.

tuner/examples/punet/punet_autotune.py
```diff
@@ -0,0 +1,181 @@
...
+    if stop_after_phase == libtuner.ExecutionPhases.compile_dispatches:
+        return
+
+    print("Benchmarking compiled candidates...")
+    top_candidates = libtuner.benchmark_dispatches(
+        args, path_config, compiled_candidates, candidate_trackers, punet_client
+    )
+    print(f"Stored results in {path_config.output_unilog}\n")
+    if stop_after_phase == libtuner.ExecutionPhases.benchmark_dispatches:
+        return
```
ScottTodd, line 147: This could be refactored into a MapReduce or other dataflow pattern. Good opportunity to scale out across local or distributed CPUs, GPUs, and machines. Fine to focus on a single local machine at first though.

tuner/examples/punet/punet_autotune.py
kuhar: This is tracked by the progress bars already

tuner/examples/punet/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Running the Tuner
+
+### Recommended Trial Run
+For an initial trial to test the tuning loop, use:
+```
+python punet_autotune.py 1286.mlir --num-candidates=1
+```
```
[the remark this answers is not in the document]
RattataKing: It will be hard to maintain a valid sample mlir here as it could be outdated quickly when iree has new updates. Instead, I add some instruction in README for users to compile and dump the benchmark file in case they need it.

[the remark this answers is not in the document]
ScottTodd: Can you at least post a sample here or on https://gist.github.com/ for the review? If the dispatch IR is unstable, you could instead include a sample input program with something like a single matmul in it and then provide instructions or a script for working with that file.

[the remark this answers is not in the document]
RattataKing: Added more explanations in READ and added a `mmt.mlir`

tuner/examples/punet/punet_autotune.py
RattataKing: implemented by reorg the run_command()

tuner/libtuner.py
RattataKing: thanks!

tuner/libtuner.py
RattataKing: thanks! The original purpose is highlight some important info, and want to log and print those info to the console at the same time. `    # Create a formatter that dynamically adds [levelname] for ERROR and WARNING
    class CustomFormatter(logging.Formatter):
        def format(self, record):
            if record.levelno == logging.INFO:
                return f"{record.message}"
            else:
                return f"[{record.levelname}] {record.message}"`

Now downgrade other `INFO` level to `DEBUG`, and change `CRITICAL` to `INFO`.

tuner/libtuner.py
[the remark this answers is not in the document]
RattataKing: very helpful!

tuner/libtuner.py
RattataKing: Can you show me some examples how you'd want to change?

tuner/examples/punet/punet_autotune.py
RattataKing: why we want to refactor it to the MapReduce pattern? We already have multiprocessing inside of each tuning phase, and separated phases makes more sense to me, because it's good to let all candidates finish and evaluate the overall state (i.e. compilation rate, selecting top winners, etc.), and then decide whether to move to benchmarking.

tuner/libtuner.py
ScottTodd: In a dry run, you're not actually running benchmarks, so why generate any benchmark results at all? If you do for some reason need to generate results, I'd make it obvious that the results are synthetic, like `process_time/real_time_mean    DRY_RUN`.

tuner/examples/punet/punet_autotune.py
ScottTodd: You have a few steps here that are "embarrassingly parallel" (no dependencies between tasks, so they can run independently with no communication needed) and would likely benefit from running at a scale not possible on a single machine.

Multiprocessing still limits you to a single machine. MapReduce is one programming model that lets you scale out across multiple machines. You could, for example,
1. generate 100,000 candidates on a single machine
2. compile 1000 candidates each on 100 CPU machines
3. benchmark 1000 candidates each on 100 GPUs

During an optimization effort, it sounded like we were seeing different tuning results across different GPUs. We could either accept that tuning will be GPU-specific or try to instead find generally beneficial tunings across a whole fleet of devices. I would expect this sort of "search the problem space for promising solutions" tool to be possible to run both locally on a single machine and distributed.

tuner/libtuner.py
RattataKing: The existing benchmarking functions still rely on using regex parsing method, extracts numbers from result_str, and does some order sorting and reconstruct output string for display purpose. The strings generated in this function are not the final form, these are fake results that pretend to be the captured console text, and these results will be processed again later in other functions (e.g. result normalization, calibrate the data and calculate the perf improvement percent)

tuner/examples/punet/punet_autotune.py
kuhar: I'm not sure this has clear mapping to map-reduce. Specifically, the reduce step has to reconcile the differences between physical gpus. I'd say that we should consider it in the future, but this is outside of the scope of what the tuner can do today. @RattataKing could you open a tracking issue for this?

tuner/examples/punet/mmt.mlir
```diff
@@ -0,0 +1,76 @@
...
+// RUN:   %s -o %s.vmfb
+
+// To compile to for benchmarking, add:
+//  --iree-flow-export-benchmark-funcs --iree-hal-benchmark-dispatch-repeat-count=1000
+//
+// To benchmark:
+//   for i in {0..4} ; do
+//     iree-benchmark-module --device=rocm://7 --module=%s.vmfb --function="main_${i}_benchmark" --device_allocator=caching \
+//       --batch_size=1000 --benchmark_repetitions=5
+//   done
```
ScottTodd, line 16: Seems weird to pin to device 7. Any particular reason for that? (examples should be _minimal_, so just `--device=hip`)

kuhar, line 16: This readme is out of date, we should drop it

tuner/examples/punet/mmt.mlir
```diff
@@ -0,0 +1,76 @@
+// RUN: iree-compile --iree-hal-target-backends=rocm --iree-rocm-target-chip=gfx942 \
+// RUN:   --iree-rocm-link-bc=true --iree-rocm-bc-dir=/opt/rocm/amdgcn/bitcode \
+// RUN:   --iree-global-opt-propagate-transposes=true --iree-opt-outer-dim-concat=true \
+// RUN:   --iree-opt-const-eval=false --iree-codegen-gpu-native-math-precision=true --iree-rocm-waves-per-eu=2 \
+// RUN:   --iree-preprocessing-pass-pipeline='builtin.module(iree-preprocessing-transpose-convolution-pipeline)' \
+// RUN:   --iree-codegen-llvmgpu-use-vector-distribution --iree-codegen-transform-dialect-library=config.mlir \
+// RUN:   %s -o %s.vmfb
```
ScottTodd, line 7: I'd prefer for examples to be minimal. Could leave off all the optional flags here and just use `iree-compile  --iree-hal-target-backends=rocm --iree-rocm-target-chip=gfx942`

tuner/examples/punet/2.mlir
```diff
@@ -0,0 +1,80 @@
+module attributes {hal.device.targets = [#hal.device.target<"rocm", {legacy_sync}, [#hal.executable.target<"rocm", "rocm-hsaco-fb", {iree.gpu.target = #iree_gpu.target<arch = "gfx942", features = "", wgp = <compute =  fp64|fp32|fp16|int64|int32|int16|int8, storage =  b64|b32|b16|b8, subgroup =  shuffle|arithmetic, dot =  dp4xi8toi32, mma = [<MFMA_F16_16x16x16_F32>, <MFMA_F16_32x32x8_F32>, <MFMA_F8E4M3FNUZ_16x16x32_F32>, <MFMA_I8_16x16x32_I32>, <MFMA_I8_32x32x16_I32>], subgroup_size_choices = [64], max_workgroup_sizes = [1024, 1024, 1024], max_thread_count_per_workgroup = 1024, max_workgroup_memory_bytes = 65536>>, ukernels = "none"}>]>]} {
```
kuhar, line 1: Instead of checking this in, can we add the rough compilation command to the readme? This can be generated by compiling `mmt.mlir` with `--iree-hal-dump-executable-benchmarks-to=dump`

kuhar (commented):
Looks good modulo the sample mmt files.

@RattataKing could you open a tracking issue for switching to json-based benchmark processing and for evaluating mapreduce?

tuner/examples/punet/punet_autotune.py
ScottTodd: Some inspiration:
* https://pytorch.org/tutorials/beginner/hyperparameter_tuning_tutorial.html
* https://keras.io/guides/keras_tuner/distributed_tuning/

tuner/examples/punet/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Running the Tuner
+
+### [Optional] Generate a tunable mlir
+A sample `2.mlir` is provided for test run. Hoever, this file may become outdated if IREE makes changes to the MLIR format. To ensure you are working with the latest format, please follow the instructions below to compile and generate the most recent benchmark file.
+Use [`punet.sh`](https://github.com/nod-ai/sdxl-scripts/blob/main/tuning/punet.sh) to compile the sample matmul `mmt.mlir` (can also find here: [`mmt_unet.mlir`](https://github.com/nod-ai/sdxl-scripts/blob/main/tuning/mmt_unet.mlir)):
+```
+./punet.sh ./mmt.mlir -o baseline.vmfb --iree-hal-dump-executable-files-to=dump-mmt
+cp ./dump-mmt/module_main_2_dispatch_0_rocm_hsaco_fb_benchmark.mlir ./2.mlir
+```
```
ScottTodd, line 26: Yeah that `2.mlir` is going to get outdated very quickly :P Might be still useful to keep as something to point people at to show them what sort of input the tuner operates on.

I'd be careful overspecializing on external scripts that set dozens of non-default flags, especially if the goal is to turn the tuner into a general purpose tool that supports multiple hardware targets.

tuner/examples/punet/punet_autotune.py
ScottTodd: I'm confused - I still see `timeout` as part of the command here. Fine to add a comment to make this portable in the future.

tuner/libtuner.py
ScottTodd: Ah, so more issues stemming from passing strings around instead of structured data. A tale as old as time 😅 

Okay for now but it would help to refactor such that fake results are not needed.

ScottTodd (approved):
Nice! Thanks for the edits!

PR #161, opened 2024-09-03
Add test that shards, exports and runs with IREE a Conv2DLayer

sogartar:
In general we should avoid having many end-to-end tests, but I decided to add it since it was the isolation of a real problem
https://github.com/iree-org/iree/issues/18283
and I already had the test.

We would like to have at least several E2E tests of tiny models that run on every PR.


stellaraccident (approved):
Thanks for asking a test to guard the gate. Agreed we don't want much of these but some are good. 

I think this will require some additional cleanup but let's have it first. Can you create an issue to simplify sharktank e2e tests and config options and tag this commit to it before landing? I'd like to get Scott to review a few of these that have surfaced in the repo together.

sogartar:
I resolved the conflicts before merging.

PR #162, opened 2024-09-03
moe

dan-garvey:
Throwing this up as a pr to make it easier to view


sharktank/sharktank/layers/ffn_moe_block.py
```diff
@@ -0,0 +1,84 @@
...
+        ffn_down = self.ffn_down(ffn_gate * ffn_up)
+        return ffn_down
+
+
+def extract_ffn_layer(
+    merged_tensor: DefaultPrimitiveTensor, layer_name: str, expert_idx: int
+):
+
+    expert_layer_name = (
+        f"blk.{merged_tensor.name.split('.')[1]}.{layer_name}.{expert_idx}.weight"
```
dan-garvey, line 79: Would be good to add a comment explaining what an example looks like to understand why we're taking the 1 index of merged tensor

sharktank/sharktank/layers/llama_attention_block.py
dan-garvey, line 1: If we're moving this delete the other one and refactor

archana-ramalingam: I was thinking we could have a separate PR to refactor the paged, direct attention and the FFN blocks into llama, as it touches the llama codebase.

dan-garvey: No, lets just do it now, llama should be tested not write-protected.

sharktank/sharktank/layers/ffn_block.py
```diff
@@ -0,0 +1,45 @@
...
+    def __init__(
+        self,
+        theta: Theta,
+        expert_idx: Optional[int] = None,
+    ):
+        super().__init__(theta)
+        ffn_g = "ffn_gate"
+        ffn_u = "ffn_up"
+        ffn_d = "ffn_down"
+        if expert_idx is not None:
```
archana-ramalingam, line 30: We can remove the MOE aspect from this FFN, as we will only be using `ffn_moe_block.py` for MOE models. Ideally we can combine both, but for now `ffn_moe_block.py` has all the logic to handle an MOE model.

sharktank/sharktank/layers/paged_llama_attention_block.py
```diff
@@ -0,0 +1,265 @@
...
+    ):
+        super().__init__(theta)
+        self.add_module(
+            "attn_norm", RMSNormLayer(theta("attn_norm"), epsilon=rms_epsilon)
+        )
+        self.add_module("attn_q", LinearLayer(theta("attn_q")))
+        self.add_module("attn_k", LinearLayer(theta("attn_k")))
+        self.add_module("attn_v", LinearLayer(theta("attn_v")))
+        self.add_module("attn_output", LinearLayer(theta("attn_output")))
+        self.add_module(
```
archana-ramalingam, line 49: `attn_output_norm` is a Grok specific layer, will break llama models.

sharktank/sharktank/layers/rotary_embedding.py
```diff
@@ -18,6 +18,7 @@ def __init__(
         self,
         *,
         rope_dimension_count: int,
+        rope_freq_base: float,
```
archana-ramalingam, line 21: This parameter isn't present in llama2 gguf, but is explicitly mentioned for Mixtral and Grok.

sharktank/sharktank/models/llama/llama.py
```diff
@@ -137,6 +137,7 @@ def __init__(self, theta: Theta, config: LlamaModelConfig):
             "attention_embedding",
             RotaryEmbeddingLayer(
                 rope_dimension_count=hp.rope_dimension_count,
+                rope_freq_base=hp.rope_freq_base,
```
archana-ramalingam, line 140: Remove all instances of `hp.rope_freq_base`, if it's not added as an optional param.

sharktank/sharktank/examples/paged_llm_v1.py
```diff
@@ -236,12 +237,12 @@ def main():
         activation_dtype=activation_dtype,
         attention_dtype=activation_dtype,
     )
-    model = PagedLlamaModelV1(dataset.root_theta, config)
-    if args.save_intermediates_path:
```
archana-ramalingam, line 240: We would want to retain this `if ` block, it's present in the main branch.

sharktank/sharktank/layers/mixture_of_experts_block.py
```diff
@@ -0,0 +1,115 @@
...
+
+        # Create an expert mask by one hot encoding the selected top k experts
+        # used to index which expert is to be invoked for each token
+        # expert_mask: (expert_count, expert_used_count, sequence_length)
+        expert_mask = F.one_hot(top_k_experts, num_classes=self.expert_count).permute(
+            2, 1, 0
+        )
+
+        # Iterate over all experts in the model
+        for expert_idx in range(self.expert_count):
```
archana-ramalingam, line 94: Are we replacing this with slow_matmul for Grok?

sharktank/sharktank/layers/ffn_block.py
dan-garvey: I think its ok to keep separate, but I'll remove the expert index logic from this one

sharktank/sharktank/layers/paged_llama_attention_block.py
dan-garvey: And this is why it fails the test!

dan-garvey: I think we can just add it as its own layer in the new grok model, and I'll just remove it here

sharktank/sharktank/layers/mixture_of_experts_block.py
dan-garvey: I'll add this in a follow up patch that replaces the expert splitting in the experts too

sharktank/sharktank/examples/paged_llm_v1.py
dan-garvey: great catch, thank you

sharktank/sharktank/layers/ffn_moe_block.py
```diff
@@ -0,0 +1,80 @@
...
+        self,
+        h: torch.Tensor,
+    ):
+        ffn_gate = F.silu(self.ffn_gate(h))
+        ffn_up = self.ffn_up(h)
+        ffn_down = self.ffn_down(ffn_gate * ffn_up)
+        return ffn_down
+
+
+def extract_ffn_layer(
```
archana-ramalingam, line 74: Please adopt this function from [here](https://github.com/archana-ramalingam/sharktank/blob/mixtral/sharktank/sharktank/layers/ffn_moe_block.py#L74). This is necessary to split expert tensors by their names for MOE gguf models.

dan-garvey: I don't think the name of the primitiveTensor matters, do you?

archana-ramalingam: It matters when debugging/tracing theta, to know which ffn layer/expert and block this belongs to, as this is fed to theta in the FFNMOE class above. Like I said we are creating new layer names for each of 8 ffn expert layers from merged ffn expert layers.

archana-ramalingam (approved):

PR #166, opened 2024-09-04
Pin IREE to specific version

marbre:
Pins IREE checkout in the CI or obtained via FetchContent to the same specified version. Furthermore, installs the iree-compiler in the CI.

### DIFF
--- a/.github/workflows/ci_linux_x64-libshortfin.yml
+++ b/.github/workflows/ci_linux_x64-libshortfin.yml
@@ -46,6 +46,7 @@ jobs:
         repository: iree-org/iree
         path: ${{ env.IREE_REPO_DIR }}
         submodules: false
+        ref: candidate-20240904.1006
 
     - name: Initalize IREE submodules
       run : |
--- a/.github/workflows/ci_linux_x64_asan-libshortfin.yml
+++ b/.github/workflows/ci_linux_x64_asan-libshortfin.yml
@@ -96,6 +96,7 @@ jobs:
         repository: iree-org/iree
         path: ${{ env.IREE_SOURCE_DIR }}
         submodules: false
+        ref: candidate-20240904.1006
 
     - name: Initalize IREE submodules
       run : |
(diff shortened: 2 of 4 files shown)


.github/workflows/ci_linux_x64-libshortfin.yml
```diff
@@ -46,6 +47,7 @@ jobs:
         repository: iree-org/iree
         path: ${{ env.IREE_REPO_DIR }}
         submodules: false
+        ref: ${{ env.IREE_REF }}
```
ScottTodd, line 50: Any particular reason to store this in an environment variable instead of inlining it?

```suggestion
        ref: candidate-20240904.1006
```

Unless using it in multiple places, I'd prefer to save the one line of code :D

.github/workflows/ci_linux_x64_asan-libshortfin.yml
```diff
@@ -96,6 +97,7 @@ jobs:
         repository: iree-org/iree
         path: ${{ env.IREE_SOURCE_DIR }}
         submodules: false
+        ref: ${{ env.IREE_REF }}
```
ScottTodd, line 100: Same question here

ScottTodd (approved):

.github/workflows/ci_linux_x64-libshortfin.yml
marbre: The intention was to make it more convenient if this (as well as other vars) to have them grouped and near to the top if those need to be updated. But no strong feeling.

ScottTodd: Gotcha. I see your point, but I also like to having some of that local to the `actions/checkout` step. Not a big deal with way.

marbre: I removed the environment variable and moved it to the action :)

ScottTodd (approved):

PR #167, opened 2024-09-04
Alias gguf tensors instead of copy

rsuderman:
Using `torch.as_tensor` we can alias the tensor rather than copy during gguf file loading. This avoids duplicating the entire tensor contents when tracing torch programs which substrantially decreases memory usage on large models.

e.g. LLaMa 70b decreased memory allocation from 60+GB to 2 GB for tensors.


stellaraccident (approved):
Good find.

PR #168, opened 2024-09-05
Disable strictness for export of llama

rsuderman:
Strictness validates correctness but this results in loading the tensors to memory. Disabling helps with export speed.


stellaraccident (approved):
I feel like there's an issue to file here on the PyTorch side. It isn't clear to me that the overhead ballooning is intended on their side, and if you can describe the issue/have a simplified reproducer, it might help them.

PR #169, opened 2024-09-05
initial grok

dan-garvey:
Initial grok work, also does some refactoring


archana-ramalingam:
I have more fixes on my branch that I am testing on Grok/llama to push.

sharktank/sharktank/layers/configs/llm_configs.py
```diff
@@ -162,3 +113,62 @@ def _optional_int_prop(p: dict[str, Any], name: str, default_value: int) -> int:
         return int(value)
     except ValueError as e:
         raise ValueError(f"Property '{name}' expected to be an int and was not") from e
+
+
```
archana-ramalingam, line 126: Adding  LlamaModelConfig to config file make sense, but having create_kv_cache() function in a config file feels odd. Is there a better way to refactor this?

sharktank/sharktank/layers/configs/llm_configs.py
```diff
@@ -41,37 +42,51 @@ class LlamaHParams:
...
     expert_used_count: int
+    model_variant: str
 
     @staticmethod
     def from_gguf_props(p: dict[str, Any]):
+        name_prefix = "llama"
+        # These shouldnt be required for llama.
+        default_attn_dim = None
+        default_rope_count = None
+        if "grok.attention.head_count" in p:
```
KyleHerndon, line 53: Let's branch on the gguf value `general.architecture == "grok"` so that we can support more single headed models than just llama and more multiheaded models than just grok.

archana-ramalingam: I have this already in my stream of commits. Will push in the next 30 mins.

dan-garvey:
I think the failing test is flaky, I'll TAL

sharktank/sharktank/layers/configs/llm_configs.py
archana-ramalingam: I moved `create_kv_cache` function to utils folder, as that seemed to be the best place for it.

archana-ramalingam:
Posted a comment [here](https://github.com/nod-ai/SHARK-Platform/commit/4816c9349c8862bdb73bf075a3baac577e9df21a#r147163987), doesn't show up on the main page.

dan-garvey:
> Posted a comment [here](https://github.com/nod-ai/SHARK-Platform/commit/4816c9349c8862bdb73bf075a3baac577e9df21a#r147163987), doesn't show up on the main page.

lets chat at meeting

sharktank/sharktank/utils/create_cache.py
dan-garvey, line 1: I did say I didn't mind if we did this, but looking at it now, all the args are just values from the config anyway. I feel like separating them I think just makes our code harder to follow. 

Not a big deal either way, but I'd personally prefer we just drop this commit. @KyleHerndon what do you think?

KyleHerndon: I think this is the right way to set things up. The bad argument is that llama.cpp has a similar setup, and we're referencing them for accuracy baselines. The better argument goes something like:

The KV cache is not ontologically related to the config, it just (currently) exclusively uses parameters from it. In the near future, we will want things like sharding, at which point the KV Cache might need additional args (number of devices, for example, which is not a ModelConfig parameter, but something like an ExecutionConfig parameter).

This change doesn't feel related to grok as much a code refactor, which I think is fine to include in a bigger patch for efficiency but might make things harder to follow when looking at commit history. Otherwise, I find the code just as easy to follow, if not slightly easier because I think it is better organized.

archana-ramalingam: I don't mind either way, but do agree with Kyle that this keeps it better organized. We can consult Rob or Stella to see if there is a better way to do this.

dan-garvey: not necessary, if you two agree thats more than enough for me

dan-garvey:
looks good to me, @KyleHerndon @archana-ramalingam any final changes you think are needed?

sharktank/sharktank/examples/paged_llm_v1.py
```diff
@@ -233,7 +233,6 @@ def main():
 
     device = torch.device(args.device) if args.device else None
     activation_dtype = getattr(torch, args.activation_dtype)
-    attention_dtype = getattr(torch, args.attention_dtype)
```
dan-garvey, line 240: nice catch

KyleHerndon (approved):

archana-ramalingam (approved):
I tested Llama and Grok models again and it's LGTM.

sogartar:
@dan-garvey, the [error](https://github.com/nod-ai/SHARK-Platform/actions/runs/11040261276/job/30667825889?pr=169#step:6:334).
Seems to be related to calling `__mul__` on the `DefaultPrimitiveTensor`.
This fails:
```python
output = weight * to(output, weight.dtype)
```

but this is fine
```python
output = elementwise(torch.mul, weight, to(output, weight.dtype))
```

`__mul__`'s implementation is just
```python
    def __mul__(self, rhs):
        from ..ops import elementwise

        return elementwise(torch.mul, self, rhs)
```

sogartar:
When calling `torch.export.export` with `strict=False`, the error is not present, so it is about the export checks.

sogartar:
Instead of calling the `*` operator I changed the code to call a custom member function in `weight`

```python
output = weight.my_mul(to(output, weight.dtype))
```

```python
class InferenceTensor(ABC):

....

    def my_mul(self, rhs):
        from ..ops import elementwise

        return elementwise(torch.mul, self, rhs)
```

This is essentially the same code, but PyTorch does something special about binary operators because '+' also suffers from the same problem.

Also when running with env vars
```
TORCH_LOGS="+dynamo"
TORCHDYNAMO_VERBOSE=1
```

The tracer reports
```
V0926 06:19:31.456000 140060607524864 torch/_dynamo/symbolic_convert.py:798] [0/0] [__trace_bytecode] TRACE LOAD_FAST results []
V0926 06:19:31.456000 140060607524864 torch/_dynamo/symbolic_convert.py:798] [0/0] [__trace_bytecode] TRACE LOAD_CONST 0 [TupleVariable()]
V0926 06:19:31.456000 140060607524864 torch/_dynamo/symbolic_convert.py:798] [0/0] [__trace_bytecode] TRACE BINARY_SUBSCR None [TupleVariable(), ConstantVariable()]
V0926 06:19:31.456000 140060607524864 torch/_dynamo/symbolic_convert.py:798] [0/0] [__trace_bytecode] TRACE RETURN_VALUE None [TensorVariable()]
V0926 06:19:31.456000 140060607524864 torch/_dynamo/symbolic_convert.py:2807] [0/0] DONE INLINING <code object __call__ at 0x7f618b39f470, file "/home/bpetkant/ws/sharktank/repo/sharktank/sharktank/ops/_registry.py", line 196>
V0926 06:19:31.456000 140060607524864 torch/_dynamo/symbolic_convert.py:798] [0/0] [__trace_bytecode] TRACE BINARY_MULTIPLY None [UserDefinedObjectVariable(), TensorVariable()]
V0926 06:19:31.456000 140060607524864 torch/_dynamo/symbolic_convert.py:814] [0/0] empty checkpoint
V0926 06:19:31.456000 140060607524864 torch/_dynamo/symbolic_convert.py:2796] [0/0] FAILED INLINING <code object rms_norm_default at 0x7f618b2126b0, file "/home/bpetkant/ws/sharktank/repo/sharktank/sharktank/ops/default_impls.py", line 308>
```

It is doing something special about `BINARY_MULTIPLY`.

sogartar:
I opened an [issue](https://github.com/pytorch/pytorch/issues/136757) with PyTorch. I think that may be a bug there.

PR #171, opened 2024-09-06
Place args for all gather/reduce on devices before the op to avoid CSE and excessive copying

sogartar:
We are encoding the device/shard information in the flow.tensor.transfer/transfer_to_logical_device
operation. Then if we do an all-gather or an all-reduce, CSE is happy to collapse the expressions into one. This would result in the all-gather/all-reduce being performed on one device and then the result is copied to the rest. We want each device to do the all-gather/all-reduce.

There is no easy way to test the desired effect, but at least we test for correctness on the PyTorch level.

This change adds the all_reduce op that is currently not used anywhere.
Here is expanded the elementwise op to support a variable number of tensor arguments.

### DIFF
--- a/sharktank/sharktank/ops/_registry.py
+++ b/sharktank/sharktank/ops/_registry.py
@@ -114,7 +114,8 @@ class AllOfExprsVariadic(BoolTypeExpr):
                 return False
             exprs = self._exprs
             if len(types) > len(exprs):
-                exprs = exprs + [exprs[-1]] * (len(types) - len(self._exprs))
+                # pad with the trailing expression.
+                exprs = exprs + ([exprs[-1]] * (len(types) - len(self._exprs)))
             return all([e(t) for e, t in zip(exprs, types)])
 
         super().__init__(expr)
--- a/sharktank/sharktank/ops/_registry.py
+++ b/sharktank/sharktank/ops/_registry.py
@@ -18,6 +18,7 @@ from ..types import PrimitiveTensor, QuantizedTensor
 
 __all__ = [
     "AllOfExprs",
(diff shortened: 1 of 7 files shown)


sharktank/sharktank/ops/_registry.py
```diff
@@ -87,6 +89,37 @@ def expr(*types: type):
...
+        if len(exprs) == 0:
+            raise ValueError("At least one expression is required.")
+        self._exprs = list(exprs)
+
+        def expr(*types: type):
+            if len(types) < len(self._exprs):
+                return False
+            exprs = self._exprs
+            if len(types) > len(exprs):
+                exprs = exprs + [exprs[-1]] * (len(types) - len(self._exprs))
```
stellaraccident, line 117: Extra parens arrive the list multiplication expression? Reads a little funny to me like this.

sharktank/sharktank/ops/default_impls.py
```diff
@@ -80,6 +81,20 @@ def elementwise_binary(operator, x, y):
     return operator(x, y)
 
 
+@elementwise.override(
```
stellaraccident, line 84: Can you ask a comment with some examples of what this overload matches?

stellaraccident (approved):
Thanks

sharktank/sharktank/ops/_registry.py
sogartar: I added extra parenthesis.

sharktank/sharktank/ops/default_impls.py
sogartar: I added documentation and reworked the function to not use recursion.

PR #175, opened 2024-09-09
[sharktank] Export Attention IRs for LLMs

archana-ramalingam:
Export Attention IRs for paged causal and non causal LLMs.


sogartar:
We can't use directly torch ops with our tensor types.
https://github.com/nod-ai/SHARK-Platform/actions/runs/11148274191/job/30984526213?pr=175#step:6:123
In this example `torch.concat` would need to be substituted with [ `sharktank.ops.cat`](https://github.com/nod-ai/SHARK-Platform/blob/b0c2399cc8256fb916462dbe04e69883eb25ef34/sharktank/sharktank/ops/signatures.py#L102).
At some instances we may be lacking the equivalent op or it may have no implementation for certain tensor types, like `ReplicatedTensor`. In this case we have to add it. For some tensor types this may be difficult, but in most cases it is easy. I have been meaning to make a general implementation when all tensors are replicated across devices, but have not gotten to it. I will add the `cat` overload for the replicated tensor.

sharktank/sharktank/layers/kv_cache.py
```diff
@@ -363,21 +363,34 @@ def write_timestep(
...
+        transformer_block = torch.full(
+            (bs, partition_count), transformer_block_index, device=device
+        )
+        page_offset = page_offset.repeat(1, partition_count)
+        partitions = partitions.repeat(bs, 1)
+
+        indices = (page_id, transformer_block, partitions, page_offset)
+        page_table.index_put_(indices=indices, values=cache_partitions)
+
+        return
```
sogartar, line 393: Is this supposed to be a functionally equivalent transformation that removes the loops?

archana-ramalingam: Yes, that is used for decode and has been tested with a llama model. And L434 -449 for prefill.

sogartar: I found the issue in the test arguments, but I need to add sharded variants for a few more new ops like `torch.gather`.

sharktank/sharktank/export_layer/export_paged_attention.py
```diff
@@ -0,0 +1,415 @@
...
+from sharktank.types import *
+
+from sharktank.models.llama.testing import *
+from sharktank.layers import causal_llm
+
+# TODO: Should be using a base class with the protocol supported.
+from ..models.llama.llama import LlamaModelConfig, PagedLlamaAttentionBlock
+
+
+def paged_attention(
```
sogartar, line 28: Why is this function together with `run_llama` can not share the code with [paged_llama_attention_block.py](https://github.com/nod-ai/SHARK-Platform/blob/9f3f70f7f5018702abfa0dbb4954630179abb768/sharktank/sharktank/layers/paged_llama_attention_block.py)? They seem to be doing mostly the same things.

sharktank/sharktank/layers/kv_cache.py
```diff
@@ -363,21 +363,34 @@ def write_timestep(
...
-                    page_id,
-                    torch.tensor([transformer_block_index], device=device),
-                    torch.tensor([partition_index], device=device),
-                    page_offset.unsqueeze(0),
-                )
-                page_table.index_put_(indices=indices, values=cache_partition[i, 0])
+
+        partition_count = len(cache_partitions)
+
+        cache_partitions_list = cache_partitions
```
sogartar, line 369: This looks unused.

sharktank/sharktank/export_layer/export_paged_attention.py
archana-ramalingam: This script is exporting a boiled down version of only the `sdpa` part of the paged attention found here -[paged_llama_attention_block.py](https://github.com/nod-ai/SHARK-Platform/blob/9f3f70f7f5018702abfa0dbb4954630179abb768/sharktank/sharktank/layers/paged_llama_attention_block.py).

sharktank/sharktank/layers/kv_cache.py
archana-ramalingam: Good catch, removed.

sharktank/sharktank/export_layer/export_paged_attention.py
sogartar: OK, so instead of doing flash attention, it is calling `torch.nn.functional.scaled_dot_product_attention`. Probably, in another PR we can have `sharktank.layers.PagedLlamaAttentionBlock` be configurable to accommodate for that scenario. This will avoid the code duplication.

sogartar (approved):

sharktank/tests/layers/sharded_paged_kv_cache_test.py
```diff
@@ -151,7 +151,7 @@ def testWriteTimestep(self):
         cache_partitions = [
             torch.rand(
                 self.batch_size,
-                self.block_seq_len * self.block_seq_stride,
+                1,
```
rsuderman, line 154: Any reason we decreased this to one?

rsuderman (approved):

sharktank/tests/layers/sharded_paged_kv_cache_test.py
archana-ramalingam: This was @sogartar changes to fix the shard tests issues after scatter fusion.

sogartar: It was wrong. The expectation is that this dimension should be always size 1. That is at least the comment at the [function defintion](https://github.com/nod-ai/SHARK-Platform/blob/8e074b45362737787e5ad1d467f0286976490e4f/sharktank/sharktank/layers/kv_cache.py#L348).

PR #177, opened 2024-09-10
[libshortfin] Python build ergonomics.

stellaraccident:
* Move bindings/python -> python
* Rework setup.py so that it can be the main driver for Python based dev:
  * Add env vars SHORTFIN_IREE_SOURCE_DIR, SHORTFIN_ENABLE_ASAN, SHORTFIN_DEV_MODE, SHORTFIN_RUN_CTESTS
  * Update README with python based dev flow commands
  * Change build dir to build/cmake (from build/b)
* Removes a redundant build of IREE in the main CI (was not doing anything in bundled mode).

With this, I can change my dev process to be entirely based on `pip install` with a fully editable build and the ability to manipulate cmake post build. This better aligns with how users will consume it.


stellaraccident:
@marbre for post review

PR #181, opened 2024-09-11
[libshortfin] Initial implementation of LLM inference server.

stellaraccident:
This is very much a first draft that needs a fair bit of work to turn into a final form. As the first real user of `libshortfin`'s core APIs, a number of rough spots were worked out even to get it to this first stage.

Just a few things that need attention:

* Sampling is just a greedy decode loop right now, but this can be easily extended because it is done per item and is just linear Python (async).
* Scaffolding for multi device in a single process is in place but we're still working on the models to mate up with it.
* A lot of configuration is hardcoded.
* The batcher is a toy. The aim is to implement something like sglang's cache aware batcher and radix attention in its place. Given that, no effort was taken to make the batcher more than write-once code.
* There are many opportunities to use different shortfin fibers, workers and executors to better balance the host-side activity. This currently just runs async on a single worker for the moment.


libshortfin/tests/apps/llm/components/tokenizer_test.py
```diff
@@ -0,0 +1,70 @@
...
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import pytest
+
+
+@pytest.fixture
+def bert_tokenizer():
+    import shortfin_apps.llm.components.tokenizer as tokenizer
```
monorimet, line 12: I love that we have this :)

libshortfin/python/shortfin_apps/llm/components/cache.py
rsuderman, line 1: Genuinely happy to see the page cache / management. This was something I had wanted to add to the previous implementation.

monorimet (approved):
I don't have any deep insights on this yet, I've left a few nit comments for typos but you already have noted the areas needing most improvement.

I see some of this could be abstracted out and reused for the next inference server implementation, but some of it, like the batcher, might not have a very substantial base class to carve out (and we might want to resolve some of your bugaboos about current implementation before doing so regardless.

Anyway, this seems to cover all the points I have in my head, and there's lots for us to iterate on. Thanks.

libshortfin/python/shortfin_apps/llm/components/generate.py
```diff
@@ -0,0 +1,176 @@
...
+            await exec.done
+
+            # Prefill result.
+            token = sfnp.argmax(exec.result_logits)
+            token_int = token.items[0]
+
+            # Decode loop.
+            # TODO: Use correct eot token from config.
+            exec.start_position = len(self.input_token_ids) - 1
+            while token_int != 128001:
```
rsuderman, line 60: Just tagging to remind in case you missed the TODO

libshortfin/tests/apps/llm/components/tokenizer_test.py
stellaraccident: xtensor has random number generators. just need to expose them to Python probably.

stellaraccident:
> Overall this looks good. I noticed a few TODOs which I assume are still WIP but the structure felt right to me. Only high level thought was having some utility / abstraction for data transfer. The direct invocations of the IREE h2d or d2h commands felt slightly out of place (but thats just a nit).

Yeah, I swallowed my own objections while typing those transfers. To do them right needs more system configuration, which I don't have yet (the way you stage transfers and allocation is system and use case specific), so I just did the least bad thing and wrote it out long hand for now. That's why today, I'm building out the system config layer more -- that is where you root fixes for things like this.

libshortfin/python/shortfin_apps/llm/components/generate.py
stellaraccident: Will come back through. There are a lot of things that need to come from config and are not right now.

PR #182, opened 2024-09-11
[libshortfin] Fix build for GCC

Groverkss:
This patch changes two things:

- Removes API attributes from forward declerations. This generates warnings in GCC and is generally not recommended: https://stackoverflow.com/questions/44591568/when-forward-declaring-classes-should-i-use-attribute-defines-there-too

- Change int -> size_t to not assume signed -> unsigned conversion.


stellaraccident (approved):
Thanks. I had wondered what the right thing there was and flipped a coin.

libshortfin/src/shortfin/local/scope.cc
```diff
@@ -72,7 +72,7 @@ Device *Scope::raw_device(std::string_view name) const {
   return it->second;
 }
 
-Device *Scope::raw_device(int ordinal) const {
+Device *Scope::raw_device(std::size_t ordinal) const {
   if (ordinal < 0 || ordinal >= devices_.size()) {
```
marbre, line 76:
```suggestion
  if (ordinal >= devices_.size()) {
```

PR #183, opened 2024-09-11
[libshortfin] Fix CMake bundled iree build

Groverkss:
https://github.com/nod-ai/sharktank/pull/177 broke building iree as a bundled dep. This is due to differences between `FetchContent_MakeAvailable` and `FetchContent_Populate`, where the former does `add_subdirectory` on the downloaded target while the latter doesnt. This patch sets IREE flags before the call to `FetchContent_MakeAvailable`, so we have the right flags enabled for `add_subdirectory`.

### DIFF
--- a/libshortfin/CMakeLists.txt
+++ b/libshortfin/CMakeLists.txt
@@ -126,18 +126,20 @@ endif()
 
 ## iree runtime
 
-# Set IREE build flags
-set(IREE_BUILD_COMPILER OFF)
-set(IREE_BUILD_TESTS OFF)
-set(IREE_BUILD_SAMPLES OFF)
-# Disable missing submodules error because we are only building the runtime.
-set(IREE_ERROR_ON_MISSING_SUBMODULES OFF)
-# Only enable local_sync/local_task/hip drivers for now.
-set(IREE_HAL_DRIVER_DEFAULTS OFF)
-set(IREE_HAL_DRIVER_LOCAL_SYNC ON)
-set(IREE_HAL_DRIVER_LOCAL_TASK ON)
-if(SHORTFIN_SYSTEMS_AMDGPU)
-  set(IREE_HAL_DRIVER_HIP ON)
+if (SHORTFIN_IREE_SOURCE_DIR OR SHORTFIN_BUNDLE_DEPS)
+  # Set IREE build flags, if we are building from source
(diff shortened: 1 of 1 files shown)


libshortfin/CMakeLists.txt
```diff
@@ -126,37 +126,38 @@ endif()
 
 ## iree runtime
 
-if (NOT SHORTFIN_IREE_SOURCE_DIR AND SHORTFIN_BUNDLE_DEPS)
+# Set IREE build flags
+set(IREE_BUILD_COMPILER OFF)
```
marbre, line 130: You could guard this by adding `if (SHORTFIN_IREE_SOURCE_DIR OR SHORTFIN_BUNDLE_DEPS)` but this isn't strictly necessary,

marbre (approved):

libshortfin/CMakeLists.txt
```diff
@@ -126,37 +126,38 @@ endif()
...
+set(IREE_BUILD_TESTS OFF)
+set(IREE_BUILD_SAMPLES OFF)
+# Disable missing submodules error because we are only building the runtime.
+set(IREE_ERROR_ON_MISSING_SUBMODULES OFF)
```
stellaraccident, line 134: Side note: I would love to get rid of this what and instead have a macro in iree that checks a submodule belt it is used. This flag has been a footgun for integration.

libshortfin/CMakeLists.txt
```diff
@@ -126,37 +126,38 @@ endif()
...
     GIT_SHALLOW TRUE
+    SYSTEM
+    EXCLUDE_FROM_ALL
   )
-  FetchContent_GetProperties(iree)
-  if(NOT iree_POPULATED)
-    FetchContent_MakeAvailable(iree)
+  FetchContent_GetProperties(shortfin_iree)
+  if(NOT shortfin_iree_POPULATED)
+    FetchContent_MakeAvailable(shortfin_iree)
```
stellaraccident, line 159: I changed this because cmake in Dev mode was giving a word salad warning about using it like it was. Thanks for untangling it...

stellaraccident (approved):

libshortfin/CMakeLists.txt
Groverkss: done

PR #184, opened 2024-09-11
[libshortfin] Fix non-editable install

Groverkss:
https://github.com/nod-ai/sharktank/pull/177 Broke building non-editable packages using `pip install .` . Non python files are not copied by package_dir and need to be explicitly copied to build_lib. The approach taken is same as IREE: https://github.com/iree-org/iree/blob/main/runtime/setup.py#L371-L390

### DIFF
--- a/libshortfin/setup.py
+++ b/libshortfin/setup.py
@@ -207,6 +207,26 @@ class CMakeBuildPy(_build_py):
                     cwd=CMAKE_BUILD_DIR,
                 )
 
+            # Copy non-python binaries generated during the build.
+            target_dir = os.path.join(
+                os.path.abspath(self.build_lib), "_shortfin_default"
+            )
+
+            print(f"Building in target: {target_dir}")
+            os.makedirs(target_dir, exist_ok=True)
+            print("Copying build to target.")
+            if os.path.exists(target_dir):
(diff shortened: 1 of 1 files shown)


stellaraccident:
Fine to land but the way that iree does this is wrong too as editable doesn't work. 

I think we're missing a thing but would need to fiddle with it. 

Let's make one of the CI actions so a non editable install and one editable so we can refactor with a bit more confidence.

marbre:
> Let's make one of the CI actions so a non editable install and one editable so we can refactor with a bit more confidence.

As building libshortfin is quite fast it would be good anyway to do multiple builds with different configurations (e.g. get IREE as a bundled dep, provide the source, use system deps,...). I will add this to my list  for the release build ci.

Groverkss:
> Fine to land but the way that iree does this is wrong too as editable doesn't work.
> 
> I think we're missing a thing but would need to fiddle with it.
> 
> Let's make one of the CI actions so a non editable install and one editable so we can refactor with a bit more confidence.

I think editable is working for me with this. If the build is editable, `self.build_lib` is set to a temporary directory where the lib is built and the target is copied there. So, it essentially becomes a no-op. The editable_finder still finds the correct target. Is there some other problem that i'm missing?

PR #187, opened 2024-09-11
[libshortfin] Fix and Improve static build ergonomics

Groverkss:
Changes two things:

- Instead of changing the name of the rule in cmake magic, specify it explicitly in src/CMakeLists.txt . Without this, we cannot set target options and static build doesn't work, because the rule name is changed.
- If SHORTFIN_BUILD_STATIC and SHORTFIN_BUILD_DYNAMIC both are specified, choose SHORTFIN_BUILD_STATIC, since SHORTFIN_BUILD_DYNAMIC is default.

### DIFF
--- a/libshortfin/CMakeLists.txt
+++ b/libshortfin/CMakeLists.txt
@@ -43,10 +43,19 @@ set(SHORTFIN_IREE_SOURCE_DIR "" CACHE FILEPATH "Path to IREE source")
 # Options for building static or dynamic libraries.
 option(SHORTFIN_BUILD_STATIC "Builds static libraries" OFF)
 option(SHORTFIN_BUILD_DYNAMIC "Builds dynamic libraries" ON)
-cmake_dependent_option(SHORTFIN_LINK_DYNAMIC "Links internal binaries against static libshortfin.a" ON "SHORTFIN_BUILD_DYNAMIC" OFF)
 if(NOT SHORTFIN_BUILD_STATIC AND NOT SHORTFIN_BUILD_DYNAMIC)
   message(FATAL_ERROR "One of SHORTFIN_BUILD_STATIC or SHORTFIN_BUILD_DYNAMIC must be ON")
 endif()
+# Since dynamic build is the default option, if someone specifies they want
+# a static build, and both options are on, we should build a static build.
+if (SHORTFIN_BUILD_STATIC)
+  if (SHORTFIN_BUILD_DYNAMIC)
+    message(STATUS "Both SHORTFIN_BUILD_DYNAMIC and SHORTFIN_BUILD_STATIC specified, choosing static build")
+  endif()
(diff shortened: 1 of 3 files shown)


libshortfin/src/CMakeLists.txt
```diff
@@ -29,4 +29,7 @@ shortfin_public_library(
     ${_INIT_INTERNAL_DEPS}
 )
 
-set_target_properties(shortfin PROPERTIES VERSION ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR} SOVERSION ${SOVERSION})
+set_target_properties(${SHORTFIN_LINK_LIBRARY_NAME}
+                       PROPERTIES
+                       VERSION ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}
+                       SOVERSION ${SOVERSION})
```
marbre, line 35: I am not sure we want to set a soversion on a static libray.

libshortfin/CMakeLists.txt
```diff
@@ -43,10 +43,19 @@ set(SHORTFIN_IREE_SOURCE_DIR "" CACHE FILEPATH "Path to IREE source")
...
 option(SHORTFIN_BUILD_DYNAMIC "Builds dynamic libraries" ON)
-cmake_dependent_option(SHORTFIN_LINK_DYNAMIC "Links internal binaries against static libshortfin.a" ON "SHORTFIN_BUILD_DYNAMIC" OFF)
 if(NOT SHORTFIN_BUILD_STATIC AND NOT SHORTFIN_BUILD_DYNAMIC)
   message(FATAL_ERROR "One of SHORTFIN_BUILD_STATIC or SHORTFIN_BUILD_DYNAMIC must be ON")
 endif()
+# Since dynamic build is the default option, if someone specifies they want
+# a static build, and both options are on, we should build a static build.
+if (SHORTFIN_BUILD_STATIC)
+  if (SHORTFIN_BUILD_DYNAMIC)
+    message(STATUS "Both SHORTFIN_BUILD_DYNAMIC and SHORTFIN_BUILD_STATIC specified, choosing static build")
```
marbre, line 53: I tend to rather exit with a `FATAL_ERROR` and let the user fix the configuration instead of making assumptions here (similar to the above that one must be set).

marbre: Looking closer to the CMake config, it seems desired that we can build both libraries. This wouldn't possible withbut is also prevented by this PR in general.

marbre (changes requested):
I think it is desirable to be able to build the static as well as the dynamic library and we shouldn't drop this. However, we cannot set properties on one of the variant if it isn't build. I send  out a PR to change this separately and would suggest to close this.

Groverkss:
Superseeded by https://github.com/nod-ai/sharktank/pull/189

PR #189, opened 2024-09-12
[libshortfin] Only set (so)version on dynamic libs

marbre:
With this the static as well as the dynamic library can be build at the same time. However, the version / so version should only be set on the dynamic library and only if that library is build.

Supersedes #187.

### DIFF
--- a/libshortfin/src/CMakeLists.txt
+++ b/libshortfin/src/CMakeLists.txt
@@ -28,5 +28,3 @@ shortfin_public_library(
     shortfin_systems_host
     ${_INIT_INTERNAL_DEPS}
 )
-
-set_target_properties(shortfin PROPERTIES VERSION ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR} SOVERSION ${SOVERSION})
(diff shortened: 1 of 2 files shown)


libshortfin/src/CMakeLists.txt
```diff
@@ -29,4 +29,6 @@ shortfin_public_library(
     ${_INIT_INTERNAL_DEPS}
 )
 
-set_target_properties(shortfin PROPERTIES VERSION ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR} SOVERSION ${SOVERSION})
+if(SHORTFIN_BUILD_DYNAMIC)
+  set_target_properties(shortfin PROPERTIES VERSION ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR} SOVERSION ${SOVERSION})
```
Groverkss, line 33: Do you  think we should set this in shortfin_public_library function defination instead?

marbre: I can't remember to have seen a version set on a static library, so I think we don't need to it to the function. But we can later on if we want to set the version number there.

Groverkss: Sorry, maybe i miscommunicated. I meant maybe we can set it here: https://github.com/nod-ai/sharktank/blob/main/libshortfin/build_tools/cmake/shortfin_library.cmake#L39

marbre: Great suggestion! I updated and renamed the patch accordingly.

Groverkss (approved):

PR #192, opened 2024-09-16
[libshortfin] Fix illegal nested lock access in Message destruction.

stellaraccident:
* When Message instances had lifetime management transferred to Python, destruction of the mutex would happen while it was held, causing an assert. I believe this was only happening in some process level GC scenarios before and was being masked. But it is also possible that it is the cause of our spurious crashes.
* While this fix made sense once I found it, I also took the change to get clang `-Wthread-safety` analysis working and wired up to a few of the trickiest pieces of thread safety code.
* This revealed real bugs and design issues that are now fixed.

### DIFF
--- a/libshortfin/src/shortfin/local/messaging.h
+++ b/libshortfin/src/shortfin/local/messaging.h
@@ -119,10 +119,6 @@ class SHORTFIN_API Message {
 
  protected:
   mutable iree::slim_mutex lock_;
-  // Guard a scope with the fine grained lock.
-  // iree::slim_mutex_lock_guard lock_guard() const {
-  //   return iree::slim_mutex_lock_guard(lock_);
-  // }
   // Manual retain and release. Callers must assume that the Message is no
   // longer valid after any call to Release() where they do not hold a known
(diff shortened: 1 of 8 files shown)


libshortfin/src/shortfin/local/messaging.h
```diff
@@ -104,10 +118,11 @@ class SHORTFIN_API Message {
   };
 
  protected:
+  mutable iree::slim_mutex lock_;
   // Guard a scope with the fine grained lock.
-  iree::slim_mutex_lock_guard lock_guard() const {
-    return iree::slim_mutex_lock_guard(lock_);
-  }
+  // iree::slim_mutex_lock_guard lock_guard() const {
```
rsuderman, line 123: Should this just be deleted now?

rsuderman (approved):

libshortfin/src/shortfin/local/messaging.h
stellaraccident: Yes, think I forgot to push. Thanks.

PR #193, opened 2024-09-17
[libshortfin] Add device_array.view

Groverkss:
Currently only restricted to row-major dense array views.

### DIFF
--- a/libshortfin/python/array_binding.cc
+++ b/libshortfin/python/array_binding.cc
@@ -99,30 +99,33 @@ py::object CreateMappingObject(mapping **out_cpp_mapping) {
   return py_mapping;
 }
 
-device_array PyDeviceArrayView(device_array &array, py::list offsets,
-                               py::list sizes) {
+device_array PyDeviceArrayView(device_array &array, py::args keys) {
   size_t rank = array.shape().size();
   Dims c_offsets(rank, 0);
   Dims c_sizes(array.shape_container());
-  if (offsets.size() > rank || sizes.size() > rank) {
+
+  if (keys.size() > rank) {
     throw std::invalid_argument(
         "Cannot create view into device_array greater than its rank");
   }
-  for (size_t idx = 0; py::handle key : offsets) {
-    if (py::isinstance<iree_device_size_t>(key)) {
+
+  for (size_t idx = 0; py::handle key : keys) {
+    if (py::isinstance<py::slice>(key)) {
+      // Slice key.
+      auto slice = py::cast<py::slice>(key);
+      auto [start, stop, step, length] = slice.compute(c_sizes[idx]);
+      if (step != 1) {
+        throw std::logic_error("view does not support strided slices");
+      }
+      c_offsets[idx] = start;
+      c_sizes[idx] = length;
+    } else if (py::isinstance<iree_device_size_t>(key)) {
       // Integer key.
       c_offsets[idx] = py::cast<iree_device_size_t>(key);
+      c_sizes[idx] = 1;
     } else {
-      throw std::logic_error("Cannot cast offset to size_t");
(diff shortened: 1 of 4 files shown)


libshortfin/python/array_binding.cc
```diff
@@ -30,6 +30,14 @@ static const char DOCSTRING_ARRAY_FILL[] = R"(Fill an array with a value.
 Equivalent to `array.storage.fill(pattern)`.
 )";
 
+static const char DOCSTRING_ARRAY_VIEW[] =
+    R"(Create a view of an array.
+
+Either integer indices or slices can be passed to the view() method to create
```
stellaraccident, line 36: Afaict, this statement isn't correct since it is taking an explicit offset/sizes args vs a list that can either be int or slice. The latter is more compatible with prevailing Python norms, fwiw.

libshortfin/tests/api/array_test.py
```diff
@@ -121,3 +121,13 @@ def test_xtensor_types(scope, dtype, code, py_value, expected_str):
     r = repr(ary)
     print("__repr__ =", r)
     assert expected_str in r, f"Expected '{expected_str}' in '{r}'"
+
+def test_view(lsys, device):
+    async def main():
+        src = sfnp.device_array(device, [2, 4], sfnp.uint8)
+        src.fill(b"\0\1\2\3")
+
+        view1 = src.view(offsets=[0, 1], sizes=[1, 3])
```
stellaraccident, line 130: Some more checking or boundary cases would be good.

libshortfin/python/array_binding.cc
Groverkss: int/slice makes sense. I switched back to that.

libshortfin/tests/api/array_test.py
Groverkss: Added some more tests

Groverkss:
The CI failure seems to be weird. It seems to be pointing at the array not being filled properly. @marbre Do you know if this is a CI flake?

stellaraccident:
There's nothing funny with scheduling there or anything. All synchronous code. 

I'd throw an assert after the fill to verify that is working. It's all pretty new code.

libshortfin/tests/api/array_test.py
```diff
@@ -121,3 +121,34 @@ def test_xtensor_types(scope, dtype, code, py_value, expected_str):
...
+        # Row indexing
+        ([0], "{{0, 1, 2, 3}}"),
+        # Sliced indexing
+        ([1, slice(2, 4)], "{{2, 3}}"),
+        ([slice(1, 2), slice(2, 4)], "{{2, 3}}"),
+    ],
+)
+def test_view(scope, keys, expected_str):
+    src = sfnp.device_array.for_host(scope.device(0), [4, 4], sfnp.uint8)
+    src.fill(b"\0\1\2\3")
```
stellaraccident, line 140: Actually, I misspoke: `fill` is asynchronous, so you are observing the value before the queue has processed the commands. This needs some more ergonomic work: we generally don't want to be async filling a host buffer like this for this kind of case. And I'm missing synchronous waits in the Python API, which would be the most convenient here.

Easiest thing would be if the items= API was available.

Best thing to do for now is to write this async like the tests above. Look at `test_fill_copy_to_for_transfer`. Note the `await device` before observing the result.

There will be better ergonomics for this eventually but no need to wait for them for a test.

stellaraccident (approved):
I left a note on how to fix the test. If that works, lgtm and can land.

libshortfin/tests/api/array_test.py
Groverkss: ah, added an await to fix it.

PR #195, opened 2024-09-18
[libshortfin] Fix python stub generation.

stellaraccident:
This brings pylance back to life for the native bindings. It was giving confusing error squiggles about not being able to use types that were variables, but the root cause was that the backing stubs were not generated properly for all of the module.

Includes a couple of other typing tweaks that were found to be helpful.

### DIFF
--- a/libshortfin/python/CMakeLists.txt
+++ b/libshortfin/python/CMakeLists.txt
@@ -45,33 +45,33 @@ function(shortfin_python_stubs build_type)
 
 endfunction()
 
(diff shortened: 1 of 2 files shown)


stellaraccident:
> LGTM!
> 
> I was looking at the same thing today. Nanobind has a —recursive flag to do this automatically too, that i was using manually. Sadly the cmake support for it never landed [wjakob/nanobind#463](https://github.com/wjakob/nanobind/pull/463)

Yeah, been on my list for a few days too. One more tweak coming in.

PR #199, opened 2024-09-19
Prefill tests for llama 2 7b and llama 3.1 8b

aviator19941:
Paged and direct kv cache prefill tests for llama 2 7b and llama 3.1 8b.

### DIFF
--- a/sharktank/sharktank/examples/paged_llm_v1.py
+++ b/sharktank/sharktank/examples/paged_llm_v1.py
@@ -140,6 +140,26 @@ class Batch:
             while len(block_ids_row) < needed_blocks:
                 block_ids_row.append(self.parent.alloc_page())
 
+    def compute_prefill_logits(
+        self,
+        model,
+        # [bs, batch_seq_len]
+        tokens: torch.Tensor,
+        *,
+        # [1, 1, batch_seq_len, batch_seq_len]
+        attention_mask: torch.Tensor,
+        # [bs, batch_seq_len // block_seq_stride]
+        seq_block_ids: torch.Tensor,
+        cache_state: list[torch.Tensor],
+    ):
+        logits = model.prefill(
+            tokens,
+            attention_mask=attention_mask,
+            seq_block_ids=seq_block_ids,
+            cache_state=cache_state,
+        )
+        return logits
+
     def prefill(self):
         model = self.parent.model
         attention_mask = model.attention_mask(
@@ -150,7 +170,7 @@ class Batch:
         trace_tensor("prefill.token_ids", self.token_ids)
         trace_tensor("prefill.seq_block_ids", seq_block_ids_tensor)
         trace_tensor("prefill.attention_mask", attention_mask)
-        self.logits = model.prefill(
+        logits = model.prefill(
             self.token_ids,
             attention_mask=attention_mask,
             seq_block_ids=seq_block_ids_tensor,
@@ -161,7 +181,7 @@ class Batch:
         # TODO: Normalize the output of extract_tokens_from_logits into
         # tensor [bs, 1].
         tokens = torch.tensor(
-            model.extract_tokens_from_logits(self.logits, self.seq_lens)
+            model.extract_tokens_from_logits(logits, self.seq_lens)
         ).unsqueeze(1)
         print(f":: Prefill results:\n{tokens.tolist()}")
         self.add_result_token(tokens)
--- a/sharktank/sharktank/examples/paged_llm_v1.py
+++ b/sharktank/sharktank/examples/paged_llm_v1.py
@@ -150,7 +150,7 @@ class Batch:
         trace_tensor("prefill.token_ids", self.token_ids)
         trace_tensor("prefill.seq_block_ids", seq_block_ids_tensor)
         trace_tensor("prefill.attention_mask", attention_mask)
-        logits = model.prefill(
+        self.logits = model.prefill(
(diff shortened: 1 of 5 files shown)


sharktank/sharktank/examples/paged_llm_v1.py
```diff
@@ -145,7 +145,7 @@ def prefill(self):
         trace_tensor("prefill.token_ids", self.token_ids)
         trace_tensor("prefill.seq_block_ids", seq_block_ids_tensor)
         trace_tensor("prefill.attention_mask", attention_mask)
-        logits = model.prefill(
+        self.logits = model.prefill(
```
ScottTodd, line 148: I don't think this value should be persisted outside of the `prefill` function, especially if `decode` is not also updating the value. Can you revert these `logits` -> `self.logits` changes?

Did the test need to read this value? Can you use another method to get the data?

sharktank/tests/models/llama/prefill_tests.py
```diff
@@ -0,0 +1,149 @@
...
+        config = self.createConfigModel(kv_cache_type)
+        model = PagedLlamaModelV1(self.dataset.root_theta, config)
+        generator = TorchGenerator(model, self.tokenizer_config)
+        batch = generator.begin_batch(self.prompts)
+        batch.prefill()
+
+        bs, *_ = batch.logits.shape
+        assert len(batch.seq_lens) == bs
+        greedy_token_logit = 0.0
+        for b, seq_len in enumerate(batch.seq_lens):
```
archana-ramalingam, line 33: Do we need a for loop here, if we only require the last token logit?

sharktank/sharktank/utils/hf_datasets.py
```diff
@@ -247,6 +247,40 @@ def alias_dataset(from_name: str, to_name: str):
     ),
 ).alias_to("mixtral_8x7b_q8_0_gguf")
 
+Dataset(
```
archana-ramalingam, line 250: While you are at it, might as well add all llama2 and llama3 variants available on amd-shark HF repo, if we support them on sharktank.

sharktank/tests/models/llama/prefill_tests.py
```diff
@@ -0,0 +1,149 @@
...
+            greedy_token_logit, golden_prefill_token_logit, rtol=rtol, atol=atol
+        )
+
+
+class Llama7BTest(BaseLlamaTest):
+    def setUp(self):
+        default_arguments = {
+            "hf_dataset": "llama2_7B_f16",
+            "tokenizer-config-json": Path("./llama2-7b/tokenizer_config.json"),
+            "prompt": ["I believe the meaning of life is"],
```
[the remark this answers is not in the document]
aviator19941: Theoretically a single prompt within tolerances should be good enough for prefill. Decode will yield different results from llama.cpp though.

sharktank/sharktank/utils/hf_datasets.py
aviator19941: I will add them to the amd-shark HF repo later.

sharktank/tests/models/llama/prefill_tests.py
aviator19941: Done

rsuderman (approved):

sharktank/tests/models/llama/llama_cpp_instructions.md
```diff
@@ -0,0 +1,13 @@
...
+git checkout llama_comparison
+cmake -B build
+cmake --build build --config Release
+
+### How to run llama.cpp
+huggingface-cli download meta-llama/Meta-Llama-3.1-70B --local-dir /home/avsharma/Meta-Llama-3.1-70B
+python convert_hf_to_gguf.py --outtype f16 --outfile Llama-3.1-70B-f16.gguf ../Meta-Llama-3.1-70B/
+
+To predict the prefill token, use `--n-predict 1` and to predict the first decode token, use `--n-predict 2`
+./build/bin/llama-cli -m Llama-3.1-70B-f16.gguf -p "I believe the meaning of life is" --threads 1 --temp 0 --n-predict 1 --no-warmup
```
ScottTodd, line 13: markdown formatting https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax

`````suggestion
### How to build llama.cpp
```bash
git clone https://github.com/aviator19941/llama.cpp.git
cd llama.cpp/
git checkout llama_comparison
cmake -B build
cmake --build build --config Release
```

### How to run llama.cpp
```bash
huggingface-cli download meta-llama/Meta-Llama-3.1-70B --local-dir /home/avsharma/Meta-Llama-3.1-70B
python convert_hf_to_gguf.py --outtype f16 --outfile Llama-3.1-70B-f16.gguf ../Meta-Llama-3.1-70B/

To predict the prefill token, use `--n-predict 1` and to predict the first decode token, use `--n-predict 2`
```bash
./build/bin/llama-cli -m Llama-3.1-70B-f16.gguf -p "I believe the meaning of life is" --threads 1 --temp 0 --n-predict 1 --no-warmup
```
`````

You may also want to use portable paths (e.g. `~/` instead of `/home/avsharma/`) and the upstream repo instead of your fork (https://github.com/ggerganov/llama.cpp instead of https://github.com/aviator19941/llama.cpp)

PR #201, opened 2024-09-19
[libshortfin] Add missing scheduler dependency.

### LINKED ISSUE #178: [libshortfin] Subprocess example tests spuriously fail
tests/examples/async_test.py subprocesses randomly/spuriously fail with an ABORT. In debug python builds, this manifests as an assert on some low level IO code trying to write to stdout. In production builds, it just aborts. We sometimes see corrupted output from these tests too (unprintable characters that show up in logs).

Marking as skip with a reference to this bug so we can see when it is happening without blocking the build.

stellaraccident:
As written, command buffers could be scheduled out of order (with signal < deps). We need a more enlightened scheduler soon, but for the moment, this makes it correct.

Also sets the IREE runtime IREE_TASK_SCOPE_FLAG_ABORT_ON_FAILURE in debug builds, which will cause an eager assert-like crash. Since in our usage, this is always a critical error, it makes sense to treat it like an assert (and I would like to add finer grained control in the future). In release builds, it will propagate as a soft error, but it is basically fatal as it will lock the accounting structures into a persistent error state.

Fixes #178

### DIFF
--- a/libshortfin/src/shortfin/local/systems/host.cc
+++ b/libshortfin/src/shortfin/local/systems/host.cc
@@ -26,6 +26,16 @@ HostCPUSystemBuilder::Deps::Deps(iree_allocator_t host_allocator) {
   iree_task_executor_options_initialize(&task_executor_options);
   iree_hal_task_device_params_initialize(&task_params);
   iree_task_topology_initialize(&task_topology_options);
+
+#ifndef NDEBUG
+  // TODO: In normal IREE programs, this is exposed as --task_abort_on_failure.
+  // It is a critical debug feature as it forces an eager program crash at
+  // the point encountered vs as a later, rolled up async status. Since it
+  // guards things that are API usage bugs in how we are using the runtime,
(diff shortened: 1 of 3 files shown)


libshortfin/src/shortfin/local/systems/host.cc
```diff
@@ -26,6 +26,16 @@ HostCPUSystemBuilder::Deps::Deps(iree_allocator_t host_allocator) {
   iree_task_executor_options_initialize(&task_executor_options);
   iree_hal_task_device_params_initialize(&task_params);
   iree_task_topology_initialize(&task_topology_options);
+
+#ifndef NDEBUG
```
benvanik, line 30: good idea

benvanik (approved):

libshortfin/src/shortfin/local/systems/host.cc
stellaraccident: I usually hate those kind of add-ons to NDEBUG, but for us, this really is like an assert so feel ok about it.

PR #202, opened 2024-09-20
[punet] Add support to export fp8 punet variant

nithinsubbiah:
This patch adds support to export punet with fp8 quantized conv/linear layers.


stellaraccident (approved):
Thanks. We're going to want to sync this and simplify with the Brevitas side soon... much more special casing, and this will be a monster. But fine for the moment.

PR #203, opened 2024-09-20
[libshortfin] Rework array mapping python bindings.

stellaraccident:
After a bit of use, I decided that some of the array API around fill and mappings needed some rework:

* Clarified that device_array.fill is asynchronous.
* Implemented `fill` and `items` on `mapping`, which matches the semantics of what they do (i.e. accessing mappable contents in a synchronous fashion).
* Implemented `mapping.fill` to operate either on a buffer pattern or primitive types.
* Made `device_array.items` an alias for `mapping.items`, mainly for convenience.

### DIFF
--- a/libshortfin/python/array_binding.cc
+++ b/libshortfin/python/array_binding.cc
@@ -35,7 +35,7 @@ update the contents.
 Equivalent to `array.storage.fill(pattern)`.
 )";
 
-static const char DOCSTRING_MAPPING_ITEMS[] =
+static const char DOCSTRING_ARRAY_ITEMS[] =
     R"(Convenience shorthand for map(...).items)";
 
 static const char DOCSTRING_ARRAY_MAP[] =
--- a/libshortfin/python/array_binding.cc
+++ b/libshortfin/python/array_binding.cc
@@ -27,20 +27,36 @@ Equivalent to `dest_array.storage.copy_from(source_array.storage)`.
 
 static const char DOCSTRING_ARRAY_FILL[] = R"(Fill an array with a value.
 
+Note that `fill` is asynchronous and may not be visible immediately. For immediate
+manipulation of host visible arrays, assign to the `items` property or use the
+`map(discard=True)` to get a mapping object which can be used to directly
+update the contents.
+
 Equivalent to `array.storage.fill(pattern)`.
 )";
 
-static const char DOCSTRING_ARRAY_ITEMS[] =
-    R"(Access contents as a Python array.
+static const char DOCSTRING_MAPPING_ITEMS[] =
+    R"(Convenience shorthand for map(...).items)";
 
-When reading this attribute, an array.array will be constructed with the
-contents of the shortfin device_array. This supports a subset of element types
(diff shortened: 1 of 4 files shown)


libshortfin/tests/api/array_storage_test.py
```diff
@@ -150,16 +150,57 @@ async def main():
...
+
+    lsys.run(main())
+
+
+def test_mapping_fill10(lsys, device):
+    async def main():
+        src = sfnp.storage.allocate_host(device, 20)
+        with src.map(discard=True) as m:
+            m.fill(b"9876543210")
+        assert bytes(src.map(read=True)) == b"98765432109876543210"
```
Groverkss, line 203: Can we parameterize these?

libshortfin/tests/api/array_test.py
```diff
@@ -123,6 +124,43 @@ def test_xtensor_types(scope, dtype, code, py_value, expected_str):
...
+    ],
+)
+def test_typed_fill(scope, dtype, value):
+    ary = sfnp.device_array.for_host(scope.device(0), [2, 4], dtype)
+    with ary.map(discard=True) as m:
+        m.fill(value)
+    readback = ary.items.tolist()
+    assert readback == [value] * 8
+
+
```
Groverkss, line 163: Can we have some tests for map(read=True, write=True) as well?

libshortfin/python/array_binding.cc
```diff
@@ -254,65 +411,73 @@ void BindArray(py::module_ &m) {
           [](storage &self, bool read, bool write, bool discard) {
             int access = 0;
             if (read) access |= IREE_HAL_MEMORY_ACCESS_READ;
-            if (write) access |= IREE_HAL_MEMORY_ACCESS_WRITE;
+            if (write || discard) access |= IREE_HAL_MEMORY_ACCESS_WRITE;
             if (discard) access |= IREE_HAL_MEMORY_ACCESS_DISCARD;
```
Groverkss, line 415: Should we have an assert/warning for read=True as well as discard=True? Since contents are undefined, reading seems like undefined behavior.

Groverkss (approved):
lgtm, with some suggestions

libshortfin/tests/api/array_storage_test.py
stellaraccident: Done.

libshortfin/python/array_binding.cc
stellaraccident: According to Ben, this is actually a valid distinction. There are configurations where knowing that the memory should only be writeable is different from read/write, even with discard. Something about knowing that it is write-only might allow the memory system to configure the caches in a way that penalizes read. So read|write|discard would be used to signal that you intend to read the contents after you've written them and need that to be efficient.

libshortfin/tests/api/array_test.py
stellaraccident: Extended the test to add a read modify write stanza.

PR #205, opened 2024-09-20
[shortfin] Fix nd array view indexing bug.

stellaraccident:
We were incorrectly truncating the view on inner index access. Used the newer APIs for data manipulation to construct a test which better validates. Kept the original test, even though it is not comprehensive.

Also adds arg names to the array constructor.

### DIFF
--- a/libshortfin/src/shortfin/array/array.cc
+++ b/libshortfin/src/shortfin/array/array.cc
@@ -152,6 +152,10 @@ device_array device_array::view(Dims &offsets, Dims &sizes) {
       has_stride = true;
     }
 
+    // Since we are only narrowing a dense, row major slice, as we traverse
+    // the dims, we are narrowing the memory view at each step by advancing
+    // the beginning based on the requested offset and pulling the end in
+    // by the difference in size.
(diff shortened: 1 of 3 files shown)


Groverkss (approved):

libshortfin/src/shortfin/array/array.cc
```diff
@@ -152,9 +152,9 @@ device_array device_array::view(Dims &offsets, Dims &sizes) {
       has_stride = true;
     }
 
-    new_dims[i] = slice_size;
     start_offset += row_stride * slice_offset;
-    span_size = row_stride * slice_size;
+    span_size -= row_stride * (new_dims[i] - slice_size);
```
Groverkss, line 156: I had to think a lot on what math this is exactly doing. Can you write a comment of how this math computes the slice?

stellaraccident: // Since we are only narrowing a dense, row major slice, as we traverse
    // the dims, we are narrowing the memory view at each step by advancing
    // the beginning based on the requested offset and pulling the end in
    // by the difference in size.

PR #206, opened 2024-09-21
[shortfin] Add argmax host op.

stellaraccident:
Adds `shortfin.array.argmax`, following closely to numpy conventions.

In the C++ implementation, this forced completion of some APIs around accessing shortfin arrays via xtensor. 

C++ based users will just use the `device_array::map_xtensor*` helpers to access xtensor expressions directly.

For Python, we export select eager op kernels to match common situations. The intent here is to just be able to do some lightweight algebra and data manipulation common in host-side ML (without deps).

### DIFF
--- a/libshortfin/python/array_host_ops.cc
+++ b/libshortfin/python/array_host_ops.cc
@@ -1,0 +1,77 @@
+// Copyright 2024 Advanced Micro Devices, Inc.
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "./lib_ext.h"
+#include "./utils.h"
+#include "shortfin/array/api.h"
+#include "shortfin/support/logging.h"
+#include "xtensor/xsort.hpp"
+#include "xtl/xhalf_float.hpp"
+
+using namespace shortfin::array;
+
+namespace shortfin::python {
+
+namespace {
+
+static const char DOCSTRING_ARGMAX[] =
+    R"(Returns the indices of the maximum values along an axis.
+
+See documentation for numpy.argmax which this operator emulates.
+
+Implemented dtypes: float16, float32.
+)";
+
+}  // namespace
+
+#define SF_UNARY_COMPUTE_CASE(dtype_name, cpp_type) \
+  case DType::dtype_name():                         \
+    return compute.template operator()<cpp_type>()
+
+void BindArrayHostOps(py::module_ &m) {
+  m.def(
+      "argmax",
+      [](device_array &input, int axis, std::optional<device_array> out,
+         bool keepdims, bool device_visible) {
+        if (axis < 0) axis += input.shape().size();
+        if (axis < 0 || axis >= input.shape().size()) {
+          throw std::invalid_argument(
+              fmt::format("Axis out of range: Must be [0, {}) but got {}",
+                          input.shape().size(), axis));
+        }
+        if (out && (out->dtype() != DType::int64())) {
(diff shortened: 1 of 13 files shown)


libshortfin/src/shortfin/array/dims.h
```diff
@@ -95,6 +137,10 @@ class SHORTFIN_API InlinedDims {
       std::fill(dims_.inline_dims.begin(), dims_.inline_dims.end(), value);
     }
   }
+  template <std::contiguous_iterator BeginTy, typename EndTy>
+  InlinedDims(BeginTy begin, EndTy end) {
```
Groverkss, line 141: nit: assert end > begin

libshortfin/tests/api/array_ops.py
```diff
@@ -0,0 +1,82 @@
...
+    assert out.view(3).items.tolist() == [0] * 16
+
+    # out= keepdims variant
+    out = sfnp.device_array(device, [4, 16, 1], dtype=sfnp.int64)
+    sfnp.argmax(src, keepdims=True, out=out)
+    assert out.shape == [4, 16, 1]
+    assert out.view(0).items.tolist() == [127] * 16
+    assert out.view(1).items.tolist() == [0] * 16
+    assert out.view(2).items.tolist() == [127] * 16
+    assert out.view(3).items.tolist() == [0] * 16
```
Groverkss, line 68: Can we have an example of argmax with axis=0?

libshortfin/python/array_host_ops.cc
```diff
@@ -0,0 +1,77 @@
...
+
+namespace {
+
+static const char DOCSTRING_ARGMAX[] =
+    R"(Returns the indices of the maximum values along an axis.
+
+See documentation for numpy.argmax which this operator emulates.
+
+Implemented dtypes: float16, float32.
+)";
```
Groverkss, line 26: Can we add a note somewhere that out must be / will be int64? It's not obvious here. Also, would be good to document what divise_visible means.

Also, maybe differing opinion, personally, I prefer the docs to not delegate semantics to something else. I don't think argmax docs are going to change much, can we just copy them here from numpy and mention the semantics mimic numpy?

Also, does nanobind allow specifying parameter specific docs? I like that numpy docs has a parameter section for each operator and has per parameter docs.

Groverkss (approved):

Groverkss:
> On the C++ side, this forced completion of some APIs around accessing shortfin arrays via xtensor. C++ based code will just use xtensor directly vs having a built-in op library.

> For Python, we export select eager op kernels to match common situations. The intent here is to just be able to do some lightweight algebra and data manipulation common in host-side ML (without deps).

What does the c++ / python side mean? Does it mean libshortfin c++ side and libshortfin python side, or do you mean host argmax vs device argmax? I'm a bit confused because I just see argmax bindings simply calling the c++ implementation.

stellaraccident:
> > On the C++ side, this forced completion of some APIs around accessing shortfin arrays via xtensor. C++ based code will just use xtensor directly vs having a built-in op library.
> 
> > For Python, we export select eager op kernels to match common situations. The intent here is to just be able to do some lightweight algebra and data manipulation common in host-side ML (without deps).
> 
> What does the c++ / python side mean? Does it mean libshortfin c++ side and libshortfin python side, or do you mean host argmax vs device argmax? I'm a bit confused because I just see argmax bindings simply calling the c++ implementation.

I'll make it clearer. What I meant was that if coding against shortfin and needing to do host side array manipulation in C++, users can just bail to xtensor directly and don't need type erased op-by-op wrappers like we provide for Python. If you're in C++, that will be superior because the xtensor expression trees use compile time code generation that you lose if you erase element types.

If coding as a python user and needing to do array manipulation, we wrap up some set of these ops so basic things can be done without more deps.

PR #210, opened 2024-09-24
Build fixes for libshortfin using Windows/MSVC.

ScottTodd:
Fixes for various warnings and errors observed when building with MSVC on my local dev machine:
* Implicit conversion from double to float
* `D:\dev\projects\SHARK-Platform\libshortfin\src\shortfin/array/dims.h(118): error C2440: 'reinterpret_cast': cannot convert from '__int64' to 'shortfin::array::InlinedDims<iree_hal_dim_t,std::allocator<iree_host_size_t>>::const_iterator::difference_type'`
* Missing symbols / includes
* `D:\dev\projects\SHARK-Platform\libshortfin\src\shortfin/local/messaging.h(143): warning C4910: 'shortfin::local::TypedFuture<shortfin::local::Message::Ref>': '__declspec(dllexport)' and 'extern' are incompatible on an explicit instantiation`
* Implicit/explicit type conversions with `string` and `c_str` in https://en.cppreference.com/w/cpp/filesystem/path

I'm still seeing `STATUS_DLL_NOT_FOUND` style errors when building with `-DSHORTFIN_BUILD_DYNAMIC=ON`: 
```
    Error running test executable.

      Path: 'D:/dev/projects/SHARK-Platform/libshortfin/build/cmake/default/src/shortfin/support/shortfin_support_test.exe'
      Working directory: 'D:/dev/projects/SHARK-Platform/libshortfin/build/cmake/default/src/shortfin/support'
      Result: Exit code 0xc0000135
```


libshortfin/src/shortfin/local/messaging.h
```diff
@@ -140,7 +140,7 @@ class SHORTFIN_API Message {
 };
 
 // Future specialization for Message::Ref.
-extern template class TypedFuture<Message::Ref>;
+template class TypedFuture<Message::Ref>;
```
stellaraccident, line 143: This is intentional. Does MSVC not support extern templates?

ScottTodd: https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4910?view=msvc-170`
> The explicit template instantiation named _<identifier>_ is modified by both the `__declspec(dllexport)` and `extern` keywords. However, these keywords are mutually exclusive. The `__declspec(dllexport)` keyword means instantiate the template class, while the `extern` keyword means do not automatically instantiate the template class.

These two types were already declared using `SHORTFIN_API`, i.e. `__declspec(dllexport)`:

https://github.com/nod-ai/SHARK-Platform/blob/61eacac59317f2b2db1e54778ca8a850db6f2734/libshortfin/src/shortfin/local/async.h#L160-L162 https://github.com/nod-ai/SHARK-Platform/blob/61eacac59317f2b2db1e54778ca8a850db6f2734/libshortfin/src/shortfin/support/api.h#L13-L27

stellaraccident (approved):
Not completely sure I fully grok the extern template thing, but if this gets us to something that works in CI, it can be poked on further if needed.

ScottTodd:
It works on my location Windows machine, at least. We can refactor the https://github.com/nod-ai/SHARK-Platform/blob/main/.github/workflows/ci_linux_x64-libshortfin.yml and other workflows into a matrix that runs across platforms / compilers / python versions (cc @marbre )... or just add a parallel `ci_windows_x64-shortfin.yml`.

PR #217, opened 2024-09-26
[libshortfin] Add additional auto documentation

marbre:
Auto-docs further members and classes for which docstrings are present.


stellaraccident (approved):
Thanks!

PR #219, opened 2024-09-26
[libshortfin] Add a workflow with disabled GIL

marbre:
This adds a seperate, and therefore temporary, workflow to run with a free-threaded Python which should be embedded into a matrix strategy.

To fully enable testing without the GIL, it is required to declare that `_shortfin_default.lib` can either be safley run without it or to run with `PYTHON_GIL=0`.

### DIFF
--- a/.github/workflows/ci_linux_x64_nogil-libshortfin.yml
+++ b/.github/workflows/ci_linux_x64_nogil-libshortfin.yml
@@ -1,0 +1,97 @@
+# Copyright 2024 Advanced Micro Devices, Inc.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_linux_x64_nogil-libshortfin.yml
```diff
@@ -0,0 +1,97 @@
+# Copyright 2024 Advanced Micro Devices, Inc.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+name: CI - libshortfin - GIL disabled
```
stellaraccident, line 7: Can we call this something like `Python 3.13 Freethreaded`

stellaraccident (approved):

.github/workflows/ci_linux_x64_nogil-libshortfin.yml
marbre: Of course :)

PR #220, opened 2024-09-26
Add test that exports to MLIR a small sharded Llama model

sogartar:
Add test that exports to MLIR a small sharded Llama model

The decode step requires exporting in non-strict torch mode due to
https://github.com/pytorch/pytorch/issues/135061

This export required to extend the registration functionality of our custom
tensor types by provinding `flatten_with_keys_fn`. This is also required
to bump the PyTorch version to >=2.4 for other export tests.

The export to MLIR fails with
TypeError: Unsupported torch type conversion for !torch.vtensor<[3,1,7],complex<f32>>
Needs further debugging.
Detailed error [here](https://github.com/user-attachments/files/17163454/sharded-llama-export-to-mlir-failure.txt).

### DIFF
--- a/sharktank/tests/models/llama/sharded_llama_test.py
+++ b/sharktank/tests/models/llama/sharded_llama_test.py
@@ -57,14 +57,13 @@ class ShardedLlamaTest(unittest.TestCase):
             config=self.config,
             vocab_size=self.vocabulary_size,
         )
+        self.prefill_seq_lens = torch.tensor(
+            [14, 9, self.block_seq_stride - 1], dtype=torch.int32
+        )
 
     def make_prefill_args(self, model: PagedLlamaModelV1) -> Dict[str, Any]:
-        seq_lens = torch.randint(
-            high=self.config.hp.context_length + 1, size=[self.batch_size]
-        )
-        seq_lens[self.batch_size - 1] = self.config.hp.context_length
         batch_seq_len = round_up_to_multiple_of(
-            int(torch.max(seq_lens)), model.cache.pad_sequence_stride
+            int(torch.max(self.prefill_seq_lens)), model.cache.pad_sequence_stride
         )
         token_ids = torch.randint(
             low=0,
@@ -72,7 +71,9 @@ class ShardedLlamaTest(unittest.TestCase):
             size=[self.batch_size, batch_seq_len],
             dtype=torch.int32,
         )
-        attention_mask = model.attention_mask(model.input_mask(seq_lens, batch_seq_len))
+        attention_mask = model.attention_mask(
+            model.input_mask(self.prefill_seq_lens, batch_seq_len)
+        )
         seq_block_ids = torch.arange(
             self.batch_size * batch_seq_len // self.config.block_seq_stride
         ).view(self.batch_size, -1)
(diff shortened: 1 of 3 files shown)


sharktank/tests/models/llama/sharded_llama_test.py
```diff
@@ -5,122 +5,119 @@
...
-            attention_mask=attention_mask,
-            seq_block_ids=seq_block_ids,
-            cache_state=sharded_cache_state,
-        )
-        # The errors are quite high, but for float64 both errors drop to < 1e-12.
-        # The numerics are probably correct.
-        torch.testing.assert_close(
-            sharded_prefill_result, expected_prefill_result, atol=1e-3, rtol=1e-2
+    def make_decode_args(self, model: PagedLlamaModelV1) -> Dict[str, Any]:
+        seq_lens = torch.randint(
```
rsuderman, line 106: Be specific in random testing where possible. If we want to check variable sequence length we should pick a set of numbers instead of randomizing.

sharktank/tests/models/llama/sharded_llama_test.py
```diff
@@ -5,122 +5,119 @@
...
-        # Verify decode step.
         decode_token_ids = torch.randint(
             low=0,
-            high=vocabulary_size,
-            size=[batch_size, 1],
+            high=self.vocabulary_size,
+            size=[self.batch_size, 1],
             dtype=torch.int32,
         )
         decode_seq_lens = torch.randint(
```
rsuderman, line 119: Same about above. We would expect the decode to be one more than the sequence length of the encode. We should maintain that behavior.

rsuderman:
Could you include a link to the export failure in a gist? We should be able to know why the export is failing.

sharktank/tests/models/llama/sharded_llama_test.py
rsuderman, line 1: Given we have a sharded attention block test, lets keep both around and renamed the original one to `sharded_attention_test.py`. It would be better to increase test coverage rather than replace. We can always delete it in the future.

rsuderman (changes requested):

sharktank/tests/models/llama/sharded_llama_test.py
sogartar: We are setting the random seed at the test setup stage, so it should be consistent between runs. Isn't the torch random generator going to be stable between releases?

sharktank/tests/models/llama/sharded_llama_test.py
sogartar: I actually had the name wrong on my previous PR. This test checks the whole model. Not just the attention. I just refactored a lot to not replicate code in the export test.

sharktank/tests/models/llama/sharded_llama_test.py
sogartar: About the sequence length I think that is the case, but the cache is being regenerated again. In this test it really does not matter if we get plausible numbers, just that the numbers are close.

sogartar:
> Could you include a link to the export failure in a gist? We should be able to know why the export is failing. There is a good chance we are successfully generating IR just failing during the conversion from `torch` to `linalg` (this would match the issue that Kyle saw).

The test calls just `shark_turbine.aot.exporter.export`, which I think does not compile with IREE. It returns an object that can then compile the IR. Are there conversion to `linalg` before that?

sogartar:
@rsuderman is this your PR https://github.com/llvm/torch-mlir/pull/3738 that may fix the type conversion issue?
```
TypeError: Unsupported torch type conversion for !torch.vtensor<[3,1,7],complex>
```

sharktank/tests/models/llama/sharded_llama_test.py
sogartar: I changed them to fixed values.

sharktank/tests/models/llama/sharded_llama_test.py
sogartar: I made them `prefill_seq_lens + 1`.

sogartar:
> Could you include a link to the export failure in a gist?
I added a link with full error in the description.

sogartar:
I tried with the changes from https://github.com/llvm/torch-mlir/pull/3738, but it does not solve the type conversion issue.

PR #221, opened 2024-09-26
[libshortfin] Build fixes for gcc 12.3.0 and 13.2.0.

ScottTodd:
Fixes https://github.com/nod-ai/SHARK-Platform/issues/215

### DIFF
--- a/libshortfin/src/shortfin/local/messaging.h
+++ b/libshortfin/src/shortfin/local/messaging.h
@@ -189,10 +189,11 @@ class SHORTFIN_API Queue : public std::enable_shared_from_this<Queue> {
   void Close();
 
  protected:
+  Queue(Options options);
+
  private:
   // Queues can only be created as shared by the System.
   static QueuePtr Create(Options options);
-  Queue(Options options);
   mutable iree::slim_mutex lock_;
   Options options_;
   // Backlog of messages not yet sent to a reader. Messages are pushed on the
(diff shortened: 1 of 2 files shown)


libshortfin/src/shortfin/local/messaging.h
```diff
@@ -189,10 +189,11 @@ class SHORTFIN_API Queue : public std::enable_shared_from_this<Queue> {
   void Close();
 
  protected:
+  Queue(Options options);
+
  private:
```
ScottTodd, line 194: If we want to keep this constructor private, we could also rework the anonymous namespace used for `QueueCreator`, which is already tagged `friend struct QueueCreator`.

ScottTodd: * https://stackoverflow.com/questions/19872920/is-it-possible-to-friend-a-class-in-an-unnamed-namespace-in-c
* https://stackoverflow.com/questions/63397661/gcc-doesnt-like-to-make-friends-with-anonymous-namespace-forward-declarations

ScottTodd: Found a different fix: https://stackoverflow.com/a/63398702

Apparently `friend struct QueueCreator` was an "elaborated type specifier", and that is not needed here since the name is not ambiguous. gcc is happier with just `friend QueueCreator`.

marbre (approved):
Tested locally, thanks for fixing!

PR #226, opened 2024-09-26
[shortfin] Fix issues with free threaded Python builds.

stellaraccident:
* Fixes a method signature issue that manifested on 3.13 generally (not FT related).
* Bumps nanobind to a recent HEAD commit to pick up free threaded support.
* Adds nanobind option to enable free threading if building for a CPython with it enabled.
* Adds a README stanza advising on how to acquire a free threaded CPython.
* Makes fastapi tests skip if deps not met (not yet available / hard to install).
* Adds an LSAN exclusion for something unrelated that has snuck in.
* Sets FT CMAKE_BUILD_TYPE=Debug in the CI.

Note that on the large NUMA system I was testing on, the CPU test which creates an executor on each NUMA node for all processes exceeded the default file handle ulimit, requiring it to be increased. I assume that the FT CPython internally uses more synchronization handles and it just happened to go over budget. This resulted in fixing https://github.com/iree-org/iree/pull/18609, which was causing an assert to be hit in this specific RESOURCE_EXHAUSTED scenario.

### DIFF
--- a/.github/workflows/ci_linux_x64_nogil-libshortfin.yml
+++ b/.github/workflows/ci_linux_x64_nogil-libshortfin.yml
@@ -79,6 +79,7 @@ jobs:
         mkdir ${{ env.LIBSHORTFIN_DIR }}/build
         cd ${{ env.LIBSHORTFIN_DIR }}/build
         cmake -GNinja \
+          -DCMAKE_BUILD_TYPE=Debug \
           -DCMAKE_C_COMPILER=clang-18 \
           -DCMAKE_CXX_COMPILER=clang++-18 \
           -DCMAKE_LINKER_TYPE=LLD \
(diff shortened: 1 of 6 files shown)


.github/workflows/ci_linux_x64_nogil-libshortfin.yml
```diff
@@ -79,6 +79,7 @@ jobs:
         mkdir ${{ env.LIBSHORTFIN_DIR }}/build
         cd ${{ env.LIBSHORTFIN_DIR }}/build
         cmake -GNinja \
+          -DCMAKE_BUILD_TYPE=Debug \
```
ScottTodd, line 82: Any particular reason for `Debug` here and only here? Or can this be reverted?

stellaraccident: Yes - we test on the CI with Debug/Asserts. This job was added as a release build... not sure why.

PR #233, opened 2024-09-27
[sharktank] Evaluation - Add Perplexity test

archana-ramalingam:
Add Perplexity test for LLM evaluation


sharktank/sharktank/utils/load_llm.py
```diff
@@ -0,0 +1,222 @@
...
+        model = self.parent.model
+        attention_mask = model.attention_mask(
+            model.input_mask(self.seq_lens, self.token_ids.shape[1])
+        )
+        seq_block_ids_tensor = self.pad_block_ids()
+        # print(f":: Invoke prefill:")
+        trace_tensor("prefill.token_ids", self.token_ids)
+        trace_tensor("prefill.seq_block_ids", seq_block_ids_tensor)
+        trace_tensor("prefill.attention_mask", attention_mask)
+        # print("prefill.token_ids", self.token_ids)
```
rsuderman, line 156: It could be handy to have a `self.debug` that conditionally prints all of these. Obviously not necessary just would be handy.

sharktank/sharktank/utils/tokenizer.py
```diff
@@ -76,9 +91,10 @@ class _TransformersTokenizer(InferenceTokenizer):
         def __init__(self, t: AutoTokenizer):
             self._t = t
 
-        def _encode(self, texts: list[str]) -> list[list[int]]:
+        def _encode(self, texts: list[str], add_start_token: bool) -> list[list[int]]:
             results = t.batch_encode_plus(
                 texts,
+                add_special_tokens=add_start_token,
```
rsuderman, line 97: I was struggling to find this param. Thanks!

sharktank/tests/evaluate/perplexity.py
```diff
@@ -0,0 +1,205 @@
...
+            self.batch = self.generator.begin_eval_batch(
+                token_batch=token_batch,
+                seq_lens_batch=seq_lens_batch,
+                bs=self.bs,
+            )
+
+            if is_first_token:
+                is_first_token = False
+                self.cache_state = self.batch.prefill()
+                # print(self.batch.detokenize())
```
rsuderman, line 121: Just another example in these prints. If we have a config setting we can optionally print these which would be handy.

rsuderman (approved):
Just some minor optional changes.

sharktank/tests/evaluate/perplexity.py
archana-ramalingam: Yes, the prints are here for debugging, I will cleanup once I validate the run. Having a config setting to print is actually a great idea.

sharktank/sharktank/utils/tokenizer.py
archana-ramalingam: Yes, here I am trying to not add a start token, but by default it's enabled. Couple of other variations, llama.cpp and HF, also pad on the left rather than the right (like we do) for their perplexity evaluation. Bunch of knobs to tune for sure.

sharktank/tests/evaluate/perplexity_test.py
```diff
@@ -0,0 +1,345 @@
...
+
+class PerplexityTest(unittest.TestCase):
+    def test_llama3_8B_f16_decomposed(self):
+
+        # Llama 3.1 8B decomposed
+
+        llama_8b_f16_gguf_path = "/data/extra/models/llama3.1_8B/llama8b_f16.gguf"
+        llama_8b_f16_tokenizer_path = (
+            "/data/extra/models/llama3.1_8B/tokenizer_config.json"
+        )
```
ScottTodd, line 21: Unit tests failed on this PR and after merge:
* https://github.com/nod-ai/SHARK-Platform/actions/runs/11359973189/job/31596941530
* https://github.com/nod-ai/SHARK-Platform/actions/runs/11360375745/job/31598023494

---

Any files required to run a test should be either
* included in the repository (if small enough)
* downloaded (and cached) on demand as part of the test
* downloaded (and cached) ahead of time via a script

As this is, this test will only run on a machine that has already run some unknown, undocumented setup steps.

.github/workflows/eval_test.yaml
```diff
@@ -0,0 +1,60 @@
...
+          # Note: We install in three steps in order to satisfy requirements
+          # from non default locations first. Installing the PyTorch CPU
+          # wheels saves multiple minutes and a lot of bandwidth on runner setup.
+          pip install --no-compile -r pytorch-cpu-requirements.txt
+          pip install --no-compile -f https://iree.dev/pip-release-links.html --src deps \
+            -e "git+https://github.com/iree-org/iree-turbine.git#egg=iree-turbine"
+          pip install --no-compile -r requirements.txt -e sharktank/ shortfin/
+
+      - name: Run perplexity test
+        run:  pytest sharktank/tests/evaluate/perplexity_test.py
```
ScottTodd, line 60: If more tests are going in this category, we could use pytest marks or some other filtering to pick up the list of tests to run. As this is now, the new tests are running in multiple workflows (on every commit _and_ here on a nightly schedule).

requirements.txt
```diff
@@ -7,6 +7,7 @@ onnx==1.15.0
 huggingface-hub==0.22.2
 transformers==4.40.0
 sentencepiece==0.2.0
+datasets==3.0.0
```
ScottTodd, line 10: We should steer towards putting requirements in the subproject requirements files instead of this top level file, especially if this is a test-only requirement

* Existing requirements file specific to sharktank: https://github.com/nod-ai/SHARK-Platform/blob/main/sharktank/requirements.txt
* Test requirements file for shortfin: https://github.com/nod-ai/SHARK-Platform/blob/main/shortfin/requirements-tests.txt

PR #234, opened 2024-09-27
[shortfin] Refactor to avoid changing directories

marbre:
Instead of changing into the directories, the working directories are now set in the jobs.

### DIFF
--- a/.github/workflows/ci_linux_x64-libshortfin.yml
+++ b/.github/workflows/ci_linux_x64-libshortfin.yml
@@ -57,8 +57,8 @@ jobs:
         ref: candidate-20240904.1006
 
     - name: Initalize IREE submodules
+      working-directory: ${{ env.IREE_REPO_DIR }}
       run : |
-        cd ${{ env.IREE_REPO_DIR }}
         git submodule update --init --depth 1 -- third_party/benchmark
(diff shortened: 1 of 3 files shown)


stellaraccident (approved):
Thanks. Better.

.github/workflows/ci_linux_x64-libshortfin.yml
```diff
@@ -80,32 +80,33 @@ jobs:
     - name: Build shortfin (full)
       run: |
         mkdir ${{ env.LIBSHORTFIN_DIR }}/build
-        cd ${{ env.LIBSHORTFIN_DIR }}/build
         cmake -GNinja \
+          -S${{ env.LIBSHORTFIN_DIR }} \
+          -B${{ env.LIBSHORTFIN_DIR }}/build \
```
ScottTodd, line 85: You can also set `working-directory` for a step (or set a default for an entire job). Running all steps from `shortfin/` could simplify this further.

ScottTodd (approved):

.github/workflows/ci_linux_x64-libshortfin.yml
marbre: Oh, nice! Refactored accordingly.

ScottTodd (approved):

PR #235, opened 2024-09-27
[shortfin] Update README

marbre:
Updates the native C++ build instructions (`pip install` finishes "successfully" even if the project was only configured and no libs were built) and simplifies the instructions on how to get a free-threaded Python.


stellaraccident (approved):
Thanks!

PR #237, opened 2024-09-27
Make sharded Llama export test also compile to IREE module and verify numerics

sogartar:
Verifies the IREE module numerical accuracy compared to execution with PyTorch.
The prefill step result has very low absolute accuracy of around `1e-2` for FP32.
The resulting cache state of prefill is way off.
The decode step accuracy is also completely off.

This test is market as skipped until https://github.com/iree-org/iree/pull/18663 is merged. Without it the IREE compilation will crash.


sogartar:
This PR is waiting on https://github.com/iree-org/iree/pull/18663.
@rsuderman should I mark it xfail and merge it like that. Then reenable when the fix gets merged?

sharktank/tests/models/llama/sharded_llama_test.py
```diff
@@ -18,6 +27,92 @@
 import torch
 from copy import deepcopy
 from shark_turbine.aot import FxProgramsBuilder, export
+import iree.runtime
+from pathlib import Path
+
+
+def get_iree_devices(device_count: int) -> List[iree.runtime.HalDevice]:
+    hal_driver = iree.runtime.get_driver("local-task")
```
IanNod, line 35: might be better to have the backend as an input arg instead of hardcoded "local-task" so that we can easily expand testing to add other backend support like rocm/hip.

sharktank/tests/models/llama/sharded_llama_test.py
```diff
@@ -18,6 +27,92 @@
...
+            res.extend(
+                [
+                    prepare_iree_module_function_args([shard], [device])[0]
+                    for shard, device in zip(arg.shards, devices)
+                ]
+            )
+        elif isinstance(arg, (InferenceTensor, torch.Tensor)):
+            res.append(
+                iree.runtime.asdevicearray(
+                    devices[0], unbox_tensor(arg).to("cpu").numpy()
```
IanNod, line 105: Why do we need to unbox the tensor args here? Will become an issue for quantized tensors

sharktank/tests/models/llama/sharded_llama_test.py
```diff
@@ -237,5 +348,86 @@ def _(model, *args, **kwargs) -> torch.Tensor:
...
+                module=iree_module,
+                vm_context=vm_context,
+            )
+            prefill_iree_result = iree_to_torch(*prefill_iree_result)
+            assert len(prefill_iree_result) == 1
+            expected_prefill_result = sharded_model.prefill(**sharded_prefill_args)
+            # TODO: Although, not entirely wrong, investigate why this accuracy is that
+            # low for fp32.
+            torch.testing.assert_close(
+                prefill_iree_result[0], expected_prefill_result, atol=0.0011, rtol=0.013
```
IanNod, line 389: This is probably fine but I think I would rather see stricter tolerances that we should be able to meet, and xfail until it works as expected.

IanNod (approved):
Couple nit comments but looks good to me

sharktank/tests/models/llama/sharded_llama_test.py
sogartar: I added a parameter.

sharktank/tests/models/llama/sharded_llama_test.py
sogartar: It can be `DeafultPrimitiveTensor`. I make the type check more specific.

sharktank/tests/models/llama/sharded_llama_test.py
sogartar: Make sense. I changed to default tolerance.

sharktank/tests/models/llama/sharded_llama_test.py
```diff
@@ -5,10 +5,20 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 import unittest
-from typing import Any, Dict, List, Tuple
+from typing import Any, List, Tuple, Union
+import typing
```
rsuderman, line 9: Import `OrderedDict` from typing above instead of needing the module.

sharktank/tests/models/llama/sharded_llama_test.py
```diff
@@ -123,17 +228,19 @@ def make_decode_args(self, model: PagedLlamaModelV1) -> Dict[str, Any]:
...
+                ("start_positions", start_positions),
+                ("seq_block_ids", seq_block_ids),
+                ("cache_state", cache_state),
+            ]
+        )
 
     def make_equal_unsharded_and_sharded_decode_args(
         self, model: PagedLlamaModelV1, sharded_model: PagedLlamaModelV1
-    ):
+    ) -> Tuple[typing.OrderedDict[str, Any], typing.OrderedDict[str, Any]]:
```
rsuderman, line 243: See comment above about importing `OrderedDict` from `typing` instead of `typing` itself

rsuderman (changes requested):

sharktank/tests/models/llama/sharded_llama_test.py
ScottTodd: Better yet, you can skip the imports for standard collections: https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections. https://peps.python.org/pep-0585/. (Since Python 3.9)

See also https://stackoverflow.com/a/62033243

sogartar: I thought it would clash with the `collections.OrderedDict`, but I guess the one from `typing` is fine for that too.

ScottTodd: I think you can get by with
```python
from typing import Any, Union
from collections import OrderedDict
```

Then use the standard collection `list, tuple` and `OrderedDict` in your typing hints.

rsuderman (approved):

PR #240, opened 2024-09-30
[shortfin] Add extras_require for optional deps.

ScottTodd:
Progress on https://github.com/nod-ai/SHARK-Platform/issues/130

These optional dependencies can be installed with e.g. `pip install {INDEX URL HERE} shortfin[server]`.

* I'm not sure which names we want and with which packages. How many is too many? 5 seems like a lot.
* YMMV with Python 3.13t (free-threaded). Packages that don't yet distribute binaries for 3.13t may need to be built from source, which might require rust or other local tools.

We can share these requirements with `requirements-*.txt` files using this technique, if we want: https://github.com/nod-ai/SHARK-Platform/blob/72665fc0e3e8d1ed3b4e7eacdfa9c493e2ba0852/sharktank/setup.py#L46-L57 https://github.com/nod-ai/SHARK-Platform/blob/72665fc0e3e8d1ed3b4e7eacdfa9c493e2ba0852/sharktank/setup.py#L100-L105

Docs: https://setuptools.pypa.io/en/latest/userguide/dependency_management.html

### DIFF
--- a/shortfin/setup.py
+++ b/shortfin/setup.py
@@ -391,7 +391,8 @@ setup(
             "tokenizers",
         ],
         "onnx": [
-            "onnx",
+            "iree-runtime",
+            "iree-compiler[onnx]",
         ],
         "torch": [
             # Note that this pulls in `torch>=2.3.0`, which then pulls in all of CUDA.
--- a/shortfin/setup.py
+++ b/shortfin/setup.py
@@ -381,6 +381,27 @@ setup(
         [CMakeExtension("_shortfin_default.lib")]
         + ([CMakeExtension("_shortfin_tracy.lib")] if ENABLE_TRACY else [])
     ),
+    extras_require={
+        "testing": [
+            "pytest",
+            "requests",
+        ],
+        "apps": [
+            "dataclasses-json",
+            "tokenizers",
+        ],
+        "onnx": [
+            "onnx",
+        ],
+        "torch": [
+            # Note that this pulls in `torch>=2.3.0`, which then pulls in all of CUDA.
+            "iree-turbine",
+        ],
+        "server": [
+            "fastapi",
+            "uvicorn",
+        ],
+    },
     cmdclass={
         "build": CustomBuild,
         "build_ext": NoopBuildExtension,


shortfin/setup.py
```diff
@@ -381,6 +381,27 @@ def populate_built_package(abs_dir):
...
+            "dataclasses-json",
+            "tokenizers",
+        ],
+        "onnx": [
+            "onnx",
+        ],
+        "torch": [
+            # Note that this pulls in `torch>=2.3.0`, which then pulls in all of CUDA.
+            "iree-turbine",
+        ],
```
ScottTodd, line 399: May want `onnx` to pull in `iree-compiler` and `iree-runtime` directly, then either have `torch` get them indirectly via `iree-turbine` or directly as well.

marbre: That feels a little bit like a circular dependency as `iree-compiler[onnx]` pulls in onnx even though needed for `iree-import-onnx`:
https://github.com/iree-org/iree/blob/c86b621665ad8b898c0d0e987a45267f78eb9c3f/compiler/setup.py#L468

marbre:
> * I'm not sure which names we want and with which packages. How many is too many? 5 seems like a lot.

I don't have strong feelings as well and I am not sure about what names we really want to have. But I agree that 5 seems like a lot, especially as 

> * YMMV with Python 3.13t (free-threaded). Packages that don't yet distribute binaries for 3.13t may need to be built from source, which might require rust or other local tools.

We actually don't want to pull in some of the deps if installing 3.13t at all. Within my early experiments ONNX failed to install which is why I added `[requirements-tests-nogil.txt](https://github.com/nod-ai/SHARK-Platform/blob/main/shortfin/requirements-tests-nogil.txt)`.

> We can share these requirements with `requirements-*.txt` files using this technique, if we want:

There might be even more options if we think about moving modernizing the project (https://packaging.python.org/en/latest/guides/modernize-setup-py-project/) and moving dependencies to our `pyproject.toml` (see https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#dependencies-and-requirements)  as long as this works in combination with a setup.py at all. But that can also be something for a follow up.


Side note: For nighties (as long as we don't publish via PyPI) we could add direct URL dependencies (see https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#direct-url-dependencies). However, as such packages cannot be published on PyPI this is nothing we want for releases.

shortfin/setup.py
ScottTodd: Oh yeah, we could put `iree-compiler[onnx]` here instead of `onnx` on its own. I can test that.

ScottTodd:
@stellaraccident we discussed this a bit today. Seems like we'll want a meta-package instead of putting more deps in shortfin itself?

I was referencing this from https://github.com/nod-ai/SHARK-Platform/issues/130 when I put together this draft PR:

> #### Python deps
> `shortfin` does not have any Python deps for minimal functionality, but it is expected that a number of deps will be useful for various applications of it, and it may provide features based on these deps in an optional fashion that people can use if helpful/available:
> 
> * pytest (for testing)
> * numpy
> * torch
> * uvicorn (or other ASGI web server)
> * iree-turbine (for torch compilation)
>   
>   * Transitive deps on iree-compiler and iree-runtime
> * onnx (for onnx compilation)
> * etc

ScottTodd:
> > * I'm not sure which names we want and with which packages. How many is too many? 5 seems like a lot.
> 
> I don't have strong feelings as well and I am not sure about what names we really want to have. But I agree that 5 seems like a lot

Um: https://pypi.org/project/transformers/
> Provides-Extra: accelerate, agents, all, audio, benchmark, codecarbon, deepspeed, deepspeed-testing, dev, dev-tensorflow, dev-torch, flax, flax-speech, ftfy, integrations, ja, modelcreation, natten, onnx, onnxruntime, optuna, quality, ray, retrieval, ruff, sagemaker, sentencepiece, serving, sigopt, sklearn, speech, testing, tf, tf-cpu, tf-speech, tiktoken, timm, tokenizers, torch, torch-speech, torch-vision, torchhub, video, vision

https://github.com/huggingface/transformers/blob/d7950bff82b18c823193d17d72188c5e46d06c83/setup.py#L264-L421

Well there's an upper limit to keep in mind :P

ScottTodd:
Here's what a meta package could look like instead of (or in addition to) these extras for `shortfin` itself: https://github.com/nod-ai/SHARK-Platform/pull/244

ScottTodd:
Closing this in favor of a meta-package approach.

PR #241, opened 2024-10-01
[shortfin] Fix mobilenet_server example by filling mapped storage.

ScottTodd:
I saw this error trying to run this example:
```
RuntimeError: Async exception on <Worker 'device-hostcpu:0:0@0'>): '_shortfin_default.lib.array.storage' object has no attribute 'data'
```

I looked around for and API that would allow for copying from `array.array` (https://docs.python.org/3/library/array.html) into shortfin's `array.storage` (https://shortfin.readthedocs.io/en/latest/reference.html#shortfin_default.lib.array.storage) and only found `map()`.

Quite possible that I'm missing something, or there is an easier interface that isn't fully plumbed through to Python and the documentation yet.

Logs of this running: https://gist.github.com/ScottTodd/da8c1d6805584490c46e4d2086404a6b


shortfin/examples/python/mobilenet_server/inference_system.py
```diff
@@ -40,7 +40,12 @@ async def run(self):
             # just writing to the backing storage is the best we have API
             # support for. Generally, APIs on storage should be mirrored onto
             # the array.
-            self.host_staging.storage.data = request.raw_image_data
+            # TODO: Easier to use API for writing into the storage
```
stellaraccident, line 43: Grrg. Yeah, these low level transfer methods need an upgrade. Up until the most recent commit, I didn't have the device level metadata I needed to determine the right way to load data on to the device (i.e. map, transfer, chunk, etc). Now that I have that, need to design a more ergonomic API.

With that said, I believe that `mapping.fill(buffer)` will do what you want if the buffer is the size of the target (i.e. it will copy x1 vs tile). There are some tests that exercise that.

So I think you can say:

```
with self.host_staging.storage.map(write=True, discard=True) as m:
  m.fill(request.raw_image_data)
```

(assuming sizes line up)

stellaraccident (approved):

PR #247, opened 2024-10-03
Refresh metadata in sharktank/setup.py.

ScottTodd:
* Drop author email for consistency with shortfin (we could also pick a mailing list or more up to date contact email)
* Update URL following repository rename

More will be needed as we add a dep on `shortfin` and resume regular publishing of this `sharktank` package to https://pypi.org/project/sharktank/


sogartar (approved):
The CI workflows would also need an update.

PR #248, opened 2024-10-03
[tuner] Account for different CWD

kuhar:
Since the default invocation is now in the repo root dir, look for the config files in the same dir as the specified input.

Also add tuning directories to gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,8 +25,11 @@ wheelhouse
 *.whl
 *.venv
 
-#Model artifacts
+# Model artifacts
 *.pt
 *.safetensors
 *.gguf
 *.vmfb
+
+# Tuner artifacts
+tuning_*/
--- a/tuner/examples/punet/punet_autotune.py
+++ b/tuner/examples/punet/punet_autotune.py
@@ -122,8 +122,12 @@ def main():
 
     path_config = libtuner.PathConfig()
     # We expect the configs to be in the same dir as the input.
-    path_config.global_config_prolog_mlir = input_dir / path_config.global_config_prolog_mlir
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -25,8 +25,11 @@ wheelhouse
...
 
-#Model artifacts
+# Model artifacts
 *.pt
 *.safetensors
 *.gguf
 *.vmfb
+
+# Tuner artifacts
+tuning_*/
```
ScottTodd, line 35: I'd prefer to put this in `tuner/.gitignore`, with the artifacts similarly scoped, so each subproject is more self-contained. As we're setting up packaging for each subproject I don't want the line between packages to blur like this.

kuhar: The thing with the current module setup is that it needs to start at the root of the project: `python -m tuner.examples.dispatch tuner/examples/dispatch/benchmark.mlir --num-candidates=20`, so this guards against any dirs created by running examples.

Maybe there's a better way to set up the module path, but I don't know how to do it.

ScottTodd: One option is to put a nested `SHARK-Platform/tuner/tuner/` directory, as with https://github.com/nod-ai/SHARK-Platform/tree/main/sharktank/sharktank and https://github.com/nod-ai/SHARK-Platform/tree/main/shortfin/python/shortfin

kuhar: OK, let me try that instead...

ScottTodd: That's top of my mind since https://github.com/nod-ai/SHARK-Platform/pull/244#pullrequestreview-2344357447

> Python projects have an annoying way of repeating primary directory names...

PR #251, opened 2024-10-04
[tuner] Reorganize tuner directories

kuhar:
Add a subdirectory for the tuner library. This is so that examples do not have to be executed from the repo root and risk polluting it with temporary files.


kuhar:
Suggested by @ScottTodd in https://github.com/nod-ai/SHARK-Platform/pull/248

archana-ramalingam (approved):
Approved, this aligns with our future sharktank packaging path.

PR #252, opened 2024-10-04
Add python3.12 requirement to README

monorimet:
We use the `walk_up` method of PurePath from pathlib that was introduced in python3.12.

I suppose we could also just implement walk_up() manually so we can support more python versions. Not sure if any other pieces will require 3.12


ScottTodd (approved):
This is fine for now. As we start directing users to the published packages we'll want to refresh this whole README file.

monorimet:
Closing since this is no longer relevant.

PR #256, opened 2024-10-05
[tuner] Use JSON for benchmark output

mihaescuvlad:
### Notes
- Adds `extract_benchmark_from_run_result` method to help in fetching the "benchmarks" data
- Updates `IREEBenchmarkResult` model to reflect that the result is no longer stored as string but as a list of benchmarks
- Updates the parsing of dispatches and models to read from Json

### Testing
- Updates tests to verify that `get_mean_time` functions as expected
- Updates tests to verify that Json data is properly parsed and processed

### DIFF
--- a/tuner/tuner/libtuner.py
+++ b/tuner/tuner/libtuner.py
@@ -36,6 +36,7 @@ from dataclasses import dataclass, field
 from typing import Type, Optional, Callable, Iterable, Any
 import pickle
 import random
+import json
 from abc import ABC, abstractmethod
 import iree.runtime as ireert
 from . import candidate_gen
@@ -236,19 +237,47 @@ class ParsedDisptachBenchmarkResult:
 class IREEBenchmarkResult:
     # Default format follows output of iree-benchmark-module
     candidate_id: int
-    result_str: str
 
-    def get_mean_time(self) -> Optional[float]:
-        if not self.result_str:
+    # A list of dictionaries, each representing a benchmark result
+    # Each dictionary contains fields like: aggregate_name: string, real_time: float, cpu_time: float, time_unit: str, repetitions: int, etc.
+    result_json: list[dict[str, Any]]
+
+    def get_mean_time_us(self) -> Optional[float]:
+        """Compute the mean time (in microseconds) for all of the benchmarks"""
+        if not self.result_json:
             return None
-        pattern = r"process_time/real_time_mean\s+([\d.]+)\s\w{2}"
-        match = re.search(pattern, self.result_str)
-        if not match:
-            return None
-        try:
-            return float(match.group(1))
-        except ValueError:
-            return None
+
+        mean_benchmark = self.find_mean_benchmark(self.result_json)
+
+        if mean_benchmark:
+            real_time = mean_benchmark.get("real_time")
+            time_unit = mean_benchmark.get("time_unit")
+
+            if real_time is not None:
+                return self.unit_to_microseconds(real_time, time_unit)
+
+        return None
+
+    @staticmethod
+    def find_mean_benchmark(result_json: list[dict[str, Any]]) -> Optional[dict]:
+        for benchmark in result_json:
+            if benchmark.get("aggregate_name") == "mean":
+                return benchmark
+
+        return None
+
+    @staticmethod
+    def unit_to_microseconds(real_time: float, time_unit: str) -> float:
+        unit_conversions = {
+            "s": 1e6,
+            "ms": 1e3,
+            "us": 1,
+            "ns": 1e-3,
+        }
+
+        assert time_unit in unit_conversions, f"Unsupported time unit: {time_unit}"
+
+        return real_time * unit_conversions[time_unit]
 
 
 def generate_display_DBR(candidate_id: int, mean_time: float) -> str:
@@ -619,6 +648,26 @@ def multiprocess_progress_wrapper(
     return results
 
 
+def extract_benchmark_from_run_result(
+    run_result: RunResult,
+) -> Optional[list[dict[str, Any]]]:
+    """Extract the benchmark from the result JSON"""
+    if run_result.process_res and run_result.process_res.stdout:
+        try:
+            result_json = json.loads(run_result.process_res.stdout)
+
+            return result_json.get("benchmarks", None)
+        except json.JSONDecodeError as e:
+            handle_error(
+                condition=True,
+                msg=f"Failed to parse JSON from stdout: {e}",
+                error_type=ValueError,
+                exit_program=True,
+            )
+
+    return None
+
+
 def numerical_sort_key(path: Path) -> tuple[int | float, str]:
     """
     Define a sort key function that splits the filename into a numeric and a string part.
@@ -896,9 +945,9 @@ def parse_dispatch_benchmark_results(
                 incomplete_list.append(candidate_id)
             continue
 
-        res_str = process_res.stdout
-        res = IREEBenchmarkResult(candidate_id, res_str)
-        benchmark_time = res.get_mean_time()
+        res_json = extract_benchmark_from_run_result(benchmark_result.run_result)
+        res = IREEBenchmarkResult(candidate_id, res_json)
+        benchmark_time = res.get_mean_time_us()
         assert benchmark_time is not None
         candidate_trackers[candidate_id].first_benchmark_time = benchmark_time
         candidate_trackers[
@@ -1185,9 +1234,9 @@ def parse_model_benchmark_results(
                     baseline_time = None
                 continue
 
-            result_str = process_res.stdout
-            res = IREEBenchmarkResult(candidate_id, result_str)
-            benchmark_time = res.get_mean_time()
+            result_json = extract_benchmark_from_run_result(task_result.run_result)
+            res = IREEBenchmarkResult(candidate_id, result_json)
+            benchmark_time = res.get_mean_time_us()
             assert benchmark_time is not None
 
             # Record baseline benchmarking result and skip rest processes
@@ -1328,7 +1377,7 @@ def benchmark_models(
     )
 
 
-def summerize_top_candidates(
+def summarize_top_candidates(
     path_config: PathConfig, candidate_trackers: list[CandidateTracker]
 ):
     dump_list = []
--- a/tuner/tuner/libtuner_test.py
+++ b/tuner/tuner/libtuner_test.py
@@ -6,11 +6,12 @@
 
 import argparse
 import pytest
+import json
 from unittest.mock import call, patch, MagicMock
 from . import libtuner
 
 """
-Usage: python -m pytest test_libtuner.py
+Usage: python -m pytest libtuner_test.py
 """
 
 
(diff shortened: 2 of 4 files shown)


tuner/tuner/libtuner_test.py
```diff
@@ -145,6 +147,24 @@ def test_parse_dispatch_benchmark_results():
         "2\tMean Time: 200.0\n",
         "Candidate 3 not completed",
     ]
+    expected_json_output = {
+        "context": {
+            "date": ANY,
+            "num_cpus": ANY,
+            "mhz_per_cpu": ANY,
+            "cpu_scaling_enabled": ANY,
```
mihaescuvlad, line 155: Should I add a comment explaining that these fields are dynamically generated, therefore they are ignored using `ANY`?

tuner/tuner/libtuner.py
```diff
@@ -696,6 +725,51 @@ def append_to_file(lines: list[str], filepath: Path, title: str = "") -> None:
...
+def preprocess_benchmarks_for_serialization(benchmarks: list[dataclass]) -> list[dict]:
+    """Preprocess benchmarks to make them JSON serializable."""
+    preprocessed = []
+    for benchmark in benchmarks:
+        benchmark_dict = asdict(benchmark)
+        for key, value in benchmark_dict.items():
+            if isinstance(value, Path):
+                benchmark_dict[key] = str(value)
+        preprocessed.append(benchmark_dict)
+    return preprocessed
```
mihaescuvlad, line 759: Should I document this somewhere?

It seems quite ambiguous currently because one would need to know the context of the task or read through the code to know that the method is only for serializing benchmarks for json

tuner/tuner/libtuner.py
```diff
@@ -390,6 +407,18 @@ def parse_arguments() -> argparse.Namespace:
...
+        choices=[x.value for x in OutputFormats],
+        default=OutputFormats.json,
+        help="The benchmark output file format",
+    )
+    general_args.add_argument(
+        "--benchmark-out-path",
+        type=str,
+        default="bm_output.json",
+        help="The benchmark output file path/name",
+    )
```
mihaescuvlad, line 421: Should these be renamed to contain `_` instead of `-`?
I've noticed that they are named using `_` in other parts of the project

tuner/tuner/libtuner.py
```diff
@@ -347,6 +360,10 @@ class ExecutionPhases(str, Enum):
     benchmark_models = "benchmark-models"
 
 
+class OutputFormats(str, Enum):
+    json = "json"
```
mihaescuvlad, line 364: Should I follow the Google Coding Standards and make this enum's values CapWords as well as the other enums in `libtuner.py`? I didn't use the CapWord style because it seemed out of place compared to the other enums in the libtuner

Also, should we keep the enum? I understood that we will only treat json, but this could help with extending support in the future, should I just remove it because it's not needed?

kuhar: I don't think we need it

tuner/tuner/libtuner.py
kuhar: We shouldn't add flag for this and instead use the right defaults

tuner/tuner/libtuner.py
```diff
@@ -696,6 +725,51 @@ def append_to_file(lines: list[str], filepath: Path, title: str = "") -> None:
         file.write("\n")
 
 
+def get_benchmark_context(benchmarks: Optional[dataclass]) -> dict:
```
kuhar, line 728: `benchmarks` should be the actual class name, not `dataclass`

tuner/tuner/libtuner.py
```diff
@@ -696,6 +725,51 @@ def append_to_file(lines: list[str], filepath: Path, title: str = "") -> None:
...
+    """Gathers system information for the context section of the JSON output."""
+    cpu_info = psutil.cpu_freq()
+    num_cpus = psutil.cpu_count(logical=True)
+
```
kuhar, line 732: We don't need this

tuner/tuner/libtuner.py
```diff
@@ -696,6 +725,51 @@ def append_to_file(lines: list[str], filepath: Path, title: str = "") -> None:
...
+    if benchmark_times:
+        mean_time = sum(benchmark_times) / len(benchmark_times)
+    else:
+        mean_time = None
+
+    context = {
+        "date": datetime.now().isoformat(),
+        "num_cpus": num_cpus,
+        "mhz_per_cpu": cpu_info.max if cpu_info else "N/A",
+        "cpu_scaling_enabled": cpu_info is not None,
```
kuhar, line 744: We don't need these

tuner/tuner/libtuner.py
```diff
@@ -696,6 +725,51 @@ def append_to_file(lines: list[str], filepath: Path, title: str = "") -> None:
...
+    """Gathers system information for the context section of the JSON output."""
+    cpu_info = psutil.cpu_freq()
+    num_cpus = psutil.cpu_count(logical=True)
+
+    benchmark_times = [benchmark.benchmark_time_in_seconds for benchmark in benchmarks]
+
+    if benchmark_times:
+        mean_time = sum(benchmark_times) / len(benchmark_times)
+    else:
+        mean_time = None
```
kuhar, line 738: we can simplify this a bit:
```suggestion
    mean_time = None
    if benchmark_times:
        mean_time = sum(benchmark_times) / len(benchmark_times)
```

tuner/tuner/libtuner.py
```diff
@@ -696,6 +725,51 @@ def append_to_file(lines: list[str], filepath: Path, title: str = "") -> None:
...
+    for benchmark in benchmarks:
+        benchmark_dict = asdict(benchmark)
+        for key, value in benchmark_dict.items():
+            if isinstance(value, Path):
+                benchmark_dict[key] = str(value)
+        preprocessed.append(benchmark_dict)
+    return preprocessed
+
+
+def save_to_json(benchmarks: list[dataclass], filepath: Path) -> None:
```
kuhar, line 762: Is this for testing?

tuner/tuner/libtuner_test.py
```diff
@@ -161,8 +181,18 @@ def test_parse_dispatch_benchmark_results():
...
+
+    with open(path_config.output_json) as file:
+        json_output = json.load(file)
+
+    assert json_output == expected_json_output
+
 
 def test_parse_model_benchmark_results():
+    path_config = libtuner.PathConfig()
+    object.__setattr__(path_config, "output_json", "output.json")
```
kuhar, line 194: Tests shouldn't write files in the current working directory. We can create a temporary directory for tests artifacts if needed.

tuner/tuner/libtuner.py
```diff
@@ -1319,16 +1407,18 @@ def benchmark_models(
             initializer_inputs=(worker_context_queue,),
         )
 
-    dump_list = parse_model_benchmark_results(
-        candidate_trackers, candidate_results, baseline_results
+    parsed_model_results, dump_list = parse_model_benchmark_results(
+        candidate_trackers, candidate_results, baseline_results, path_config
     )
 
+    save_to_json(parsed_model_results, path_config.output_json)
```
kuhar, line 1414: Why do we save the json? I think previously the output wasn't saved to a file?

tuner/requirements-dev.txt
```diff
@@ -1,2 +1,3 @@
 pre-commit==3.8.0
 virtualenv==20.13.0
+psutil==6.0.0
```
kuhar, line 3: What do we need this for?

kuhar (changes requested):

tuner/tuner/libtuner_test.py
```diff
@@ -108,12 +108,15 @@ def test_parse_dispatch_benchmark_results():
...
 
     mock_result_1 = MagicMock()
-    mock_result_1.run_result.process_res.stdout = "process_time/real_time_mean 100.0 us"
+    mock_json_1 = {"benchmarks": [{"real_time": 100.0}]}
+    mock_result_1.run_result.process_res.stdout = json.dumps(mock_json_1)
     mock_result_1.candidate_id = 1
     mock_result_2 = MagicMock()
-    mock_result_2.run_result.process_res.stdout = "process_time/real_time_mean 200.0 us"
+    mock_json_2 = {"benchmarks": [{"real_time": 200.0}]}
```
kuhar, line 118: We should also have a test cast with a smaller number like: `"real_time": 3.4591828516259519e-02`

kuhar (commented):
Nice, this looks good but we should also account for the time unit

tuner/tuner/libtuner_test.py
```diff
@@ -6,11 +6,12 @@
 
 import argparse
 import pytest
-from unittest.mock import call, patch, MagicMock
+import json
+from unittest.mock import ANY, call, patch, MagicMock
```
kuhar, line 10: We don't need `ANY` anymore

tuner/tuner/libtuner.py
```diff
@@ -236,20 +237,59 @@ class ParsedDisptachBenchmarkResult:
 class IREEBenchmarkResult:
     # Default format follows output of iree-benchmark-module
     candidate_id: int
-    result_str: str
+    result_json: list
```
kuhar, line 240: Can we specify the element type?

tuner/tuner/libtuner.py
```diff
@@ -236,20 +237,59 @@ class ParsedDisptachBenchmarkResult:
...
 
     def get_mean_time(self) -> Optional[float]:
```
kuhar, line 242:
```suggestion
    def get_mean_time_us(self) -> Optional[float]:
```

tuner/tuner/libtuner.py
```diff
@@ -236,20 +237,59 @@ class ParsedDisptachBenchmarkResult:
...
-        if not self.result_str:
-            return None
-        pattern = r"process_time/real_time_mean\s+([\d.]+)\s\w{2}"
-        match = re.search(pattern, self.result_str)
-        if not match:
-            return None
-        try:
-            return float(match.group(1))
-        except ValueError:
+        """Compute the mean time for all of the benchmarks - The result will be in ms"""
```
kuhar, line 243: Let's use microseconds. This unit is more realistic for the kind of dispatches we care about

```suggestion
        """Compute the mean time (in microseconds) for all of the benchmarks"""
```

tuner/tuner/libtuner.py
```diff
@@ -236,20 +237,59 @@ class ParsedDisptachBenchmarkResult:
...
+        if not self.result_json:
             return None
 
+        total_time = 0.0
+        count = 0
+
+        for benchmark in self.result_json:
+            real_time = benchmark.get("real_time")
+            time_unit = benchmark.get("time_unit", "ms")
```
kuhar, line 252: What does the second parameter do here?

tuner/tuner/libtuner.py
```diff
@@ -236,20 +237,59 @@ class ParsedDisptachBenchmarkResult:
...
+                        condition=True,
+                        msg=f"Invalid benchmark JSON: {e}",
+                        error_type=ValueError,
+                        exit_program=True,
+                    )
+
+        if count > 0:
+            return total_time / count
+
+        return None
```
kuhar, line 272: We don't have to keep track of the count:
```suggestion
       if not self.result_json:
         return None
         
       return total_time_ms / len(self.result_json)
```

tuner/tuner/libtuner.py
```diff
@@ -236,20 +237,59 @@ class ParsedDisptachBenchmarkResult:
...
+            "ns": 1e-6,
+        }
+
+        if time_unit not in unit_conversions:
+            handle_error(
+                condition=True,
+                msg=f"Unsupported time unit: {time_unit}",
+                error_type=ValueError,
+                exit_program=True,
+            )
```
kuhar, line 289: This can be an assert -- the list of time units should be well known.

tuner/tuner/libtuner.py
```diff
@@ -236,20 +237,59 @@ class ParsedDisptachBenchmarkResult:
...
+                    )
+
+        if count > 0:
+            return total_time / count
+
+        return None
+
+    def convert_time_to_miliseconds(self, real_time: float, time_unit: str) -> float:
+        """Converts the real_time to ms"""
+        unit_conversions = {
```
kuhar, line 276: I'd call this: `unit_to_miliseconds`

tuner/tuner/libtuner.py
```diff
@@ -236,20 +237,59 @@ class ParsedDisptachBenchmarkResult:
...
+                        exit_program=True,
+                    )
+
+        if count > 0:
+            return total_time / count
+
+        return None
+
+    def convert_time_to_miliseconds(self, real_time: float, time_unit: str) -> float:
+        """Converts the real_time to ms"""
```
kuhar, line 275: It's obvious what the function does based on the signature
```suggestion
```

tuner/tuner/libtuner.py
```diff
@@ -236,20 +237,59 @@ class ParsedDisptachBenchmarkResult:
...
+
+        if time_unit not in unit_conversions:
+            handle_error(
+                condition=True,
+                msg=f"Unsupported time unit: {time_unit}",
+                error_type=ValueError,
+                exit_program=True,
+            )
+
+        return float(real_time) * unit_conversions[time_unit]
```
kuhar, line 291: real_time is already a flow, no need to convert

tuner/tuner/libtuner_test.py
kuhar: This hasn't been resolved

tuner/tuner/libtuner.py
mihaescuvlad: I found it defined like this in the documentation: `get(key, default=None).` I assumed it sets a default value in case of a missing field. If needed, I can remove it.

kuhar: The value should always be there, we don't need defaults

tuner/tuner/libtuner.py
```diff
@@ -236,20 +237,48 @@ class ParsedDisptachBenchmarkResult:
 class IREEBenchmarkResult:
     # Default format follows output of iree-benchmark-module
     candidate_id: int
-    result_str: str
+    result_json: list[dict[str, Any]]
```
kuhar, line 240: Could you add a comment explaining what the content is?

tuner/tuner/libtuner.py
```diff
@@ -236,20 +237,48 @@ class ParsedDisptachBenchmarkResult:
...
+                        error_type=ValueError,
+                        exit_program=True,
+                    )
+
+        if not self.result_json:
             return None
 
+        return total_time / len(self.result_json)
+
+    def unit_to_microseconds(self, real_time: float, time_unit: str) -> float:
```
kuhar, line 270: this can be a `@staticmethod`

tuner/tuner/libtuner_test.py
```diff
@@ -57,34 +58,61 @@ def test_collision_handler():
...
-        candidate_id=2,
-        result_str="process_time/real_time_mean 123.45 us, process_time/real_time_mean 246.78 us",
+        candidate_id=7, result_json=invalid_real_time_json
     )
-    assert res.get_mean_time() == 123.45
+    assert res.get_mean_time_us() == 0.0
+
+    # Invalid json: empty dictionary
+    res = libtuner.IREEBenchmarkResult(candidate_id=8, result_json={})
+    assert res.get_mean_time_us() == None
```
kuhar, line 106:
```suggestion
    assert res.get_mean_time_us() is None
```

kuhar (commented):
Overall looks good! Just a few minor comments. I will try it out locally before approving.

tuner/tuner/libtuner.py
```diff
@@ -236,20 +237,52 @@ class ParsedDisptachBenchmarkResult:
...
             return None
-        try:
-            return float(match.group(1))
-        except ValueError:
+
+        total_time = 0.0
+
+        for benchmark in self.result_json:
+            real_time = benchmark.get("real_time")
+            time_unit = benchmark.get("time_unit")
```
kuhar, line 254: We should extract the time only from one result that contains: `"aggregate_name" : "mean"` instead of averaging all entries.

kuhar (changes requested):

kuhar (approved):
I ran this on my machine and it works great. Thank you for this contribution!

PR #257, opened 2024-10-07
Add dumping of reproducer for toy sharded llama test

sogartar:
Add pytest CLI option for overriding the temporary test directory. If the option is provided dump MLIR, VMFB, function call arguments and results. Move numerical checks at the end of the test.
Add some basic functionality to run multi-device IREE functions with tracing.

### DIFF
--- a/sharktank/tests/models/llama/sharded_llama_test.py
+++ b/sharktank/tests/models/llama/sharded_llama_test.py
@@ -5,115 +5,40 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 import unittest
-from typing import Any, List, Tuple, Union, OrderedDict
-import collections.abc
+import pytest
+from typing import Any, List, Tuple, OrderedDict
 from sharktank.models.llama.llama import LlamaModelConfig, PagedLlamaModelV1
 import sharktank.ops as ops
 from sharktank.types import (
     unbox_tensor,
-    ShardedTensor,
-    DefaultPrimitiveTensor,
     Dataset,
-    AnyTensor,
 )
 from sharktank.models.llama.testing import make_random_llama_theta
 from sharktank.models.llama.sharding import shard_theta
 from sharktank.layers.configs import LlamaHParams
 from sharktank.utils.math import round_up_to_multiple_of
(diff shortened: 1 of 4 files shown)


sogartar:
The CI needs some updates to the `iree-turbine` packages.

sharktank/tests/models/llama/sharded_llama_test.py
```diff
@@ -346,91 +286,105 @@ def _(model, *args, **kwargs) -> torch.Tensor:
...
+            vm_context=vm_context,
+            driver=iree_driver,
+            trace_path_prefix=path_prefix if dump_enabled else None,
+        )
+        decode_iree_result = iree_to_torch(*decode_iree_result)
+        expected_decode_result = call_torch_module_function(
+            module=sharded_model,
+            function_name="decode",
+            kwargs=sharded_decode_args,
+            trace_path_prefix=f"{path_prefix}expected_" if dump_enabled else None,
```
rsuderman, line 359: You should use `os.path.join` for combining

sharktank/tests/models/llama/sharded_llama_test.py
```diff
@@ -346,91 +286,105 @@ def _(model, *args, **kwargs) -> torch.Tensor:
...
+        prefill_iree_result = iree_to_torch(*prefill_iree_result)
+        assert len(prefill_iree_result) == 1
+        expected_prefill_result = call_torch_module_function(
+            module=sharded_model,
+            function_name="prefill",
+            kwargs=sharded_prefill_args,
+            trace_path_prefix=f"{path_prefix}expected_" if dump_enabled else None,
+        )
+        prefill_iree_cache_state_shards = prefill_iree_args[
+            -self.config.tensor_parallelism_size - 1 :
```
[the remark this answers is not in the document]
sogartar: The last `self.config.tensor_parallelism_size` number of arguments is the cache state.

sharktank/tests/models/llama/sharded_llama_test.py
sogartar: This prefix is interpreted as a string, not as a directory. This allows to prefix not only with a directory, but a part of the filename as well.

PR #265, opened 2024-10-10
Initial commit of SD inference server.

monorimet:


shortfin/python/shortfin_apps/sd/components/service.py
```diff
@@ -0,0 +1,330 @@
...
+            f"  page_cache={self.page_cache}\n"
+            f")"
+        )
+
+
+########################################################################################
+# Batcher
+########################################################################################
+
+import math
```
stellaraccident, line 91: Import at the top

shortfin/python/shortfin_apps/sd/components/service.py
```diff
@@ -0,0 +1,330 @@
...
+                self.submit(StrobeMessage())
+
+    async def run(self):
+        strober_task = asyncio.create_task(self._background_strober())
+        reader = self.batcher_infeed.reader()
+        while item := await reader():
+            self.strobe_enabled = False
+            if isinstance(item, InferenceExecRequest):
+                phase = item.phase
+                if phase == InferencePhase.PREP:
```
stellaraccident, line 139: I'm not entirely sure we need to do fine grained batching like this for this model. You could just have one exec process that said:

```
if req.needs_prep:
  await do_something()

if req.needs_encode:
  await do_something()
```

That would put some constraints on basically the batch needing to be dense across all steps but that is probably ok.  Then the batcher here is really just about lumping multiple client requests together. And if they happen to be disjoint (i.e. some need prep and others don't), then you are potentially inefficient. But I'm not sure that's really a valid case. If it is, you could just group likes together here vs enforcing the structure everywhere.

shortfin/python/shortfin_apps/sd/components/service.py
```diff
@@ -0,0 +1,330 @@
...
+    def board_postprocesses():
+        pass
+
+
+########################################################################################
+# Inference Executors
+########################################################################################
+
+
+class InferenceExecProcessPrep(sf.Process):
```
stellaraccident, line 213: If you took my advice above, there would just be one InferenceExec and it would do if req.need_*: inside.

stellaraccident (commented):
Starting to look reasonable. I left one comment of substance/design to consider.

shortfin/python/shortfin_apps/sd/components/service.py
monorimet: I think this would mean we get rid of the style of resetting the request with a given inference phase, instead, we write program outputs directly to the request (not its payload attribute) and let the request keep track of its own completion with the `needs_<job>`?

monorimet: thought about this a bit more...

Regarding disjoint requests, while I don't see a client batch coming with disjoint phases as a common thing, I do imagine that a single client request may end up being disjoint in that some inference phases will be compatible for a batched inference process and others will not.
For example, we recieve a client request with 8 individual image generations which all have the same inference "parameters", i.e., they should run as batch size 8 and be coupled throughout inference execution, except they all ask for a different number of unet iterations.
This would mean that the denoising phase needs to be distributed among different program invocations, even if we do something fancy for the denoising loop and break it up into one program invocation per unet iteration. In this example, that might look like:
Do bs8 unet until one request's steps are completed
Do bs7 "
and so on...
which seems difficult unless we get really flexible about batch sizes in SD compiled modules.
Assuming we want to share work where possible, i.e., to be able to join/unjoin batches at different phases of the image generation process, is there a better way to go about it besides doing the fine-grained (phased) batching? I think we would lose quite a bit of efficiency if we had to split up batches just because they are incompatible for some subsection of the model structure -- generation parameters like inference steps, image output size, and CFG-free mode would all be examples of request parameterizations that require different programs at some point, but could be batched for other parts of the image generation process.

shortfin/python/shortfin_apps/sd/components/generate.py
```diff
@@ -0,0 +1,146 @@
...
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import asyncio
+import io
+import logging
+
+import shortfin as sf
+import shortfin.array as sfnp
+
+# TODO: Have a generic "Responder" interface vs just the concrete impl.
```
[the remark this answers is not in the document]
monorimet: https://github.com/nod-ai/SHARK-Platform/issues/272

shortfin/python/shortfin_apps/sd/components/service.py
monorimet: I think regardless of whether we do the finer-grained batching, it might not be best practice to split these processes up into components. I think it makes the implementation a bit tighter -- this single process structure accounts for any reasonable extension of the SD pipeline, barring expansion into significantly larger workflows like creative upscaling, but at that point we'd be out of implementation scope, I'd say. For now.

monorimet:
A few notes:
- I took a transformers dep when setting up the CLIP tokenizers -- perhaps this can be worked around at some point.
- Kept the fine batching of inference phases, but that structure can be adapted easily from here if we need to tweak it.
- Inference code just has all the plumbing and won't actually do inference yet -- this will take me a few hours yet
- README lists diffusers req, this may not be needed
- Batcher needs to be a bit smarter to make use of the batched exec/request structure. Current impl assumes dense batches for client requests.

monorimet:
NOTE: currently the serving forces use of the first fiber in the `GenerateService`'s `fibers` attribute. Plumbing that fiber selection is needed.

shortfin/python/shortfin_apps/sd/components/tokenizer.py
```diff
@@ -0,0 +1,85 @@
+# Copyright 2024 Advanced Micro Devices, Inc.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+from pathlib import Path
+
+from transformers import CLIPTokenizer, BatchEncoding
```
stellaraccident, line 9: Yeah, we're going to want to see how to not need this. But whatever it takes for now.

shortfin/python/shortfin_apps/sd/components/service.py
```diff
@@ -0,0 +1,609 @@
...
+            req.image_array = np.frombuffer(image_array, dtype=dtype).reshape(
+                *image_shape
+            )
+        return
+
+    async def _postprocess(self, device, requests):
+        # Process output images
+        for req in requests:
+            # TODO: reimpl with sfnp
+            permuted = (
```
stellaraccident, line 600: Can you file me issues to create host ops for this stuff? Easy.

shortfin/python/shortfin_apps/sd/components/generate.py
```diff
@@ -0,0 +1,132 @@
...
+            self.gen_req.steps[self.index],
+            self.gen_req.guidance_scale[self.index],
+            self.gen_req.seed[self.index],
+        )
+        self.client.batcher.submit(exec)
+        await exec.done
+
+        exec.reset(InferencePhase.ENCODE)
+        self.client.batcher.submit(exec)
+        await exec.done
```
stellaraccident, line 63: I still think we have this broken up into way too many phases at the high level here. It will block a lot of pipelining to do it this way. It would be better to add the needed phases to one request object and submit it at once: then have the executor perform each step in sequence vs batching/waiting for each step on the outside like this.

Fine to submit as-is but something we will want to invert.

stellaraccident (approved):
A lot we can make better here, but would rather land/iterate and get some CI on it before doing more surgery.

Thanks for the first cut.

shortfin/python/shortfin_apps/sd/components/generate.py
monorimet: I agree! This is much clearer now that it’s laid out and loading/batching enhancements become more tangible.

I will follow up with a separate patch

PR #266, opened 2024-10-11
[tuner] Fix mfma constructor arguments

kuhar:


kuhar:
No test because I don't want to either mock the solver or assert a specific solver output (which is generally not stable across z3 releases)

saienduri (approved):

PR #267, opened 2024-10-11
Refactor llama / mixtral / grok for shared features

rsuderman:
Many of these features can toggle between depending on architecture. Replumbing the configurations separately allows better reuse and understanding of how models vary between eachother.

grok uses a softcap, plumbing a value enables `sc * tanh( v / sc)` grok has some hardcoded values that have better representations, e.g. `sqrt(6144)` and `sqrt(3)`.

output normalization is optional but used by mixtral. Presence of the tensor is sufficient for performing the normalization.

### DIFF
--- a/sharktank/sharktank/export_layer/export_moe.py
+++ b/sharktank/sharktank/export_layer/export_moe.py
@@ -10,7 +10,7 @@ import torch.nn.functional as F
 from iree.turbine.aot import *
 
 from sharktank.models.llama.testing import make_moe_block_theta, make_rand_torch
-from sharktank.layers.mixture_of_experts_block import PreGatherMoeBlock
+from sharktank.layers.mixture_of_experts_block import MoeBlock
 from ..utils import cli
 
 
@@ -49,7 +49,7 @@ def main():
 
     bs = args.batch_size
 
-    model = PreGatherMoeBlock(
+    model = MoeBlock(
         theta=make_moe_block_theta()("blk.0"),
         expert_count=8,
         expert_used_count=2,
--- a/sharktank/sharktank/export_layer/export_moe.py
+++ b/sharktank/sharktank/export_layer/export_moe.py
@@ -5,7 +5,10 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 import torch
+import torch.nn.functional as F
+
 from iree.turbine.aot import *
+
 from sharktank.models.llama.testing import make_moe_block_theta, make_rand_torch
 from sharktank.layers.mixture_of_experts_block import PreGatherMoeBlock
 from ..utils import cli
@@ -51,7 +54,7 @@ def main():
         expert_count=8,
         expert_used_count=2,
         rms_epsilon=1e-5,
-        use_grok=args.use_grok,
+        moe_activation=F.gelu if args.use_grok else F.silu,
     )
     fxb = FxProgramsBuilder(model)
     input = make_rand_torch((bs, 32, 6144))
--- a/sharktank/sharktank/export_layer/export_moe.py
+++ b/sharktank/sharktank/export_layer/export_moe.py
@@ -40,8 +40,8 @@ def main():
         action="store_true",
     )
     parser.add_argument(
-        "--use-grok",
-        help="Enable to export Grok model's version of MOE block",
+        "--use-gelu",
+        help="Enable to use gelu for moe activation",
         action="store_true",
     )
 
@@ -54,7 +54,7 @@ def main():
         expert_count=8,
         expert_used_count=2,
         rms_epsilon=1e-5,
-        moe_activation=F.gelu if args.use_grok else F.silu,
+        moe_activation=F.gelu if args.use_gelu else F.silu,
     )
     fxb = FxProgramsBuilder(model)
     input = make_rand_torch((bs, 32, 6144))
--- a/sharktank/sharktank/layers/paged_llama_attention_block.py
+++ b/sharktank/sharktank/layers/paged_llama_attention_block.py
@@ -37,7 +37,8 @@ class PagedLlamaAttentionBlock(ThetaLayer):
         head_dim: int,
         head_count_kv: int,
         rms_epsilon: float,
-        use_grok: Optional[bool] = False,
+        attention_scale: Optional[float] = None,
+        softcap: Optional[float] = None,
     ):
         super().__init__(theta)
 
@@ -46,7 +47,8 @@ class PagedLlamaAttentionBlock(ThetaLayer):
         self.head_count = head_count
         self.head_dim = head_dim
         self.head_count_kv = head_count_kv
-        self.use_grok = use_grok
+        self.attention_scale = attention_scale
+        self.softcap = softcap
 
         self.add_module(
             "attn_norm", RMSNormLayer(theta("attn_norm"), epsilon=rms_epsilon)
@@ -56,7 +58,12 @@ class PagedLlamaAttentionBlock(ThetaLayer):
         self.add_module("attn_v", LinearLayer(theta("attn_v")))
         self.add_module("attn_output", LinearLayer(theta("attn_output")))
 
-        if self.use_grok:
+        if theta.optional_tensor("attn_output_norm") is None:
+            self.add_module(
+                "attn_output_norm",
+                torch.nn.Identity(),
+            )
+        else:
             self.add_module(
                 "attn_output_norm",
                 RMSNormLayer(theta("attn_output_norm"), epsilon=rms_epsilon),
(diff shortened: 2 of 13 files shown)


sharktank/sharktank/models/grok/grok.py
```diff
@@ -93,16 +95,15 @@ def __init__(self, theta: Theta, config: LlamaModelConfig):
...
-            self.attn_blocks.append(
-                PreGatherMoeBlock(
+            self.moe_blocks.append(
+                MoeBlock(
                     theta("blk", n),
                     expert_count=hp.expert_count,
                     expert_used_count=hp.expert_used_count,
                     rms_epsilon=hp.attention_layer_norm_rms_epsilon,
-                    use_grok=True,
+                    activation=F.gelu,
```
archana-ramalingam, line 106: Should this be `moe_activation=F.gelu` as defined in the `MoeBlock`?

sharktank/sharktank/models/grok/grok.py
```diff
@@ -122,33 +123,32 @@ def prefill(
         self._assert_device(seq_block_ids)
         self._assert_device(*cache_state, dtype=self.activation_dtype)
         h = self.token_embedding(tokens)
-        h *= 78.38367176906169
+        h *= math.sqrt(h.shape[-1])
```
archana-ramalingam, line 126: This requires `import math`

dan-garvey: better as torch? I don't think math traces well

IanNod: Do we not have prefill here testing export? If not that needs to be added

sharktank/sharktank/models/grok/grok.py
```diff
@@ -200,34 +200,34 @@ def decode(
...
+                h,
+                start_positions=start_positions,
+                embedding=self.attention_embedding,
+                embedding_batch_mask=embedding_batch_mask,
+                attention_mask=attention_mask,
+                cache_state=cache_state,
+                seq_block_ids=seq_block_ids,
+                xk_temp=xk_temp,
+                xv_temp=xv_temp,
+                softcap=30.0,
```
IanNod, line 223: would be good to have comments that identifying where these hard coded numbers come from

KyleHerndon: Would you want like a link to the llama.cpp or grok model github line? We're literally just copying from the publicly available implementation of grok. There's some logic relating to quantization behind it, but we didn't derive it ourselves so I don't know that we could give a "real" explanation.

sharktank/sharktank/models/grok/grok.py
rsuderman: Fixed.

sharktank/sharktank/models/grok/grok.py
rsuderman: We should try to make faked versions of each model that is locally exportable. @KyleHerndon can you look into making faked `theta` parameters for the model? It can just be a single layer with smallers tensors.

sharktank/sharktank/export_layer/export_moe.py
```diff
@@ -46,12 +49,12 @@ def main():
...
     bs = args.batch_size
 
-    model = PreGatherMoeBlock(
+    model = MoeBlock(
         theta=make_moe_block_theta()("blk.0"),
         expert_count=8,
         expert_used_count=2,
         rms_epsilon=1e-5,
-        use_grok=args.use_grok,
+        moe_activation=F.gelu if args.use_grok else F.silu,
```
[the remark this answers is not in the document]
rsuderman: Done.

sharktank/sharktank/models/grok/grok.py
rsuderman: Added a link to where we pull the hardcoded from.

IanNod: Yeah, link to the public implementations would be good. Just so someone not familiar with our codebase has some clue as to where these are coming from

sharktank/sharktank/models/grok/grok.py
```diff
@@ -122,33 +123,32 @@ def prefill(
         self._assert_device(seq_block_ids)
         self._assert_device(*cache_state, dtype=self.activation_dtype)
         h = self.token_embedding(tokens)
-        h *= 78.38367176906169
+        h *= tanh.sqrt(h.shape[-1])
```
archana-ramalingam, line 126: You mean `torch.tanh`?

sharktank/sharktank/models/grok/grok.py
```diff
@@ -123,7 +123,7 @@ def prefill(
         self._assert_device(seq_block_ids)
         self._assert_device(*cache_state, dtype=self.activation_dtype)
         h = self.token_embedding(tokens)
-        h *= math.sqrt(h.shape[-1])
+        h *= tanh.sqrt(h.shape[-1])
```
archana-ramalingam, line 126: You mean `torch.sqrt()`

archana-ramalingam: Also, everywhere we use `torch.sqrt()`, we need to make sure the input is in tensor form and not an `int`.

sharktank/sharktank/layers/paged_llama_attention_block.py
```diff
@@ -156,7 +156,7 @@ def repeat_kv(x: torch.Tensor) -> torch.Tensor:
 
         attn_weights = ops.matmul(xq, keys.transpose(2, 3))
         if self.attention_scale is None:
-            attn_weights = attn_weights / math.sqrt(self.head_dim)
+            attn_weights = attn_weights / torch.sqrt(self.head_dim)
```
archana-ramalingam, line 159: Remove `import math` as it's unused now.

sharktank/sharktank/models/grok/grok.py
archana-ramalingam: [Here](https://github.com/nod-ai/SHARK-Platform/blob/main/sharktank/sharktank/models/llama/testing.py) are the theta generators for attention/FFN/MOE blocks. We already have attention and prefill tests, might need to add export component in there.

sharktank/sharktank/models/grok/grok.py
```diff
@@ -200,34 +200,34 @@ def decode(
...
 
         h = self.token_embedding(tokens)
-        h *= 78.38367176906169
+        h *= torch.sqrt(h.shape[-1])
         self.trace_tensor("grok.token_embedding", h)
 
         # Iterate over attention blocks.
-        for block_idx, block in enumerate(self.attn_blocks):
+        for block_idx, (attn_block, moe_block) in enumerate(
+            self.attn_blocks, self.moe_blocks
```
archana-ramalingam, line 208: ` zip(self.attn_blocks, self.moe_blocks)`

sharktank/sharktank/layers/paged_llama_attention_block.py
```diff
@@ -147,16 +154,16 @@ def repeat_kv(x: torch.Tensor) -> torch.Tensor:
...
+
         # Flash attention.
-        if not self.use_grok:
-            attn_weights = ops.matmul(xq, keys.transpose(2, 3)) / math.sqrt(
-                self.head_dim
-            )
-        elif self.use_grok:
-            attn_weights = ops.matmul(xq, keys.transpose(2, 3))
-            attn_weights = 30.0 * torch.tanh(
-                attn_weights * (0.08838834764831845 / 30.0)
```
archana-ramalingam, line 158: Should we also hardcode the `attention_scale = 0.08838834764831845` in `grok.py`?

sharktank/sharktank/models/grok/grok.py
```diff
@@ -200,34 +200,34 @@ def decode(
...
+                h,
+                start_positions=start_positions,
+                embedding=self.attention_embedding,
+                embedding_batch_mask=embedding_batch_mask,
+                attention_mask=attention_mask,
+                cache_state=cache_state,
+                seq_block_ids=seq_block_ids,
+                xk_temp=xk_temp,
+                xv_temp=xv_temp,
+                softcap=30.0,  # https://github.com/xai-org/grok-1/blob/7050ed204b8206bb8645c7b7bbef7252f79561b0/model.py#L864
```
archana-ramalingam, line 223: Does `softcap` belong to `PagedLlamaAttentionBlock()` and not `decode()`.

rsuderman: Whoops, should have put it in the initializer. Fixed.

sharktank/sharktank/layers/paged_llama_attention_block.py
rsuderman: So the secret is `0.08838834764831845` is not a random hardcoded value, its actually `1.0 / sqrt(128)` and for `grok` `head_dim == 128`. So that normalization is already handled by the default attention scaling above.

sharktank/sharktank/models/grok/grok.py
rsuderman: changed back to `math.sqrt`. For any of the sqrt of dims we need to use the math variant. All of these are constants though so this should be fine.

sharktank/sharktank/models/grok/grok.py
rsuderman: Fixed.

IanNod (approved):
Looks good to me

archana-ramalingam (approved):

PR #269, opened 2024-10-11
Add ci_windows_x64-libshortfin.yml

marbre:
Adds a Windows CI for shortfin. Building passes (with several warnings) but not all ctests and pytests do. As soon as we have passing tests, the CI configuration should be merged with the Linux CI into a matrix configuration.

### DIFF
--- a/.github/workflows/ci_windows_x64-libshortfin.yml
+++ b/.github/workflows/ci_windows_x64-libshortfin.yml
@@ -1,0 +1,95 @@
+# Copyright 2024 Advanced Micro Devices, Inc.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+name: CI - shortfin - Windows
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+      - main
+    paths:
+      - '.github/workflows/ci_windows_x64-libshortfin.yml'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_windows_x64-libshortfin.yml
```diff
@@ -0,0 +1,95 @@
...
+        ref: candidate-20240904.1006
+
+    - name: Initalize IREE submodules
+      working-directory: ${{ env.IREE_REPO_DIR }}
+      run : |
+        git submodule update --init --depth 1 -- third_party/benchmark
+        git submodule update --init --depth 1 -- third_party/cpuinfo/
+        git submodule update --init --depth 1 -- third_party/flatcc
+        git submodule update --init --depth 1 -- third_party/googletest
+        git submodule update --init --depth 1 -- third_party/hip-build-deps/
```
ScottTodd, line 63: > Building passes (with several warnings) but not all ctests and pytests do. As soon as we have passing tests, the CI configuration should be merged with the Linux CI into a matrix configuration.

This is fine for now, but we should definitely de-dup using a matrix (across at least operating system and Python version), so we aren't copy/pasting all this boilerplate. We can xfail or skip tests based on operating system using a few different code patterns (see https://docs.pytest.org/en/stable/how-to/skipping.html for some)

ScottTodd (approved):

.github/workflows/ci_windows_x64-libshortfin.yml
marbre: Yes, I fully agree.

PR #274, opened 2024-10-11
[sharktank] Add test for sharded rotary table

rsuderman:
We should be able to validate the sharded rotary table via comparison with the unsharded version. This runs the sharded and unsharded implementations, asserting near identical results.

### DIFF
--- a/sharktank/tests/layers/sharded_rotary_embedding_test.py
+++ b/sharktank/tests/layers/sharded_rotary_embedding_test.py
@@ -1,0 +1,58 @@
+# Copyright 2024 Advanced Micro Devices, Inc.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+
+import torch
+
+from sharktank.layers import RotaryEmbeddingLayer
+from sharktank import ops
+from sharktank.types import (
+    ShardedTensor,
+    SplitPrimitiveTensor,
+    unbox_tensor,
+)
+
+import unittest
+from typing import List, Optional
+import os
+
+
+def test_sharded_rotary_table():
+    bs = 4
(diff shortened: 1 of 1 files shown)


sharktank/tests/layers/sharded_rotary_embedding_test.py
```diff
@@ -0,0 +1,58 @@
...
+        rope_dimension_count=rope_dims,
+        max_seqlen=max_seqlen,
+        rope_freq_base=rope_freq_base,
+        tensor_parallelism_size=4,
+    )
+    sq, sk = shard_layer(xq=xq, xk=xk, start_index=0)
+
+    # Gathering and unboxing should yield the same results
+    sq = ops.all_gather(sq)
+    sk = ops.all_gather(sk)
```
sogartar, line 53: We actually have `ops.unshard` that would return the corresponding unsharded tensor.

At some point may be worth implementing unbox_tensor for sharded types through it.

sogartar (approved):

sharktank/tests/layers/sharded_rotary_embedding_test.py
rsuderman: We should avoid throwing more uses of `unbox_tensor` in general. I mostly used it in this case to quickly throw everthing together. `ops.unshard` is the right path in this case.

sharktank/tests/layers/sharded_rotary_embedding_test.py
```diff
@@ -0,0 +1,56 @@
...
+
+def test_sharded_rotary_table():
+    bs = 4
+    rope_dims = 16
+    heads = 8
+    max_seqlen = 128
+    rope_freq_base = None
+
+    # First we setup and get the default rotary embedding layer
+    xq = torch.rand((bs, max_seqlen, heads, rope_dims), dtype=torch.float)
```
sogartar, line 31: You may seed the torch RNG to make sure we get deterministic test results.

PR #275, opened 2024-10-11
Add device selection to shortfin llm demo

renxida:


renxida:
good idea. working it into the pytest infra.

PR #276, opened 2024-10-14
Add sharded paged attention test

sogartar:
Verify that the sharded Llama paged attention block behaves in PyTorch as the unsharded variant.
The fp32 accuracy seems low and this test is xfailed. The fp64 accuracy is fine.


sharktank/sharktank/layers/sharding.py
```diff
@@ -0,0 +1,38 @@
...
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+"""Specifications describing how layers are sharded."""
+
+from ..types.sharding import *
+
+
+class PagedLlamaAttentionBlockSharding(ThetaLayerSharding):
```
rsuderman, line 12: This should stay in `models` as its more of a model specific sharding.

sharktank/tests/layers/sharded_paged_llama_attention_block.py
```diff
@@ -0,0 +1,162 @@
...
+            head_count_kv=self.head_count_kv,
+            rms_epsilon=self.rms_epsilon,
+        )
+        embedding_module = RotaryEmbeddingLayer(
+            rope_dimension_count=self.rope_dimension_count,
+            max_seqlen=self.max_seqlen,
+            rope_freq_base=self.rope_freq_base,
+        )
+
+        input_tensor = make_rand_torch(
```
rsuderman, line 108: Declare your inputs before creating each layer.

rsuderman (approved):

sharktank/sharktank/layers/sharding.py
sogartar: I moved it back.

sharktank/tests/layers/sharded_paged_llama_attention_block.py
sogartar: I moved them.

PR #280, opened 2024-10-15
[llama] Update kv cache to have read/write functions

rsuderman:
Made the interfaces of both caches line up. This allows us to interface with the caches via their utility functions instead of modifying the model behavior. Some roughness still exists in their parameters but the irrelevant details are ignored for each implementation.

Need to still add some slicing / ignoring on the page ids to make more flexible.

### DIFF
--- a/sharktank/sharktank/layers/kv_cache.py
+++ b/sharktank/sharktank/layers/kv_cache.py
@@ -115,19 +115,27 @@ class DirectKVCache(BaseKVCache):
 
         Each tensor has shape: [bs, sl, attn_head_count, attn_head_dim]
         """
-        shards = [[torch.empty(
-                [bs, self.seq_length, self.attn_head_count // self.shard_count, self.attn_head_dim],
+        allocations = [
+            torch.empty(
+                [
+                    bs,
+                    self.seq_length,
+                    self.attn_head_count,
+                    self.attn_head_dim,
+                ],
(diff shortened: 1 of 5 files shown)


sharktank/sharktank/layers/kv_cache.py
```diff
@@ -113,15 +115,109 @@ def allocate(self, *, bs: int) -> list[torch.Tensor]:
...
+        return [
+            ops.reshard_split(allocation, dim=2, count=self.shard_count)
+            for allocation in allocations
+        ]
+
+    def read(
+        self,
+        state: list[Union[torch.Tensor, SplitPrimitiveTensor]],
+        *,
+        read_into_partitions: list[Union[torch.Tensor, SplitPrimitiveTensor]],
```
sogartar, line 144: Why are we not reading into this?

sogartar (changes requested):

dan-garvey:
We needed this so badly, thanks Rob

sharktank/sharktank/layers/kv_cache.py
rsuderman: There is no value in placing in a destination block. Ultimately this should fuse with the attention block (this is the premise of paged attention fusion)

dan-garvey (approved):
LGTM

sogartar (approved):

PR #283, opened 2024-10-16
Ignore dir containing perplexity test from presubmit workflow

IanNod:
Perplexity is to run nightly but being pulled into presubmit CI. Adding ignore here to fix

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,4 +60,4 @@ jobs:
       - name: Run sharktank tests
         if: ${{ !cancelled() }}
         run: |
-          pytest -n 4 sharktank/
+          pytest -n 4 sharktank/ --ignore=sharktank/tests/evaluate/


.github/workflows/ci.yaml
```diff
@@ -60,4 +60,4 @@ jobs:
       - name: Run sharktank tests
         if: ${{ !cancelled() }}
         run: |
-          pytest -n 4 sharktank/
+          pytest -n 4 sharktank/ --ignore=sharktank/tests/evaluate/
```
ScottTodd, line 63: We should find another fix instead of patching this github actions workflow. The new tests break local testing workflows too, and developers shouldn't need to pass this extra pytest option or do other manual setup to run unit tests.

Simpler to just revert https://github.com/nod-ai/SHARK-Platform/pull/233 to get the CI passing again. We should drive that towards always passing and flip it to a required check.

IanNod: https://github.com/nod-ai/SHARK-Platform/pull/233 is intended as a nightly test as it will be very slow, so we won't want that running on the presubmit ci. Good point on breaking local testing workflows though, should we just move the new test to some separate dir instead?

archana-ramalingam: Reverting here - https://github.com/nod-ai/SHARK-Platform/pull/285

ScottTodd: Some ideas:
* https://stackoverflow.com/a/43938191
* https://docs.pytest.org/en/stable/how-to/mark.html
* https://pypi.org/project/pytest-manual-marker/

(first option looks nice)

PR #284, opened 2024-10-16
Benchmark Llama 3.1 f16 and fp8 with CI

aviator19941:
Adds pytests for f16 and fp8 with the CI. Currently Llama 3.1 8B f16 is the only test that fully benchmarks through. Llama 3.1 8B fp8, Llama 3.1 70B f16/fp8, and Llama 3.1 405B f16/fp8 tests are marked as XFAIL for now.

### DIFF
--- a/.github/workflows/ci-llama.yaml
+++ b/.github/workflows/ci-llama.yaml
@@ -28,6 +28,8 @@ jobs:
         shell: bash
     env:
       PIP_CACHE_DIR: "${{ github.workspace }}/.pip-cache"
+      VENV_DIR: ${{ github.workspace }}/.venv
+      SHARK_PLATFORM_REPO_ROOT: ${{ github.workspace }}
     steps:
       - name: "Setting up Python"
         id: setup_python
@@ -47,8 +49,6 @@ jobs:
 
       - name: Install pip deps
         run: |
-          python -m venv --prompt sharktank .venv
-          source .venv/bin/activate
           python -m pip install --no-compile --upgrade pip
           # Note: We install in three steps in order to satisfy requirements
           # from non default locations first. Installing the PyTorch CPU
--- a/.github/workflows/ci-llama.yaml
+++ b/.github/workflows/ci-llama.yaml
@@ -1,4 +1,4 @@
-name: Integration Tests
+name: Llama Benchmarking Tests
 
 on:
   workflow_dispatch:
@@ -16,11 +16,11 @@ concurrency:
 
 jobs:
   test_llama:
-    name: "Integration Tests - llama"
+    name: "Llama Benchmarking Tests"
     strategy:
       matrix:
         version: [3.11]
-        os: [ubuntu-latest, windows-latest]
+        os: [llama-mi300]
       fail-fast: false
     runs-on: ${{matrix.os}}
     defaults:
@@ -68,3 +68,9 @@ jobs:
 
       - name: Run llama test
         run: pytest sharktank/tests/models/llama/benchmark-tests.py
+
+      - name: Upload llama executable files
+        uses: actions/upload-artifact@v4
+        with:
+          name: llama-files
+          path: ${{ github.workspace }}/files/
--- a/.github/workflows/ci-llama.yaml
+++ b/.github/workflows/ci-llama.yaml
@@ -1,0 +1,70 @@
+name: Integration Tests
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Weekdays at 13:00 UTC = 05:00 PST / 06:00 PDT.
+    - cron: "5 4 * * 1-5"
+
+concurrency:
+  # A PR number if a pull request and otherwise the commit hash. This cancels
+  # queued and in-progress runs for the same PR (presubmit) or commit
+  # (postsubmit). The workflow name is prepended to avoid conflicts between
+  # different workflows.
+  group: ${{ github.workflow }}-${{ github.event.number || github.sha }}
+  cancel-in-progress: true
+
+jobs:
+  test_llama:
+    name: "Integration Tests - llama"
+    strategy:
+      matrix:
+        version: [3.11]
+        os: [ubuntu-latest, windows-latest]
+      fail-fast: false
+    runs-on: ${{matrix.os}}
+    defaults:
+      run:
+        shell: bash
+    env:
+      PIP_CACHE_DIR: "${{ github.workspace }}/.pip-cache"
+    steps:
+      - name: "Setting up Python"
+        id: setup_python
+        uses: actions/setup-python@v3
+        with:
+          python-version: ${{matrix.version}}
+
+      - name: "Checkout Code"
+        uses: actions/checkout@v3
+
+      - name: Cache Pip Packages
+        uses: actions/cache@v4
+        id: cache-pip
+        with:
+          path: ${{ env.PIP_CACHE_DIR }}
+          key: pip-${{ steps.setup_python.outputs.python-version }}-${{ hashFiles('*requirements.txt') }}
+
+      - name: Install pip deps
+        run: |
+          python -m venv --prompt sharktank .venv
+          source .venv/bin/activate
+          python -m pip install --no-compile --upgrade pip
+          # Note: We install in three steps in order to satisfy requirements
+          # from non default locations first. Installing the PyTorch CPU
+          # wheels saves multiple minutes and a lot of bandwidth on runner setup.
+          pip install --no-compile -r pytorch-cpu-requirements.txt
+          pip install --no-compile -f https://iree.dev/pip-release-links.html --src deps \
+            -e "git+https://github.com/iree-org/iree-turbine.git#egg=iree-turbine"
+          pip install --no-compile -r requirements.txt -e sharktank/
+
+          # Try with the latest nightly releases, not what iree-turbine pins.
+          # We could also pin to a known working or stable version.
+          # This should eventually stabilize. Do the best we can for now.
+          pip install -f https://iree.dev/pip-release-links.html --upgrade \
+            iree-compiler \
+            iree-runtime \
+            "numpy<2.0"
+
+      - name: Run llama test
+        run: pytest sharktank/tests/models/llama/benchmark-tests.py
--- a/.github/workflows/ci-llama.yaml
+++ b/.github/workflows/ci-llama.yaml
@@ -2,7 +2,6 @@ name: Llama Benchmarking Tests
 
 on:
   workflow_dispatch:
-  pull_request:
   schedule:
     # Weekdays at 9:00 AM UTC = 2:00 AM PST.
     - cron: "0 9 * * 1-5"
--- a/.github/workflows/ci-llama.yaml
+++ b/.github/workflows/ci-llama.yaml
@@ -2,6 +2,7 @@ name: Llama Benchmarking Tests
 
 on:
   workflow_dispatch:
+  pull_request:
   schedule:
     # Weekdays at 9:00 AM UTC = 2:00 AM PST.
     - cron: "0 9 * * 1-5"
--- a/.github/workflows/ci-llama.yaml
+++ b/.github/workflows/ci-llama.yaml
@@ -3,8 +3,8 @@ name: Llama Benchmarking Tests
 on:
   workflow_dispatch:
   schedule:
-    # Weekdays at 13:00 UTC = 05:00 PST / 06:00 PDT.
-    - cron: "5 4 * * 1-5"
+    # Weekdays at 12:00 AM UTC = 05:00 PST / 06:00 PDT.
+    - cron: "0 0 * * 1-5"
 
 concurrency:
   # A PR number if a pull request and otherwise the commit hash. This cancels
@@ -67,7 +67,7 @@ jobs:
             "numpy<2.0"
 
       - name: Run llama test
-        run: pytest sharktank/tests/models/llama/benchmark-tests.py
+        run: pytest sharktank/tests/models/llama/benchmark-tests.py -v -s --longrun
 
       - name: Upload llama executable files
         uses: actions/upload-artifact@v4
--- a/.github/workflows/ci-llama.yaml
+++ b/.github/workflows/ci-llama.yaml
@@ -21,9 +21,8 @@ jobs:
     strategy:
       matrix:
         version: [3.11]
-        os: [llama-mi300]
       fail-fast: false
-    runs-on: ${{matrix.runs-on}}
+    runs-on: llama-mi300
     defaults:
       run:
         shell: bash
--- a/.github/workflows/ci-llama.yaml
+++ b/.github/workflows/ci-llama.yaml
@@ -32,6 +32,10 @@ jobs:
       VENV_DIR: ${{ github.workspace }}/.venv
       SHARK_PLATFORM_REPO_ROOT: ${{ github.workspace }}
     steps:
+      - name: Get Current Date
+        id: date
+        run: echo "::set-output name=date::$(date +'%Y-%m-%d')"
+
       - name: "Setting up Python"
         id: setup_python
         uses: actions/setup-python@v3
@@ -74,4 +78,4 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: llama-files
-          path: ${{ github.workspace }}/files/
+          path: ${{ github.workspace }}/${{ steps.date.outputs.date }}
--- a/.github/workflows/ci-llama.yaml
+++ b/.github/workflows/ci-llama.yaml
@@ -23,14 +23,13 @@ jobs:
         version: [3.11]
         os: [llama-mi300]
       fail-fast: false
-    runs-on: ${{matrix.os}}
+    runs-on: ${{matrix.runs-on}}
     defaults:
       run:
         shell: bash
     env:
       PIP_CACHE_DIR: "${{ github.workspace }}/.pip-cache"
       VENV_DIR: ${{ github.workspace }}/.venv
-      SHARK_PLATFORM_REPO_ROOT: ${{ github.workspace }}
     steps:
       - name: Get Current Date
         id: date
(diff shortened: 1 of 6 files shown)


.github/workflows/ci-llama.yaml
```diff
@@ -0,0 +1,70 @@
+name: Integration Tests
+
+on:
```
saienduri, line 3: Do we want to run on every PR?

.github/workflows/ci-llama.yaml
```diff
@@ -0,0 +1,70 @@
...
+  group: ${{ github.workflow }}-${{ github.event.number || github.sha }}
+  cancel-in-progress: true
+
+jobs:
+  test_llama:
+    name: "Integration Tests - llama"
+    strategy:
+      matrix:
+        version: [3.11]
+        os: [ubuntu-latest, windows-latest]
```
saienduri, line 23: This is where we determine what machine to run on. Change to `llama-mi300`. That will bring in the 30F runner

.github/workflows/ci-llama.yaml
```diff
@@ -0,0 +1,70 @@
...
+          # Try with the latest nightly releases, not what iree-turbine pins.
+          # We could also pin to a known working or stable version.
+          # This should eventually stabilize. Do the best we can for now.
+          pip install -f https://iree.dev/pip-release-links.html --upgrade \
+            iree-compiler \
+            iree-runtime \
+            "numpy<2.0"
+
+      - name: Run llama test
+        run: pytest sharktank/tests/models/llama/benchmark-tests.py
```
saienduri, line 70: new line

sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -51,6 +51,12 @@ def main():
         help="Enables strictness during export",
         action="store_true",
     )
+    parser.add_argument(
```
saienduri, line 54: Is there somewhere this argument is being used?

.github/workflows/ci-llama.yaml
IanNod: No, we should have this just nightly

sharktank/tests/models/llama/benchmark-tests.py
```diff
@@ -0,0 +1,310 @@
...
+        self.iree_benchmark_module(
+            "0",
+            self.output_vmfb_fp8,
+            self.irpa_path_fp8,
+            self.iree_run_decode_args_fp8,
+            self.repo_root,
+        )
+
+    @pytest.mark.xfail
+    def testBenchmark8B_f16_Non_Decomposed(self):
```
saienduri, line 216: Instead of marking the whole test as xfail, can we xfail the particular step it fails on? That way we can track as we make progress or regress with certain configs

.github/workflows/ci-llama.yaml
IanNod, line 70: this will likely be picked up by the presubmit ci.yaml. Should probably add something like this to avoid https://stackoverflow.com/questions/33084190/default-skip-test-unless-command-line-parameter-present-in-py-test/43938191#43938191

.github/workflows/ci-llama.yaml
```diff
@@ -0,0 +1,70 @@
+name: Integration Tests
```
saienduri, line 1: You can change the name here because we already have an Integration Tests. Maybe something more specific to llama?

sharktank/tests/models/llama/benchmark-tests.py
```diff
@@ -0,0 +1,310 @@
...
+
+
+class BenchmarkLlama3_1_8B_f16(BaseBenchmarkTest):
+    def setUp(self):
+        # TODO: add numpy files to Azure and download from it
+        self.repo_root = os.getenv("SHARK_PLATFORM_REPO_ROOT")
+        artifacts_dir = "/data/extra/models/llama3.1_8B/"
+        self.irpa_path = artifacts_dir + "llama8b_f16.irpa"
+        self.irpa_path_fp8 = artifacts_dir + "llama8b_fp8.irpa"
+        self.output_mlir = self.repo_root + "llama8b_f16.mlir"
```
IanNod, line 137: output files should be temporary. We don't want them lasting or being picked up accidentally on subsequent runs

sharktank/tests/models/llama/benchmark-tests.py
```diff
@@ -0,0 +1,310 @@
...
+        self.iree_benchmark_module(
+            "0",
+            self.output_vmfb,
+            self.irpa_path,
+            self.iree_run_decode_args,
+            self.repo_root,
+        )
+
+    @pytest.mark.xfail
+    def testBenchmark8B_fp8_Non_Decomposed(self):
```
IanNod, line 277: can you add TP1 70B and TP8 405B as well?

sharktank/tests/models/llama/benchmark-tests.py
```diff
@@ -0,0 +1,310 @@
...
+        self.output_mlir = self.repo_root + "llama8b_f16.mlir"
+        self.output_json = self.repo_root + "llama8b_f16.json"
+        self.output_vmfb = self.repo_root + "llama8b_f16.vmfb"
+        self.output_mlir_fp8 = self.repo_root + "llama8b_fp8.mlir"
+        self.output_json_fp8 = self.repo_root + "llama8b_fp8.json"
+        self.output_vmfb_fp8 = self.repo_root + "llama8b_fp8.vmfb"
+        self.iree_compile_args = [
+            "--iree-hal-target-backends=rocm",
+            "--iree-hip-target=gfx942",
+            f"--iree-hal-dump-executable-files-to={self.repo_root}/files/llama",
```
IanNod, line 146: Would be ideal if we could upload the dispatches somewhere easy for codegen to access

sharktank/tests/models/llama/benchmark-tests.py
```diff
@@ -0,0 +1,310 @@
...
+            output_mlir_path,
+            "--output-config",
+            output_json_path,
+        ]
+        if attention_kernel in ["decomposed", "torch_sdpa"]:
+            export_args.append("--attention-kernel")
+            export_args.append(attention_kernel)
+
+        cmd = subprocess.list2cmdline(export_args)
+        return cmd
```
ScottTodd, line 43: This can be separate from this PR, but we should use an in-memory API for all model exporting/importing workflows, rather than fork into a subprocess.

Users shouldn't need to chain together `python -m sharktank.examples.` and  `iree-compile ...` commands. We can aim for something like https://docs.vllm.ai/en/latest/getting_started/quickstart.html#offline-batched-inference
```python
llm = LLM(model="facebook/opt-125m")
outputs = llm.generate(prompts, sampling_params)
```

(that's as minimal as it gets - we'll want to pass options like the compilation target though)

sharktank/tests/models/llama/benchmark-tests.py
ScottTodd, line 1: nit: other files in this repo use underscores instead of dashes in their file names

sharktank/tests/models/llama/benchmark-tests.py
saienduri: We can upload that directory as a github artifact at the end of every run. (https://github.com/actions/upload-artifact)

archana-ramalingam:
Updating the branch with changes from `main` will fix the second CI test.

sharktank/tests/models/llama/benchmark-tests.py
aviator19941: Added it as a last step in the yaml file

sharktank/tests/models/llama/benchmark-tests.py
aviator19941: Done

sharktank/tests/models/llama/benchmark-tests.py
aviator19941: Added a function to remove the output files at the end of each test

.github/workflows/ci-llama.yaml
aviator19941: Done

sharktank/tests/models/llama/benchmark-tests.py
aviator19941: I can do this in a follow-up PR.

sharktank/tests/models/llama/benchmark_tests.py
```diff
@@ -0,0 +1,779 @@
...
+        ]
+        benchmark_args += args
+        cmd = subprocess.list2cmdline(benchmark_args)
+        logging.getLogger().info(f"Launching run command:\n" f"cd {cwd} && {cmd}")
+        proc = subprocess.run(cmd, shell=True, stdout=sys.stdout, cwd=cwd)
+        return_code = proc.returncode
+        if return_code != 0:
+            raise Exception(f"{cmd} failed to run")
+
+    def cleanup_output_files(
```
IanNod, line 141: probably better to use tempfile like done here https://github.com/nod-ai/SHARK-Platform/blob/0c2e965c3ffe723db2fe2be9193c6d45fe558dbe/sharktank/tests/types/tensors_test.py#L71. Should better handle potential multiple async runs effectively corrupting the data in these effectively hardcoded paths

sharktank/tests/models/llama/benchmark_tests.py
```diff
@@ -0,0 +1,779 @@
...
+        self.iree_benchmark_module(
+            "0",
+            output_vmfb,
+            self.irpa_path,
+            self.iree_run_decode_args,
+            self.repo_root,
+        )
+        self.cleanup_output_files(output_mlir, output_json, output_vmfb)
+
+    @pytest.mark.skip(reason="TODO: Need to plumb through attention_kernel")
```
IanNod, line 247: why skip instead of xfail?

sharktank/tests/models/llama/benchmark_tests.py
```diff
@@ -0,0 +1,779 @@
...
+        )
+        # benchmark decode
+        self.iree_benchmark_module(
+            "0",
+            output_vmfb,
+            self.irpa_path,
+            self.iree_run_decode_args,
+            self.repo_root,
+        )
+        self.cleanup_output_files(output_mlir, output_json, output_vmfb)
```
IanNod, line 245: To more easily handle regressions and such I think we really want to upload and store the model IR, vmfb, and dispatch dump so we can triage exactly what this benchmark used.

sharktank/tests/models/llama/benchmark_tests.py
```diff
@@ -0,0 +1,779 @@
...
+
+        except subprocess.CalledProcessError as e:
+            print(f"Error removing files: {e}")
+
+
+class BenchmarkLlama3_1_8B(BaseBenchmarkTest):
+    def setUp(self):
+        # TODO: add numpy files to Azure and download from it
+        self.repo_root = os.getenv("SHARK_PLATFORM_REPO_ROOT")
+        artifacts_dir = "/data/extra/models/llama3.1_8B/"
```
IanNod, line 165: likely need at least a comment that this is system specific for A30F. Anybody trying to replicate locally will need to know how to recreate. Better would probably be adding an arg to pass this path that defaults to this for the runner

sharktank/tests/models/llama/benchmark_tests.py
```diff
@@ -0,0 +1,779 @@
...
+        self.iree_benchmark_module(
+            "0",
+            output_vmfb,
+            self.irpa_path,
+            self.iree_run_decode_args,
+            self.repo_root,
+        )
+        self.cleanup_output_files(output_mlir, output_json, output_vmfb)
+
+    @pytest.mark.xfail
```
IanNod, line 283: nit: would be good to comment why we have the xfails for any newcomers not familiar with what isn't functional yet

sharktank/tests/models/llama/benchmark_tests.py
aviator19941: Because the test is still able to run successfully since it is still able to export an IR and compile and benchmark fully through, however, it is not exporting the correct IR yet. The attention_kernel support needs to be added for it to execute as expected.

sharktank/tests/models/llama/benchmark_tests.py
aviator19941: Makes sense, I will save the temporary files into the github actions file then.

sharktank/tests/models/llama/benchmark_tests.py
ScottTodd: General workflow tip for tests / CI: start by making tests runnable by a developer, _then_ teach a CI machine how to follow those steps. We should never have a hardcoded path like this (and this will only work on systems that support such a file path, so no Windows).

sharktank/tests/models/llama/benchmark_tests.py
IanNod: We should raise an error when calling the attention_kernel flag with torch_sdpa until that is plumbed through and then we can xfail here

sharktank/tests/models/llama/benchmark_tests.py
aviator19941: We want to save the output files to track any regressions right? So, do we still need tempfile? I think `os.mkdir()` would make more sense to use in this case, especially when specifying the directory path of the artifiacts to upload.

IanNod: you can upload the output artifacts before tempfile gets rid of them. I don't think we want to keep the files on the CI system permanently. os.mkdir could still have issues with multiple simultaneous instances running interfering with each other unless you are very careful with it

sharktank/tests/models/llama/benchmark_tests.py
IanNod: I was kind of thinking of just having it upload to sharkpublic since that would be easy to access for those who want them. Not sure if github actions file would be better or not

sharktank/tests/models/llama/benchmark_tests.py
aviator19941: How can I upload the output artifacts before tempfile gets rid of them?

IanNod: https://github.com/nod-ai/SHARK-TestSuite/blob/95b35cc076618ef79b7f84dc4117f4528451e5e6/e2eshark/_run_helper.py#L61 for example uploading files. You should just need to do that before closing the tempfile or directory that is created

saienduri: Do we prefer sharkpublic? We will be uploading a bunch of artifacts there every night, and it might be hard to correlate which run correlates to which files. With github actions, we can just click on a run and see it's respective output artifacts without cluttering the storage

IanNod: Does github actions work well for downloading things like vmfbs? If so, then that would work well.

ScottTodd: How much data are we talking about here? Cloud storage isn't cheap. Using GitHub artifacts with GitHub workflows is nice since it doesn't require separate privileges to maintain, but then you're limited by GitHub's retention policies.

If you want to keep historical data in an organized way, please design a database early. I've seen too many benchmarking systems without that structure and it makes analysis tricky.

For this PR itself, I'd try to keep the scope small and checkpoint the work, rather than get hung up on all these discussions. Fine to land and iterate a bit so things don't pile up.

saienduri: Yeah let's start with github artifacts without a temp dir (it won't persist between runs anyways) and iterate from there

sharktank/tests/models/llama/benchmark-tests.py
aviator19941: Done

.github/workflows/ci-llama.yaml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Install pip deps
+        run: |
+          python -m pip install --no-compile --upgrade pip
+          # Note: We install in three steps in order to satisfy requirements
+          # from non default locations first. Installing the PyTorch CPU
+          # wheels saves multiple minutes and a lot of bandwidth on runner setup.
+          pip install --no-compile -r pytorch-cpu-requirements.txt
+          pip install --no-compile -f https://iree.dev/pip-release-links.html --src deps \
+            -e "git+https://github.com/iree-org/iree-turbine.git#egg=iree-turbine"
+          pip install --no-compile -r requirements.txt -e sharktank/
```
archana-ramalingam, line 59: After updating main changes, if your CI fails, just update this: `pip install --no-compile -r requirements.txt -r sharktank/requirements-tests.txt -e sharktank/`, there are deps used by perplexity tests (merged on Friday) and any sharktank ci fails if the dep is not detected (even though perplexity tests are skipped). 
Not sure if there is a better way to handle this.

aviator19941: Ok added

archana-ramalingam (approved):

sharktank/tests/models/llama/benchmark_tests.py
```diff
@@ -0,0 +1,837 @@
...
+        self.temp_dir_8b = Path(self.dir_path_8b)
+        self.temp_dir_8b.mkdir(parents=True, exist_ok=True)
+        self.iree_compile_args = [
+            "--iree-hal-target-backends=rocm",
+            "--iree-hip-target=gfx942",
+        ]
+        self.prefill_args_f16 = artifacts_dir + "prefill_args"
+        self.decode_args_f16 = artifacts_dir + "decode_args"
+        self.prefill_args_fp8 = artifacts_dir + "prefill_args_fp8"
+        self.decode_args_fp8 = artifacts_dir + "decode_args_fp8"
```
ScottTodd, line 198: Path manipulation should use https://docs.python.org/3/library/pathlib.html

These would then look like
```python
self.prefill_args_f16 = artifacts_dir / "prefill_args"
```

pathlib handles operating system path differences while direct string manipulation like you have here will not. For example, paths on Windows use `\` instead of `/`. The pathlib code treats the `/` _operator_ as an abstract path join operation (similar to https://docs.python.org/3/library/os.path.html#os.path.join) that will do the right thing when eventually converting to a concrete operating system specific path.

saienduri:
Looks like artifacts are not being uploaded as part of the latest run?

saienduri (approved):
Nice, everything works!

PR #290, opened 2024-10-16
WIP Fix in-place tensor mutation when sharding

sogartar:
Add option to construct sharded tensors without inserting device placement operations.

Add a simple test for in-place addition
t += 1

When unflattening a sharded tensor during exporting insert in-place device placements, which get translated to function argument affinity attributes.


sogartar:
This depends on https://github.com/iree-org/iree-turbine/pull/220

PR #296, opened 2024-10-17
Add a flag to select between nondecomposed and decomposed attention

KyleHerndon:


sharktank/sharktank/export_layer/export_paged_attention.py
```diff
@@ -193,6 +193,12 @@ def main():
         action="store_true",
     )
 
+    parser.add_argument(
```
dan-garvey, line 196: I think we need to move this to CLI to enable re-use with eager.

sharktank/sharktank/ops/default_impls.py
```diff
@@ -405,7 +405,7 @@ def rms_norm_default(x, weight, *, epsilon: float) -> Tensor:
     variance = x.pow(2).mean(-1, keepdim=True)
     output = x * elementwise(torch.rsqrt, variance + epsilon)
     # The cast here is to match the hf implementation, affects numerics
-    output = weight * to(output, weight.dtype)
+    output = unbox_tensor(weight) * to(output, weight.dtype)
```
dan-garvey, line 408: I don't think this breaks anything but why do we need it?

dan-garvey (commented):
minor comments

sharktank/sharktank/export_layer/export_paged_attention.py
dan-garvey: can leave this as a todo if you want

sharktank/sharktank/ops/default_impls.py
KyleHerndon: When exporting, torch gets upset when the multiplication has one argument of a `sharktank` tensor and one of a `torch` tensor type. The `to` op always returns a torch tensor, so this fixes them both to be `torch` tensors. Almost all the functions in `default_impls.py` are calling `unbox_tensor` on their inputs. I'm not sure why this patch turned up this particular issue, but it was in fact failing before I added this.

KyleHerndon: Actually, using that fix breaks other things. Changing it to an explicit `ops.elementwise(torch.mul, x, y)` works better.

dan-garvey (approved):

PR #297, opened 2024-10-18
Include a region in attention template

nithinsubbiah:
Fix the attention template to include a region (https://github.com/iree-org/iree/pull/18728). Also some code quality improvements


sharktank/sharktank/kernels/batch_matmul_transpose_b.py
```diff
@@ -80,7 +82,7 @@ def generate(self, ksel: KernelSelection, kb: KernelBuilder):
         spec_sig = f"L{a_ident}_R{b_ident}"
         template_file = "batch_matmul_transpose_b.mlir"
         target_function_name = f"sharktank_batch_matmul_transpose_b_{spec_sig}"
-        cst_zero = "0." if "f" in str(accum_type) else "0"
+        cst_zero = "0" if isinstance(accum_type, IntegerType) else "0."
```
stellaraccident, line 85: MLIR types aren't Python types. For reasons, you actually want to ask MLIR whether accum_type is an MLIR isa of IntegerType. You do this by saying `IntegerType.isinstance(accum_type)`. How you have it will work sometimes but not all the time (it will work if someone has already cast the MLIR type to the Python `IntegerType` wrapper).

It is best to think of these as `IntegerFidget` or something. They are their own thing that is in no way related to Python types.

stellaraccident (changes requested):

sharktank/sharktank/kernels/batch_matmul_transpose_b.py
nithinsubbiah: Got it, changed the instance checking to use MLIR's method

rsuderman (approved):

PR #299, opened 2024-10-21
Finish end-to-end sharded workflow

rsuderman:
This include some minor fixes for exporting llama style LLMs

### DIFF
--- a/sharktank/sharktank/examples/export_paged_llm_v1.py
+++ b/sharktank/sharktank/examples/export_paged_llm_v1.py
@@ -19,6 +19,7 @@ from ..models.llama.llama import LlamaModelConfig, PagedLlamaModelV1
 from ..models.llama.sharding import shard_theta
 from ..models.mixtral.mixtral import *
 from ..models.grok.grok import *
+from .. import ops
 
 
 def main():
@@ -34,7 +35,7 @@ def main():
     parser.add_argument(
         "--output-config",
         help="Output file path for exported config file",
-        default="tmp/batch_llama_v1.json",
+        default="/tmp/batch_llama_v1.json",
     )
     parser.add_argument(
         "--bs",
@@ -79,6 +80,11 @@ def main():
     llama_config.kv_cache_type = "direct" if args.bs == [1] else "paged"
     llama_config.attention_kernel = args.attention_kernel
 
+    # This is a bit gross and should be changed in the future. Best Idea I had so far.
+    attn_q_weight = dataset.root_theta.tensor("blk")["0"]["attn_q"]["weight"]
+    if isinstance(attn_q_weight, SplitPrimitiveTensor):
+        llama_config.tensor_parallelism_size = attn_q_weight.shard_count
+
     if llama_config.hp.expert_count:
         if llama_config.hp.model_arch == "grok":
             model = PagedGrokModelV1(dataset.root_theta, llama_config)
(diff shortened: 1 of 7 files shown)


sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -61,7 +62,11 @@ def main():
     llama_config = LlamaModelConfig(hp)
     llama_config.use_hf = False
     llama_config.static_tables = False  # Rely on the compiler for hoisting tables.
-    llama_config.kv_cache_type = "direct" if args.bs == [1] else "paged"
+    llama_config.kv_cache_type = "paged"
+
+    attn_q_weight = dataset.root_theta.tensor("blk")["0"]["attn_q"]["weight"]
+    if isinstance(attn_q_weight, SplitPrimitiveTensor):
+        llama_config.tensor_parallelism_size = attn_q_weight.shard_count
```
sogartar, line 69: This looks a bit brittle. Can dataset.properties carry the tensor_parralelism_size? If it is too much trouble I guess we can live with it.

sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -121,26 +126,32 @@ def generate_batch_prefill(bs: int):
...
+                seq_lens,
+                seq_block_ids,
+                [shard._data for shard in cache_state[0].shards],
+            ),
             dynamic_shapes=dynamic_shapes,
             strict=args.strict,
         )
-        def _(model, tokens, seq_lens, seq_block_ids, cache_state):
+        def _(model, tokens, seq_lens, seq_block_ids, cs):
+            cs = [SplitPrimitiveTensor(ts=cs, shard_dim=cache_state[0].shard_dim)]
```
sogartar, line 146: Do we always what a sharded cache state?

sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -121,26 +126,32 @@ def generate_batch_prefill(bs: int):
...
         print(f"Exporting prefill_bs{bs}")
 
         @fxb.export_program(
             name=f"prefill_bs{bs}",
-            args=(tokens, seq_lens, seq_block_ids, cache_state),
+            args=(
+                tokens,
+                seq_lens,
+                seq_block_ids,
+                [shard._data for shard in cache_state[0].shards],
```
sogartar, line 140: We have the sharded tensor types registered in PyTorch and can be a part of function signatures. They will be auto-flattened/unflattened during export.
Then you would not need to reconstruct the cache from the pieces latter.

sogartar (changes requested):

sharktank/sharktank/examples/export_paged_llm_v1.py
sogartar: It is fine to leave it like that, not a big deal.

rsuderman: I reworked the helpers for packing / unpacking. Ultimately we should writing some export tooling to automate this. Anything sharded should generate the affinities, etc.

sharktank/sharktank/examples/export_paged_llm_v1.py
rsuderman: Reworked to only happen for when sharded. We should consider making this constructor a helper so that when shard count is 1 it just returns the tensor.

sharktank/sharktank/examples/export_paged_llm_v1.py
rsuderman: So the problem with adding it to `data.properties`  is that its not a default part of `gguf` and we are already plumbing it via the `LLamaHParams`. I attempted to wire it further down but it just turned into spaghetti coding.

Long term we want to load from the same `irpa` file so we will want to re-evaluate this for now.

sogartar (approved):

sogartar:
I see the toy-sized sharded Llama test is failing due to the change in the returned tensor type. @rsuderman would you like me to fix it?

sharktank/sharktank/models/llama/llama.py
```diff
@@ -132,25 +132,6 @@ def prefill(
         self._assert_device(seq_block_ids)
         self._assert_device(*cache_state, dtype=self.activation_dtype)
 
-        if self.config.tensor_parallelism_size > 1:
```
sogartar, line 135: It would be nice to leave asserts for the expected argument types.

rsuderman: We want to avoid bleeding too much sharding code into the modelling work. Too much mixing make it difficult for straight modelling people to work through.

rsuderman:
> I see the toy-sized sharded Llama test is failing due to the change in the returned tensor type. @rsuderman would you like me to fix it?

Fixed the sharding types being pumped through

PR #300, opened 2024-10-21
Add exporting with argument device affinities

sogartar:
We don't have handling of device affinties for our sharded tensor types when they are the arguments of exported functions. Although, they can be specified explicitly when exporting, this change adds the ability to hide this and deduces the affinties from the tensor types.

### DIFF
--- a/sharktank/sharktank/export.py
+++ b/sharktank/sharktank/export.py
@@ -1,0 +1,156 @@
+# Copyright 2024 Advanced Micro Devices, Inc.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+from typing import Callable, Any
+import torch
+from iree.turbine.aot import DeviceAffinity, FxProgramsBuilder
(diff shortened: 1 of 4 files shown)


sogartar:
This requires https://github.com/iree-org/iree-turbine/pull/231.

sharktank/sharktank/export.py
```diff
@@ -0,0 +1,156 @@
...
+            return flat_fn(*flat_args)
+
+        amended_kwargs = dict(**transitive_kwargs)
+        if "name" not in amended_kwargs or amended_kwargs["name"] is None:
+            amended_kwargs["name"] = f.__name__
+        return fx_builder.export_program(
+            module_fn_with_flat_signature,
+            *transitive_args,
+            args=flat_args,
+            argument_device_affinities=argument_device_affinities,
```
rsuderman, line 152: Wait, where does this actually plumb through? I don't see this as part of the PR

rsuderman (changes requested):

sharktank/sharktank/export.py
rsuderman: I see, the changes are dependent on https://github.com/iree-org/iree-turbine/pull/231/files
Lets focus on that and shelve these changes. Once `export` has the right parameter we won't need to make the sharktank changes.

rsuderman (approved):

PR #304, opened 2024-10-22
Numpy Support + Debug ops

stbaione:
# Links

Addresses this issue: [(shortfin) Need debug-print support for f16 arrays and other types not supported array.array types](https://github.com/nod-ai/SHARK-Platform/issues/292)

# Description

This PR adds support for directly converting device_arrays to numpy arrays. We also implement some common functions to be used in the debug context, to achieve better visibility into the values of a device_array at runtime.

We implement the [ArrayProtocol](https://numpy.org/doc/stable/reference/arrays.interface.html) when numpy is available to make casting between device_arrays and numpy arrays easier.

We define a module named `shortfin.array.nputils` to access these features, which is only available in public API if numpy is installed.

# Supported Functions

- debug_dump_array: Dump contents of array to debug logs
- debug_fill_array: Convert to np array, and fill with specific values
- _find_mode: Find the mode of an np.array, along a specified dimension
- debug_log_tensor_stats: Log the following stats for a tensor:
    - NaN Count
    - Shape & dtype
    - Min
    - Max
    - Mode
    - First 10 Elements
    - Last 10 Elements


stbaione:
> Looked into it a bit.
> 
> Seems like we have a debug environment variable used by setup.py (which contains settings used by `pip install -e .`)
> 
> https://github.com/nod-ai/SHARK-Platform/blob/main/shortfin/setup.py#L209

Updated to throw:

`EnvironmentError(
        "Debug module should only be imported in Debug mode (SHORTFIN_CMAKE_BUILD_TYPE=Debug)"
)`

if Debug module is imported without proper env value.

Added patch in tests to ensure env is set properly when debug tests are run, which does make the tests a little less pretty unfortunately. Had to move the `debug_array` import into each individual test for it to honor the patched value.

stellaraccident:
> > Looked into it a bit.
> > Seems like we have a debug environment variable used by setup.py (which contains settings used by `pip install -e .`)
> > https://github.com/nod-ai/SHARK-Platform/blob/main/shortfin/setup.py#L209
> 
> Updated to throw:
> 
> `EnvironmentError( "Debug module should only be imported in Debug mode (SHORTFIN_CMAKE_BUILD_TYPE=Debug)" )`
> 
> if Debug module is imported without proper env value.
> 
> Added patch in tests to ensure env is set properly when debug tests are run, which does make the tests a little less pretty unfortunately. Had to move the `debug_array` import into each individual test for it to honor the patched value.
> 
> Also added a line to skip debug tests if numpy is not installed, which occurred in one of the CI run failures.

We should not tie this to debug *builds*. Generally, the availability of features that aid in building shortfin applications should always be available. You only want to gate on the debug build type for things like asserts, expensive checks, etc.

shortfin/python/shortfin/debug/array.py
```diff
@@ -0,0 +1,106 @@
...
+    logger.debug(np_array)
+
+
+async def fill_array(tensor: sfnp.device_array, fill_value: int | float):
+    np_array = await to_np(tensor)
+    np_array.fill(fill_value)
+    return np_array
+
+
+def find_mode(arr: np.ndarray, axis=0, keepdims=False):
```
stellaraccident, line 51: I don't think this should be in shortfin itself. Should just be a fidget of numpy code used when needed. If it is being used all the time, can add to the public api later.

shortfin/python/shortfin/debug/array.py
```diff
@@ -0,0 +1,106 @@
...
+
+import numpy as np
+
+from shortfin import array as sfnp
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+
+
+async def to_np(tensor: sfnp.device_array):
```
stellaraccident, line 11: You're going to want to base this on implementing the array protocol on the backing C++ types. Basically, [here](https://github.com/nod-ai/SHARK-Platform/blob/861ead52a26528974878041617b0e09ee58528fe/shortfin/python/array_binding.cc#L491C3-L491C21), you want to implement the [Array Protocol](https://numpy.org/doc/stable/reference/arrays.interface.html) anywhere the buffer protocol is available. Whereas the buffer protocol just always returns a bytes view, the array protocol is rich enough to let you expose any data type directly as a numpy array (i.e. fully typed if [it is a typed mapping](https://github.com/nod-ai/SHARK-Platform/blob/main/shortfin/python/array_binding.cc#L335) or untyped/bytes otherwise).

Then with that basic support in, you will have a better time writing utility fidgets.

shortfin/python/shortfin/debug/array.py
```diff
@@ -0,0 +1,106 @@
+import logging
+
+import numpy as np
```
stellaraccident, line 3: Let's instead put this under shortfin.array.np or something like that. And then we only include that in the API if numpy is installed.

stellaraccident (changes requested):
Ok, looking at the whole thing, let's just make the primitives part of the regular API but only exposed if numpy is available. I think that `shortfin.array.nputils` or something like that is fine. Let's prefix functions intended for debugging with `debug_` or something vs having a completely separate debugging namespace/setup.

stbaione:
> Ok, looking at the whole thing, let's just make the primitives part of the regular API but only exposed if numpy is available. I think that `shortfin.array.nputils` or something like that is fine. Let's prefix functions intended for debugging with `debug_` or something vs having a completely separate debugging namespace/setup.

Working on the suggested changes now. Will reach out if I hit any blockers

shortfin/python/shortfin/debug/array.py
stbaione: Made `find_mode` a private method

shortfin/python/shortfin/debug/array.py
stbaione: Reimplemented by optionally implementing ArrayProtocol for device_array, if numpy is present

shortfin/python/shortfin/debug/array.py
stbaione: Moved to shortfin.array.nputils

stbaione:
- Rewrote to implement ArrayProtocol on device_array, if numpy is available, which allows direct `np.array(device_array)` instead of needing to have a function with some custom logic.
- Moved functions to `shortfin.array.np_utils`. Is imported and publicly available only when numpy is installed

shortfin/python/array_binding.cc
```diff
@@ -495,117 +495,212 @@ void BindArray(py::module_ &m) {
...
+    device_array_class
+        .def("__array__",
+             [refs](device_array &self) {
+               py::object np = py::module_::import_("numpy");
+
+               auto &table = refs->element_type_array_type_code_table;
+               auto it = table.find(self.dtype());
+               if (it == table.end()) {
+                 throw std::invalid_argument(fmt::format(
+                     "Python array.array type code not know for dtype "
```
renxida, line 630:
```suggestion
                     "Python array.array type code not known for dtype "
```

shortfin/python/array_binding.cc
```diff
@@ -495,117 +495,212 @@ void BindArray(py::module_ &m) {
...
+            shapeList.append(shape[i]);
+          }
+          py::tuple shape_tuple(py::cast(shapeList));
+          interface["shape"] = shape_tuple;
+
+          auto &table = refs->element_type_array_type_code_table;
+          auto it = table.find(self.dtype());
+          if (it == table.end()) {
+            throw std::invalid_argument(
+                fmt::format("Python array.array type code not know for dtype "
```
renxida, line 672:
```suggestion
                fmt::format("Python array.array type code not known for dtype "
```

shortfin/python/shortfin/array/_nputils.py
```diff
@@ -0,0 +1,75 @@
...
+    if len(np_array_no_nan) > 0:
+        mode = _find_mode(np_array_no_nan)[0]
+        logger.debug(f"Min (excluding NaN): {np_array_no_nan.min()}")
+        logger.debug(f"Max (excluding NaN): {np_array_no_nan.max()}")
+        logger.debug(f"Mean (excluding NaN): {np_array_no_nan.mean()}")
+        logger.debug(f"Mode (excluding NaN): {mode}")
+        logger.debug(f"First 10 elements: {np_array_no_nan.flatten()[:10]}")
+        logger.debug(f"Last 10 elements: {np_array_no_nan.flatten()[-10:]}")
+    else:
+        logger.warning(f"  All values are NaN")
```
renxida, line 75: type hints for return values & some doc strings would be nice

I expect these functions to be used a lot in somebody's IDE

Like
```python
def debug_log_tensor_stats(tensor: sfnp.device_array) -> None:
"""
Log NaN count, shape, min/max/mean/mode, and sample values for device array. Also works if the array is an array.array or np.ndarray.
"""
    ...
```

renxida (changes requested):
mostly looks good, love having tests. Small readability / usability changes. Stella mentioned she'll get back to this later so feel free to merge after her approval.

shortfin/python/array_binding.cc
stbaione: Fixed Typo

shortfin/python/array_binding.cc
stbaione: Fixed Typo

shortfin/python/shortfin/array/_nputils.py
stbaione: Added docstring and type hints for all return values of functions in _nputils.py

shortfin/python/shortfin/array/__init__.py
```diff
@@ -83,3 +85,10 @@
     "fill_randn",
     "RandomGenerator",
 ]
+
+# Import nputils if numpy is present.
+np_present = importlib.util.find_spec("numpy") is not None
+if np_present:
+    from . import _nputils as nputils
+
+    __all__.append("nputils")
```
stellaraccident, line 94: I have a feeling that this will not play well with static type analyzers, but I guess we will find out.

shortfin/python/array_binding.cc
```diff
@@ -495,117 +495,212 @@ void BindArray(py::module_ &m) {
...
+          if (it == table.end()) {
+            throw std::invalid_argument(
+                fmt::format("Python array.array type code not known for dtype "
+                            "{}: Cannot access items",
+                            self.dtype().name()));
+          }
+
+          auto typeString = py::str();
+          if (it->first == DType::float16()) {
+            typeString = py::str(np.attr("dtype")(np.attr("float16")));
```
stellaraccident, line 679: This is a long way around to just get the literal "float16". That's basically never going to change or need this indirection. If you just change this to a string literal, you can avoid the entire numpy import and even remove the conditional definition.

stellaraccident (commented):
Thanks. Can you have a look at the comment in array_binding.cc and remove the numpy import if you agree?

shortfin/python/array_binding.cc
stbaione: Replaced with a literal string, and removed the conditional definition. Also removed `__array__` function... not sure why I thought that was required, but everything works fine without it.

stellaraccident (approved):

renxida (approved):

PR #306, opened 2024-10-22
[sharktank] Evaluation - Add Perplexity test for vmfb

archana-ramalingam:
Add Perplexity test for vmfb


sharktank/conftest.py
```diff
@@ -121,11 +121,34 @@ def pytest_addoption(parser):
     )
 
     parser.addoption(
-        "--baseline-perplexity-score-json",
+        "--baseline-perplexity-scores",
         type=Path,
         action="store",
-        default="sharktank/tests/evaluate/baseline_perplexity_scores.json",
-        help="Llama3.1 8B & 405B model baseline perplexity scores json",
+        default="/home/aramalin/SHARK-Platform/sharktank/tests/evaluate/baseline_perplexity_scores.json",
```
IanNod, line 127: Should avoid hardcoded absolute paths

sharktank/conftest.py
```diff
@@ -121,11 +121,34 @@ def pytest_addoption(parser):
...
+        help="Llama3.1 8B & 405B model baseline perplexity scores",
+    )
+
+    parser.addoption(
+        "--llama3-8b-f16-vmfb-path",
+        type=Path,
+        action="store",
+        default="/data/extra/models/llama3.1_8B/llama8b_f16.vmfb",
```
IanNod, line 135: We should be compiling new vmfb's within this test to catch regressions, not testing precompiled ones which it looks like we are doing now

sharktank/sharktank/evaluate/perplexity_vmfb.py
```diff
@@ -0,0 +1,402 @@
...
+    def get_prompts(self):
+        test_prompts = load_dataset("wikitext", "wikitext-2-raw-v1", split="test")[
+            "text"
+        ]
+        num_test_prompts = 219
+
+        random.seed(0)
+        test_prompts = random.sample(test_prompts, num_test_prompts)
+
+        # Ignore prompts that are: empty, less than 20 tokens or a title.
```
IanNod, line 149: Why are we ignoring < 20 tokens or title?

sharktank/sharktank/evaluate/perplexity_vmfb.py
```diff
@@ -0,0 +1,402 @@
...
+    tensor_parallelism_size,
+):
+    perplexity = Perplexity(
+        torch_device=torch_device,
+        iree_device=iree_device,
+        kv_cache_type=kv_cache_type,
+        tensor_parallelism_size=tensor_parallelism_size,
+    )
+
+    # perplexity.load_model(tokenizer, vmfb_path, weight_path)
```
IanNod, line 350: remove commented code

sharktank/conftest.py
```diff
@@ -121,11 +121,34 @@ def pytest_addoption(parser):
...
+    parser.addoption(
+        "--llama3-405b-f16-vmfb-path",
+        type=Path,
+        action="store",
+        default="/data/extra/models/llama3.1_405B/llama405b_fp16.vmfb",
+        help="Llama3.1 405b fp16 vmfb path, defaults to 30F CI system path",
+    )
+
+    parser.addoption(
+        "--iree-device",
```
IanNod, line 148: Not clear to me what device we are running for the CI, where are we setting that?

sharktank/sharktank/evaluate/perplexity_vmfb.py
archana-ramalingam: Perplexity scores are too high for incomplete sentences like titles or if tokens < 20. Especially in the future, if we add sliding window and run the whole golden wiki dataset, this is helpful filtering.

sharktank/conftest.py
archana-ramalingam: Will add `--device='hip://0'` flag in yaml to vmfb pytest. The vmfb is compiled with `--iree-hip-target=gfx942`, so it needs to run on MI300.

.github/workflows/ci_eval.yaml
```diff
@@ -58,5 +59,51 @@ jobs:
             -e "git+https://github.com/iree-org/iree-turbine.git#egg=iree-turbine"
           pip install --no-compile -r requirements.txt -r sharktank/requirements-tests.txt -e sharktank/
 
-      - name: Run perplexity test
-        run:  pytest -n 4 -v -s sharktank/tests/evaluate/perplexity_test.py  --longrun
+      - name: Run perplexity test with vmfb
+        run:  pytest -n 4 -v -s sharktank/tests/evaluate/perplexity_vmfb_test.py --iree-device='hip://7' --longrun
```
IanNod, line 63: Why -n 4? This is running on a system with something like 192 cores so we could expand that quite a bit to maybe speed runtime up.

.github/workflows/ci_eval.yaml
```diff
@@ -1,6 +1,7 @@
 name: Evaluation Tests
 
 on:
+  pull_request:
```
IanNod, line 4: is this temporary for testing?

.github/workflows/ci_eval.yaml
```diff
@@ -58,5 +59,51 @@ jobs:
             -e "git+https://github.com/iree-org/iree-turbine.git#egg=iree-turbine"
           pip install --no-compile -r requirements.txt -r sharktank/requirements-tests.txt -e sharktank/
 
-      - name: Run perplexity test
-        run:  pytest -n 4 -v -s sharktank/tests/evaluate/perplexity_test.py  --longrun
+      - name: Run perplexity test with vmfb
+        run:  pytest -n 4 -v -s sharktank/tests/evaluate/perplexity_vmfb_test.py --iree-device='hip://7' --longrun
+
+  # test_perplexity_torch:
```
IanNod, line 65: commented test?

sharktank/conftest.py
```diff
@@ -81,11 +81,19 @@ def pytest_addoption(parser):
     )
 
     parser.addoption(
-        "--llama3-8b-f16-gguf-path",
+        "--llama3-8b-json-path",
         type=Path,
         action="store",
-        default="/data/extra/models/llama3.1_8B/llama8b_f16.gguf",
-        help="Llama3.1 8b gguf model path, defaults to 30F CI system path",
+        default="/data/extra/models/llama3.1_8B/llama8b_test.json",
```
IanNod, line 87: hardcoded path's in a test are not good to have. Can you at least add a todo to resolve in a later pr?

.github/workflows/ci_eval.yaml
archana-ramalingam, line 65: Will uncomment this job once I test vmfb version, as the jobs run sequentially in unordered fashion.

sharktank/sharktank/evaluate/perplexity_vmfb.py
```diff
@@ -0,0 +1,449 @@
...
+logger.setLevel(log_levels["info"])
+
+logger.root.handlers[0].setFormatter(
+    logging.Formatter(fmt="\n%(levelname)s:%(name)-8s %(message)s")
+)
+
+__all__ = ["Perplexity", "run_perplexity"]
+
+
+class Perplexity:
```
IanNod, line 52: There seems to be a fair amount of repeated code between this class and Perplexity_torch. Can be done in a later PR but can we combine the two?

sharktank/sharktank/utils/export_artifacts.py
```diff
@@ -0,0 +1,142 @@
...
+            logger.error("Error exporting mlir: ", return_code)
+        else:
+            logger.info(f"Exported to mlir successfully: {mlir_path}")
+
+    def compile_to_vmfb(
+        self,
+        mlir_path,
+        vmfb_path,
+    ):
+        compile_flags = ["--iree-hip-target=" + self.iree_hip_target]
```
IanNod, line 91: Fine for now but we do not want to hardcode for a specific backend like this in our tests. Eventually we will want to run across multiple backends so likely want this flag controlled

sharktank/tests/evaluate/perplexity_vmfb_test.py
```diff
@@ -0,0 +1,306 @@
...
+            current_perplexity["mean_perplexity"],
+            delta=self.delta,
+            msg=f"Current perplexity deviates baseline by {perplexity_difference}",
+        )
+
+    @pytest.mark.xfail(
+        reason="Non-decomposed attention is not supported yet",
+    )
+    @longrun
+    def test_llama3_8B_fp8_non_decomposed(self):
```
IanNod, line 136: nit: non-decomposed should be the default eventually, so probably no need to add the _non_decomposed part here

.github/workflows/ci_eval.yaml
archana-ramalingam: Yes

.github/workflows/ci_eval.yaml
archana-ramalingam: Like commented [above ](https://github.com/nod-ai/SHARK-Platform/pull/306#discussion_r1817084138) will remove after testing vmfb version.

IanNod (approved):

sharktank/sharktank/evaluate/perplexity_vmfb.py
```diff
@@ -0,0 +1,453 @@
...
+            seq_lens_batch=self.seq_lens_batch,
+            bs=self.bs,
+        )
+
+        seq_block_ids = self.batch.pad_block_ids()
+        prefill_logits = self.runner.ctx.modules.module[f"prefill_bs{self.bs}"](
+            token_batch,
+            self.seq_lens_batch,
+            seq_block_ids,
+            self.batch.cache_state[0].to(torch.float16),
```
sogartar, line 218: @archana-ramalingam, I think this may make a copy.
First the .to(...) may make a copy.
Then before the making the actual call to the IREE module function it will make a copy to the device if we are not targeting the CPU.

sogartar: You would need to first explicitly make a `iree.runtime.DeviceArray` and then copy back to the cache state after the call.

I have also added some IREE related functionality [here](https://github.com/nod-ai/SHARK-Platform/blob/main/sharktank/sharktank/utils/iree.py). We should unify this aspect to reduce code duplication.

PR #309, opened 2024-10-22
Export KV cache and remove flow.tensor.transfer on Tensor creation

rsuderman:
We should not be assuming a `flow.tensor.transfer` on tensor creation. This assumes that the tensor needs to be transfered to the necessary device which can mess up in place operations.


sharktank/sharktank/export_layer/export_kv_cache.py
```diff
@@ -0,0 +1,125 @@
...
+    )
+    def _(model, state, partition_0, write_page_ids: torch.Tensor) -> torch.Tensor:
+        old_state = state
+        if args.sharding > 1:
+            state = [SplitPrimitiveTensor(ts=state, shard_dim=alloc[0].shard_dim)]
+            partition_0 = SplitPrimitiveTensor(ts=partition_0, shard_dim=2)
+            write_page_ids = replicate(write_page_ids, count=args.sharding)
+        cache.write(
+            state,
+            cache_partitions=[partition_0, partition_0],
```
sogartar, line 108: Is this intentional to pass the same tensor for both k and v?

sogartar (approved):

sharktank/sharktank/export_layer/export_kv_cache.py
rsuderman: In this case yes just to minimize the function signature. In short, I wanted to be lazy :S

PR #310, opened 2024-10-23
[shortfin] Implements RandomGenerator and fill_randn.

### LINKED ISSUE #261: (sfnp) random number generation
For Stable Diffusion we want an alternative to `torch.random.manual_seed` / `torch.rand` to generate noisy image latents, e.g.:

```
def get_rand_latents(self, seed: int):
    generator = torch.manual_seed(seed)
    rand_sample = torch.randn(
        (
            self.batch_size,
            4,
            self.height // 8,
            self.width // 8,
        ),
        generator=generator,
        dtype=self.latents_dtype,
    )
    return rand_sample
```

ramblings: there's usually some expectation of determinism vs. existing txt2img solutions, given identical inputs, but the denoising entrypoint won't care about where the noisy latents come from (so client can bring their own latents if it matters).

stellaraccident:
For this, I opted to only support in-place variants since in shortfin, we generally operate directly on arrays vs forcing copies or having a lot of options in constructors for managing the device assignment. Otherwise, the definition of randn matches torch.randn.

Fixes #261.

### DIFF
--- a/shortfin/python/array_host_ops.cc
+++ b/shortfin/python/array_host_ops.cc
@@ -8,6 +8,7 @@
 #include "./utils.h"
 #include "shortfin/array/api.h"
 #include "shortfin/support/logging.h"
+#include "xtensor/xrandom.hpp"
 #include "xtensor/xsort.hpp"
 #include "xtl/xhalf_float.hpp"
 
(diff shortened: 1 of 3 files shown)


monorimet (approved):
LGTM!!

PR #313, opened 2024-10-23
Make sharded in-place ops actually in-place

sogartar:
We would insert transfers to devices on each sharded tensor
construction, which is wrong for results of in-place ops.
    
This fixes the issue for most ops except some cases of elementwise ops.
It relies on the change https://github.com/iree-org/iree-turbine/pull/239
that harmonizes the behavior of tensor transfers between eager and compiled
execution.

This fixes the cache updates for the toy-sized sharded Llama model.
Minus the smallish numerical inaccuracies, which are probably to be expected.

Added atentry/atexit hooks to the signature dispatcher.
I used them during debugging. They should be useful in the future as well.

### DIFF
--- a/sharktank/sharktank/layers/kv_cache.py
+++ b/sharktank/sharktank/layers/kv_cache.py
@@ -207,7 +207,9 @@ class PagedKVCache(BaseKVCache):
                 )
                 for shard in page_slab.shards
             ]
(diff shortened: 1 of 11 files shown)


sogartar:
I am closing this since @rsuderman came up with a better approach. See #309.

PR #317, opened 2024-10-23
Add simple cache unit test and move server test to proper location

renxida:
Also reorganizes server test to test multiple batch sizes.


kumardeepakamd (commented):
It is expected to fail at this point? I do not see any harm in getting in and then fixing the shortfin failure.

renxida:
yeah it should fail once we have amdgpu on ci, until we fix shortfin's kvcache problem

kumardeepakamd (approved):
Okay to merge

PR #318, opened 2024-10-23
Harmonize ops/tests with non-in-place transfer even for eager

sogartar:
This makes the code compatible with the change
https://github.com/iree-org/iree-turbine/pull/239

Avoid transfers in all gather/reduce for same shard/device tensors. This removes one unnecessary transfer per device.


rsuderman (approved):

sogartar:
For the sharded toy-size Llama export test I am getting a silent compilation failure during ConvertToStreamPass (iree-stream-conversion).

sogartar:
I also see some constants getting embedded into the IR. This is maybe the rotary embedding table. At least one of them.

PR #325, opened 2024-10-24
Simple KV Cache Access Tests

renxida:
This test file closely mimics the kvcache indexing, reading, and writing for open-llama-3b-v2-f16.gguf.

It's a simple test to verify that shortfin can:
- Create device arrays
- Map and fill them
- Invoke a VMFB and pass arguments to it
- Properly retain changes made by the VMFB to device arrays provided as arguments


kumardeepakamd (approved):
LG

PR #326, opened 2024-10-24
Add device affinity to sharded tensors

rsuderman:
This adds some default device affinities to loaded sharded tensors. This still needs to be updated for a configurable allocation but enables our current models to run on per-tensor parallelism.

### DIFF
--- a/sharktank/sharktank/ops/sharded_impls.py
+++ b/sharktank/sharktank/ops/sharded_impls.py
@@ -41,7 +41,7 @@ def all_gather_split(
     shards = [
         cat(
             [
-                transfer_to_logical_device(shard, i) if j != i else shard
+                transfer_to_logical_device(shard, i)
                 for j, shard in enumerate(input.shards)
             ],
             dim=dim,
@@ -62,7 +62,7 @@ def all_reduce_split_or_unreduced(
         elementwise(
             torch.add,
             *[
-                transfer_to_logical_device(shard, i) if j != i else shard
+                transfer_to_logical_device(shard, i)
                 for j, shard in enumerate(input.shards)
             ],
         )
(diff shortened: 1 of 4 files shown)


sharktank/sharktank/types/tensors.py
```diff
@@ -1135,10 +1138,21 @@ def create(
     ) -> "InferenceTensor":
         shard_count = int(extra_properties["shard_count"])
         try:
-            ts = raw_tensors[""]
+            # We have to do this to avoid exporting as part of the `mlir` blob:
```
sogartar, line 1141: Maybe this will fix what I have been observing that some constants appear in the MLIR.

sogartar (approved):

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -41,7 +41,7 @@ def all_gather_split(
     shards = [
         cat(
             [
-                transfer_to_logical_device(shard, i) if j != i else shard
+                transfer_to_logical_device(shard, i)
```
sogartar, line 44: Why did you remove this optimization?

sharktank/sharktank/types/tensors.py
rsuderman: Yes. The copying semantic drops the externalization for flow tensor transfer. I think we should ultimately drop the transfer in the ReplicateTensor construction for a `torch.Tensor`. I honestly wish we could drop it altogether but materialized constants are going to be an issue we need to track down.

sharktank/sharktank/ops/sharded_impls.py
rsuderman: It causes compilation issues with the model sadly. It happens during some fusion passes so its likely a compiler bug but we can reintroduce it later on.

PR #327, opened 2024-10-24
Split sharded Llama dataset exporting and loading in export scripts

sogartar:
Separate the 2 steps. We need exported irpa files for the IREE module anyway.

### DIFF
--- a/sharktank/sharktank/examples/export_paged_llm_v1.py
+++ b/sharktank/sharktank/examples/export_paged_llm_v1.py
@@ -72,18 +72,14 @@ def main():
     dataset = cli.get_input_dataset(args)
 
     hp = configs.LlamaHParams.from_gguf_props(dataset.properties)
-    llama_config = LlamaModelConfig(hp)
-    if args.tensor_parallelism_size > 1:
-        dataset.root_theta = shard_theta(dataset.root_theta, llama_config)
-    llama_config.use_hf = False
-    llama_config.static_tables = False  # Rely on the compiler for hoisting tables.
-    llama_config.kv_cache_type = "direct" if args.bs == [1] else "paged"
-    llama_config.attention_kernel = args.attention_kernel
-
-    # This is a bit gross and should be changed in the future. Best Idea I had so far.
-    attn_q_weight = dataset.root_theta.tensor("blk")["0"]["attn_q"]["weight"]
-    if isinstance(attn_q_weight, SplitPrimitiveTensor):
-        llama_config.tensor_parallelism_size = attn_q_weight.shard_count
+    llama_config = LlamaModelConfig(
+        hp,
+        tensor_parallelism_size=args.tensor_parallelism_size,
+        use_hf=False,
+        static_tables=False,  # Rely on the compiler for hoisting tables.
+        kv_cache_type="direct" if args.bs == [1] else "paged",
+        attention_kernel=args.attention_kernel,
+    )
 
     if llama_config.hp.expert_count:
         if llama_config.hp.model_arch == "grok":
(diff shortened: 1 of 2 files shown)


sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -72,18 +72,14 @@ def main():
...
-    llama_config.kv_cache_type = "direct" if args.bs == [1] else "paged"
-    llama_config.attention_kernel = args.attention_kernel
-
-    # This is a bit gross and should be changed in the future. Best Idea I had so far.
-    attn_q_weight = dataset.root_theta.tensor("blk")["0"]["attn_q"]["weight"]
-    if isinstance(attn_q_weight, SplitPrimitiveTensor):
-        llama_config.tensor_parallelism_size = attn_q_weight.shard_count
+    llama_config = LlamaModelConfig(
+        hp,
+        tensor_parallelism_size=args.tensor_parallelism_size,
```
rsuderman, line 77: This will not work as it assumes that the passed arg is guaranteed to match the `irpa` file. We should look into plumming sharding into the saved hyper parameters then extract. I also dislike the hack in line `84` however the solution is not to introduce mismatches.

rsuderman (changes requested):

sharktank/sharktank/examples/export_paged_llm_v1.py
sogartar: OK, I will add it to the exported irpa hyper parameters.

sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -59,31 +59,21 @@ def main():
...
-    llama_config.kv_cache_type = "direct" if args.bs == [1] else "paged"
-    llama_config.attention_kernel = args.attention_kernel
-
-    # This is a bit gross and should be changed in the future. Best Idea I had so far.
-    attn_q_weight = dataset.root_theta.tensor("blk")["0"]["attn_q"]["weight"]
-    if isinstance(attn_q_weight, SplitPrimitiveTensor):
-        llama_config.tensor_parallelism_size = attn_q_weight.shard_count
+    llama_config = LlamaModelConfig(
+        hp,
+        tensor_parallelism_size=dataset.properties["tensor_parallelism_size"],
```
rsuderman, line 71: Add a check prior for when `tensor_parallelism_size` has no value and in those cases default to `1`. We should maintain that the old non-sharded `irpa` files still work.

rsuderman (changes requested):

sharktank/sharktank/examples/export_paged_llm_v1.py
sogartar: I fixed it.

sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -59,31 +59,23 @@ def main():
...
-    llama_config.kv_cache_type = "direct" if args.bs == [1] else "paged"
-    llama_config.attention_kernel = args.attention_kernel
-
-    # This is a bit gross and should be changed in the future. Best Idea I had so far.
-    attn_q_weight = dataset.root_theta.tensor("blk")["0"]["attn_q"]["weight"]
-    if isinstance(attn_q_weight, SplitPrimitiveTensor):
-        llama_config.tensor_parallelism_size = attn_q_weight.shard_count
+    llama_config = LlamaModelConfig(
+        hp,
+        tensor_parallelism_size=dataset.properties["tensor_parallelism_size"]
```
rsuderman, line 71: Just declare it outside of `llama_config`. Giant comprehensions are bad for readability.

rsuderman (changes requested):

sharktank/sharktank/examples/export_paged_llm_v1.py
sogartar: Done

sogartar: Done

rsuderman (approved):

PR #328, opened 2024-10-24
Fix different --bs=1 behavior

renxida:
https://xilinx.slack.com/archives/C079V23F7CP/p1729805729161599

A lot of us newcomers are running into issues where we expect a bs=1 vmfb to be a drop in replacement for a bs=4 vmfb, but it's not.

I think we should do a different cmdline arg for serving local.

### DIFF
--- a/sharktank/sharktank/examples/export_paged_llm_v1.py
+++ b/sharktank/sharktank/examples/export_paged_llm_v1.py
@@ -21,6 +21,12 @@ from ..models.mixtral.mixtral import *
 from ..models.grok.grok import *
 from .. import ops
 
+import logging
+
+logging.basicConfig(
+    level=logging.INFO, format="%(asctime)s - %(name)s - %(levelname)s - %(message)s"
+)
+
 
 def main():
     from ..utils import cli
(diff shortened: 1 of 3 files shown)


sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -37,6 +43,20 @@ def main():
         help="Output file path for exported config file",
         default="/tmp/batch_llama_v1.json",
     )
+    parser.add_argument(
+        "--kv_cache_type",
+        help="Type of KV cache to use (direct or paged). Default is paged for performant serving. For local use, you may need to set this to direct.",
+        type=str,
+        choices=["direct", "paged"],
+        default="paged",
```
dan-garvey, line 51: duplicate

dan-garvey (changes requested):
Lets move this flag to the cli.py file so it can be shared and not repeated. Also, iirc paged is probably not required in the bs1 case?

sharktank/sharktank/examples/export_paged_llm_v1.py
ScottTodd, line 1: My general feeling with this script is that it is due for a rewrite. That might just involve splitting into separate scripts, one entry point for each model / variant, or something more like a fully self-contained model import pipeline from HF/GGUF --> MLIR --> VMFB+IRPA.

What users will expect is a simple process to follow for off-the-shelf and custom models that produces files which are ready for serving via shortfin. Whatever we need to do with cache types, batch sizes, etc. to get no errors, good accuracy, and good performance should all be handled for the user on the default path. For power users, we can expose tuning and customization options.

gpetters-amd: > or something more like a fully self-contained model import pipeline from HF/GGUF --> MLIR --> VMFB+IRPA.

Are you envisioning something like turbine-models' export scripts or something more akin to an SD pipeline class? I've refactored some of the export_paged_llm_v1 code into a dirty export script I've been working off of that I can clean up if that's the direction we want to move toward.

renxida:
Superceded by #330 thanks Dan!

sharktank/sharktank/examples/export_paged_llm_v1.py
ScottTodd: I'm thinking of something between what we have today and https://docs.vllm.ai/en/latest/getting_started/quickstart.html#offline-batched-inference

image

or https://nvidia.github.io/TensorRT-LLM/quick-start-guide.html
image

PR #331, opened 2024-10-25
(shortfin-apps)(sd) Rework batching and structuring of image generation requests.

monorimet:
This change also includes an update to the README for the SD server.

The motivation behind this patch is to enable pipelining efforts and keep execution seamless and linear. It also makes the batcher actually work, instead of just letting any requests batch with a prayer for density in program config / invocation parameters.

It also:
- pipes through a simple option for specifying device indices 
- cleans up a bit of the client generation request struct.
- adds a bs8 example request.


stellaraccident (approved):
Looking good!

PR #332, opened 2024-10-25
[shortfin] Implement fiber local program forking.

stellaraccident:
API changes:

* sf.Program(..., fiber=) no longer takes a fiber. The fiber is now moved to the invocation (i.e. function(..., fiber=)
* sf.Program now requires an explicit devices= (as opposed to implicitly inferring this from the fiber). The prior behavior can be achieved by sf.Program(, devices=fiber.raw_devices).

Semantic changes:

* Programs are now created with an isolation level (NONE, PER_FIBER, PER_CALL), defaulting to PER_FIBER. The prior behavior was like NONE.
* In PER_FIBER mode, only one in-flight call to a program can happen at a time on a single fiber. When a program is used across fibers, each fiber will get their own context fork (which is a light-weight clone of the context that shares immutable data).
* In PER_CALL mode, backing contexts are pooled on each calling fiber and each call gets their own context. This supports multiple concurrent invocations to a context at the expense of each call having its own mutable state (i.e. this is only suitable for immutable programs).

### DIFF
--- a/shortfin/python/lib_ext.cc
+++ b/shortfin/python/lib_ext.cc
@@ -248,8 +248,9 @@ void PyAddProgramInvocationArg(py::capsule &inv_capsule, py::handle arg) {
 }
 
 local::ProgramInvocation::Future PyFunctionCall(local::ProgramFunction &self,
-                                                py::args args) {
-  auto inv = self.CreateInvocation();
+                                                py::args args,
+                                                local::Fiber &fiber) {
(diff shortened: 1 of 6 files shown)


monorimet (approved):
We will want to bump the required IREE version for shortfin to include the context forking support.
Otherwise, LGTM.

stellaraccident:
> We will want to bump the required IREE version for shortfin to include the context forking support. Otherwise, LGTM.

Someone already did it a few days ago. Came in with a branch merge.

PR #338, opened 2024-10-26
fix missing 1st token from prefill in inference response

renxida:
The 1st output token is generated by prefill. It was missing.


kumardeepakamd (approved):

renxida:
will be superceded in another off-by-1 fix pr

PR #340, opened 2024-10-26
Revise and split requirements files

marbre:
Moves dependencies defined in the root `requirements.txt` to `sharktank/` and splits out test only deps to `requirements-tests.txt` file. Dependencies only used for development / in the CI are moved to `requirements-dev.txt` and the root requirements file is now used to pull in all the deps. Furthermore, some no longer used dependencies are removed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,30 +1,4 @@
-# Runtime deps.
-gguf==0.6.0
-numpy==1.26.3
-onnx==1.15.0
-
-# Model deps.
-huggingface-hub==0.22.2
-transformers==4.40.0
-sentencepiece==0.2.0
-
-# It is expected that you have installed a PyTorch version/variant specific
-# to your needs, so we only include a minimum version spec.
-# TODO: Use a versioned release once 2.3.0 drops.
-torch>=2.3.0.dev1
-
-# Used for managing pre-commit flows.
-pre-commit
-
-# Type checking
-mypy==1.8.0
-types-requests==2.31.0.20240125
-
-# Testing
-parameterized
-pytest==8.0.0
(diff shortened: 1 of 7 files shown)


marbre:
This will need some further work but hopefully is the right step to untangle all the deps.

requirements.txt
```diff
@@ -1,30 +1,4 @@
...
-pytest==8.0.0
-pytest-xdist==3.5.0
-
-# Serving deps.
-fastapi==0.112.2
-uvicorn==0.30.6
+-r sharktank/requirements.txt
+-r sharktank/requirements-tests.txt
+-r shortfin/requirements-tests.txt
+-r requirements-dev.txt
```
ScottTodd, line 4: Interesting... top level requirements that pulls in all the others. That seems like a good idea. I hope it works out in practice :P

sharktank/setup.py
```diff
@@ -99,7 +99,6 @@ def initialize_options(self):
     extras_require={
         "testing": [
             f"pytest{get_version_spec('pytest')}",
-            f"pytest-xdist{get_version_spec('pytest-xdist')}",
```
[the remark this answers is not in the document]
marbre: `pytest-xdist` isn't used, as `pytest -n` is only called in the CI. Thus I would rather go with the minimal requirements instead of pulling in non-strictly required packages.

requirements.txt
marbre: Our CI says yes ;) Well I hope to get rid off it at some point to be honest but this would mean to refactor the CI at the same time whereas this relaxes the required changes a little and let us iterate more easily.

ScottTodd (approved):

PR #341, opened 2024-10-26
Addition of booleans is currently wrong in iree-compile

rsuderman:
Addition of booleans is performing xor which breaks causal mapping.

### DIFF
--- a/sharktank/sharktank/layers/causal_llm.py
+++ b/sharktank/sharktank/layers/causal_llm.py
@@ -127,9 +127,10 @@ class BaseCausalLMModel(ThetaLayer):
         dtype = self.attention_dtype
         _, batch_seq_len = input_mask.shape
         causal_mask = causal_context_mask[:, :, :batch_seq_len, :batch_seq_len]
-        boolean_mask = causal_mask + input_mask[:, None, None, :]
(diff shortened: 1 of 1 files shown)


sharktank/sharktank/layers/causal_llm.py
```diff
@@ -95,10 +95,9 @@ def input_mask(
 
     def decode_attention_mask(self, boolean_input_mask: torch.Tensor):
         dtype = self.attention_dtype
-        numeric_mask = torch.zeros_like(boolean_input_mask, dtype=dtype)
-        numeric_mask.masked_fill_(
-            boolean_input_mask, self._maximally_negative_value(dtype)
-        )
+        numeric_mask = torch.where(
```
stellaraccident, line 98: I'm not opposed to a war, but it seems like we should fix something here at it's core... Or at least file an issue

rsuderman: Going to write a fix for this right now. Just need to land the workaround to avoid people being blocked.

renxida (approved):

rsuderman:
Upstream fix here: https://github.com/llvm/torch-mlir/pull/3820

PR #343, opened 2024-10-27
(shortfin-sd) add e2e test + fixes for batched requests

monorimet:
Put some railings around concurrent program invocations, fix batched image responses.

This lets us do more without relying on iree vm context forking, ensuring our shared resources are being locked when appropriate. When context forking support lands, we should enable a knob to increase the number of inference proc concurrencies per fiber.

I'm not too sure about the use cases for these large input batches... we'll need to stress test a stream of many client requests, as that's a more likely scenario.

image


monorimet:
Switched SD scheduler module sourcing over to a version that enables dynamic num_inference_steps by use of static caches for small, data-dependent values.

This generally reduces the cases for which we consider requests disjoint.

Example of what this allows without any extra headache:
image

.github/workflows/ci-sdxl.yaml
```diff
@@ -0,0 +1,103 @@
...
+  cancel-in-progress: true
+
+env:
+  IREE_REPO_DIR: ${{ github.workspace }}/iree
+  LIBSHORTFIN_DIR: ${{ github.workspace }}/shortfin/
+
+jobs:
+  build-and-test:
+    name: Build and test
+    runs-on: ubuntu-24.04
```
monorimet, line 40: @saienduri What will it take to get this .yml spinning on a MI300x?

monorimet: I noticed some other jobs running on a llama MI300 runner, can we do the same for SDXL?

saienduri: Yeah we can. Do we want SPX? Also how long would a job take (seems pretty quick) and how often would it run?

monorimet: SPX is ok to start, we can parametrize a bit once we get larger scenario tests/benchmarks running nightly. The e2e test here should take under a minute to complete, but we will want a place with room to grow. maybe a small suite (splat weights, small scenarios) for pre-submit and a slow suite (like mlperf CI) that runs nightly?

marbre (commented):
Quick question, how does `.github/workflows/ci-sdxl.yaml` differ from `.github/workflows/ci_linux_x64-libshortfin.yml`? Without a diff seems pretty similar.

monorimet:
> Quick question, how does `.github/workflows/ci-sdxl.yaml` differ from `.github/workflows/ci_linux_x64-libshortfin.yml`? Without a diff seems pretty similar.

They are essentially the same in current form, save for the pytest command. The idea is that this will soon include model exports from sharktank and perhaps run on different triggers and/or runners.

marbre:
> > Quick question, how does `.github/workflows/ci-sdxl.yaml` differ from `.github/workflows/ci_linux_x64-libshortfin.yml`? Without a diff seems pretty similar.
> 
> They are essentially the same in current form, save for the pytest command. The idea is that this may soon pull in model exports from sharktank and perhaps run on different triggers and/or runners.

We've testing on AMDGPU on our list. But rather than duplicating `.github/workflows/ci_linux_x64-libshortfin.yml` a matrix strategy which allows to run (at least) on different runners seems more suitable :)

.github/workflows/ci-sdxl.yaml
saienduri: Ok cool, free to get started with this label: `mi300-sdxl-kernel`. This is shared with the kernel benchmarking repo, but other than a nightly there isn't too much traffic there. When we are ready for a nightly here, we can find a good time to schedule it

saienduri:
I think this sdxl testing suite will be growing in what it tests (API calls, exports, etc.) which are sdxl specific. Seems fine to be in its standalone file

shortfin/python/shortfin_apps/sd/components/generate.py
```diff
@@ -95,12 +96,8 @@ async def run(self):
...
             # TODO: stream image outputs
             logging.debug("Responding to one shot batch")
-            out = io.BytesIO()
-            result_images = [p.result_image for p in gen_processes]
-            for idx, result_image in enumerate(result_images):
-                out.write(result_image)
-                # TODO: save or return images
-                logging.debug("Wrote images as bytes to response.")
-            self.responder.send_response(out.getvalue())
+            response_data = {"images": [p.result_image for p in gen_processes]}
```
stellaraccident, line 99: A couple of notes to self here: a) Need to get tracing tags up to this level so we can see specific overheads, b) need to add a general purpose IO task queue so we can easily offload stuff like this to a side-thread

shortfin/python/shortfin_apps/sd/components/service.py
```diff
@@ -48,15 +50,17 @@ def __init__(
...
         self.workers = []
         self.fibers = []
+        self.locks = []
         for idx, device in enumerate(self.sysman.ls.devices):
             for i in range(self.procs_per_device):
                 worker = sysman.ls.create_worker(f"{name}-inference-{device.name}-{i}")
                 fiber = sysman.ls.create_fiber(worker, devices=[device])
                 self.workers.append(worker)
                 self.fibers.append(fiber)
+                self.locks.append(asyncio.Lock())
```
stellaraccident, line 63: huh - didn't know this worked yet.

shortfin/python/shortfin_apps/sd/components/service.py
```diff
@@ -514,17 +525,17 @@ async def _denoise(self, device, requests):
                 "".join([f"\n  {i}: {ary.shape}" for i, ary in enumerate(unet_inputs)]),
             )
             await device
-            (noise_pred,) = await fns["unet"](*unet_inputs)
+            (noise_pred,) = await fns["unet"](*unet_inputs, fiber=self.fiber)
             await device
```
stellaraccident, line 529: We should be able to take all of these explicit device waits out after my patch (#350) lands. We could use some more framework tests on dependencies to be sure but should be ok (all of my examples to you guys had waits in them so as to not trip you up while things were getting worked out).

Leave in for now and we can come back.

shortfin/python/shortfin_apps/sd/components/service.py
```diff
@@ -591,5 +603,8 @@ async def _postprocess(self, device, requests):
             # TODO: reimpl with sfnp
             permuted = np.transpose(req.image_array, (0, 2, 3, 1))[0]
             cast_image = (permuted * 255).round().astype("uint8")
-            req.result_image = Image.fromarray(cast_image).tobytes()
+            image_bytes = Image.fromarray(cast_image).tobytes()
```
stellaraccident, line 606: Another thing that should go to an IO thread pool.

shortfin/tests/apps/sd/e2e_test.py
```diff
@@ -0,0 +1,201 @@
...
+            img = bytes_to_img(item.encode("utf-8"), idx, width, height)
+            imgs.append(img)
+
+    except requests.exceptions.RequestException as e:
+        print(f"Error sending the request: {e}")
+
+    return imgs, response.status_code
+
+
+def find_free_port():
```
stellaraccident, line 185: This and the above has been copied a couple of times now. Probably need a test_utils of some kind.

stellaraccident (approved):
Looks good. Fair bit of low hanging fruit, but let's get this in/tested and then grind on some tracing and perf improvements.

monorimet:
FYI @marbre @stellaraccident I have changed the sharktank unit tests to always trigger since we require the job to pass. If we do not trigger a required workflow, it will prevent the PRs from merging normally.

An alternative is to create a duplicate workflow with the same appearance that auto-passes, fulfilling the requirement without running unnecessary jobs. I wasn't able to get this working and it seems hacky.

ScottTodd:
> FYI @marbre @stellaraccident I have changed the sharktank unit tests to always trigger since we require the job to pass. If we do not trigger a required workflow, it will prevent the PRs from merging normally.
> 
> An alternative is to create a duplicate workflow with the same appearance that auto-passes, fulfilling the requirement without running unnecessary jobs. I wasn't able to get this working and it seems hacky.

In general, please make such changes in their own small PRs so we can review changes more easily and have a clean commit history for each project component. This PR has lots of commits touching different areas.

For this particular issue (required checks not being compatible with `paths` filters, I've only found a few workflows to be effective:
1. Use path filters but not required checks. This puts the responsibility on authors and reviewers to check for workflow results before merging. It also makes it difficult to trust auto-merge, since that will not wait for non-required checks.
2. Always run all checks. This can waste CI resources (particularly if jobs are slow or use expensive runners) but is predictable.
3. Always run the checks, but move the path filters into a workflow, like https://github.com/iree-org/iree/blob/main/.github/workflows/ci.yml, https://github.com/iree-org/iree/blob/main/.github/workflows/setup.yml, and https://github.com/iree-org/iree/blob/main/build_tools/github_actions/configure_ci.py

renxida:
This broke shortfin build for me, @vinayakdsci , and Stephen.

Looks like CMAKE doesn't support commit hashes when they're a few commits away from HEAD.

PR #345, opened 2024-10-27
[tuner] Generalizes Configuration dataclass

mihaescuvlad:
### Notes:
- Splits `Configuration` into `LLVMGPUConfiguration` & `LLVMCPUConfiguration`
- Adds `device_target` to the `MlirRegex` dataclass
- Adds `*ConfigurationBuilder` for abstracting the building of configurations
- Adds `ConfigurationBuilderFactory` to fetch the proper configuration builder based on target architecture

### Tests:
N/A


tuner/tuner/candidate_gen.py
```diff
@@ -187,20 +187,232 @@ def __str__(self) -> str:
...
 
 @dataclass
-class Configuration:
+class BaseConfiguration:
     subgroup_size: int
     workgroup_size: list[int]
-    intrinsic: MfmaIntrinsic
     tile_sizes: list[int]
     subgroup_m_count: int
     subgroup_n_count: int
```
kuhar, line 195: I think that all the base configuration needs is tile sizes? For example on CPU this would be the only thing to tune. `subgroup_m_count` and `subgroup_n_count` are specific to `LLVMGPUVectorDistribute`

tuner/tuner/candidate_gen.py
```diff
@@ -187,20 +187,232 @@ def __str__(self) -> str:
...
+
+
+@dataclass
+class LLVMCPUConfiguration(BaseConfiguration):
+    def get_pipeline_config(self) -> str:
+        return ""
+
+
+@dataclass
+class BaseConfigurationBuilder:
```
kuhar, line 293: Why do we need these builders?

kuhar (commented):
I think the key consideration should be how to generate the constraints. I think this should belong to each configuration: given a problem size and dispatch kind, it should know how to generate constraints and perform the lookup in the model generated by Z3 to apply the assignment

tuner/tuner/candidate_gen.py
mihaescuvlad: That was one of the only ways of creating the configurations that I could think of while trying to avoid hacks. I tried using a strategy pattern before, but it was slightly less clear than using builders + factory. It's still not a perfect approach though, because the `generate_solutions` method still loses some of the clarity. I'm considering moving that method inside of the configuration dataclasses and getting rid of the builder while keeping a factory for creating the configurations

PR #346, opened 2024-10-28
[sharktank] Revise `setup.py` file

marbre:
* Move `version_info.json` to sharktank subdir
* Pulls in the local requirements file for sharktank


sharktank/setup.py
```diff
@@ -13,7 +13,7 @@
 
 THIS_DIR = Path(__file__).resolve().parent
 REPO_DIR = THIS_DIR.parent
-VERSION_INFO_FILE = REPO_DIR / "version_info.json"
+VERSION_INFO_FILE = THIS_DIR / "version_info.json"
```
ScottTodd, line 16: shortfin is using this too:
https://github.com/nod-ai/SHARK-Platform/blob/ffb146bdb3442c1d6d3a7417f334181a5834d117/shortfin/setup.py#L143-L144

Should we have a separate source of version information for each package, or require that they share a version?

marbre: Yes, let's have a separate one please to be able to raise the patch level individually. But I am fine with having it more similar to shortfin. Will adopt accordingly.

ScottTodd: I'm confused by git/github here. Am I missing something about your updates since my comment?

shortfin/setup.py should also be updated and a second version_info.json should be added

marbre: Seems more likely that I am confused. I assumed that `SOURCE_DIR` you pointed to would refer to `shortfin/` and that the way to get the version in shortfin works (as is). If this is the case shortfin is indeed missing a `version_info.json`. If that's not the case my assumption that `SOURCE_DIR` refers to `shortfin` is plainly wrong. Let me take a second look...

ScottTodd: Heh. We're both confused 😅. At least things should be more clear once we have the two files.

marbre: Okay, iterated further on the setup.py files and altered the PR / title description accordingly.

ScottTodd: Oooh wait, I'd forgotten about https://github.com/nod-ai/SHARK-Platform/blob/main/.github/workflows/build_packages.yml. That may need some updates too, but we'll be reworking it with new version information anyways.

What that does now is it generates a single `version_info.json` file for shortfin to use then extracts that file into the `shortfin/` directory.

What we should end up with is a local file that determines the `major.minor.patch` then the workflows will append a suffix into that version.

marbre: Okay, let me roll back the changes applied to shortfin. We can harmonize these in a follow up.

marbre:
This should be in a state that it can be landed as is. Happy if you could stamp @ScottTodd.

ScottTodd (approved):

PR #348, opened 2024-10-28
[shortfin] Make error copyable.

stellaraccident:
* Just eagerly serializes the exception. Trying to defer this makes the type non copyable and is a dubious optimization given that we never use this type for flow control.
* Should fix MSVC warnings and ill-defined behavior there.


ScottTodd:
Thanks!

PR #350, opened 2024-10-29
[shortfin] Finishes plumbing ProgramIsolation through to the Python API.

stellaraccident:
* Fixes an egregious bug on wait fence setup (that was causing invocations to always have an empty wait set) discovered while testing these more advanced modes.
* Adds `isolation=` to `Program` constructor and invocation APIs (for fine-grained control).
* Adds `ProgramIsolation` enum.
* Adds `Program.isolation` property.
* Adds `ProgramFunction.isolation` property.
* Overhauls the mobilenet invocation test to exercise several newly available concurrency variants.


stellaraccident:
Select traces (in a debug build - timing not representative):

test_invoke_mobilenet_parallel_per_call: https://drive.google.com/file/d/11woACOZUxSVlJqRS-IAU94WgiDJHWxGR/view?usp=sharing
test_invoke_mobilenet_multi_fiber_per_fiber: https://drive.google.com/file/d/1B7OLJZBRtZALfWS-NpHYou3ghlYHdKF3/view?usp=sharing

These represent the two easiest ways to fork work. The first shows the effect of same-fiber, concurrent host. The second shows multi-fiber, fully overlapped. All of these tests run on one host thread.

monorimet (approved):
Thanks, this makes it much more tangible. LGTM

monorimet:
Trying this on SDXL.

Any idea why this program invocation might fail this way when using PER_CALL program isolation? CLIP works but the SDXL scheduler (doesn't use params, lightweight fns) gives a runtime error:
```
[2024-10-29 08:49:36.101] [error] [service.py:307] Fatal error in image generation
Traceback (most recent call last):
  File "/home/eagarvey/SHARK-Platform/shortfin/python/shortfin_apps/sd/components/service.py", line 296, in run
    await self._denoise(device=device0, requests=self.exec_requests)
  File "/home/eagarvey/SHARK-Platform/shortfin/python/shortfin_apps/sd/components/service.py", line 494, in _denoise
    (latents, time_ids, timesteps, sigmas) = await fns["init"](
                                                   ^^^^^^^^^^^^
ValueError: iree/runtime/src/iree/base/allocator.c:22: INVALID_ARGUMENT; allocator has no control routine
```
Also seems to happen if I do multiple fibers per device and run as PER_FIBER, but only on the scheduler function...

stellaraccident:
> Trying this on SDXL.
> 
> Any idea why this program invocation might fail this way when using PER_CALL program isolation? CLIP works but the SDXL scheduler (doesn't use params, lightweight fns) gives a runtime error:
> 
> ```
> [2024-10-29 08:49:36.101] [error] [service.py:307] Fatal error in image generation
> Traceback (most recent call last):
>   File "/home/eagarvey/SHARK-Platform/shortfin/python/shortfin_apps/sd/components/service.py", line 296, in run
>     await self._denoise(device=device0, requests=self.exec_requests)
>   File "/home/eagarvey/SHARK-Platform/shortfin/python/shortfin_apps/sd/components/service.py", line 494, in _denoise
>     (latents, time_ids, timesteps, sigmas) = await fns["init"](
>                                                    ^^^^^^^^^^^^
> ValueError: iree/runtime/src/iree/base/allocator.c:22: INVALID_ARGUMENT; allocator has no control routine
> ```
> 
> Also seems to happen if I do multiple fibers per device and run as PER_FIBER, but only on the scheduler function...

Something nefarious going on - seems like a null getting in somewhere. Do you have a conservative path to land this that I can debug on?

stellaraccident:
As discussed, going to land this but not opt the uses into the new feature until it is ready.

PR #351, opened 2024-10-29
Fix shortfin gibberish tokens and add cpu_llm_server_test.py

renxida:
Rob and I found some weirdness in paged_llm_v1 that was inconsistent with how shortfin worked.

We hacked a solution for now but we really need to fix paged_llm_v1 by cleaning up the seq_lens and start_positions and adding some CLEAR documentation as to EXACTLY what they do.

Currently we're able to make the llm count.

Right now the context length is effectively 16 tokens. By adding cache allocation, we'll soon be able to actually talk with the LLM.


renxida:
To test these changes, run: https://gist.github.com/renxida/d40e14b1c9b4606279d19899bcce5e9a

shortfin/python/shortfin_apps/llm/components/generate.py
```diff
@@ -54,10 +54,12 @@ async def run(self):
...
             token_int = token.items[0]
 
+            self.append_token(token_int)
             # Decode loop.
-            # TODO: Use correct eot token from config.
             exec.start_position = len(self.input_token_ids) - 1
-            while token_int != 128001:
+            # TODO: Use correct eot token from config.
+            # while token_int != 128001:
+            for i in range(15):
```
kumardeepakamd, line 62: needs a proper fix to iterate over a variable number of tokens. okay to land and iterate.

kumardeepakamd (approved):

Issue #342, opened 2024-10-26
(shortfin-sd) choice of device visibility specifier impacts program loading performance


monorimet:
I've noticed a discrepancy in module load/init performance across different methods of setting amdgpu system builder device visibility.

Observed as follows:
 - Setting env var `HIP_VISIBLE_DEVICES=0` -> initialization is lightning quick. SD server starts in ~2 seconds.
 - Setting env var `SHORTFIN_AMDGPU_VISIBLE_DEVICES=0` -> initialization takes anywhere from 10 to 20 seconds.
 - Using the CLI option `--device_ids=0` -> initialization takes anywhere from 10 to 20 seconds.

AWoloszyn:
My first instinct is that this is not something we can easily fix in the runtime.  (We can open a bug against HIP for this). In general allocating memory (which is done during load) updates the page tables for *all* hip devices visible to the application (even if they are not being used), so the cost of page table updates increases. I would expect that to be more or less linear with the number of devices.

AWoloszyn:
Now that loads are typically done via file reads instead of mapping memory this should largely be solved in the latest version of shortfin. @monorimet Let me know if you are still running into the issue.

PR #360, opened 2024-10-29
(shortfin-sd) Adds program isolation optionality and fibers_per_device.

monorimet:


monorimet:
This sets HIP_VISIBLE_DEVICES from the .yaml -- it can be exposed as a pytest option as well (it's an option for the server CLI) but this is easiest and most obvious. https://github.com/nod-ai/SHARK-Platform/issues/342 describes why this method was chosen for the runner (it's a little faster to init)

.github/workflows/ci-sdxl.yaml
```diff
@@ -99,4 +99,4 @@ jobs:
       working-directory: ${{ env.LIBSHORTFIN_DIR }}
       run: |
         ctest --timeout 30 --output-on-failure --test-dir build
-        pytest tests/apps/sd/e2e_test.py -v -s --system=amdgpu
+        HIP_VISIBLE_DEVICES=0 pytest tests/apps/sd/e2e_test.py -v -s --system=amdgpu
```
ScottTodd, line 102: > This sets HIP_VISIBLE_DEVICES from the .yaml -- it can be exposed as a pytest option as well (it's an option for the server CLI) but this is easiest and most obvious. #342 describes why this method was chosen for the runner (it's a little faster to init)

I think @saienduri also did something with `HIP_VISIBLE_DEVICES` for having multiple github actions runner instances on the same node, with one runner per GPU.

monorimet: This doesn't seem to be restricting the visible devices otherwise -- test failed due to a known multi-gpu bug.

monorimet: I think it's better to move this inside anyway, if anything, to test the server CLI option.

shortfin/python/shortfin_apps/sd/components/service.py
```diff
@@ -50,17 +59,20 @@ def __init__(
...
+        self.fiber_status = []
         for idx, device in enumerate(self.sysman.ls.devices):
-            for i in range(self.procs_per_device):
+            for i in range(self.fibers_per_device):
                 worker = sysman.ls.create_worker(f"{name}-inference-{device.name}-{i}")
                 fiber = sysman.ls.create_fiber(worker, devices=[device])
                 self.workers.append(worker)
                 self.fibers.append(fiber)
-                self.locks.append(asyncio.Lock())
+                self.fiber_status.append(0)
```
monorimet, line 75: I'm not in love with this, but haven't dreamed up a different way about it yet.

stellaraccident: Yeah, we're missing a data structure. If I read this right, you just want to be able to pick a free fiber, right? There should really be some kind of a Pool or something which we don't have yet, but you can fake it with a simple fiber idle list: all available fibers go on the idle_list. Then when you need one, you pop and have the fiber put itself back when done. Something like that. You'd typically use a data structure that can yield if none are available but I think you are somehow never managing to underflow here? The underflow blocking could be faked today with a Queue or something like that.

stellaraccident (approved):
Ok, let's give this a go and see where it takes us. I think we'll end up landing on a simpler set of options but we can break it down then.

.github/workflows/ci-sdxl.yaml
saienduri: @monorimet can you try using ROCR_VISIBLE_DEVICES. That's what we've been using in IREE and SHARK-TestSuite

PR #361, opened 2024-10-29
Expand testing to cover Python 3.11

marbre:
Further replaces `PurePath.relative_to` as the `walk_up` parameter was added with Python 3.12 and the old behavior defaults to `False`.

Progress on https://github.com/nod-ai/SHARK-Platform/issues/357

### DIFF
--- a/.github/workflows/ci_linux_x64-libshortfin.yml
+++ b/.github/workflows/ci_linux_x64-libshortfin.yml
@@ -38,6 +38,9 @@ jobs:
   build-and-test:
     name: Build and test
     runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        python-version: ["3.11", "3.12"]
 
     steps:
     - name: Install dependencies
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_linux_x64-libshortfin.yml
```diff
@@ -38,6 +38,9 @@ jobs:
   build-and-test:
     name: Build and test
     runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        python-version: ["3.11", "3.12"]
```
ScottTodd, line 43: Is anything stopping us from expanding all the way to 3.10 and 3.13 too? Fine to go incrementally too.

ScottTodd (approved):

.github/workflows/ci_linux_x64-libshortfin.yml
marbre: The tests hang on 3.10 (see https://github.com/marbre/SHARK-Platform/actions/runs/11582552347/job/32245832805) and intended to not push a workflow that fails by default ;)

3.13 is already tested in a different workflow with free-threading support but I will add it here as well. However, this should be _without_ free-threading as this feature is requested but seems it was not shipped yet (see https://github.com/actions/setup-python/issues/771).

marbre: I skip 3.13 in this workflow for now as some pre-build wheels are not available and those fail building without further effort:
>  ERROR: Failed building wheel for onnx

PR #363, opened 2024-10-29
Fix Llama Benchmarking 8b f16 tests to run on presubmit

aviator19941:
Modify all the changes merged into the sharktank repo to make Llama Benchmarking test work.

### DIFF
--- a/.github/workflows/ci-llama.yaml
+++ b/.github/workflows/ci-llama.yaml
@@ -8,6 +8,7 @@ name: Llama Benchmarking Tests
 
 on:
   workflow_dispatch:
+  pull_request:
   schedule:
     # Weekdays at 5:00 AM UTC = 10:00 PM PST.
     - cron: "0 5 * * 1-5"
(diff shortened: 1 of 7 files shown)


dan-garvey (approved):
I think eventually we will want to move some of these to nightly and only run some on presubmit, but this looks good for now, assuming it passes ci = D.

dan-garvey:
@sogartar fyi

.github/workflows/ci-llama.yaml
```diff
@@ -8,6 +8,7 @@ name: Llama Benchmarking Tests
 
 on:
   workflow_dispatch:
+  pull_request:
   schedule:
     # Weekdays at 5:00 AM UTC = 10:00 PM PST.
     - cron: "0 5 * * 1-5"
```
ScottTodd, line 14: Why run on both `pull_request` and `schedule`? Usually we would run on `pull_request` and `push` _or_ `schedule`, not this combination.

See https://iree.dev/developers/general/github-actions/#workflow-triggers for conventions we use in IREE

PR #368, opened 2024-10-29
VMFB buffer access test

renxida:
This test mimics the kvcache indexing, reading, and writing in open-llama-3b-v2-f16.gguf.

  It's a simple test to verify that shortfin can:
  - Create device arrays
  - Map and fill them
  - Invoke a VMFB and pass arguments to it
  - Properly retain changes made by the VMFB to device arrays provided as arguments


renxida:
used to be #325, which was reverted because it broke CI

renxida:
Compared to #325 this has 2 code changes:
- no more np dependency
- skipped for windows
and some name / docstring /comment changes.

kumardeepakamd (commented):
We should try to get it fixed for windows before merging.

kumardeepakamd (approved):
Thanks for enabling it for Windows too.

PR #370, opened 2024-10-30
Set paths to not run on unrelated PRs

marbre:


.github/workflows/ci-llama.yaml
```diff
@@ -9,6 +9,10 @@ name: Llama Benchmarking Tests
 on:
   workflow_dispatch:
   pull_request:
+    paths:
+      - '.github/workflows/ci-llama.yaml'
+      - 'sharktank/**'
+      - '*requirements*.txt'
   schedule:
```
ScottTodd, line 16: I left a comment on https://github.com/nod-ai/SHARK-Platform/pull/363#discussion_r1822842474. I'd rather see this run on _either_ `schedule` _or_ `pull_request, push`.

The workflow name also isn't consistent with others (`ci-llama` but `Llama Benchmarking Tests` name doesn't sort along with the others at https://github.com/nod-ai/SHARK-Platform/actions/workflows/ci-llama.yaml)

.github/workflows/ci-llama.yaml
```diff
@@ -9,6 +9,10 @@ name: Llama Benchmarking Tests
 on:
   workflow_dispatch:
   pull_request:
+    paths:
+      - '.github/workflows/ci-llama.yaml'
+      - 'sharktank/**'
+      - '*requirements*.txt'
```
ScottTodd, line 15: I think for the moment we should actually remove all these path filters from the workflows and start switching more checks to be required. Too many developers are ignoring CI results and relying on auto-merge, wasting multiple developer-hours a week on reverts and fix-forwards.

We could build a system like https://iree.dev/developers/general/github-actions/#opt-in-for-presubmit-jobs to get the best of both worlds (filtering _and_ path filters), since GitHub's default implementation is completely broken :|

ScottTodd (changes requested):

.github/workflows/ci-llama.yaml
ScottTodd: Here's what I mean with sorting: 
image

.github/workflows/ci-llama.yaml
marbre: I generally agree, but don't have a good feeling yet which workflows would be a good choice to be run on every PR. However, I clearly see the point. Closing this therefore.

.github/workflows/ci-llama.yaml
ScottTodd: We're going to have more models to test and benchmark, so I'd like us to get better organized about how we define workflows.

.github/workflows/ci-llama.yaml
ScottTodd: Every workflow that runs in less than 30 minutes.

PR #373, opened 2024-10-30
CPU LLM Integration Test

stbaione:
- Adds integration test for CPU LLM Server (will make sure all CIs are passing before merge)
- Small fix in `shortfin/python/shortfin_apps/llm/components/service.py` to re-enable llm serving after recent changes


shortfin/tests/apps/llm/cpu_llm_server_test.py
```diff
@@ -0,0 +1,156 @@
...
+        # Export model if it doesn't exist
+        mlir_path = os.path.join(tmp_dir, "model.mlir")
+        config_path = os.path.join(tmp_dir, "config.json")
+        if not os.path.exists(mlir_path) or not os.path.exists(config_path):
+            bs_string = ",".join(map(str, BATCH_SIZES))
+            subprocess.run(
+                [
+                    "python",
+                    "-m",
+                    "sharktank.examples.export_paged_llm_v1",
```
ScottTodd, line 55: At the moment, sharktank and shortfin are self-contained subprojects. The CI workflows build one _or_ the other, not both.

This is an integration test that uses both, so it should live outside of `shortfin/`. Two ideas for that:
* A new top level `tests/` directory for integration tests
* A new top level `shark_platform/` directory for the combination of all packages, like what I started on in https://github.com/nod-ai/SHARK-Platform/pull/244

Either way, we'll need a new workflow file instead of the existing `ci.*shortfin.*yml` or `ci.*sharktank.*yml` files.

See also https://github.com/nod-ai/SHARK-Platform/pull/284#discussion_r1803610173

> This can be separate from this PR, but we should use an in-memory API for all model exporting/importing workflows, rather than fork into a subprocess.

shortfin/tests/apps/llm/cpu_llm_server_test.py
```diff
@@ -0,0 +1,156 @@
...
+                    f"--gguf-file={model_path}",
+                    f"--output-mlir={mlir_path}",
+                    f"--output-config={config_path}",
+                    f"--bs={bs_string}",
+                ],
+                check=True,
+            )
+        # Compile model if it doesn't exist
+        vmfb_path = os.path.join(tmp_dir, "model.vmfb")
+        if not os.path.exists(vmfb_path):
```
ScottTodd, line 65: Caching in tests can be tricky. If the user updates the compiler, they might want the test to use the newer version of the tool.

Please be very deliberate here.
* Which files are going to persistent paths and which are going to temporary paths?
* Should the developer running this test be able to control the cache behavior, e.g. by explicitly clearing cached files at the start or end of the test?
* Log when using a cached version
* Log when compiling (especially if compilation takes more than 10 seconds)
* Log the full command being run

shortfin/tests/apps/llm/cpu_llm_server_test.py
```diff
@@ -0,0 +1,156 @@
...
+            f"--tokenizer={os.path.join(model_test_dir, 'tokenizer.json')}",
+            f"--model_config={os.path.join(model_test_dir, 'edited_config.json')}",
+            f"--vmfb={os.path.join(model_test_dir, 'model.vmfb')}",
+            f"--parameters={os.path.join(model_test_dir, 'open-llama-3b-v2-f16.gguf')}",
+            f"--device={settings['device']}",
+        ]
+    )
+    # Wait for server to start
+    time.sleep(2)
+    yield server_process
```
ScottTodd, line 114: Sketchy timeout. Might be flaky on CI. Is there a more reliable way to wait for startup?

shortfin/tests/apps/llm/cpu_llm_server_test.py
```diff
@@ -0,0 +1,156 @@
...
+        "return_logprob": False,
+        "logprob_start_len": -1,
+        "top_logprobs_num": 0,
+        "return_text_in_logprobs": False,
+        "stream": False,
+    }
+    print("Prompt text:")
+    print(data["text"])
+    BASE_URL = "http://localhost:8000"
+    response = requests.post(f"{BASE_URL}/generate", headers=headers, json=data)
```
ScottTodd, line 136: Is this port 8000 configurable? What happens if the port is in use? Will the test fail with a good error message?

renxida:
https://github.com/nod-ai/SHARK-Platform/blob/fb4be2774500a6773e0db71ead7a039a0ca6f83e/.github/workflows/ci-sharktank.yml

Looks like we should maybe paste together the sharktank and shortfin ci ymls to make a new environment to do integration testing on.

@ScottTodd need your opinion on this.

stbaione:
> https://github.com/nod-ai/SHARK-Platform/blob/fb4be2774500a6773e0db71ead7a039a0ca6f83e/.github/workflows/ci-sharktank.yml
> 
> Looks like we should maybe paste together the sharktank and shortfin ci ymls to make a new environment to do integration testing on.
> 
> @ScottTodd need your opinion on this.

Reached out to @ScottTodd this morning. Am currently working on setting up a workflow file

renxida:
oh hell yeah : D

ScottTodd (changes requested):

stbaione:
@ScottTodd Thanks for the feedback! Working on requested changes and setting up CI workflow

shortfin/tests/apps/llm/cpu_llm_server_test.py
stbaione: That's a good catch. Maybe a good route could be:

Before server starts we use `socket` to attempt to bind to the default port (8000). If that's unsuccessful, we increment the port until it finds an available one. Once we find an available port, we use that to start the server and proceed through the rest of the tests. Should probably set a limit on how many increments is allowed before failing, in case there's ever an error with a runner host that affects ability to bind to ANY ports.

We still add a good error message, just in case in the window between verifying a port's availability and starting the server, that port becomes unavailable.

shortfin/tests/apps/llm/cpu_llm_server_test.py
stbaione: That's a good point. Looking at it again, I think it'd be "safe" to cache `open-llama-3b-v2-f16.gguf` and `tokenizer.json`.

Unless I'm wrong, I don't think those files will change, and take up the bulk of the time when downloading collaterals to run the test.

That leaves the files produced from `sharktank.examples.export_paged_llm_v1` and `iree-compile` uncached, which seems like it could add some value as an e2e test and catch any potential errors in those two necessary commands.

stbaione:
Pushing in pieces to keep better track of feedback

.github/workflows/ci_linux_x64_nogil-libshortfin.yml
```diff
@@ -98,6 +98,6 @@ jobs:
     - name: Run shortfin Python tests (full)
       working-directory: ${{ env.LIBSHORTFIN_DIR }}
       run: |
-        pytest -s --ignore=tests/examples/fastapi_test.py --ignore=tests/apps/llm/components/cache_test.py --ignore=tests/apps/sd
+        pytest -s --ignore=tests/examples/fastapi_test.py --ignore=tests/apps/llm/components/cache_test.py --ignore=tests/apps/llm/cpu_llm_server_test.py --ignore=tests/apps/sd
```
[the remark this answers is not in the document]
stbaione: Yeah, admittedly, I was just following what seemed to be in place when I added it. I removed this for now, and will look for a better alternative for skipping. I'm thinking something like:

`@pytest.mark.skipif(hasattr(sys, "_is_gil_enabled") and not sys._is_gil_enabled(), reson="Test is not supported for no-gil python")`

shortfin/tests/apps/llm/cpu_llm_server_test.py
```diff
@@ -0,0 +1,156 @@
+import pytest
```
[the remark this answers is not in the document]
stbaione: License header added

shortfin/tests/apps/llm/cpu_llm_server_test.py
```diff
@@ -0,0 +1,156 @@
...
+import tempfile
+import shutil
+
+from transformers import AutoTokenizer
+
+BATCH_SIZES = [1, 4]
+cpu_settings = {
+    "device_flags": ["-iree-hal-target-backends=llvm-cpu"],
+    "device": "local-task",
+}
```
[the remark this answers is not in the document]
stbaione: Added `--iree-llvmcpu-target-cpu=host` to cpu_settings

shortfin/tests/apps/llm/cpu_llm_server_test.py
```diff
@@ -0,0 +1,156 @@
...
+    tmp_dir = tmp_path_factory.mktemp("cpu_llm_server_test")
+    try:
+        # Download model if it doesn't exist
+        model_path = os.path.join(tmp_dir, "open-llama-3b-v2-f16.gguf")
+        if not os.path.exists(model_path):
+            subprocess.run(
+                f"huggingface-cli download --local-dir {tmp_dir} SlyEcho/open_llama_3b_v2_gguf open-llama-3b-v2-f16.gguf",
+                shell=True,
+                check=True,
+            )
```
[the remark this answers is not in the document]
stbaione: Looking into this now

shortfin/tests/apps/llm/cpu_llm_server_test.py
```diff
@@ -0,0 +1,156 @@
...
+import subprocess
+import time
+import requests
+import os
+import json
+import uuid
+import tempfile
+import shutil
+
+from transformers import AutoTokenizer
```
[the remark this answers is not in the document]
stbaione: Added a `pytest.importorskip("transformers")` line above import

shortfin/tests/apps/llm/cpu_llm_server_test.py
stbaione: Moved test to toplevel `tests` directory. Will fix existing CIs and implement new CI after responding to code feedback

shortfin/tests/apps/llm/cpu_llm_server_test.py
stbaione: Changed to using `/health` route of server to verify that it's started properly

shortfin/tests/apps/llm/cpu_llm_server_test.py
stbaione: Created an `available_port` fixture to find an available dynamically, instead of having one hardcoded

shortfin/tests/apps/llm/cpu_llm_server_test.py
ScottTodd: @stellaraccident do you have an opinion on where integration tests using both shortfin and sharktank should live?

* `tests/`
* `shark_platform/tests` (meta package _and_ integration tests?)

shortfin/tests/apps/llm/cpu_llm_server_test.py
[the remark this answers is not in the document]
stbaione: Parameterization is added for model and tokenizer inputs

shortfin/tests/apps/llm/cpu_llm_server_test.py
ScottTodd: Oooh wait. I already wrote a test and workflow for this, but that was using the old version of shortfin: https://github.com/nod-ai/SHARK-Platform/pull/70

We should replace https://github.com/nod-ai/SHARK-Platform/blob/main/build_tools/integration_tests/llama_export_compile_serve.sh and https://github.com/nod-ai/SHARK-Platform/blob/main/.github/workflows/test.yaml with this new code using the latest APIs :P

That workflow has been failing (since it wasn't updated to use the new APIs): https://github.com/nod-ai/SHARK-Platform/actions/workflows/test.yaml

ScottTodd: Prior discussion about paths was https://github.com/nod-ai/SHARK-Platform/pull/70#discussion_r1655281497

> Let's put these scripts in build_tools/integration_tests. Python projects end up using the tests/ folder for more unit-testy things.

stbaione: Moved test to `./build_tools/integration_tests/cpu_llm_server_test.py`, removed `test.yaml` and `llama_export_and_compile_serve.sh`

stbaione: @ScottTodd Workflow file added and passing. 

In regard to: 
```See also https://github.com/nod-ai/SHARK-Platform/pull/284#discussion_r1803610173```

This integration test was in part spurred by a change being implemented in the sharktank export script, which caused a failure when using exported models to launch the shortfin llm server. So the export script, in this case, would be considered part of the "e2e" flow for testing the llm server.

With that in mind does it still make sense to switch to using `vllm.LLM` or is this a broader suggestion that maybe we shouldn't be using the export script as part of the "golden" flow?

File: .github/workflows/ci_linux_x64_nogil-libshortfin.yml (hunk 5 above)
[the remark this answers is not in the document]
Comment (stbaione) [reply .github/workflows/ci_linux_x64_nogil-libshortfin.yml]: Test has been moved to another location, and removed from this line

File: shortfin/tests/apps/llm/cpu_llm_server_test.py (hunk 1 above)
Comment (ScottTodd) [reply shortfin/tests/apps/llm/cpu_llm_server_test.py]: > With that in mind does it still make sense to switch to using `vllm.LLM` or is this a broader suggestion that maybe we shouldn't be using the export script as part of the "golden" flow?

The current export script is a developer tool built for a single model that has rough edges. We can improve or create something new. Users need something better either way.

Comment (stbaione) [reply shortfin/tests/apps/llm/cpu_llm_server_test.py]: Gotcha, does this make sense for a subsequent PR? It sounds like a separate task from this one.

I know that raises some caution of the "can getting kicked down the road", so maybe we could create an issue for this right now and I can discuss in sync tomorrow to get it assigned?

Comment (ScottTodd) [reply shortfin/tests/apps/llm/cpu_llm_server_test.py]: Yep, I just want to make sure we keep focused on providing a good user experience. We should have easy to use scripts and test _those_.

Comment (stbaione) [reply shortfin/tests/apps/llm/cpu_llm_server_test.py]: Issue created here: https://github.com/nod-ai/SHARK-Platform/issues/402

File: shortfin/tests/apps/llm/cpu_llm_server_test.py (hunk 2 above)
Comment (stbaione) [reply shortfin/tests/apps/llm/cpu_llm_server_test.py]: Added cachine for model parameter and tokenizer files, based on HF_HOME dir

File: shortfin/tests/apps/llm/cpu_llm_server_test.py [hunk 10]
```diff
@@ -0,0 +1,156 @@
...
+
+BATCH_SIZES = [1, 4]
+cpu_settings = {
+    "device_flags": ["-iree-hal-target-backends=llvm-cpu"],
+    "device": "local-task",
+}
+gpu_settings = {
+    "device_flags": ["-iree-hal-target-backends=rocm", "--iree-hip-target=gfx1100"],
+    "device": "hip",
+}
```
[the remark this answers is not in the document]
stbaione: Was thinking we could do this with two env variables: IREE_HAL_TARGET_BACKENDS and IREE_HIP_TARGET?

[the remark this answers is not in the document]
ScottTodd: Probably just the `IREE_HIP_TARGET` variable.

@saienduri do we have a standard env var we set? I see `ROCM_CHIP` in a few spots:
* https://github.com/nod-ai/SHARK-TestSuite/blob/ac2d83e65a9fff56ff96437a01f471bae958766f/.github/workflows/test_iree.yml#L48C7-L48C16
* https://github.com/iree-org/iree/blob/20c83470b4186944e14c67488d295f3f0f33ecce/.github/workflows/pkgci_regression_test.yml#L194-L195

also `IREE_HIP_TEST_TARGET_CHIP` used by ctest https://github.com/iree-org/iree/blob/20c83470b4186944e14c67488d295f3f0f33ecce/.github/workflows/pkgci_test_amd_mi250.yml#L33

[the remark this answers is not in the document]
saienduri: Yeah usually `ROCM_CHIP`

.github/workflows/ci-sharktank-and-shortfin.yml
```diff
@@ -70,31 +70,5 @@ jobs:
             iree-runtime \
             "numpy<2.0"
 
-      - name: Run llama test
-        run: ./build_tools/integration_tests/llama_export_compile_serve.sh
-
-  test_punet:
```
renxida, line 76: Am I right to understand that llama tests makes the punet one unnecessary?

.github/workflows/ci-sharktank-and-shortfin.yml
```diff
@@ -70,31 +70,5 @@ jobs:
             iree-runtime \
             "numpy<2.0"
 
-      - name: Run llama test
-        run: ./build_tools/integration_tests/llama_export_compile_serve.sh
```
renxida, line 74: A little curious why we put this into build_tools instead of its own tests dir

If it's only because `llama_export_compile_serve.sh` is in there, we should probably make its own tests dir

I i know the org has the convention of putting temporary build / test / run scripts in `build_tools` and i don't think this is an obvious and intuitive place to be finding `integration_tests` in

It would make more sense to put this into a top level `tests` or `integration_tests` dir

ignore the above if you've received feedback from other people that caused you to delete the top level tests directory

renxida (changes requested):

.github/workflows/ci-sharktank-and-shortfin.yml
stbaione: Thread discussing the removal of this script:

https://github.com/nod-ai/SHARK-Platform/pull/373#discussion_r1823450607

.github/workflows/ci-sharktank-and-shortfin.yml
stbaione: Thread discussing moving the test to `build_tools/integrations_tests`: https://github.com/nod-ai/SHARK-Platform/pull/373#discussion_r1823458520

.github/workflows/ci-sharktank-and-shortfin.yml
```diff
@@ -70,31 +70,5 @@ jobs:
...
-          pip install --no-compile -f https://iree.dev/pip-release-links.html --src deps \
-            -e "git+https://github.com/iree-org/iree-turbine.git#egg=iree-turbine"
-          pip install --no-compile -r requirements.txt -r sharktank/requirements-tests.txt -e sharktank/ shortfin/
-
-      - name: Run punet tests
-        run: |
-          source ${VENV_DIR}/bin/activate
-          pytest -v sharktank/ -m model_punet
+      - name: Run shortfin LLM Server Integration Test
+        run: pytest -v build_tools/integration_tests/cpu_llm_server_test.py
```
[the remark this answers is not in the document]
stbaione: Replaced with just integration_tests directory

.github/workflows/ci-sharktank-and-shortfin.yml
renxida: gotcha i missed that ty!

renxida (approved):

.github/workflows/ci-sharktank-and-shortfin.yml
```diff
@@ -4,13 +4,14 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-name: Integration Tests
+name: CI - sharktank and shortfin
```
ScottTodd, line 7: I'd call this `CI - shark-platform` and rename the file `ci-shark-platform.yml`

.github/workflows/ci-sharktank-and-shortfin.yml
```diff
@@ -70,31 +70,5 @@ jobs:
...
-          pip install --no-compile -f https://iree.dev/pip-release-links.html --src deps \
-            -e "git+https://github.com/iree-org/iree-turbine.git#egg=iree-turbine"
-          pip install --no-compile -r requirements.txt -r sharktank/requirements-tests.txt -e sharktank/ shortfin/
-
-      - name: Run punet tests
-        run: |
-          source ${VENV_DIR}/bin/activate
-          pytest -v sharktank/ -m model_punet
+      - name: Run shortfin LLM Server Integration Test
+        run: pytest -v build_tools/integration_tests/
```
ScottTodd, line 74: Could keep this name generic, for when other tests are added to this folder.

```suggestion
      - name: Run integration tests
        run: pytest -v build_tools/integration_tests/
```

Or split into subfolders per component... but we can solve that when N > 1.

```yml
      - name: Run LLM integration tests
        run: pytest -v build_tools/integration_tests/llm

      - name: Run SD integration tests
        run: pytest -v build_tools/integration_tests/sd
```

build_tools/integration_tests/cpu_llm_server_test.py
```diff
@@ -0,0 +1,223 @@
...
+        }
+        with open(edited_config_path, "w") as f:
+            json.dump(config, f)
+        yield hf_home, tmp_dir
+    finally:
+        shutil.rmtree(tmp_dir)
+
+
+@pytest.fixture(scope="module")
+def available_port(port=8000, max_port=8100):
```
ScottTodd, line 117: These fixtures / helper functions could move into a `conftest.py` or `test_utils.py` to share with other integration tests. Fine to keep here for now.

.github/workflows/ci-sharktank-and-shortfin.yml
ScottTodd, line 74: Does the test log anything while it is running? The output is pretty sparse for a test that takes over 3 minutes: https://github.com/nod-ai/SHARK-Platform/actions/runs/11670527836/job/32494907067?pr=373#step:6:1

Could run with `pytest --log-cli-level=info` or some of the other flags: https://docs.pytest.org/en/stable/how-to/usage.html

I'd like to see logs for:
* Artifact downloading (including cache hit/miss)
* Program import (sharktank)
* Program execution (shortfin), from both the client and server

build_tools/integration_tests/cpu_llm_server_test.py
```diff
@@ -0,0 +1,223 @@
...
+    ],
+    indirect=True,
+)
+def test_llm_server(llm_server, available_port):
+    # Here you would typically make requests to your server
+    # and assert on the responses
+    assert llm_server.poll() is None
+    output = do_generate("1 2 3 4 5 ", available_port)
+    print(output)
+    assert output.startswith("6 7 8")
```
ScottTodd, line 223: If we were to parameterize this test over the prompt and response too, how much of the model download, import, and compile would be cached between test cases?

(Totally fine to start with the one test case running on every commit, mostly wondering if we can scale this to dozens of test cases running nightly or more frequently)

ScottTodd (commented):
Nice! Mostly LGTM. A few comments about how this could continue to evolve.

.github/workflows/ci-sharktank-and-shortfin.yml
ScottTodd: Note that pytest will consume log output by default on success. The `--log-cli-level=info` flag will show the captured output.

See also https://docs.pytest.org/en/stable/how-to/logging.html

build_tools/integration_tests/cpu_llm_server_test.py
stbaione: That's a good idea. Currently, the model file and tokenizer file downloaded from hugging face would be cached between runs and re-used.

As of right now, the `mlir` and `vmfb` file would still be regenerated for subsequent runs and saved in another temporary dir

.github/workflows/ci-sharktank-and-shortfin.yml
stbaione: Updated to `CI - shark-platform`

.github/workflows/ci-sharktank-and-shortfin.yml
stbaione: I went with the latter. Really like that setup. Will make it much easier to track where the error is and whether or not it has to do with your PR (i.e. I make an SD change unrelated to LLM)

stbaione: Added logging throughout the run:

```bash
------------------------------------------------------------------------------------------------------------ live log setup -------------------------------------------------------------------------------------------------------------
INFO     conftest:conftest.py:33 Preparing model artifacts...
INFO     conftest:conftest.py:47 Preparing model_path: /tmp/pytest-of-amd/pytest-53/cpu_llm_server_test0/open-llama-3b-v2-f16.gguf..
INFO     conftest:conftest.py:49 Downloading model SlyEcho/open_llama_3b_v2_gguf open-llama-3b-v2-f16.gguf from Hugging Face...
INFO     conftest:conftest.py:57 Model downloaded to /tmp/pytest-of-amd/pytest-53/cpu_llm_server_test0/open-llama-3b-v2-f16.gguf
INFO     conftest:conftest.py:63 Preparing tokenizer_path: /tmp/pytest-of-amd/pytest-53/cpu_llm_server_test0/tokenizer.json...
INFO     conftest:conftest.py:65 Downloading tokenizer openlm-research/open_llama_3b_v2 from Hugging Face...
INFO     conftest:conftest.py:70 Tokenizer saved to /tmp/pytest-of-amd/pytest-53/cpu_llm_server_test0/tokenizer.json
INFO     conftest:conftest.py:78 Exporting model with following settings:
  MLIR Path: /tmp/pytest-of-amd/pytest-53/cpu_llm_server_test0/model.mlir
  Config Path: /tmp/pytest-of-amd/pytest-53/cpu_llm_server_test0/config.json
  Batch Sizes: 1,4
INFO     conftest:conftest.py:96 Model successfully exported to /tmp/pytest-of-amd/pytest-53/cpu_llm_server_test0/model.mlir
INFO     conftest:conftest.py:100 Compiling model to /tmp/pytest-of-amd/pytest-53/cpu_llm_server_test0/model.vmfb
INFO     conftest:conftest.py:111 Model successfully compiled to /tmp/pytest-of-amd/pytest-53/cpu_llm_server_test0/model.vmfb
INFO     conftest:conftest.py:126 Saving edited config to: /tmp/pytest-of-amd/pytest-53/cpu_llm_server_test0/edited_config.json

INFO     conftest:conftest.py:127 Config: {
  "module_name": "module",
  "module_abi_version": 1,
  "max_seq_len": 2048,
  "attn_head_count": 32,
  "attn_head_dim": 100,
  "prefill_batch_sizes": [
    1,
    4
  ],
  "decode_batch_sizes": [
    1,
    4
  ],
  "transformer_block_count": 26,
  "paged_kv_cache": {
    "block_seq_stride": 16,
    "device_block_count": 256
  }
}
INFO     conftest:conftest.py:130 Model artifacts setup successfully
INFO     conftest:conftest.py:140 Finding available port in range 8000-8100...
INFO     conftest:conftest.py:149 Found available port: 8000
INFO     conftest:conftest.py:184 Starting LLM server...
INFO     conftest:conftest.py:158 Waiting for server to start at http://localhost:8000...
INFO     conftest:conftest.py:163 Server successfully started
------------------------------------------------------------------------------------------------------------- live log call -------------------------------------------------------------------------------------------------------------
INFO     cpu_llm_server_test:cpu_llm_server_test.py:33 Generating request...
INFO     cpu_llm_server_test:cpu_llm_server_test.py:46 Prompt text:
INFO     cpu_llm_server_test:cpu_llm_server_test.py:47 1 2 3 4 5 
INFO     cpu_llm_server_test:cpu_llm_server_test.py:50 Generate endpoint status code: 200
INFO     cpu_llm_server_test:cpu_llm_server_test.py:52 Generated text:
INFO     cpu_llm_server_test:cpu_llm_server_test.py:84 6 7 8 8 8 8 8 8 
```

build_tools/integration_tests/cpu_llm_server_test.py
stbaione: Good idea, moved setup functions to `conftest.py`

.github/workflows/ci-shark-platform.yml
```diff
@@ -4,13 +4,14 @@
...
 
 on:
   workflow_dispatch:
-  schedule:
-    # Weekdays at 13:00 UTC = 05:00 PST / 06:00 PDT.
-    - cron: "5 4 * * 1-5"
+  pull_request:
+  push:
+    branches:
+      - main
```
ScottTodd, line 14: @saienduri is `nodai-amdgpu-mi250-x86-64` available enough for the `pull_request` trigger here? Runners matching that label are currently busy running on-demand jobs in E2ESHARK Test Suite: https://github.com/nod-ai/SHARK-TestSuite/actions/runs/11673948346/job/32505667904

This job itself takes ~5 minutes today, but we shouldn't be queueing for 1h+.

ScottTodd (commented):
LGTM once we decide on which triggers (`pull_request`, `push`, `schedule`) are appropriate.

ScottTodd (approved):

.github/workflows/ci-shark-platform.yml
stbaione: Sounds good and makes sense as we scale these integration tests in the future. What needs to happen for us to enable that?

ScottTodd:
Something about this new test started failing overnight. The test might be flaky, or it might not be hermetic enough (e.g. by downloading dependencies that aren't pinned to specific versions).

https://github.com/nod-ai/SHARK-Platform/actions/workflows/ci-shark-platform.yml?query=branch%3Amain

The workflow initially succeeded at this run/commit, but then failed on a retry: https://github.com/nod-ai/SHARK-Platform/actions/runs/11704581609

stbaione:
> Something about this new test started failing overnight. The test might be flaky, or it might not be hermetic enough (e.g. by downloading dependencies that aren't pinned to specific versions).
> 
> https://github.com/nod-ai/SHARK-Platform/actions/workflows/ci-shark-platform.yml?query=branch%3Amain
> 
> The workflow initially succeeded at this run/commit, but then failed on a retry: https://github.com/nod-ai/SHARK-Platform/actions/runs/11704581609

@ScottTodd 

Yeah, for some reason the LLM Server started returning faulty responses:

```text
INFO     cpu_llm_server_test:cpu_llm_server_test.py:46 Prompt text:
INFO     cpu_llm_server_test:cpu_llm_server_test.py:47 1 2 3 4 5 
INFO     cpu_llm_server_test:cpu_llm_server_test.py:50 Generate endpoint status code: 200
INFO     cpu_llm_server_test:cpu_llm_server_test.py:52 Generated text:
INFO     cpu_llm_server_test:cpu_llm_server_test.py:84 6000 1 10000 10000 1000
```

Which caused an assertion failure. I'm seeing the exact same output for a failure here: https://github.com/nod-ai/SHARK-Platform/actions/runs/11706659706/job/32604263601?pr=435

The action is not pinned to a specific version for IREE, so my initial thought is that may be the culprit:

```bash
pip install --no-compile -f https://iree.dev/pip-release-links.html --src deps \
            -e "git+https://github.com/iree-org/iree-turbine.git#egg=iree-turbine"
```

Others specify an installation candidate:

```text
- name: Checkout IREE repo
      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
      with:
        repository: iree-org/iree
        submodules: false
        ref: candidate-20241104.1068
```

In a way this does highlight that there may be a bug between shortfin and latest IREE, but it shouldn't block development. I used the installation instructions from the README for installing IREE as opposed to a specific candidate to catch stuff like this, but that's kind of annoying from a developer perspective.

I'm thinking we pin it to a specific version and we'll catch these kinds of issues when we bump IREE.

PR #383, opened 2024-10-30
Introduce CausalLMModelABC interface

sogartar:
We do not have a clearly defined interface for LMs. Decode and prefill have different signature when exporting to IREE. Here is added a new ABC CausalLMModelABC that makes a distinction between the two variants. The BaseCausalLMModel provides a default implementation for the new prefill_from_seq_lens and decode_from_seq_lens methods.

The export script export_paged_llm_v1 does too much in its exported functions. It computes the attention mask then. It shards its arguments and unshards its result. This change lets it be a thiner wrapper around the new interface functions.

Make paged_llm_v1.TorchGenerator use the new interface methods.

### DIFF
--- a/sharktank/sharktank/models/llama/llama.py
+++ b/sharktank/sharktank/models/llama/llama.py
@@ -7,7 +7,7 @@
 from typing import Optional
 
 from dataclasses import dataclass
-from typing import Union
+from typing import Any, Union
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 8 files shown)


sharktank/sharktank/models/llama/llama.py
```diff
@@ -27,7 +27,7 @@
 ################################################################################
 
 
-class PagedLlamaModelV1(BaseCausalLMModel):
+class PagedLlamaModelV1(BaseCausalLMModel, CausalLMModelABC):
```
rsuderman, line 30: Double inheritance almost always a bad idea especially considering these are both LLM interfaces. Can we have an alternative option? Why can't we expand this functionality in the existing BaseCausalLMModel?

rsuderman (changes requested):

PR #391, opened 2024-10-31
Add block seq len and attention/activation dtype args to LLM export script

sogartar:
Some upcoming sharded tests don't use the default values.

### DIFF
--- a/sharktank/sharktank/examples/export_paged_llm_v1.py
+++ b/sharktank/sharktank/examples/export_paged_llm_v1.py
@@ -44,6 +44,12 @@ def main():
         default="4",
     )
     parser.add_argument(
+        "--block-seq-stride",
+        help="Block sequence stride for a paged KV cache.",
+        type=int,
+        default=LlamaModelConfig.default_block_seq_stride,
+    )
+    parser.add_argument(
         "--verbose",
         help="Include verbose logging",
         action="store_true",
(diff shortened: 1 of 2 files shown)


sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -73,11 +89,14 @@ def main():
     )
     llama_config = LlamaModelConfig(
         hp,
+        block_seq_stride=args.block_seq_stride,
         tensor_parallelism_size=tensor_parallelism_size,
         use_hf=False,
         static_tables=False,  # Rely on the compiler for hoisting tables.
         kv_cache_type="direct" if args.bs == [1] else "paged",
         attention_kernel=args.attention_kernel,
+        attention_dtype=serialized_name_to_dtype(args.attention_dtype),
```
rsuderman, line 98: This depends on the other PR. It should be included there and not bleed into this PR

sharktank/sharktank/layers/configs/llm_configs.py
```diff
@@ -121,7 +121,8 @@ class LlamaModelConfig:
 
     # Block sequence stride for a paged KV cache. This must divide evenly
     # into the context length.
-    block_seq_stride: int = 16
+    default_block_seq_stride: ClassVar[int] = 16
```
rsuderman, line 124: Why do we need `ClassVar `here? Seeing the `default_block_seq_stride` default feels bad.

rsuderman (changes requested):

sharktank/sharktank/examples/export_paged_llm_v1.py
sogartar: I already merged the other PR.

PR #392, opened 2024-10-31
In LLM export make dynamic dims compatible with non-defaults

sogartar:
Some tests for toy-sized sharded Llama that I want to adapt to use this script have non-default values. The export chokes on them.


dan-garvey (approved):
Makes sense to me as long as it passes

PR #393, opened 2024-10-31
Add class that implements CausalLMModelABC but is backed by an IREE module

sogartar:
This is meant only for testing. It may not be performant as it marshals tensors and uses the IREE Python bindings.

### DIFF
--- a/sharktank/sharktank/layers/testing.py
+++ b/sharktank/sharktank/layers/testing.py
@@ -4,10 +4,20 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+from typing import Union
+
 import torch
 from ..types.theta import Theta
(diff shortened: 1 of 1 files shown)


sogartar:
This PR depends on #383.

sharktank/sharktank/layers/testing.py
```diff
@@ -43,3 +53,106 @@ def make_llama_attention_block_theta(
             ),
         }
     )
+
+
+class CausalLMIreeModel(BaseCausalLMModel, CausalLMModelABC):
```
rsuderman, line 58: Dependent on the double inheritance PR you have. This should have a cleaner design rather than extend with double inheritance.

rsuderman (changes requested):

sharktank/sharktank/layers/testing.py
sogartar: I removed `CausalLMModelABC`.

PR #394, opened 2024-10-31
Add IREE numerics test for Llama 3.1 8B FP16 TP8

sogartar:
Introduce a Llama 3.1 8B FP16 TP8 test that appears to not have good numerical accuracy. It is compared to an fp64 unsharded torch variant to ensure that the reference is of high accuracy.

Refactor the sharded Llama tests. Increase code reuse and use the TorchGenerator in the toy-sized tests. Use the shard_llm_dataset and export_paged_llm_v1 scripts in the test flow to increase their test coverage.


sogartar:
This PR depends on #383, #384, #386, #390, #391, #392, #393.

Issue #401, opened 2024-10-31
Exporting & serving llama3 8b f16 errs on shortfin


renxida:
# Error:
```
RuntimeError: ValueError: shortfin_iree-src/runtime/src/iree/modules/hal/utils/buffer_diagnostics.c:221: INVALID_ARGUMENT; tensor shape dimension 1 mismatch; expected 1048576 but have 4194304; expected shape `256x1048576`, actual shape `256x4194304`; while invoking native function hal.buffer_view.assert; while calling import; 
```

<details>
```
[2024-10-31 12:13:38.339] [info] [service.py:390] INVOKE ProgramFunction(prefill_bs1$async: 0rrrrrr_r): 
  0: [1, 16]
  1: [1]
  2: [1, 1]
  3: [256, 4194304]
[2024-10-31 12:13:38.339] [error] [service.py:415] Fatal error in prefetch invocation
Traceback (most recent call last):
  File "/home/xidaren2/SHARK-Platform/shortfin/python/shortfin_apps/llm/components/service.py", line 396, in run
    (logits,) = await fn(*args, fiber=self.fiber)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: ValueError: shortfin_iree-src/runtime/src/iree/modules/hal/utils/buffer_diagnostics.c:221: INVALID_ARGUMENT; tensor shape dimension 1 mismatch; expected 1048576 but have 4194304; expected shape `256x1048576`, actual shape `256x4194304`; while invoking native function hal.buffer_view.assert; while calling import; 
[ 0] bytecode module.prefill_bs1$async:3530 /home/xidaren2/xshortfin/llama3_8b/export/model.mlir:293:294
/home/xidaren2/SHARK-Platform/shortfin/src/shortfin/support/iree_helpers.h:315: UNKNOWN; Unhandled exception: Traceback (most recent call last):
  File "/home/xidaren2/SHARK-Platform/shortfin/python/_shortfin/asyncio_bridge.py", line 79, in _sf_maybe_run
  File "/home/xidaren2/miniforge3/envs/env1/lib/python3.12/asyncio/events.py", line 103, in _run
  File "/home/xidaren2/SHARK-Platform/shortfin/python/shortfin_apps/llm/components/generate.py", line 78, in run
RuntimeError: Async exception on <Worker 'default-inference'>): argmax(): incompatible function arguments. The following argument types are supported:
[...]

PR #406, opened 2024-11-01
Fix cache config inconsistency between shortfin and sharktank

### LINKED ISSUE #405: Mismatches between config.json exported by export_paged_llm_v1.py and expected by shortfin
The following edits were required to make llama3 8b fp16 work:
```
config["attn_head_count"] = 8 # 8 instead of 32
config["paged_kv_cache"] = {}
config["paged_kv_cache"]["block_seq_stride"] = config["block_seq_stride"]
del config["block_seq_stride"]
config["paged_kv_cache"]["device_block_count"] = 256
```
There are 2 main problems:
- the `attn_head_count` should be set to `attention_head_count_kv` from export_paged_llm_v1 and not `attention_head_count`. This should be fixed in sharktank, at least by including both attention head counts
- kvcache params should be in config["paged_kv_cache"]

Really need integration tests between sharktank and shortfin.

renxida:
Fixes #405 and #401 

Up till now we have been patching the configs to reorganize it after exporting from sharktank, but for llama3 8b (which has grouped attention), the config doesn't specify grouping, so this has to be fixed.

Might be a good idea to rename all attn to attention for consistency. But I want this PR to just fix the issue at hand.


renxida:
image


Proof it works.

Tested with script:

https://gist.github.com/renxida/2023ec7f9931c697879762c9330de001

sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -98,7 +98,10 @@ def generate_params_json(hp, prefill_bs: list[int], decode_bs: list[int]):
             "prefill_batch_sizes": prefill_bs,
             "decode_batch_sizes": decode_bs,
             "transformer_block_count": hp.block_count,
-            "block_seq_stride": llama_config.block_seq_stride,
+            "paged_kv_cache": {
+                "block_seq_stride": llama_config.block_seq_stride,
+                "attn_head_count_kv": hp.attention_head_count_kv,
```
archana-ramalingam, line 103: Just FYI, in sharktank's [llm_configs.py](https://github.com/nod-ai/SHARK-Platform/blob/main/sharktank/sharktank/layers/configs/llm_configs.py#L67), hp.attention_head_count_kv==hp.attention_head_count.

archana-ramalingam: If you req vmfb and irpa for shortfin, is there a reason we rely on `generate_params_json` params instead of the `irpa` params? There are ways to load & access irpa params directly in sharktank. Do we want/have something similar in shortfin?

renxida: Oh ty! i should probably take a deeper look then to figure this part out.

renxida:
Replaced by #487

PR #411, opened 2024-11-01
(shortfin-sd) Adds iree.build artifact fetching.

monorimet:
Also adds two _slow_ tests for testing larger SDXL server loads that will not trigger in any workflows yet.

This is missing a few things:
 - ad-hoc artifacts fetching (e.g. someone inits the server with only 1024x1024 and wants to fetch and load modules for other output shapes ad-hoc when requested by client)
 - compile integrate (currently pulls precompiled vmfbs and weights)
 
 Builder isn't very smart yet, and will just blindly try to download composed filenames from sharkpublic buckets.
 It should eventually cover:
  - compile (short-term)
  - export (short/medium-term)


shortfin/python/shortfin_apps/sd/components/config_struct.py
```diff
@@ -59,11 +65,13 @@ class ModelParams:
     # Classifer free guidance mode. If set to false, only positive prompts will matter.
     cfg_mode = True
 
-    # DTypes (basically defaults):
+    # DTypes (not necessarily weights precision):
     clip_dtype: sfnp.DType = sfnp.float16
     unet_dtype: sfnp.DType = sfnp.float16
     vae_dtype: sfnp.DType = sfnp.float16
 
+    use_i8_punet: bool = False
```
monorimet, line 73: really don't like this but the module expects fp16 I/O...

shortfin/python/shortfin_apps/sd/components/builders.py
```diff
@@ -0,0 +1,162 @@
...
+import os
+import shortfin.array as sfnp
+
+from shortfin_apps.sd.components.config_struct import ModelParams
+
+this_dir = os.path.dirname(os.path.abspath(__file__))
+parent = os.path.dirname(this_dir)
+default_config_json = os.path.join(parent, "examples", "sdxl_config_i8.json")
+
+dtype_to_filetag = {
```
monorimet, line 12: maybe we rename the artifacts to match sfnp.Dtype attributes instead of doing little workarounds like this for old naming conventions... trying to align it

shortfin/python/shortfin_apps/sd/components/config_struct.py
monorimet: Maybe a specific place for IO dtype is in order. One (`*_dtype`) is used for instructing device array creation, and the other (`use_i8_punet`) is used when inferring artifact names.

monorimet:
NOTE: needs rebase on main once https://github.com/nod-ai/SHARK-Platform/pull/413 lands

shortfin/python/shortfin_apps/sd/components/builders.py
```diff
@@ -0,0 +1,164 @@
...
+    vmfb_bucket = SDXL_BUCKET + "vmfbs/"
+    vmfb_filenames = get_vmfb_filenames(model_params, target=target)
+    vmfb_urls = get_url_map(vmfb_filenames, vmfb_bucket)
+    for f, url in vmfb_urls.items():
+        fetch_http(name=f, url=url)
+    params_filenames = get_params_filenames(model_params, splat)
+    params_urls = get_url_map(params_filenames, SDXL_WEIGHTS_BUCKET)
+    ctx = executor.BuildContext.current()
+    for f, url in params_urls.items():
+        out_file = os.path.join(ctx.executor.output_dir, f)
```
monorimet, line 156: This isn't very robust. Should have a md5sum checklist fetched from the server if downloads enabled, and compare with local checklist to determine which, if any, artifacts need updating.

stellaraccident: Yeah, I hadn't yet gotten to stamp and change detection... Will in a bit. 

Do you already have file hashes stored in the bucket somewhere?

monorimet: Not yet. Right now, the vmfbs are always downloaded from a bucket versioned by date only.

stellaraccident: That can work. You basically need something to derive a stamp value from. That can come from some part of the URL.

stellaraccident: For mammoth files, a manual version of some kind can be best anyway: it can take a long time to compute a hash of such things

monorimet: Does it seem too heavyweight to keep a md5sums.json in each bucket, and have the builder generate and keep a local set of hashes for its outputs? That way we can filter  exactly what's needed before doing fetch_http?

stellaraccident: Yeah, that's the basic mechanism. We wouldn't actually compute the hash in the builder in typical use. Instead, you would tell it how to get the stamp artifact (ie. Some fixed string, a hash file, etc). If a hash file, we compute a running hash only during download and store the result, erroring if it mismatches. But just an opaque stamp value drives the up-to-date check. 

It's better for everyone if such artifacts are in write once storage (ie. The same URL produces the same content for all of time). Then the stamp is just the url, and any has checking is just for verifying the integrity of the transfer.

shortfin/python/shortfin_apps/sd/components/config_struct.py
nithinsubbiah: Drop by comment, I agree with the above. We have multiple `punet` models to support like `int8` and `fp8`, so it would be better to keep them separate

stellaraccident (approved):

PR #413, opened 2024-11-01
Update IREE ref to candidate-20241101.1065

monorimet:


ScottTodd (approved):
When we cut a stable release across projects within the next two weeks we'll want to be a bit deliberate about where we make the cut. Fine to stay close to tip of tree as patches are landing.

PR #420, opened 2024-11-05
[shortfin] Make SystemBuilder fully configuration/environment driven.

stellaraccident:
* Now `shortfin.SystemBuilder` (which was an abstract class) has a constructor. When used (vs one of the concrete subclasses), the `system_type=` keyword (or SHORTFIN_SYSTEM_TYPE env var, or `SystemBuilder.default_system_type` property) is used to drive subclass selection.
* This feature was already in the C++ API: this patch merely exposes it to Python in an ergonomic way.
* Pytest configuration and fixtures were updated to support configurable system independence where possible/straightforward. The result is that some tests now run for `--system amdgpu`.
* Pytest `--compile-flags=` is now available to also provide explicit compilation flags for tests that need to compile a binary.
* Both such tests are currently non functional on amdgpu, which needs further triage. This is ok to land because no one is passing `--compile-flags` yet, leaving time to triage these tests.

### DIFF
--- a/shortfin/README.md
+++ b/shortfin/README.md
@@ -111,10 +111,13 @@ Run platform independent tests only:
 pytest tests/
 ```
 
-Run tests including for a specific platform:
+Run tests including for a specific platform (in this example, a gfx1100 AMDGPU):
+
+(note that not all tests are system aware yet and some may only run on the CPU)
 
 ```bash
-pytest tests/ --system amdgpu
+pytest tests/ --system amdgpu \
+    --compile-flags="--iree-hal-target-backends=rocm --iree-hip-target=gfx1100"
 ```
 
 # Production Library Building
--- a/shortfin/python/array_binding.cc
+++ b/shortfin/python/array_binding.cc
@@ -428,6 +428,7 @@ void BindArray(py::module_ &m) {
           py::kw_only(), py::arg("read") = false, py::arg("write") = false,
           py::arg("discard") = false, DOCSTRING_STORAGE_MAP)
       .def(py::self == py::self)
+      .def("__len__", &storage::byte_length)
       .def("__repr__", &storage::to_s);
 
   // mapping
--- a/shortfin/python/lib_ext.cc
+++ b/shortfin/python/lib_ext.cc
@@ -1198,6 +1198,7 @@ void BindHostSystem(py::module_ &global_m) {
       m, "SystemBuilder");
   py::class_<local::systems::HostCPUSystemBuilder,
              local::systems::HostSystemBuilder>(m, "CPUSystemBuilder")
+      .def("__init__", [](py::args, py::kwargs) {})
       .def_static(
           "__new__",
           [](py::handle cls, std::optional<std::string> env_prefix,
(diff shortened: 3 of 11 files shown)


renxida (approved):
In the future, 

cpu system builder is replaced with sf.SystemBuilder()

right?

stellaraccident:
> In the future,
> 
> cpu system builder is replaced with sf.SystemBuilder()
> 
> right?

Yes, unless if in the (hopefully) rare case that we are intentionally writing a CPU only test (like in the test case where we check the API and options of the CPU builder specifically).

A lot of the tests are still CPU only but can be fixed.

stellaraccident:
@ScottTodd for post review

shortfin/README.md
```diff
@@ -111,10 +111,13 @@ Run platform independent tests only:
...
 
-Run tests including for a specific platform:
+Run tests including for a specific platform (in this example, a gfx1100 AMDGPU):
+
+(note that not all tests are system aware yet and some may only run on the CPU)
 
 ```bash
-pytest tests/ --system amdgpu
+pytest tests/ --system amdgpu \
+    --compile-flags="--iree-hal-target-backends=rocm --iree-hip-target=gfx1100"
```
ScottTodd, line 120: Clever - passing a _list_ of flags through via this new option.

We have a few tests in other places that pull `--iree-hip-target` from a `IREE_HIP_TARGET` environment variable but that needs the tests to know about the env var for that specific backend and doesn't allow for passing extra flags.

One recent example is https://github.com/nod-ai/SHARK-Platform/pull/373, with this code:
```python
CPU_SETTINGS = {
    "device_flags": [
        "-iree-hal-target-backends=llvm-cpu",
        "--iree-llvmcpu-target-cpu=host",
    ],
    "device": "local-task",
}
IREE_HIP_TARGET = os.environ.get("IREE_HIP_TARGET", "gfx1100")
gpu_settings = {
    "device_flags": [
        "-iree-hal-target-backends=rocm",
        f"--iree-hip-target={IREE_HIP_TARGET}",
    ],
    "device": "hip",
}
```

fyi @stbaione we might want to use this `--compile-flags` pattern for pytest options across more of the project, including that integration test.

One advantage to putting the flags _in the test_ vs loading them from CLI flags is that the test itself can encode xfail behavior for each set of parameters. Hopefully we won't have that many xfails in shortfin 🤞 

In yet another place, I have config files specify compile and run flags along with lists of xfails that are used to instantiate parameterized tests, like https://github.com/iree-org/iree/blob/main/tests/external/iree-test-suites/onnx_ops/onnx_ops_cpu_llvm_sync.json

stellaraccident: This will get better. I'm going to poke through enough API so that the compile flags can be completely inferred. But was too much for tonight (and in any case, you do want to be able to set them explicitly too).

stellaraccident: Presumably then, too, if worse comes to worse, XFAIL could be inferred based on what device it actually is.

stellaraccident: I don't anticipate having a lot of XFAIL'y stuff in shortfin. We should just be fixing things.

PR #427, opened 2024-11-05
(shortfin-sd) Adds compile action to sdxl model builder.

monorimet:
A few new builder-related changes in this PR:
 - Adds `--build_preference` to server arguments (choices `["compile", "precompiled"]`)
 - Isolates builder invocations by submodel for finer grained compilation flag control
 - Adds a flagfile for GFX942 that contains optimal flag settings for iree-compile by submodel.
 - Runs builder in a subprocess to prevent compile arg confusion


shortfin/python/shortfin_apps/sd/examples/sdxl_flags_gfx942.txt
```diff
@@ -0,0 +1,23 @@
+all
```
saienduri, line 1: Just want to appreciate that this a lot easier to maintain than how we were doing in SHARK-ModelDev :). Related question: Do we have a list of target architectures that we promise to support when we ship? Might be worth putting in some command line help info at least

monorimet: For V1.0 we are just on gfx942. I'll add in choices to the CLI

saienduri: Nice, sounds good!

shortfin/python/shortfin_apps/sd/components/builders.py
```diff
@@ -1,7 +1,9 @@
 from iree.build import *
+from iree.build.executor import FileNamespace
```
stellaraccident, line 2: As a follow-up, can we have a separate tree for compilation? Move this out from under sd/components. Maybe into sd/compile. May want to do more later, but it's generally idea to have a strong separation between compilation and runtime.

shortfin/python/shortfin_apps/sd/components/builders.py
```diff
@@ -142,21 +154,38 @@ def needs_update(ctx):
...
+    vmfb_name = f"{filename}_{device}-{target}.vmfb"
+    namespace = FileNamespace.BIN
+    return ctx.file(vmfb_name)
+
+
 @entrypoint(description="Retreives a set of SDXL submodels.")
 def sdxl(
     model_json=cl_arg(
-        "model_json",
+        "model-json",
```
stellaraccident, line 188: As a followup, let's be consistent on underscore or dash flags throughout the sd directory.

stellaraccident (approved):
Thanks. Let's take this as-is. I can see various ways it needs to improve, but one step at a time.

shortfin/python/shortfin_apps/sd/components/builders.py
ScottTodd: I was expecting compilation to live in sharktank, like here: https://github.com/nod-ai/SHARK-Platform/tree/main/sharktank/sharktank/models , where we already depend on `torch`, `iree-compiler`, and other packages.

Does SDXL use anything from sharktank? Just the punet model? I put some basic setup docs in the readme that tried to explain the separation: https://github.com/nod-ai/SHARK-Platform?tab=readme-ov-file#support-matrix

I guess shortfin/ makes sense for models that don't use any custom "model recipes"? For example onnx models that just want to use the build system + standard compiler flows.

monorimet: When we do builders with exports enabled it will make much more sense. Currently, we get along without the sharktank import for SDXL, and I think it's prudent to make the cut this way until we have exports prepared with iree.build (those would warrant a spot in sharktank, and their enablement would require the sharktank dep.)
At that point, they can function largely the same but be equipped with export capabilities, and the server can decide which builder to use based on which dependencies/options are present.

PR #431, opened 2024-11-06
[shortfin] Plumb through `amdgpu_tracing_level` config option.

stellaraccident:
* Corresponds to IREE runtime `--hip_tracing` flag.
* Adds `amdgpu_tracing_level=<int>` with a default of 2 to match IREE.
* Verified that the default produces traces with GPU zones and that both the Python level setting and the env var (`SHORTFIN_AMDGPU_TRACING_LEVEL`) works.
* Not a good way to automate this kind of test, so I just did manual verification.

### DIFF
--- a/shortfin/python/lib_ext.cc
+++ b/shortfin/python/lib_ext.cc
@@ -1267,6 +1267,27 @@ environment variable is searched as a fallback in all cases. Multiple paths
 can be separated by semicolons on all platforms.
 )";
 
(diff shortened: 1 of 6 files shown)


ScottTodd (approved):

stellaraccident:
Added minimal unit test of the setting itself.

ScottTodd (approved):

PR #434, opened 2024-11-06
[shortfin] A number of tweaks for dylib builds.

stellaraccident:
* When bundling deps, sets visibility and export control options.
  * Re-export versioned spdlog symbols so that clients share registry, etc.
  * Forces inline hidden visibility.
  * Enables IREE default visibility so that the API is re-exported.
* Applies a version script on Linux so that all symbols get versioned.
* Makes it possible to use IREE tracing macros in binaries that depend on the dylib.
* Enables (thin) LTO by default.
* Removes the obsolete find_package() method of depending on IREE. It is incompatible with how we have been doing runtime source builds for a long time.

Requires an IREE version bump to pick up https://github.com/iree-org/iree/pull/19068

### DIFF
--- a/shortfin/build_tools/cmake/shortfin_library.cmake
+++ b/shortfin/build_tools/cmake/shortfin_library.cmake
@@ -20,26 +20,64 @@ function(shortfin_public_library)
   cmake_parse_arguments(
     _RULE
     ""
-    "NAME"
-    "COMPONENTS"
+    "NAME;LINUX_LD_SCRIPT"
+    "SRCS;COMPONENTS;USAGE_DEPS"
     ${ARGN}
   )
+
+  # Get usage requirements from requested USAGE_DEPS and forward them from the
+  # public library. This is because the contents of the libraries stop at the
+  # public library vs propagating to callers. So we must manually control
+  # the combined usage requirements of the aggregate library.
+  set(_usage_include_directories)
+  set(_usage_compile_definitions)
+  foreach(_usage_dep _shortfin_defs ${_RULE_USAGE_DEPS})
+    get_target_property(_value ${_usage_dep} INTERFACE_INCLUDE_DIRECTORIES)
+    if(_value)
+      list(APPEND _usage_include_directories ${_value})
+    endif()
+    get_target_property(_value ${_usage_dep} INTERFACE_COMPILE_DEFINITIONS)
+    if(_value)
+      list(APPEND _usage_compile_definitions ${_value})
+    endif()
+  endforeach()
+
+  # Useful for debugging include/definition issues.
+  # message(STATUS "Public library ${_RULE_NAME}: Includes = ${_usage_include_directories}")
+  # message(STATUS "Public library ${_RULE_NAME}: Definitions = ${_usage_compile_definitions}")
+
   if(SHORTFIN_BUILD_STATIC)
     # Static library.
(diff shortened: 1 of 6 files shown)


shortfin/src/CMakeLists.txt
```diff
@@ -20,10 +20,15 @@ message(STATUS "Linking optional components '${_SHORTFIN_LIB_OPTIONAL_COMPONENTS
 shortfin_public_library(
   NAME
     shortfin
+  LINUX_LD_SCRIPT ${CMAKE_CURRENT_SOURCE_DIR}/shortfin.ld
```
marbre, line 23:
```suggestion
  LINUX_LD_SCRIPT ${libshortfin_SOURCE_DIR}/build_toold/shortfin.ld
```
Should we move `shortfin/src/shortfin.ld` to `shortfin/build_toold/shortfin.ld`?

marbre (approved):
The patch does what the PR description says, even though I don't think I've been able to fully understand every detail (yet).

I assume it is intended that also `fmt` symbols are re-exported (?).

shortfin/src/CMakeLists.txt
stellaraccident: There will eventually be multiple shared libraries, and the linker scripts are source code, so I think they live with the library.

stellaraccident:
> The patch does what the PR description says, even though I don't think I've been able to fully understand every detail (yet).
> 
> I assume it is intended that also `fmt` symbols are re-exported (?).

For the moment, yes. Basically all of this was working around a root bug where deps of the shared libraries were being linked as PUBLIC, which was causing the users of the shared library to duplicatively link them all in. That's why any of this was working at all. This patch at least gets the bundling working properly, and it can be improved in future iterations from at least a correct starting point.

shortfin/build_tools/cmake/shortfin_library.cmake
```diff
@@ -20,26 +20,64 @@ function(shortfin_public_library)
...
+  # Useful for debugging include/definition issues.
+  # message(STATUS "Public library ${_RULE_NAME}: Includes = ${_usage_include_directories}")
+  # message(STATUS "Public library ${_RULE_NAME}: Definitions = ${_usage_compile_definitions}")
+
   if(SHORTFIN_BUILD_STATIC)
     # Static library.
     shortfin_components_to_static_libs(_STATIC_COMPONENTS ${_RULE_COMPONENTS})
-    add_library("${_RULE_NAME}-static" STATIC)
+    add_library("${_RULE_NAME}-static" STATIC ${_RULE_SRCS})
+    target_compile_definitions("${_RULE_NAME}" INTERFACE
```
stellaraccident, line 53: This needs to be "${_RULE_NAME}-static"

shortfin/build_tools/cmake/shortfin_library.cmake
```diff
@@ -20,26 +20,64 @@ function(shortfin_public_library)
...
+      _SHORTFIN_USING_DYLIB
+      ${_usage_compile_definitions}
+    )
+    target_include_directories("${_RULE_NAME}" INTERFACE ${_usage_include_directories})
```
stellaraccident, line 57: And here.

stellaraccident:
Windows failure is to an IREE issue that got swallowed with this update. Will land with that broken and fix after.

PR #436, opened 2024-11-06
[sharktank] Fix numerics for perplexity with vmfb

archana-ramalingam:
Fix cache update to resolve numerics issue. 
Update cache directly via IREE's DeviceArray.


sharktank/sharktank/evaluate/perplexity_vmfb.py
```diff
@@ -74,7 +74,7 @@ def __init__(
         self.iree_hal_target_backends = iree_hal_target_backends
         self.kv_cache_type = kv_cache_type
         self.activation_dtype = torch.float32
```
sogartar, line 76: wasn't activation also float16 by default?

sogartar (changes requested):

sharktank/sharktank/evaluate/perplexity_vmfb.py
archana-ramalingam: Updated to float16.

sogartar (approved):

PR #438, opened 2024-11-06
Switch user instructions to install iree-turbine from PyPI.

ScottTodd:
Progress on https://github.com/nod-ai/SHARK-Platform/issues/294 and https://github.com/nod-ai/SHARK-Platform/issues/359.

Now that a new release of `iree-turbine` is available (https://pypi.org/project/iree-turbine/#history, https://github.com/iree-org/iree-turbine/releases/tag/v2.5.0, we can drop the requirement to use source builds or nightly packages. Developers will still likely want to use the latest versions of `iree-turbine`, `iree-compiler`, and `iree-runtime`, and that is what the CI jobs are testing. We could also add separate CI jobs that use stable versions of each package.


marbre (approved):
Nice!

PR #439, opened 2024-11-06
XFail Accuracy of CPU LLM Server Integration Test

stbaione:
This test is currently failing. See https://github.com/nod-ai/SHARK-Platform/issues/437 for details.


ScottTodd (approved):
nit: please add a bit of context to the PR description.

this would suffice:
```
This test is currently failing. See https://github.com/nod-ai/SHARK-Platform/issues/437 for details.
```
and linking to the issue like that will point between this PR and the issue (https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue)

build_tools/integration_tests/llm/cpu_llm_server_test.py
```diff
@@ -82,4 +82,8 @@ def test_llm_server(llm_server, available_port):
     assert llm_server.poll() is None
     output = do_generate("1 2 3 4 5 ", available_port)
     logger.info(output)
-    assert output.startswith("6 7 8")
+    # TODO: Remove when fixed: https://github.com/nod-ai/SHARK-Platform/issues/437
+    with pytest.xfail(
+        reason="Accuracy with Shortfin LLM Server is decreased from latest iree-compiler RC."
+    ):
+        assert output.startswith("6 7 8")
```
ScottTodd, line 89: This is fine, but here are some other ideas to consider:

```python
# add this to the top of the test file, a conftest.py, or a utils file
class AccuracyValidationException(RuntimeError):
    pass

# put this in the test body
expected_output_prefix = "6 7 8"
if not output.startswith(expected_output_prefix):
  raise AccuracyValidationException(f"Expected '{output}' to start with '{expected_output_prefix}'")

# Put this on the `def test_llm_server`
@pytest.mark.xfail(raises=AccuracyValidationException)
```

I think that would give a bit more context in the logs when the test starts passing again, or fails for a different reason

stbaione: Updated

build_tools/integration_tests/llm/cpu_llm_server_test.py
```diff
@@ -82,4 +82,8 @@ def test_llm_server(llm_server, available_port):
     assert llm_server.poll() is None
     output = do_generate("1 2 3 4 5 ", available_port)
     logger.info(output)
-    assert output.startswith("6 7 8")
+    # TODO: Remove when fixed: https://github.com/nod-ai/SHARK-Platform/issues/437
```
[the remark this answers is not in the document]
stbaione: This is nice, updated to your suggestion

ScottTodd (approved):
There is also a way to show the xfail reason in the logs even if the test xfails as expected. This is fine for now. Hopefully we don't have too many xfails for long

https://github.com/nod-ai/SHARK-Platform/actions/runs/11714371321/job/32628923321?pr=439#step:6:2

PR #440, opened 2024-11-07
Set sharktank package deps during package build reliably.

ScottTodd:
Progress on https://github.com/nod-ai/SHARK-Platform/issues/294.

We could mirror requirements  between `requirements.txt` and `pyproject.toml` or keep them dynamic. For packaging, we probably don't want to set explicit versions unless absolutely required. Developers may want to pin to a specific version to make testing consistent though.

As part of this I added a dep on `iree-turbine` to sharktank's `requirements.txt` and then switched workflows and user guides to install `sharktank` first and _then_ install a newer/local version from git if desired.


sharktank/requirements.txt
```diff
@@ -1,16 +1,21 @@
+iree-turbine
+
 # Runtime deps.
-gguf==0.6.0
-numpy==1.26.3
+gguf>=0.6.0
+numpy>=1.26.3
+
+# Needed for gguf (TODO: check why transitive dep is missing)
+sentencepiece>=0.1.98,<=0.2.0
```
ScottTodd, line 8: I think this should be set: https://github.com/ggerganov/llama.cpp/blob/5c333e014059122245c318e7ed4ec27d1085573c/gguf-py/pyproject.toml#L21-L26

Need to debug a bit more. Logs from what I've tried: https://gist.github.com/ScottTodd/07029ddd75a409934bd52098231686f2

Note to self: try installing just gguf on its own and see what happens.

sharktank/requirements.txt
ScottTodd, line 1: > We could mirror these between requirements.txt and pyproject.toml or keep them dynamic. For packaging, we don't want to set explicit versions unless absolutely required. Developers may want to pin to a specific version to make testing consistent.

@stellaraccident opinions?

ScottTodd: Industry guidance:
* https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#dependencies-and-requirements
* https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata

    > Library packagers are discouraged from using overly strict (or “locked”) dependency versions in their dependencies and optional-dependencies.

sharktank/requirements.txt
ScottTodd: Tested gguf, it's missing?

```bash
python3.11 -m venv 3.11.venv
source 3.11.venv/bin/activate
python -m pip install gguf
pip freeze
# gguf==0.10.0
# numpy==2.1.3
# PyYAML==6.0.2
# tqdm==4.67.0
```

sharktank/requirements.txt
```diff
@@ -1,7 +1,12 @@
+iree-turbine
+
 # Runtime deps.
 gguf==0.6.0
 numpy==1.26.3
```
ScottTodd, line 5: I ran into errors trying to unpin these versions (replacing `==` with `>=`). Sample logs: https://github.com/nod-ai/SHARK-Platform/actions/runs/11714432277/job/32629101709#step:6:64

Maybe one of the deps doesn't support numpy 2.0? I can play dependency-updates-whack-a-mole to try unpinning _some_ of the packages I guess...?

sharktank/requirements.txt
```diff
@@ -1,7 +1,12 @@
+iree-turbine
```
ScottTodd, line 1: > As part of this I added a dep on `iree-turbine` to sharktank's `requirements.txt` and then switched workflows and user guides to install `sharktank` first and _then_ install a newer/local version from git if desired.

@stellaraccident is this set of changes okay?

sharktank/README.md
```diff
@@ -50,7 +50,15 @@ python -m sharktank.tools.dump_gguf --hf-dataset=open_llama_3b_v2_f16_gguf
 * To build wheels for Linux:
 
     ```bash
-    sudo ./build_tools/build_linux_package.sh
+    ./build_tools/build_linux_package.sh
+    ```
+
+    That should produce
+    `build_tools/wheelhouse/sharktank-3.0.0.dev0-py3-none-any.whl`, which can
```
marbre, line 57:
```suggestion
    `build_tools/wheelhouse/sharktank-2.9.0.dev0-py3-none-any.whl`, which can
```
Want to have to current version or a generic
```suggestion
    `build_tools/wheelhouse/sharktank-{X.Y.Z}.dev0-py3-none-any.whl`, which can
```
instead?

sharktank/README.md
```diff
@@ -50,7 +50,15 @@ python -m sharktank.tools.dump_gguf --hf-dataset=open_llama_3b_v2_f16_gguf
...
+    then be installed with
+
+    ```bash
+    python3 -m pip install build_tools/wheelhouse/sharktank-3.0.0.dev0-py3-none-any.whl
```
marbre, line 61: Maybe
```suggestion
    python3 -m pip install build_tools/wheelhouse/sharktank-2.9.0.dev0-py3-none-any.whl
```
or
```suggestion
    python3 -m pip install build_tools/wheelhouse/sharktank-{X.Y.Z}.dev0-py3-none-any.whl
```
? But can also leave as is.

sharktank/requirements.txt
marbre: We probably need to, yes :/ Can help with that on Monday. Should probably fill a follow up issue to keep track.

marbre (approved):

PR #442, opened 2024-11-07
[shortfin] Add options to set allocators and logical devices per physical.

stellaraccident:
* `allocators=`, `amdgpu_allocators=` and `hostcpu_allocators=` keywords control application of allocator decorators. Adding specifically to enable `caching`.
* `amdgpu_logical_devices_per_physical_device=n` will create `n` logical HAL devices per physical/visible AMDGPU device. This is used for some forms of multi-device distribution.

Filed #443 with a disabled test which seems to tickle an IREE memory leak in `iree_hal_configure_allocator_from_specs`.


stellaraccident:
@monorimet Please post review. Getting landed for the morning perf runs.

PR #444, opened 2024-11-07
(shortfin-sd) Program initialization and logging improvements

monorimet:


monorimet:
Best found throughput number (11/6) is at .89 samples per second per gpu. More progress to come.

shortfin/dev_me.py
```diff
@@ -196,7 +197,8 @@ def configure_mode(env_info: EnvInfo, args):
         env_vars["SHORTFIN_IREE_SOURCE_DIR"] = env_info.iree_dir
     if env_info.clang_exe:
         env_vars["CC"] = env_info.clang_exe
-        env_vars["CXX"] = f"{env_info.clang_exe}++"
+        clangxx_exe = "clang++-".join(env_info.clang_exe.split("clang-"))
```
stellaraccident, line 200: Does this work for unadorned "clang" commands? Maybe drop this file from the patch and we do it separately?

shortfin/python/shortfin_apps/sd/components/manager.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, device="local-task", device_ids=None):
                         raise ValueError(f"Device id {did} could not be parsed.")
             else:
                 selected = available
-            sb = sf.amdgpu.SystemBuilder(amdgpu_visible_devices=";".join(selected))
+            sb = sf.SystemBuilder(system_type="amdgpu")
```
stellaraccident, line 37: You can just create the SystemBuilder once:

```
sb = sf.SystemBuilder(system_type="amdgpu")
sb.visible_devices = get_selected_devices(sb)
self.ls = sb.create_system()
```

(i.e. you can set properties between construction of the SystemBuilder and create_system())

Note that the property takes a normal list of strings so no need to join.

It also looks like you dropped the visible_devices thing in this patch.

shortfin/src/shortfin/local/systems/amdgpu.cc
```diff
@@ -23,6 +23,7 @@ AMDGPUSystemBuilder::AMDGPUSystemBuilder(iree_allocator_t host_allocator,
   iree_hal_hip_device_params_initialize(&default_device_params_);
   InitializeDefaultSettings();
   config_options().ValidateUndef();
+  default_device_params_.async_allocations = false;
```
stellaraccident, line 26: Can drop this now.

stellaraccident (approved):
Some nits.

shortfin/python/shortfin_apps/sd/components/manager.py
monorimet: yeah, that was for making sure the specified device IDs actually matched an available device. I don't really like it, so I'll pass in the raw args.

PR #447, opened 2024-11-07
Stop generation on EOS & Allow specifying token length on client.

renxida:
One thing I'm not sure of is how different export methods would generate different configs.

Right now this relies on tokenizer_config.json specifying which token is the EOS.


stbaione (approved):
LGTM, it doesn't seem like the sdxl test failure is related to the changes here, but would be good to double check

stbaione (approved):
LGTM

PR #456, opened 2024-11-08
Add prefill 8B f16 torch sdpa test, update tests with compile flags and tp flags, with nightly iree

aviator19941:
1. Update tests with compile flags and tp flags, with nightly iree.
2. Only use 8b f16 decomposed test to decrease the CI wait time.

### DIFF
--- a/.github/workflows/ci-llama-large-tests.yaml
+++ b/.github/workflows/ci-llama-large-tests.yaml
@@ -71,8 +71,8 @@ jobs:
 
           # Test with nightly releases, not what iree-turbine uses.
           pip install -f https://iree.dev/pip-release-links.html --upgrade \
-            iree-base-compiler==2.9.0rc20241108 \
-            iree-base-runtime==2.9.0rc20241108 \
+            iree-base-compiler==2.9.1rc20241110 \
+            iree-base-runtime==2.9.1rc20241110 \
             "numpy<2.0"
 
       - name: Run llama tests
--- a/.github/workflows/ci-llama-large-tests.yaml
+++ b/.github/workflows/ci-llama-large-tests.yaml
@@ -68,7 +68,7 @@ jobs:
           pip install --no-compile -f https://iree.dev/pip-release-links.html --src deps \
             -e "git+https://github.com/iree-org/iree-turbine.git#egg=iree-turbine"
 
-          # Test with pinned nightly releases, not what iree-turbine uses.
+          # Test with nightly releases, not what iree-turbine uses.
           pip install -f https://iree.dev/pip-release-links.html --upgrade \
             iree-base-compiler==3.0.0rc20241118 \
             iree-base-runtime==3.0.0rc20241118
--- a/.github/workflows/ci-llama-large-tests.yaml
+++ b/.github/workflows/ci-llama-large-tests.yaml
@@ -70,12 +70,12 @@ jobs:
 
 
           # Test with nightly releases, not what iree-turbine uses.
-          pip install --upgrade --pre --no-cache-dir -f https://iree.dev/pip-release-links.html \
+          pip install -f https://iree.dev/pip-release-links.html --upgrade --pre \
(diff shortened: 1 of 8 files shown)


.github/workflows/ci-llama-large-tests.yaml
```diff
@@ -68,8 +68,8 @@ jobs:
 
           # Test with pinned nightly releases, not what iree-turbine uses.
           pip install -f https://iree.dev/pip-release-links.html --upgrade \
-            iree-base-compiler==2.9.0rc20241107 \
-            iree-base-runtime==2.9.0rc20241107 \
+            iree-base-compiler \
+            iree-base-runtime \
```
ScottTodd, line 72: Keep comments up to date when changing code.

You can copy the text from other workflows that leave the version unpinned:
https://github.com/nod-ai/SHARK-Platform/blob/27b061e376276db1799f96360ca68176c4692642/.github/workflows/ci-shark-platform.yml#L65-L66

sharktank/sharktank/utils/export_artifacts.py
```diff
@@ -202,19 +202,29 @@ def compile_to_vmfb(
             f"{mlir_path}",
             f"--iree-hip-target={self.iree_hip_target}",
             f"--iree-hal-target-backends={self.iree_hal_target_backends}",
+            "--iree-dispatch-creation-enable-aggressive-fusion=true",
+            "--iree-global-opt-propagate-transposes=true",
+            "--iree-opt-aggressively-propagate-transposes=true",
+            "--iree-opt-data-tiling=false",
+            '--iree-preprocessing-pass-pipeline="builtin.module\\(util.func\\(iree-preprocessing-generalize-linalg-matmul-experimental\\)\\)"',
```
ScottTodd, line 209: This `export_artifacts` script looks generic. Don't add program-specific flags to a generic script, much less "experimental" flags. If an individual test wants to opt in to flags, add them to the test (e.g. by allowing extra args to be passed to this function).

ScottTodd (dismissed):

.github/workflows/ci-llama-large-tests.yaml
marbre: I just bumped to the latest nightly, thus this shouldn't be required (for now).

.github/workflows/ci-llama-quick-tests.yaml
ScottTodd, line 1: I see that the https://github.com/nod-ai/SHARK-Platform/actions/runs/11829302221/job/32960886459?pr=456 workflow run is queued, following 15 other jobs (7 hour queue).

Can you link to logs of the workflow running the new test here and/or in the PR description so we know that the new tests work as expected?

sharktank/tests/models/llama/benchmark_amdgpu_test.py
```diff
@@ -173,6 +180,41 @@ def testBenchmark8B_f16_Decomposed(self):
             cwd=self.repo_root,
         )
 
+    @skipif_run_8b_llama
+    def testBenchmark8B_f16_Non_Decomposed_Prefill(self):
```
ScottTodd, line 184: This skip seems inverted to me. Skip 8b_f16 if "run_8b_llama" is set?

Do we have a central list of pytest marks? Looks like https://github.com/nod-ai/SHARK-Platform/blob/main/sharktank/conftest.py

I'd like to see a developer guide for how to run each set of tests and filter based on model family, model size, hardware required, etc. Once such a list of common usage exists it should be easier to square the workflow files against that (and shard across multiple runners, decide what to run on every commit vs nightly, etc.).

.github/workflows/ci-llama-quick-tests.yaml
aviator19941: https://github.com/nod-ai/shark-ai/actions/runs/11829302221/job/32960886459?pr=456

saienduri (approved):

PR #457, opened 2024-11-08
Bump IREE version to 2.9.0rc20241108

marbre:


shortfin/requirements-iree-compiler.txt
```diff
@@ -1,4 +1,4 @@
 # Keep in sync with IREE_REF in CI and GIT_TAG in CMakeLists.txt
```
ScottTodd, line 1: Can you update this comment to `"ref: iree-" in .github/workflows/*` ?
I tried searching for `IREE_REF` and didn't find anything.

You could also introduce a workflow variable / env var / matrix value called `IREE_REF`

ScottTodd (approved):

shortfin/requirements-iree-compiler.txt
marbre: I think you wanted me to *not* introduce an `IREE_REF` variable, see https://github.com/nod-ai/SHARK-Platform/pull/166#discussion_r1744483172 ;)

ScottTodd: Ah. Well that's fine. At least make the comment here easier to follow then :P

ScottTodd (approved):

PR #473, opened 2024-11-11
Add workflow to test packages

marbre:


.github/workflows/test_packages.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Setup python
+        uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.3
+        with:
+          python-version: 3.11
+      - name: Install Python packages
+        run: |
+          python -m pip install sharktank==${{ github.event.inputs.sharktank_package_version }} \
+            --no-deps -f https://github.com/nod-ai/SHARK-Platform/releases/expanded_assets/dev-wheels
+          python -m pip install shortfin==${{ github.event.inputs.shortfin_package_version }} \
+            --no-deps -f https://github.com/nod-ai/SHARK-Platform/releases/expanded_assets/dev-wheels
```
ScottTodd, line 33: Why `--no-deps`? Imports will likely fail if the packages don't also download their dependencies.

.github/workflows/build_packages.yml
```diff
@@ -134,3 +136,18 @@ jobs:
...
+
+  test:
+    name: "Trigger test packages"
+    needs: build_packages
+    runs-on: ubuntu-24.04
+    steps:
+      - name: "Invoke workflow :: Test Packages"
+        uses: benc-uk/workflow-dispatch@e2e5e9a103e331dad343f381a29e654aea3cf8fc # v1.2.4
+        with:
+          workflow: Validate and Publish Release
```
ScottTodd, line 148: Does this "Validate and Publish Release" workflow exist? Did you meant to invoke "Test packages" instead?

.github/workflows/test_packages.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Test import sharktank
+        id: import_sharktank
+        run: |
+          python -c "import sharktank; print('Sanity check passed')"
+      - name: Test import shortfin
+        id: import_shortfin
+        run: |
+          python -c "import shortfin; print('Sanity check passed')"
```
ScottTodd, line 41: Probably want to split this into multiple jobs:

* Test shortfin
* Test sharktank
* Test both together

Each package should be able to stand on its own (at least until we make one of them depend on the other)

For the combined case, we can run tests/workflows like https://github.com/nod-ai/SHARK-Platform/blob/main/.github/workflows/ci-shark-platform.yml too.

ScottTodd (changes requested):

.github/workflows/test_packages.yml
marbre: If we want to test a thing like `from shartank import ops`, that's true. For what is tested right now I would like to avoid downloading torch and especially all the deps that get pulled in by torch.

.github/workflows/build_packages.yml
marbre: Must be `Test packages` -.-

.github/workflows/test_packages.yml
ScottTodd: We need to be testing what users will do.

marbre: I removed `--no-deps`.

PR #475, opened 2024-11-11
Add a shark-ai meta package

marbre:


packaging/shark-platform/pyproject.toml
```diff
@@ -0,0 +1,42 @@
+[build-system]
+requires = ["setuptools", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "shark-platform"
```
ScottTodd, line 6: Shall we build this as part of https://github.com/nod-ai/SHARK-Platform/blob/main/.github/workflows/build_packages.yml too?

packaging/shark-platform/pyproject.toml
```diff
@@ -0,0 +1,42 @@
...
+  "License :: OSI Approved :: Apache Software License",
+  "Programming Language :: Python :: 3",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
+]
+requires-python = ">= 3.10"
+
+version = "2.9.0"
```
ScottTodd, line 24: Could source this from a file, so the build_packages workflow can update a `_rc` version and then a single script like https://github.com/iree-org/iree/blob/main/build_tools/python_deploy/promote_whl_from_rc_to_final.py can run on all the packages when ready to promote and push to pypi.

packaging/shark-platform/README.md
ScottTodd, line 1: Could populate this with some of the text from https://github.com/nod-ai/SHARK-Platform/pull/244

ScottTodd (commented):
Nice!

packaging/shark-platform/pyproject.toml
marbre: Sure, that can definitely be done. I think if we automatically build it, we also want to auto-generate a `requirements.txt` file which pulls in the correct versions (which means the exact `X.Y.ZrcYYYYMMDD` versions for a nightly meta-package` and the `Z.Y.*` packages for a stable release). WDYT?

packaging/shark-platform/pyproject.toml
ScottTodd: I want releasing to be as automated as possible, so the developer / release engineer that is managing the process has to run a minimal number of commands and make as few decisions as possible.

* For the iree-org/iree packages, the process is to run https://github.com/iree-org/iree/blob/main/build_tools/python_deploy/pypi_deploy.sh pointed at a github release (one nightly release per day, so each release has only 1 version of each whl)
* For iree-org/iree-turbine, https://github.com/iree-org/iree-turbine/blob/main/docs/releasing.md has a sequence of commands to run: build package, test, upload to pypi. The build package step there involves a decision about the package version.
* For nod-ai/SHARK-Platform, we have scripts to build the wheels locally and dev wheels (from multiple versions) uploaded to https://github.com/nod-ai/SHARK-Platform/releases/tag/dev-wheels. We can either download from the dev-wheels or build locally, then update the package version, then publish.

(As an aside, I also want the dependency updating to be automated too - I lost most of Friday debugging dependency update issues in iree-turbine since we have no automation for that and things were broken. The release engineer should _just_ be updating packages and publishing, they shouldn't also have to debug the code)

ScottTodd: So... idk what specific recommendation I have here. We can start with a developer needing to build the meta package manually when publishing, but we should also be able to have test workflows in this project that pip install the nightly meta package and test it.

packaging/shark-platform/pyproject.toml
ScottTodd: How about we start with manually authored and run. We can iterate on the automation after a few runs through the process.

packaging/shark-platform/pyproject.toml
ScottTodd: Hmmm... could have a script that reads the release index for https://github.com/nod-ai/SHARK-Platform/releases/tag/dev-wheels and lets you pick a specific version to download, promote, and push. Could default to the latest nightly, prompt from a list, or require specifying explicitly (e.g. `2.9.0rc20241111` for today)

packaging/shark-platform/pyproject.toml
marbre: I added `build_tools/python_deploy/write_requirements.py` to write the file based on what version is provided via a previously written `version_local.json` (which we need to generate anyway).

packaging/shark-platform/pyproject.toml
marbre: I've now drafted it in a similar fashion to what I've written for `iree`. The script `build_tools/python_deploy/compute_common_version.py` computes a common version and writes it to a `version_local.json` file which is then parsed via a `setup.py`. Furthermore, the requirements file is also auto-generated as this feels more convenient. If we want to, we can simply add building the meta package to what is already existing in the CI.

build_tools/python_deploy/compute_common_version.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env python3
+# Copyright 2024 Advanced Micro Devices, Inc.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# This scripts grabs the X.Y.Z[.dev]` version identifier from the
+# sharktank and shortfin version files and computes the version
+# for the meta package.
```
ScottTodd, line 10: nit: open \` to line up with the closing \`

```suggestion
# This scripts grabs the `X.Y.Z[.dev]` version identifier from the
# sharktank and shortfin version files and computes the version
# for the meta package.
```

build_tools/python_deploy/compute_common_version.py
```diff
@@ -0,0 +1,75 @@
...
+    CURRENT_VERSION = SHORTFIN_BASE_VERSION
+
+if args.nightly_release:
+    CURRENT_VERSION += "rc" + datetime.today().strftime("%Y%m%d")
+
+if args.write_json:
+    version_local = {"package-version": CURRENT_VERSION}
+    write_version_info()
+
+print(CURRENT_VERSION)
```
ScottTodd, line 75: /s/CURRENT/COMMON ?

```suggestion
if SHARKTANK_BASE_VERSION > SHORTFIN_BASE_VERSION:
    COMMON_VERSION = SHARKTANK_BASE_VERSION
else:
    COMMON_VERSION = SHORTFIN_BASE_VERSION

if args.nightly_release:
    COMMON_VERSION += "rc" + datetime.today().strftime("%Y%m%d")

if args.write_json:
    version_local = {"package-version": COMMON_VERSION}
    write_version_info()

print(COMMON_VERSION)
```

build_tools/python_deploy/write_requirements.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python3
+# Copyright 2024 Advanced Micro Devices, Inc.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# This scripts writes the `packaging/shark-platform/requirements.txt` file
```
ScottTodd, line 8: plural -> singular

```suggestion
# This script writes the `packaging/shark-platform/requirements.txt` file
```

build_tools/python_deploy/write_requirements.py
```diff
@@ -0,0 +1,66 @@
...
+# and pins the versions of the dependencies accordingly. For nighly releases,
+#  * sharktank
+#  * shortfin
+# get pinned to the corresponding nighly version. For stable releases,
+# * iree-base-compiler
+# * iree-base-runtime
+# * iree-turbine
+# * sharktank
+# * shortfin
+# get pinned to the corresponding `X.Y.*` version.
```
ScottTodd, line 18: Ooo this is cool. I hope it does what we want/expect though. Since a nightly/prerelease package has different pins than a stable release, we can't directly promote the nightly meta package to stable. We can design the release scripting to account for that, but any developers running scripts manually will have to watch for that.

packaging/shark-platform/pyproject.toml
```diff
@@ -0,0 +1,38 @@
...
+requires = ["setuptools", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "shark-platform"
+authors = [
+  {name = "SHARK Authors"},
+]
+description = "SHARK Platform meta package"
+readme = "README.md"
```
ScottTodd, line 11: Do you know which `README.md` this is? Should we use the repository root `/README.md`, instead of `/packaging/shark-platform/README.md`?

marbre: For now it is the one side by side to the `pyproject.toml`, no idea if we can include the one in the repo root instead.

build_tools/python_deploy/write_requirements.py
marbre: Fair point. But as it is a python only package without any real content I assume we can just build a new package?

ScottTodd: Yep. Easy enough to build locally for the desired configuration, but that does introduce a point where what we push to pypi is not what we run automated tests on. Hopefully that's fine for the metadata-only package.

ScottTodd (approved):

PR #478, opened 2024-11-11
Fix path to `shark-ai` package

marbre:


ScottTodd (approved):

ScottTodd:
We can also add a workflow like https://github.com/iree-org/iree-turbine/blob/main/.github/workflows/test_build_release.yml that goes through the release steps and runs a smoketest. That should catch typos like these.

marbre:
> We can also add a workflow like https://github.com/iree-org/iree-turbine/blob/main/.github/workflows/test_build_release.yml that goes through the release steps and runs a smoketest. That should catch typos like these.

On the list :)

PR #480, opened 2024-11-12
[shortfin] Adds C++ tracing scopes.

stellaraccident:
* Manually chose a number of things that would be interesting to see in a trace and added annotations.
* QC'd with some unit test traces.

### DIFF
--- a/shortfin/src/shortfin/support/logging.h
+++ b/shortfin/src/shortfin/support/logging.h
@@ -23,6 +23,14 @@
 #define SHORTFIN_SCHED_LOG(...)
 #endif
 
+// Tracing macros. These are currently just aliases of the underlying IREE
+// macros, but we maintain the ability to redirect them in the future (i.e.
+// for certain kinds of library builds, etc).
+#define SHORTFIN_TRACE_SCOPE IREE_TRACE_SCOPE
+#define SHORTFIN_TRACE_SCOPE_NAMED(name_literal) \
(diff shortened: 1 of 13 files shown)


monorimet (approved):
LGTM.

shortfin/src/shortfin/support/logging.h
```diff
@@ -23,6 +23,14 @@
 #define SHORTFIN_SCHED_LOG(...)
 #endif
 
+// Tracing macros. These are currently just aliases of the underlying IREE
+// macros, but we maintain the ability to redirect them in the future (i.e.
+// for certain kinds of library builds, etc).
+#define SHORTFIN_TRACE_SCOPE IREE_TRACE_SCOPE
+#define SHORTFIN_TRACE_SCOPE_NAMED(name_literal) \
+  IREE_TRACE_SCOPE_NAMED(name_literal)
+#define SHORTFIN_TRACE_SCOPE_ID IREE_TRACE_SCOPE_ID
```
ScottTodd, line 32: Could put these in a new `tracing.h` instead of `logging.h`. Here seems fine given the small number of macros though.

ScottTodd (approved):

shortfin/src/shortfin/support/logging.h
stellaraccident: I'm fine with where they are for this kind of API.

PR #482, opened 2024-11-12
[shortfin] Add conversion host ops.

### LINKED ISSUE #315: (sfnp) add "round" host op
Only used in one place currently:
https://github.com/nod-ai/SHARK-Platform/blob/089f788cc4ec25cca92e23091b5fa61af3815965/shortfin/python/shortfin_apps/sd/components/service.py#L607

Since it's different than casting to int, it's probably a good idea to have a host op for rounding array elements to their nearest integer.

torch.round docs: https://pytorch.org/docs/stable/generated/torch.round.html

also related (from `round()` docs): [torch.ceil()](https://pytorch.org/docs/stable/generated/torch.ceil.html#torch.ceil), which rounds up. [torch.floor()](https://pytorch.org/docs/stable/generated/torch.floor.html#torch.floor), which rounds down. [torch.trunc()](https://pytorch.org/docs/stable/generated/torch.trunc.html#torch.trunc), which rounds towards zero.

stellaraccident:
Ops added: `convert`, `round`, `ceil`, `floor`, `trunc`

All ops were implemented to the same pattern, supporting fused conversion and output array.

Fixes #315

### DIFF
--- a/shortfin/python/array_host_ops.cc
+++ b/shortfin/python/array_host_ops.cc
@@ -38,6 +38,34 @@ Returns:
   A device_array of dtype=int64, allocated on the host and not visible to the device.
 )";
 
+static const char DOCSTRING_CONVERT[] =
+    R"(Does an elementwise conversion from one dtype to another.
+
+The same behavior exists for several conversion ops:
+
+* `convert` : element-wise conversion like a static cast.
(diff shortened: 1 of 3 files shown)


monorimet (approved):
LGTM! I'll add these into sf-sd tomorrow. Thanks.

PR #488, opened 2024-11-12
use persistent cache to store gguf weights for cpu_llm_test

renxida:


renxida:
Doesn't seem to be working. Probably need a deeper look into our ci runners clean up after each run.

PR #489, opened 2024-11-12
Add metadata to shortfin python package.

ScottTodd:
Progress on https://github.com/nod-ai/SHARK-Platform/issues/130.

This will add more information to https://pypi.org/project/shortfin/. See also https://pypi.org/project/sharktank/. Both README files are due for updates too (https://github.com/nod-ai/SHARK-Platform/issues/359).

Tested with a sanity check build of `pip install -v -e .` from `/shortfin/`. Not sure if there is a way to preview the PyPI page locally but could push to TestPyPI (https://packaging.python.org/en/latest/guides/using-testpypi/) I suppose :thinking:

PR #492, opened 2024-11-13
Add metadata to shortfin Python package

marbre:
Similar to #489, this adds mertadata to the shortfin Python package but does so by extending the `pyproject.toml` file instead of adding metadata to the `setup.py` file.

### DIFF
--- a/shortfin/pyproject.toml
+++ b/shortfin/pyproject.toml
@@ -8,6 +8,32 @@ requires = [
 ]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "shortfin"
+authors = [
+  {name = "SHARK Authors"},
+]
+description = "SHARK inference library and serving engine"
+readme = "README.md"
+license = {text = "Apache-2.0"}
+classifiers = [
+  "Development Status :: 3 - Alpha",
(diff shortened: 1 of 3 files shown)


shortfin/pyproject.toml
```diff
@@ -8,6 +8,32 @@ requires = [
 ]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "shortfin"
+authors = [
+  {name = "SHARK Authors"},
+]
+description = "SHARK inference library and serving engine"
+readme = "README.md"
```
ScottTodd, line 17: Can you also update the README text in this PR to match this description?

https://github.com/nod-ai/SHARK-Platform/blob/ce6ccf88e8f106a6a88b1946076139c011a8f5e8/shortfin/README.md?plain=1#L1

(It would be weird for a project on PyPI to have "C++ library" right at the top of the page)

ScottTodd (approved):

shortfin/pyproject.toml
marbre: Good point! Sure, I will add it to this PR prior to merging.

PR #502, opened 2024-11-13
[shortfin] Read IREE version from text file

marbre:
Instead of hardcoding the version in the CMakeLists and the workflows, a `iree_version.txt` file is introduced that is used to store the version.


.github/workflows/ci_linux_x64-libshortfin.yml
```diff
@@ -53,13 +53,19 @@ jobs:
...
+        echo "IREE_REF=iree-${IREE_VERSION}" >> $GITHUB_ENV
+
     - name: Checkout IREE repo
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       with:
         repository: iree-org/iree
         path: ${{ env.IREE_REPO_DIR }}
         submodules: false
-        ref: iree-2.9.0rc20241108
+        ref: ${{ env.IREE_REF }}
```
ScottTodd, line 68: This does limit us to only using git tags matching the `iree-` format (so any nightly releases). For source checkouts, we've pinned to specific commits before.

shortfin/CMakeLists.txt
```diff
@@ -40,7 +40,8 @@ if(NOT WIN32)
 endif()
 
 # Pins
-set(SHORTFIN_IREE_GIT_TAG "iree-2.9.0rc20241108")
+file(STRINGS ${CMAKE_CURRENT_SOURCE_DIR}/build_tools/iree_version.txt IREE_VERSION_STRING)
+set(SHORTFIN_IREE_GIT_TAG "iree-${IREE_VERSION_STRING}")
```
ScottTodd, line 44: We can keep the ability to pin to an arbitrary git ref (branch name, tag, or commit hash) by switching the workflows to omit `SHORTFIN_IREE_SOURCE_DIR` and let it use `FetchContent`. I'd rather do that workflow simplification instead.

I'd rather have our most tested path be the default path with the most streamlined setup, instead of the balancing act of cloning multiple repositories and pointing them at each other explicitly.

(and folding the workflows into a single workflow that uses a `matrix` would also save on duplicated code)

shortfin/requirements-iree-compiler.txt
```diff
@@ -1,4 +1,4 @@
-# Keep in sync with "ref: iree-" in .github/workflows/* and GIT_TAG in CMakeLists.txt
+# Keep in sync with version in shortfin/buidl_tools/iree_version.txt
```
ScottTodd, line 1: typo in here, but I think we can just say that this should be compatible with `SHORTFIN_IREE_GIT_TAG` in the CMake and have the workflows use FetchContent

PR #503, opened 2024-11-13
Use IREE versions pinned in requirements file

marbre:
Together with #502 this reduces the locations where IREE versions are pinned to two files.


ScottTodd (approved):

marbre:
I think we're right now stable in a way that we don't need to pin. Will do later if still needed.

PR #505, opened 2024-11-14
Iterate on root README.md.

ScottTodd:
Progress on https://github.com/nod-ai/SHARK-Platform/issues/359

* Relax warning at the top of the file about project status then add a new warning about sharktank status
* Swap shortfin and sharktank ordering under sub-projects (no longer alphabetical, but lead with what is most user-facing)
* Reword developer instructions
* Fix nightly IREE install instructions to use `--pre`

### DIFF
--- a/docs/nightly_releases.md
+++ b/docs/nightly_releases.md
@@ -67,7 +67,7 @@ python3.11 -m venv 3.11.venv
 source 3.11.venv/bin/activate
 
 # Install 'sharktank' package from nightly releases.
-python -m pip install sharktank -f https://github.com/nod-ai/SHARK-Platform/releases/expanded_assets/dev-wheels
+pip install sharktank -f https://github.com/nod-ai/SHARK-Platform/releases/expanded_assets/dev-wheels
 
 # Test the installation.
 python -c "from sharktank import ops; print('Sanity check passed')"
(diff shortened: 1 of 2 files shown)


docs/nightly_releases.md
```diff
@@ -67,7 +67,7 @@ python3.11 -m venv 3.11.venv
 source 3.11.venv/bin/activate
 
 # Install 'sharktank' package from nightly releases.
-python -m pip install sharktank -f https://github.com/nod-ai/SHARK-Platform/releases/expanded_assets/dev-wheels
+pip install sharktank -f https://github.com/nod-ai/SHARK-Platform/releases/expanded_assets/dev-wheels
```
amd-chrissosa, line 70: I've heard tell of a shark-ai package though don't see it yet in the nightly wheels. Not blocking but wondering when that'll be there.

amd-chrissosa (approved):

docs/nightly_releases.md
ScottTodd: https://pypi.org/project/shark-ai/

discussion about publishing to nightly: https://github.com/nod-ai/SHARK-Platform/pull/475#discussion_r1837044406

(meta package is a bit special when it comes to packaging... not sure if/how we want to publish it nightly too)

amd-chrissosa: ACK I just want whatever instructions we have here to work with the SDXL workflow. Right now if we follow this guide and the SDXL guide iree-compiler, fastapi, and uviron aren't installed. So we should either have the meta package here or ensure these deps get pulled in / ask folks to install them as prereqs when starting with the nightlies.

ScottTodd: Yep. I'll keep iterating on this over the next few days but wanted to checkpoint progress. There is still no mention of just `pip install shortfin` (or any of the other stable packages).

marbre: If we want to publish nightly releases of the shark-ai package, that's easily doable. Regarding the dependencies required for the SDXL workflow, I assume we might want to add so-call extra requires to shortfin (`pip install shortfin[{app.sdxll,sdxl,sd}]`) that pulls in directly used deps like `fastapi`. The same options can be added to the shark-ai package. Other deps like `iree-compiler`get pulled in by shark-ai already by default.

marbre (approved):

PR #506, opened 2024-11-14
Fix workflows installing nightly IREE packages to use `--pre`.

ScottTodd:
Pip will only install "pre-release" versions if `--pre` is set or the version is set explicitly: https://pip.pypa.io/en/stable/cli/pip_install/#pre-release-versions. The `rc` part of our `rcYYYYMMDD` suffix used in the nightly IREE packages is considered a "Pre-release segment" in the version string: https://packaging.python.org/en/latest/specifications/version-specifiers/#public-version-identifiers.

This should change workflow jobs like https://github.com/nod-ai/SHARK-Platform/actions/runs/11827078971/job/32954462017#step:5:163 that have been downloading `iree-base-compiler-2.9.0 iree-base-runtime-2.9.0` to downloading the latest nightly versions instead, as they intend: https://github.com/nod-ai/SHARK-Platform/blob/51cf2f45139f67e15434990173b49375998db9f3/.github/workflows/ci-sharktank.yml#L64-L67

### DIFF
--- a/.github/workflows/ci-shark-platform.yml
+++ b/.github/workflows/ci-shark-platform.yml
@@ -67,7 +67,7 @@ jobs:
           # Try with the latest IREE nightly releases, not what iree-turbine pins.
           # We could also pin to a known working or stable version.
           # This should eventually stabilize. Do the best we can for now.
-          pip install -f https://iree.dev/pip-release-links.html --upgrade \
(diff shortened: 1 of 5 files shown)


marbre (approved):
Thanks for fixing this Scott! Tested locally.

PR #513, opened 2024-11-14
Specify optional dependencies for shortfin_apps

marbre:


shortfin/pyproject.toml
```diff
@@ -34,6 +34,13 @@ dynamic = ["version"]
 Repository = "https://github.com/nod-ai/SHARK-Platform"
 Documentation = "https://shortfin.readthedocs.io/en/latest/"
 
+[project.optional-dependencies]
+apps = [
+  "transformers",
+  "dataclasses-json",
+  "pillow",
+]
```
ScottTodd, line 42: This will allow us to simplify the setup instructions https://github.com/nod-ai/SHARK-Platform/blob/3e7429544fd81938e2b1c0d8fbb098b1c871ecf7/shortfin/python/shortfin_apps/sd/README.md?plain=1#L6-L15

to just one of these:
```bash
pip install shortfin[apps]
pip install shark-ai[apps]
```

shark-ai/pyproject.toml
```diff
@@ -30,6 +30,9 @@ Repository = "https://github.com/nod-ai/SHARK-Platform"
 onnx = [
   "iree-base-compiler[onnx]",
 ]
+apps = [
+  "shortfin[apps]",
+]
```
ScottTodd, line 35: We also considered other names for `extras` (docs here: https://packaging.python.org/en/latest/specifications/dependency-specifiers/#extras).

* `sd` or `sdxl`: these are very specific, but we don't want to have an extra for every model (sd3, llama2, llama3, flux, etc.), unless there is good reason for that. The `transformers`, `dataclasses-json`, and `pillow` dependencies here aren't specific to SDXL either - they are generic ML / data / image libraries
* `examples` or `demos`: these suggest a lower support level than `apps`
* `apps`: this matches the code path and will let us eventually include other deps for other apps. We can later add more specific extras if `apps` gets too large

ScottTodd (approved):

monorimet (approved):
LGTM.

PR #514, opened 2024-11-14
[shark-ai] Temporarily drop sharktank dependency

marbre:
This drops the strict dependency to sharktank which otherwise always gets pulled in when installing the shark-ai package.


ScottTodd (approved):
Rationale:
* For the initial release, SDXL serving only depends on the `shortfin` package.
* The `sharktank` package has compatibility issues across Python and PyTorch versions, so removing it from the meta package allows us to claim broader support with less chance of something going wrong
* Users can still install `sharktank` directly
* The meta package is still useful since it includes `iree-base-compiler`, `iree-base-runtime`, and `iree-turbine` at the same `X.Y.Z` version as `shortfin`
* We can easily add `sharktank` back to the meta package in a future release

amd-chrissosa (approved):
Overall agree with Scott's feedback. We're not highlighting sharktank as part of this release.

PR #517, opened 2024-11-14
Restore RotaryEmbedding use complex numbers

rsuderman:
It appears the change away from complex numbers triggered a downstream iree failure. An inflight work to use `flow.tensor.bitcast` and restore to complex numbers appears to fix the issue. Fixing forward as we wanted to revert part of this change anyway.


renxida (approved):

stbaione (approved):
Nice!

PR #519, opened 2024-11-14
Add scripting and documentation for pushing releases to PyPI.

ScottTodd:
This scripting allows us to publish .whl files from https://github.com/nod-ai/SHARK-Platform/releases/tag/dev-wheels to [PyPI](https://pypi.org/). Here are the basic steps:

1. Download wheels for a specific pre-release (e.g. `2.9.1rc20241114`)
2. Edit the versions in the downloaded wheels to remove the `rcYYYYMMDD` suffix
3. Build the `shark-ai` meta package using the versions in those whls (NOTE: currently this uses the versions _in the source tree_, see below)
4. Upload all wheels to PyPI

Logs of this running: https://gist.github.com/ScottTodd/9c7418d5bbbebc8aea72a39bc2ac37b0 (to push 2.9.1 to PyPI)

The new `pypi_deploy.sh` script is based on the similar script we maintain in IREE: https://github.com/iree-org/iree/blob/main/build_tools/python_deploy/pypi_deploy.sh . The `README.md` file is also forked from IREE.

Known sharp edges with the publishing process:

* This currently mixes local information (versions from the source tree where this is running) with remote information (versions from the nightly release packages). Publishing nightly meta packages will let us simplify and make this safer to run from arbitrary working trees
* The local build of the shark-ai meta package copies _all_ files in `shark-ai/build_tools/wheelhouse/` to the working directory that gets sent to `twine upload *`. If there are preexisting files in that directory they will be published.
* The step that downloads releases from GitHub uses `pip download` so it can leverage PyPI's version resolution logic, but I couldn't figure out a way to download 3.13t wheels without running a 3.13t python interpreter. I happen to have 3.13t installed on my system, but we shouldn't require that in the scripting for all release engineers. We could try using the `gh` tool as in IREE, if we properly filter the download to just the version we want: https://github.com/iree-org/iree/blob/9eaa4ef7d6b439d8c444b533beddd82146578e25/build_tools/python_deploy/pypi_deploy.sh#L69-L72 (that has one release per nightly, while here we have a single release shared between all nightlies)


build_tools/python_deploy/pypi_deploy.sh
```diff
@@ -0,0 +1,124 @@
...
+function upload_wheels() {
+  # TODO: list packages that would be uploaded, pause, prompt to continue
+  echo ""
+  echo "Uploading wheels:"
+  ls
+  twine upload --verbose *
+}
+
+function build_shark_ai_meta_package() {
+  # TODO: download meta package from nightly releases instead of this
```
marbre, line 91:
```suggestion
  # TODO: download meta package from nightly releases instead of this
  #   Be aware that nightly releases pin other dependencies via the
  #   generated `requirements.txt` compared to stable releases.
```

build_tools/python_deploy/pypi_deploy.sh
```diff
@@ -0,0 +1,124 @@
...
+  ${SCRIPT_DIR}/compute_common_version.py --stable-release --write-json
+
+  echo "Writing requirements for shark-ai meta package..."
+  ${SCRIPT_DIR}/write_requirements.py
+
+  echo "Building shark-ai meta package..."
+  ${REPO_ROOT}/shark-ai/build_tools/build_linux_package.sh
+
+  # TODO: This is error-prone. We only want to publish the whl for this release.
+  #   copy instead? specify exact file name? clear directory before building?
```
marbre, line 109:
```suggestion
  #   Copy instead? Specify exact file name? Clear directory before building?
```

build_tools/python_deploy/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Version management
+
+These scripts handle versioning across packages, including considerations like
+major, minor, and patch levels (`X.Y.Z`), as well as suffixes like
+`rc20241107`:
+
+* [`compute_common_version.py`](./compute_common_version.py)
+* [`compute_local_version.py`](./compute_local_version.py)
+* [`promote_whl_from_rc_to_final.py`](./promote_whl_from_rc_to_final.py)
```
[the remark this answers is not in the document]
ScottTodd: Yep. There is overlap.

marbre (approved):

PR #527, opened 2024-11-14
Use dev_me.py for shortfin integration tests

renxida:


.github/workflows/ci-shark-platform.yml
```diff
@@ -71,5 +71,20 @@ jobs:
...
+          echo "Found IREE version: ${IREE_VERSION}"
+          # Clone IREE repository at the parent directory level. Git clone --branch supports tags.
+          cd .. && git clone --depth 1 --branch "iree-${IREE_VERSION}" https://github.com/iree-org/iree.git
+          cd SHARK-Platform
+
+      - name: Install CMake
+        uses: ssrobins/install-cmake@v1 # this action installs cmake into the action workspace
+
+      - name: Install shortfin via dev_me.py
+        run: cd shortfin && python dev_me.py
```
ScottTodd, line 87: Ideally integration tests would use built packages, so they can't assume anything about the source environment. Developer scripts could still be used for unit tests though.

renxida: oh good point. Will take the setup section and swap it into a unit test once done.

ScottTodd: IDK what you're doing with the git clone stuff, but I also want more workflows to just use the `FetchContent` code in CMake that downloads automatically. Developer flows and GitHub Actions should both be simple. Branchy bash scripts that peek at output from pip commands are not simple :P

dev_me.py appears to handle the automatic download case: https://github.com/nod-ai/SHARK-Platform/blob/778b567aa111b558fb7cf6c3fff98c79753f37ef/shortfin/dev_me.py#L29-L31

renxida: just trying to not repeatedly download code - - i was also trying to make this integration test faster

renxida: (now that i look at it iree_repo_dir already seems to cache the iree repo for the unittests; might get that over so that shortfin build doesn't have to pull iree)

renxida:
closing because didn't find the low hanging fruits i wanted to find : /

PR #530, opened 2024-11-15
[tuner][CI] Enable mypy type checking

kuhar:
For now only typecheck the tuner implementation. In the future, we should extend it to examples.


kuhar:
Proof of life: 
image

raikonenfnu (approved):

PR #538, opened 2024-11-15
Bump IREE version to 3.0.0rc20241114

marbre:
Bumps the IREE version (which is among other things used to build shortfin) to a newer release candidate / nightly version.


marbre:
If we not go with the latest nightly, this might be an alternative.

ScottTodd:
> If we not go with the latest nightly, this might be an alternative.

Diff: https://github.com/iree-org/iree/compare/iree-3.0.0rc20241114...iree-3.0.0rc20241115

Seems reasonable to use last night's (15)

marbre:
In that case I close in favor of #537.

PR #550, opened 2024-11-15
Add T5 LM v1.1 encoder

sogartar:
The encoder shares much of the underlying stack as the decoder. Here only the encoder is presented as a class.
I have not gone out of my way to strip all decoder related stuff from the stack. Things like check-pointing and dropout are stripped.

The author attribution is added to the license of the T5 model file as this seems like a derivative work. They are both Apache 2.0.

There are a few tests of the various components and 2 tests for the entire encoder for the small and xxl variants. They relay on huggingface and the models are downloaded no the fly into the cache. The tests expect the corresponding GGUF files to be already preset and available on the file system.

### DIFF
--- a/.github/workflows/ci-sharktank.yml
+++ b/.github/workflows/ci-sharktank.yml
@@ -74,7 +74,7 @@ jobs:
 
   test_with_data:
     timeout-minutes: 1000
-    name: "Data dependent tests"
+    name: "Data-dependent Tests"
     strategy:
       matrix:
         version: [3.11]
@@ -127,6 +127,6 @@ jobs:
 
       - name: Run tests
         run: |
-          pytest --longrun \
+          pytest \
             --with-t5-data \
             sharktank/tests/models/t5/t5_test.py
--- a/.github/workflows/ci-sharktank.yml
+++ b/.github/workflows/ci-sharktank.yml
@@ -70,3 +70,63 @@ jobs:
         if: ${{ !cancelled() }}
         run: |
           pytest -n 4 sharktank/
+
+
+  test_with_data:
+    timeout-minutes: 1000
+    name: "Data dependent tests"
+    strategy:
+      matrix:
+        version: [3.11]
+        runs-on: [llama-mi300x-3]
+      fail-fast: false
+    runs-on: ${{matrix.runs-on}}
+    defaults:
+      run:
+        shell: bash
+    env:
+      PIP_CACHE_DIR: "${{ github.workspace }}/.pip-cache"
+      HF_HOME: "/data/huggingface"
+      SHARK_PLATFORM_REPO_ROOT: ${{ github.workspace }}
+    steps:
+      - name: "Setting up Python"
+        id: setup_python
+        uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.0
+        with:
+          python-version: ${{matrix.version}}
+
+      - name: "Checkout Code"
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Cache Pip Packages
+        uses: actions/cache@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
+        id: cache-pip
+        with:
+          path: ${{ env.PIP_CACHE_DIR }}
+          key: pip-${{ steps.setup_python.outputs.python-version }}-${{ hashFiles('*requirements*.txt','sharktank/requirements*.txt') }}
+
+      - name: Install sharktank deps
+        run: |
+          python -m pip install --no-compile --upgrade pip
+          # Note: We install in three steps in order to satisfy requirements
+          # from non default locations first. Installing the PyTorch CPU
+          # wheels saves multiple minutes and a lot of bandwidth on runner setup.
+          pip install --no-compile -r pytorch-cpu-requirements.txt
+          pip install --no-compile -r requirements.txt -r sharktank/requirements-tests.txt -e sharktank/
+
+          # Install latest iree-tubrine.
+          pip install --no-compile -f https://iree.dev/pip-release-links.html --src deps \
+            -e "git+https://github.com/iree-org/iree-turbine.git#egg=iree-turbine"
+
+          # Try with the latest IREE nightly releases, not what iree-turbine pins.
+          # We could also pin to a known working or stable version.
+          # This should eventually stabilize. Do the best we can for now.
+          pip install -f https://iree.dev/pip-release-links.html --upgrade --pre \
+            iree-base-compiler \
+            iree-base-runtime
+
+      - name: Run tests
+        run: |
+          pytest --longrun \
+            --with-t5-data \
+            sharktank/tests/models/t5/t5_test.py
--- a/.github/workflows/ci-sharktank.yml
+++ b/.github/workflows/ci-sharktank.yml
@@ -73,7 +73,6 @@ jobs:
 
 
   test_with_data:
-    timeout-minutes: 30
     name: "Data-dependent Tests"
     strategy:
       matrix:
--- a/.github/workflows/ci-sharktank.yml
+++ b/.github/workflows/ci-sharktank.yml
@@ -73,7 +73,7 @@ jobs:
 
 
   test_with_data:
-    timeout-minutes: 1000
+    timeout-minutes: 30
     name: "Data-dependent Tests"
     strategy:
       matrix:
--- a/.github/workflows/ci_eval.yaml
+++ b/.github/workflows/ci_eval.yaml
@@ -20,6 +20,8 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.number || github.sha }}
   cancel-in-progress: true
 
+# TODO: refactor out common steps into actions/scripts.
+
 jobs:
   test_perplexity_vmfb:
     if: ${{ github.repository_owner == 'nod-ai' || github.event_name != 'schedule' }}
@@ -125,3 +127,63 @@ jobs:
 
       - name: Run perplexity test in eager mode
         run:  pytest -n 8 -v -s sharktank/tests/evaluate/perplexity_torch_test.py --longrun --llama3-8b-f16-model-path=/data/llama3.1/8b/llama8b_f16.irpa --llama3-8b-tokenizer-path=/data/llama3.1/8b/tokenizer_config.json
+
+  test_torch:
+    timeout-minutes: 1000
+    name: "Torch/Eager mode"
+    strategy:
+      matrix:
+        version: [3.11]
+        runs-on: [llama-mi300x-3]
+      fail-fast: false
+    runs-on: ${{matrix.runs-on}}
+    defaults:
+      run:
+        shell: bash
+    env:
+      PIP_CACHE_DIR: "${{ github.workspace }}/.pip-cache"
+      HF_HOME: "/data/huggingface"
+      SHARK_PLATFORM_REPO_ROOT: ${{ github.workspace }}
+    steps:
+      - name: "Setting up Python"
+        id: setup_python
+        uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.0
+        with:
+          python-version: ${{matrix.version}}
+
+      - name: "Checkout Code"
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Cache Pip Packages
+        uses: actions/cache@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
+        id: cache-pip
+        with:
+          path: ${{ env.PIP_CACHE_DIR }}
+          key: pip-${{ steps.setup_python.outputs.python-version }}-${{ hashFiles('*requirements*.txt','sharktank/requirements*.txt') }}
+
+      - name: Install sharktank deps
+        run: |
+          python -m pip install --no-compile --upgrade pip
+          # Note: We install in three steps in order to satisfy requirements
+          # from non default locations first. Installing the PyTorch CPU
+          # wheels saves multiple minutes and a lot of bandwidth on runner setup.
+          pip install --no-compile -r pytorch-cpu-requirements.txt
+          pip install --no-compile -r requirements.txt -r sharktank/requirements-tests.txt -e sharktank/
+
+          # Install latest iree-tubrine.
+          pip install --no-compile -f https://iree.dev/pip-release-links.html --src deps \
+            -e "git+https://github.com/iree-org/iree-turbine.git#egg=iree-turbine"
+
+          # Try with the latest IREE nightly releases, not what iree-turbine pins.
+          # We could also pin to a known working or stable version.
+          # This should eventually stabilize. Do the best we can for now.
+          pip install -f https://iree.dev/pip-release-links.html --upgrade --pre \
+            iree-base-compiler \
+            iree-base-runtime
+
+      - name: Run long running tests
+        run: |
+          pytest --longrun \
+            --google-t5-v1-1-small-fp32-model-path=/data/t5/small/google__t5-v1_1-small_fp32.gguf \
+            --google-t5-v1-1-xxl-fp32-model-path=/data/t5/xxl/google__t5-v1_1-xxl_fp32.gguf \
+            sharktank/tests/models/t5/t5_test.py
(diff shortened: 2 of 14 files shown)


sharktank/sharktank/layers/ffn_block.py
```diff
@@ -22,18 +23,29 @@ class FFN(ThetaLayer):
     def __init__(
         self,
         theta: Theta,
+        is_gated: bool = None,
```
archana-ramalingam, line 26: How about setting a default boolean value here instead of None.

sharktank/sharktank/layers/ffn_block.py
```diff
@@ -22,18 +23,29 @@ class FFN(ThetaLayer):
...
+        activation_fn: Optional[Callable[[AnyTensor], AnyTensor]] = None,
     ):
         super().__init__(theta)
 
-        self.add_module("ffn_gate", LinearLayer(theta("ffn_gate")))
+        self.is_gated = is_gated is None or is_gated
```
archana-ramalingam, line 31: If default value is boolean, will not need an `or` statement here. Just `self.is_gated = is_gated`

sharktank/sharktank/models/t5/t5.py
```diff
@@ -0,0 +1,1136 @@
...
+)
+from ... import ops
+from ...types.theta import Theta
+from ...layers import FFN
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class T5Config:
```
archana-ramalingam, line 35: This might need to move to `layers/configs/t5_configs.py`. Even better if moved to `llm_configs.py`

sharktank/sharktank/layers/ffn_block.py
sogartar: Having the default as None helps with propagation of the underlying default value to callers. They can easily say that their default should change with the default of the underlying function.
E.g.
```
def f(x: bool = None)
    ...

def g(x: bool = None):
    f(x)
```

sogartar: I changed the default to `True` to avoid confusion.

sharktank/sharktank/layers/ffn_block.py
sogartar: Done

sharktank/sharktank/models/t5/t5.py
sogartar: I moved it there.

sharktank/sharktank/models/t5/t5.py
```diff
@@ -0,0 +1,1067 @@
...
+        if hidden_states.dtype == torch.float16:
+            clamp_value = ops.elementwise(
+                torch.where,
+                ops.elementwise(torch.isinf, hidden_states).any(),
+                torch.finfo(hidden_states.dtype).max - 1000,
+                torch.finfo(hidden_states.dtype).max,
+            )
+            hidden_states = ops.elementwise(
+                torch.clamp, hidden_states, min=-clamp_value, max=clamp_value
+            )
```
sogartar, line 574: @archana-ramalingam, you mentioned some training remnants, do you mean these?
I am a bit reluctant to remove these as if the train was done with clamping, shouldn't inference also clamp? In general we want to run the same function as it was trained on.
The question is if we get a trained fp16 T5, would it have used this clamping during training?

.github/workflows/ci_eval.yaml
```diff
@@ -123,3 +125,63 @@ jobs:
...
+          # Try with the latest IREE nightly releases, not what iree-turbine pins.
+          # We could also pin to a known working or stable version.
+          # This should eventually stabilize. Do the best we can for now.
+          pip install -f https://iree.dev/pip-release-links.html --upgrade --pre \
+            iree-base-compiler \
+            iree-base-runtime
+
+      - name: Run long running tests
+        run: |
+          pytest --longrun \
```
IanNod, line 184: How long does this take to run to have this fall under longrun? Ideally we should be testing the models on presubmit if possible

sharktank/conftest.py
```diff
@@ -256,6 +272,16 @@ def get_model_artifacts(request: FixtureRequest):
     model_path["llama3_405b_fp8_model_path"] = set_fixture_from_cli_option(
         request, "--llama3-405b-fp8-model-path", "llama3_405b_fp8_model"
     )
+    model_path["google__t5_v1_1_small_fp32_model_path"] = set_fixture_from_cli_option(
```
IanNod, line 275: Do we need 2 separate model_path flags for small vs xxl models instead of just calling this twice from our tests?

sharktank/sharktank/ops/default_impls.py
```diff
@@ -141,36 +141,19 @@ def elementwise_binary(operator, x, y, *args, **kwargs):
...
-    (Tensor)
-    (Tensor, Tensor)
-    (int, Tensor, Tensor)
-    ```
-    """
-    res = elementwise(operator, x, y)
-    for arg in args:
-        res = elementwise(operator, res, arg)
-    return res
+def elementwise_trenary(operator, x, y, z, *args, **kwargs):
```
IanNod, line 150: Why is this change needed? Also should this be elementwise_ternary?

sharktank/sharktank/ops/signatures.py
```diff
@@ -374,6 +376,23 @@ def _gather_trampoline(
         d.fail(dispatch_args)
 
 
+def gelu_tanh_approximation(input: AnyTensor) -> AnyTensor:
```
IanNod, line 379: lots of hardcoded values here. Are those all specified by the linked paper?

sharktank/tests/models/t5/t5_test.py
```diff
@@ -0,0 +1,290 @@
...
+        [
+            "google/t5-v1_1-small",
+            "google/t5-v1_1-xxl",
+        ]
+    )
+    @longrun
+    def testV1_1Fp32CompareTorchEagerAgainstHuggingFace(self, huggingface_repo_id: str):
+        get_dataset(
+            huggingface_repo_id,
+        ).download()
```
IanNod, line 56: should we just cache the model for the CI instead of downloading each time? May end up with failures due to corrupted downloads on occasion.

sharktank/tests/ops/ops_test.py
```diff
@@ -35,26 +34,6 @@ def testBroadcastDims(self):
         assert res[1] == 2
 
 
-class ElementwiseTest(unittest.TestCase):
```
IanNod, line 38: Why are we no longer testing?

.github/workflows/ci_eval.yaml
sogartar: It takes 30 seconds. These are 2 models small and xxl. Maybe I can leave only small on presubmit.

sogartar: These presubmit tests will need to run on `llama-mi300x-3` or other machines that have the files available.

sharktank/sharktank/ops/default_impls.py
KyleHerndon: I also expect this should be InferenceTensor

.github/workflows/ci_eval.yaml
sogartar: I refactored a bit. Added option `--with-t5-data` that enables tests that depend on T5 data. Made only the XXL eager mode test longrun. Added a presubmit job that runs the `small` model test.

sharktank/sharktank/ops/default_impls.py
sogartar: Fixed the typo.
Originally I added this for convenience. Here is removed as it conflicts with dispatching to ternary ops. If the user wants binary op + folding they should do that themselves.

sharktank/tests/ops/ops_test.py
sogartar: This tests the binary op + folding, which is no longer present.

sharktank/sharktank/ops/signatures.py
sogartar: Yes, this approximation is proposed in the paper.

.github/workflows/ci_eval.yaml
```diff
@@ -20,10 +20,12 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.number || github.sha }}
   cancel-in-progress: true
 
+# TODO: refactor out common steps into actions/scripts.
+
 jobs:
   test_perplexity_vmfb:
     timeout-minutes: 1000
-    name: "IREE/vmfb"
+    name: "IREE/vmfb perplexity"
```
archana-ramalingam, line 28: Nit: I removed this earlier because it's `CI-Perplexity`.

sharktank/conftest.py
```diff
@@ -133,6 +143,20 @@ def pytest_addoption(parser):
         help="Llama3.1 405b fp8 model path",
     )
 
+    parser.addoption(
+        "--google-t5-v1-1-small-fp32-model-path",
+        type=Path,
+        default="/data/t5/small/google__t5-v1_1-small_fp32.gguf",
```
archana-ramalingam, line 149: We don't hardcode any llama model/tokenizer paths anymore here. You can pass it as an arg directly to pytest.

sharktank/sharktank/models/t5/t5.py
archana-ramalingam: Yes, this was it. Are you able to test if removing this block made a diff in the results?

.github/workflows/ci_eval.yaml
```diff
@@ -123,3 +125,63 @@ jobs:
...
+          # Try with the latest IREE nightly releases, not what iree-turbine pins.
+          # We could also pin to a known working or stable version.
+          # This should eventually stabilize. Do the best we can for now.
+          pip install -f https://iree.dev/pip-release-links.html --upgrade --pre \
+            iree-base-compiler \
+            iree-base-runtime
+
+      - name: Run long running tests
+        run: |
+          pytest
```
archana-ramalingam, line 184: I need to check if the same perplexity dataset fits for both llama and T5.

sharktank/conftest.py
sogartar: Why not have a default that allows you to `pytest sharktank/tests` if you have data at default paths?

sharktank/sharktank/layers/ffn_block.py
```diff
@@ -22,18 +23,29 @@ class FFN(ThetaLayer):
...
         self,
         theta: Theta,
+        is_gated: bool = True,
+        activation_fn: Optional[Callable[[AnyTensor], AnyTensor]] = None,
     ):
         super().__init__(theta)
 
-        self.add_module("ffn_gate", LinearLayer(theta("ffn_gate")))
+        self.is_gated = is_gated
+        self.activation_fn = activation_fn or F.silu
```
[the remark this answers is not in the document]
sogartar: Done.

.github/workflows/ci_eval.yaml
sogartar: Done.

sharktank/sharktank/models/t5/t5.py
sogartar: I have not done any float16 tests yet. Only float32 as this is the quantization of the original model from Google. I think this is an addition others made when doing fine tunning. I want to leave it in as this is also how inference works on the HF model.
This infinity probably comes from some previous calculation and if it appears it will blow up downstream results. It probably happens rarely.

.github/workflows/ci_eval.yaml
sogartar: I fixed the pytest command to be
```
pytest \
...
```

sharktank/tests/evaluate/perplexity_torch_test.py
```diff
@@ -8,10 +8,9 @@
 import pytest
 import json
 
+from sharktank.utils.testing import longrun
```
archana-ramalingam, line 11: I will add this change in my other PR, we can remove this here, if not using perplexity.

.github/workflows/ci_eval.yaml
```diff
@@ -123,3 +125,63 @@ jobs:
 
       - name: Run perplexity test in eager mode
         run:  pytest -n 8 -v -s sharktank/tests/evaluate/perplexity_torch_test.py --longrun --llama3-8b-f16-model-path=/data/llama3.1/8b/llama8b_f16.irpa --llama3-8b-tokenizer-path=/data/llama3.1/8b/tokenizer_config.json
+
```
archana-ramalingam, line 128: We can revert all changes to `ci_eval.yaml`, if not using perplexity.

sharktank/sharktank/layers/configs/llm_configs.py
```diff
@@ -179,3 +179,72 @@ class LlamaModelConfig:
     # be the difference of many gigabytes of static data being embedded in
     # the program and not.
     static_tables: bool = True
+
+
+@dataclass
+class T5Config:
+    return_dict: bool = True
```
archana-ramalingam, line 186: This is not being used elsewhere. Also, possible to stay consistent with the format `LlamaHParams` returns?

sharktank/sharktank/layers/ffn_block.py
```diff
@@ -22,18 +23,29 @@ class FFN(ThetaLayer):
     def __init__(
         self,
         theta: Theta,
+        is_gated: bool = True,
+        activation_fn: Callable[[AnyTensor], AnyTensor] = F.silu,
```
archana-ramalingam, line 27: Since we have default values for both `is_gated` and `activation_fn`. Can we make them optional? Llama models use this default values, so shouldn't break anything.

sharktank/sharktank/layers/configs/llm_configs.py
sogartar: It is possible to refactor out something the is common to Llama and T5. T5 does not have rotary embedding layer and and has `relative_buckets_count`. In general we could make dataclass hierarchy that reflects the GGUF properties. On the other hand there is already the properties dictionary if you want general. These config classes are tailored towards the model. There is some potential for code reuse. I would rather do that later.

sharktank/sharktank/layers/ffn_block.py
sogartar: Do you mean `Optional[SomeType]`?
`Optional[SomeType] == Union[SomeType, type(None)]`. They are not really optional in that sense as they can't be `None`. I had them optional in the first iteration, but you requested to not use `None` values for defaults.

.github/workflows/ci_eval.yaml
sogartar: Done.

sharktank/tests/evaluate/perplexity_torch_test.py
sogartar: I revered this change.

sharktank/sharktank/models/t5/t5.py
```diff
@@ -0,0 +1,1067 @@
...
+        relative_attention_num_buckets: int,
+        relative_attention_max_distance: int,
+        d_model: int,
+        d_kv: int,
+        num_heads: int,
+        activation_dtype: torch.dtype,
+        has_relative_attention_bias: bool = False,
+    ):
+        super().__init__()
+        self.is_decoder = is_decoder
```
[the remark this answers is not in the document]
sogartar: Yes, I decided to leave this part in if/when we add the decoder.

.github/workflows/ci-sharktank.yml
```diff
@@ -70,3 +70,63 @@ jobs:
         if: ${{ !cancelled() }}
         run: |
           pytest -n 4 sharktank/
+
+
+  test_with_data:
+    timeout-minutes: 1000
```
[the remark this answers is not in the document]
sogartar: I reduced it to 30 minutes.

sharktank/sharktank/layers/ffn_block.py
archana-ramalingam: Oh yeah, my bad, this shouldn't break llama. This looks good.

.github/workflows/ci-sharktank.yml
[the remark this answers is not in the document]
IanNod: Does this take long enough that we need to specify a timeout here at all?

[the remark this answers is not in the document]
sogartar: You are right. I removed it. The default is 6 hours. Should be plenty. We should expect this to not stall.

IanNod (approved):

sharktank/sharktank/ops/default_impls.py
sogartar: @KyleHerndon, I think this should not accept any inference tensor as we do unboxing, which would be undesirable for some tensor types. The unboxing here is fine as it is a no-op.
If someone wants an implementation for example for quantized tensors they should write one.
I guess it is about whether we want to perform potentially a very inefficient computation or to fail.

sharktank/conftest.py
sogartar: I added a comment.

PR #554, opened 2024-11-16
[tuner] Use `ir.(Integer|Float)Type` for element types

kuhar:
This follows the general goal of using the IREE/MLIR bindings instead of working with textual IR.


Groverkss (approved):
Very cool!

PR #555, opened 2024-11-17
(shortfin-sd) Cleanup fiber distribution, logging, error handling.

monorimet:
- Switches fiber distribution to a much simpler process where idle fibers are kept in a set and pop/replaced when used. For dense batches this should be near optimal, though it cycles through fibers without regard of each fiber's current load (this only matters in per_call/none isolation where we allow in-fiber concurrency -- in per_call/none isolation the fibers are put back in the idle_fibers set after boarding)

- Suppresses leak messages for shortfin nanobind objects with keep_awake (maybe fix and rollback after V3.0)
- Makes port selection/error return much more friendly
- Segments configuration and suppresses some builder outputs for now.
- Makes logging look a little better, removes some extraneous outputs.


shortfin/python/lib_ext.h
```diff
@@ -65,6 +65,7 @@ template <typename CppType, typename KeepAlivePatient, typename... Args>
 inline py::object custom_new_keep_alive(py::handle py_type,
                                         KeepAlivePatient &keep_alive,
                                         Args &&...args) {
+  py::set_leak_warnings(false);
```
stellaraccident, line 68: Please move this to the module registration function in lib_ext.cpp. it is a global. 

Agreed this thing is flaky. ASAN does a more reliable job.

shortfin/python/shortfin_apps/sd/server.py
```diff
@@ -135,7 +136,7 @@ def get_configs(args):
         f"--model={modelname}",
         f"--topology={topology_inp}",
     ]
-    outs = subprocess.check_output(cfg_builder_args).decode()
+    outs = subprocess.check_output(cfg_builder_args, stderr=subprocess.DEVNULL).decode()
```
stellaraccident, line 139: This makes debugging impossible... We really need to finish the build API and get this stuff more for free

shortfin/python/shortfin_apps/sd/server.py
```diff
@@ -158,18 +159,19 @@ def get_configs(args):
                 arglist = spec.strip("--").split("=")
                 arg = arglist[0]
                 if len(arglist) > 2:
+                    print(arglist)
```
stellaraccident, line 162: Stray print?

shortfin/python/shortfin_apps/sd/server.py
```diff
@@ -222,7 +223,14 @@ def get_modules(args, model_config, flagfile, td_spec):
             f"--iree-hip-target={args.target}",
             f"--iree-compile-extra-args={' '.join(ireec_args)}",
         ]
-        output = subprocess.check_output(builder_args).decode()
+        logger.info(f"Preparing runtime artifacts for {modelname}...")
+        logger.debug(
+            f"COMMAND LINE EQUIVALENT: "
+            + " ".join([str(argn) for argn in builder_args])
```
stellaraccident, line 229: If using an fstring, don't also concat like this.

shortfin/python/shortfin_apps/sd/server.py
```diff
@@ -222,7 +223,14 @@ def get_modules(args, model_config, flagfile, td_spec):
...
+        )
+        output = subprocess.check_output(
+            builder_args, stderr=subprocess.DEVNULL
```
stellaraccident, line 232: Ditto. Debugging not good.

stellaraccident (approved):
A number of small nits. Approving since I'm going to be out today and you can address.

shortfin/python/shortfin_apps/sd/server.py
```diff
@@ -158,18 +159,19 @@ def get_configs(args):
...
                 arg = arglist[0]
                 if len(arglist) > 2:
+                    print(arglist)
                     value = arglist[1:]
                     for val in value:
                         try:
                             val = int(val)
                         except ValueError:
-                            continue
+                            val = val
```
amd-chrissosa, line 168: Why are you doing val=val and value=value below?

amd-chrissosa (approved):
Overall took a look through it all. It's a pretty decent change to get in right before the deadline so please run flake8 / pylint against it to get some static checking in before submitting. It'll help to give extra qa coverage since you are changing some of the function inputs/outputs and you may have missed an edge case

shortfin/python/shortfin_apps/sd/server.py
monorimet: This is mostly for the device ids arg, where user can pass either device indexes (ints) or device UIDs (strings).
The logic being: if we can't convert them to ints, then we pass them as they were recieved.
doing _continue_ here breaks parsing of non-list/int arguments e.g. `isolation`.

monorimet:
@amd-chrissosa thanks for the pylint/flake8 suggestion -- didn't find any regressions but did clean up quite a few unused imports/variables.

stellaraccident:
I could have sworn we had those enabled in pre-commit. Probably running of another repo. Should fix that.

PR #557, opened 2024-11-18
Use default GitHub app token

marbre:


ScottTodd (approved):
Oh. Nice if this works. Thanks!

marbre:
> Oh. Nice if this works. Thanks!

It should and did when testing in my local fork. Also matches the documentation of the action, which says that the token _"[..] is primarily useful if you want to use your personal token (for targeting other repos, etc)."_

PR #560, opened 2024-11-18
(shortfin-sd) Validate size of downloaded artifacts.

monorimet:
This commit introduces simple filesize/content-length comparison for a simple validation of downloaded artifacts. 
While I'm not sold on this as a long-term validation solution, it is certainly better than nothing.


stellaraccident (approved):
Ok for now. We should fix this properly upstream too.

monorimet:
> Ok for now. We should fix this properly upstream too.

Thanks. I created an issue: https://github.com/nod-ai/shark-ai/issues/563

PR #561, opened 2024-11-18
(shortfin) Increase loop queue/wait limit for workers.

monorimet:
This is a relatively arbitrary limit that we increase to allow more traffic on server at a given time.


stellaraccident (approved):
Good for now. We'll replace with a more robust impl soon

amd-chrissosa (approved):

PR #565, opened 2024-11-18
Replace AttnPagedCache with BasePagedAttentionCache

renxida:
Creates space for #593 (prefix-sharing)

Coming next: #607 , which should be the last thing I do before I can check in my blocktrie implementation.

Summary of changes:
- copied over stella's cache.py and renamed it to page_pool.py
- each inference request now notifies the cache when its pages are done written to


shortfin/python/shortfin_apps/llm/components/kvcache/page_pool.py
```diff
@@ -0,0 +1,159 @@
...
+                device, page_table_shape, self.config.dtype
+            )
+            self.page_tables.append(page_table)
+
+    def acquire_free_pages(self, count: int) -> list[PageInfo] | None:
+        with self._lock:
+            available = len(self.attn_page_free)
+            if count > available:
+                return None
+            return [self.attn_page_free.pop() for _ in range(count)]
```
stbaione, line 114: Kind of a nit, but might be relevant in cases where we're acquiring a large number of pages:

```python
pages = self.attn_page_free[:count]
self.attn_page_free = self.attn_page_free[count:]
return pages
```

Removes entire slice instead of individually popping and saves the overhead of creating a new list with list comprehension. I don't think we need to explicitly remove from the end of `self.attn_page_free`, so should be fine without the pop. If we do need to explicitly remove from end, we could negative index `count`. 

Does unfortunately turn a one-liner into 3, but this might be the case where it's worth it

shortfin/python/shortfin_apps/llm/components/kvcache/page_pool.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def release_pages(self, pages: list[PageInfo]):
```
stbaione, line 116: Suppose we have 10 pages that we call `release_pages` on, but 3 of them have non-zero reference counts. What happens in that scenario? Or is that handled already by the `lock`?

shortfin/python/shortfin_apps/llm/components/kvcache/base_attention_cache.py
```diff
@@ -0,0 +1,79 @@
...
+        Given a list of tokens and pages containing KV corresponding to these tokens, make these pages available to other requests.
+
+        Associates the tokens with the pages, and mark them as done writing.
+
+        It is assumed that hereafter, the calling request will not modify these pages, at least not the positions [0:len(tokens)].
+        """
+
+        pass  # the base implementation doesn't cache unfinished requests.
+
+    def release_pages(self, tokens, pages):
```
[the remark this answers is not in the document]
renxida: bcs later we'll have to find the node in the RadixTree to release it.

shortfin/python/shortfin_apps/llm/components/kvcache/page_pool.py
renxida: By the time PagePool.release_pages is called, all the pages should have reference count zero.

I'm still struggling a little with how to do reference counting though. There's the RadixNode (in the RadixTree) which could (and probably should) have a reference count. Then there are the kvcache page which should maytbe also be reference counted. Depending on whether PageInfo->RadixNode is a one-to-one or one-to-many relationship.

shortfin/python/shortfin_apps/llm/components/kvcache/page_pool.py
renxida: I think this is a good idea when we go larger

Check this out

https://colab.research.google.com/drive/1otgqrRgSkdc1d0hQ8Zxuq85UqeuGoCQ0?usp=sharing


    Size |      Pop Method |    Slice Method |      Ratio
-------------------------------------------------------
       1 |        0.000027 |        0.000165 |       0.17
      10 |        0.000031 |        0.000147 |       0.21
     100 |        0.000055 |        0.000149 |       0.37
    1000 |        0.000157 |        0.000153 |       1.02

Popping is much faster than slicing up to some 1000 pages at a time. But beyond that slicing gets slower very fast.

renxida: hm but now that i think about it, at that point maybe we should just increase the page size.

renxida: i think size 1 is just mostly dominated by the set-up so maybe we should toss that result!

shortfin/python/shortfin_apps/llm/components/kvcache/page_pool.py
stbaione: I could see one-to-many being possible.

Check out example below, and lemme know if you agree that it's a valid case. I could see scenarios where the same node is repeated on a different branch on the radix tree. IIUC, in this scenario, nodes C and F should be pointing to the same KVCache tensors, right?

```mermaid
graph LR
    A[A - SYSTEM PROMPT]
    B[B - User: Hello!<br>Assistant: Hi!]
    C[C - User: What's your favorite color?]
    D[D - User: How are you?]
    E[E - Assistant: Good!]
    F[F - User: What's your favorite color?]

    A --> B
    B --> C
    B --> D
    D --> E
    E --> F

    C -.-> G[...]
    F -.-> H[...]
```

shortfin/tests/apps/llm/components/kvcache/page_pool_test.py
```diff
@@ -0,0 +1,57 @@
...
+import shortfin as sf
+import shortfin.host
+import shortfin.array as sfnp
+import shortfin.amdgpu
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.fixture
+def setup_pool(generic_device):
```
stbaione, line 13: Is it possible to have a simulate multi-device using `generic_device`? That would be a nice test case to have

shortfin/python/shortfin_apps/llm/components/kvcache/page_pool.py
renxida: Not for most models!

For most language models, in the sentences

He stalks his prey
The wheat stalks dry in the summer sun

"stalks" would have different KV.

There are some models that I vaguely remember would be able to share KV if a stretch of preceeding tokens are the same. I think they're called windowed models? Not sure.

shortfin/tests/apps/llm/components/kvcache/page_pool_test.py
renxida: I think it's probably possible!

We would just have to make it a list. Will create an issue for this.

renxida: https://github.com/nod-ai/shark-ai/issues/599

shortfin/python/shortfin_apps/llm/components/kvcache/base_attention_cache.py
```diff
@@ -0,0 +1,79 @@
...
+        - tokens: all the known tokens for this generation request
+        - extra_token_slots: number of kvcache slots needed in addition to the ones needed to hold the given tokens.
+
+        In the base implementation, this will just allocate all new pages, but in shared-kv implementations, we will fetch cached pages if applicable.
+
+        The pages are returned in order.
+
+        No token at idx < n_cached_token should be written to. TODO: consider enforcing this.
+        """
+        pages_needed = math.ceil(len(tokens) + extra_token_slots / self.tokens_per_page)
```
stbaione, line 56: Should this be:

```python
math.ceil((len(tokens) + extra_token_slots) / self.tokens_per_page
```

For `boardprefills`, we set extra_token_slots == 0. 

Let `len(tokens)` == 99, and `token_per_page` == 25, then `pages_needed` = `math.ceil(99 + 0 / 25)` == `math.ceil(99 + 0)` == 99. Looks like it should be 25

renxida: ack! good catch!

renxida: really odd that i was able to pass the integration test despite this. i guess we were just not spanning enough pages. Might put in another pr later to do a slightly longer generation on the integration tests.

renxida: Fixed!

stbaione (approved):

PR #570, opened 2024-11-19
SGLang Integration + Accuracy Tests, Restructure app_tests/integration_tests

stbaione:
# Description

This PR implements integration tests for the Shortfin LLM Server w/ the SGLang integration. It uses llama3-8b-instruct on GPU, which is downloaded using sharktank's `hf_datasets` script.

The tests server two purposes:
1. Test that the SGLang integration works properly at a functional level.
2. Test that the accuracy of the responses from the shortfin LLM server are consistent. 
    - We have a batch of candidate questions, with expected answers
    - We have temperature set to `1.0`, so the responses should be deterministic.
 
This test is intended to run every 4 hours, which allows for us to detect degradations in shortfin LLM output accuracy. If we do get a failure due to an accuracy degradation, there will only be a small set of shark-ai/iree commits that could be responsible.


app_tests/integration_tests/llm/sglang/conftest.py
```diff
@@ -0,0 +1,123 @@
...
+from ..utils import (
+    find_available_port,
+    start_llm_server,
+    download_with_hf_datasets,
+    export_paged_llm_v1,
+    compile_model,
+)
+
+pytest.importorskip("sglang")
+import sglang as sgl
```
renxida, line 21: typically i see this written as

sgl = pytest.importorskip("sglang")

However, I think this looks nicer because although there are concerns with repeat import, import is supposed to be idempotent anyways & this way we get a syntax highlighted "import" keyword.

app_tests/integration_tests/llm/sglang/sglang_frontend_test.py
```diff
@@ -0,0 +1,309 @@
...
+        "--iree-hal-target-backends=rocm",
+        "--iree-hip-target=gfx942",
+    ],
+    "device": "hip",
+}
+
+ACCEPTED_THRESHOLD = 0.8
+
+
+def compute_similarity(model: SentenceTransformer, sentence_1: str, sentence_2: str):
```
renxida, line 35: NICE!

If we end up needing a faster version of this CI though, we might need a smaller model / another model free approach to compute sentence similarity.

renxida (approved):

app_tests/integration_tests/llm/sglang/conftest.py
stbaione: Yeah, I had a double thought about this, but decided to keep it. You don't get any IntelliSense in the editor if you do `sgl = pytest.importorskip("sgl")`. Worth it to import it in line below, to make writing the tests much easier.

app_tests/integration_tests/llm/sglang/sglang_frontend_test.py
stbaione: Agreed, most of the other solutions required way too much setup to be able to get embeddings. But, there should be a way to do this without loading a model at all. 

I considered seeing if I could get the embedding with shortfin LLM itself, but that felt like using shortfin tokens to validate shortfin tokens

PR #577, opened 2024-11-20
[tuner] Refactor tuner flow to use unified functions for models and dispatches

Max191:
This PR refactors the tuner flow into a unified compilation + benchmark path, and adds an example test showing how to use the new flow. The PR does not remove any of the code from the old paths, because the new path requires a newer version of iree-compiler, which is not compatible with most of the tuner code yet. This PR is quite lengthy so the new flow is described below:

### Goals and Motivations ###
The purpose of this PR is first to unify the model and dispatch compilation and benchmark flow, and second to remove much of the spooky action at a distance between tuner clients and the internal tuner implementation. The first is accomplished by having a single compile and benchmark function, and the second is greatly improved by attempting to hide as much of the CandidateTracker information from the client as possible.

### Candidate Generation ###
Candidate generation uses the same constraint logic to determine a set of potential configurations, and then directly generates transform dialect specs based on the input mlir file, and the generated constraints. The TD specs are created by matching specific operation types in the input mlir file, and then creating a `transform.iree.match.cast_compatible_dag_from_root` in the TD named sequence matcher. This uses a mix of python bindings for matching root ops, and string formatting for building the TD spec, but can eventually be shifted to using exclusively python bindings.

The main difference in the new candidate generation is that only TD specs are created, instead of generating the full configured source for compilation later. The source will be stripped of compilation info and compiled with the TD spec during the compile step.

### Candidate Compilation ###
Candidates are now compiled with a single function for both models and dispatches. There are a few primary differences here:
 - Candidates are stripped of their compilation info with the `iree-codegen-strip-compilation-info` pass before compilation. This allows dispatches to be compiled in the same way as models.
 - The TuningClient has a new function called `get_iree_compile_flags`, which is used in place of `get_dispatch_compile_command` and `get_model_compile_command`. This new function hides the candidate trackers from the client, and is only meant to return some additional iree-compile flags (not including the source file path or `iree-hal-target-backends`). This simplifies the TuningClient implementations significantly, and makes it much easier to understand how to create one.
 - The compile function takes an optional argument to use as the compilation input. This overwrites whatever benchmark file was used for candidate generation during compilation. This allows using the same compile command for the full model by overwriting the input with the model ir.

### Candidate Benchmarking ###
Similar to the compilation, the `get_dispatch_benchmark_command` and `get_model_benchmark_command` are replaced by a single `get_iree_benchmark_module_flags` function. This function hides the candidate trackers from the client, and simply expects a list of extra flags for iree-benchmark-module (not including `--device`, `--module`, or `--benchmark_format`). The benchmark parsing also needed to be rewritten to match the new benchmarking command.

### DIFF
--- a/tuner/examples/test/tuner_test.py
+++ b/tuner/examples/test/tuner_test.py
@@ -1,0 +1,165 @@
+# Copyright 2024 Advanced Micro Devices, Inc
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+"""
+Sample Usage:
+
+python -m examples.test double_mmt.mlir mmt_benchmark.mlir --devices=hip://0,hip://1 --num-candidates=64
+
+
+Recommended Trial Run:
+
+python -m examples.test double_mmt.mlir mmt_benchmark.mlir --num-candidates=10
+
+
+Dry Run Test (no gpu required, not currently working):
+
+python -m examples.test double_mmt.mlir mmt_benchmark.mlir --num-candidates=64 --dry-run
+
+"""
+
+from tuner import libtuner
+from pathlib import Path, PurePath
+import argparse
+import os
+
+
+class TestTuner(libtuner.TuningClient):
+    def __init__(self):
+        self.compile_flags = [
+            "--iree-hip-target=gfx942",
+            "--compile-from=executable-sources",
+        ]
+        self.benchmark_flags = [
+            "--benchmark_repetitions=3",
+            "--benchmark_format=json",
+        ]
+
+    def get_compile_timeout_s(self) -> int:
+        return 10
+
+    def get_iree_compile_flags(self) -> list[str]:
+        return self.compile_flags
+
+    def get_benchmark_timeout_s(self) -> int:
+        return 15
+
+    def get_iree_benchmark_module_flags(
+        self,
+    ) -> list[str]:
+        return self.benchmark_flags
+
+    # TODO(Max191): Remove the following unused abstract functions once they
+    # are removed from the TuningClient definition.
+    def get_dispatch_benchmark_timeout_s(self) -> int:
+        return 0
+
+    def get_dispatch_compile_timeout_s(self) -> int:
+        return 0
+
+    def get_dispatch_compile_command(
+        self, candidate_tracker: libtuner.CandidateTracker
+    ) -> list[str]:
+        return []
+
+    def get_dispatch_benchmark_command(
+        self,
+        candidate_tracker: libtuner.CandidateTracker,
+    ) -> list[str]:
+        return []
+
+    def get_model_compile_timeout_s(self) -> int:
+        return 0
+
+    def get_model_compile_command(
+        self, candidate_tracker: libtuner.CandidateTracker
+    ) -> list[str]:
+        return []
+
+    def get_model_benchmark_timeout_s(self) -> int:
+        return 0
+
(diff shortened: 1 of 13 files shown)


Max191:
This PR is not really ready to land yet, since it requires an updated iree-compiler version. I'm mainly sending this out right now for visibility.

tuner/examples/test/tuner_test.py
```diff
@@ -0,0 +1,165 @@
...
+from tuner import libtuner
+from pathlib import Path, PurePath
+import argparse
+import os
+
+
+class TestTuner(libtuner.TuningClient):
+    def __init__(self):
+        self.compile_flags = [
+            "--iree-hip-target=gfx942",
```
kuhar, line 33: For dispatches, we don't need to provide the target as dispatches are already configured.

But we will need this for the full model,

tuner/examples/test/tuner_test.py
```diff
@@ -0,0 +1,165 @@
...
+            "--compile-from=executable-sources",
+        ]
+        self.benchmark_flags = [
+            "--benchmark_repetitions=3",
+            "--benchmark_format=json",
```
kuhar, line 38: Eventually we should use the bindings for benchmarking too. Just FYI, this is definitely outside of the scope of this PR.

tuner/pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [project]
-name = "SHARK Tuner"
+name = "SHARK-Tuner"
```
kuhar, line 2: Why?

tuner/tuner/candidate_gen.py
```diff
@@ -160,6 +212,48 @@ def apply_params(
...
         return MLIRTransformation(template, modified, embeddable)
 
+    def get_reduction_tile_sizes(self, configuration: Configuration):
+        tile_sizes = [0, 0]
+        tile_sizes.append(configuration.tile_sizes[2])
+        return tile_sizes
+
+    def get_workgroup_tile_sizes(self, configuration: Configuration):
+        tile_sizes = configuration.tile_sizes[:2]
+        tile_sizes.append(0)
```
kuhar, line 222: We will probably also have to refactor how we store tile sizes to support other pipelines

kuhar (commented):
Overall looks great, thanks for the detailed description and breakdown of the plan!

We should split it into smaller PRs and start landing 😸

tuner/examples/test/tuner_test.py
Max191: Yeah, that's actually what I did at first, but the compiler bindings did not seem to recognize many of the compiler codegen flags. I didn't dig too deep into it, but I think we may need to expose more flags to the compiler api in IREE before we can do this for compilation. And then since I was already using subprocess for compilation, I just did the same for benchmarking for consistency.

tuner/pyproject.toml
Max191: The space was causing my pip install of the tuner to fail for some reason, so I just made it a dash. I forgot to remove this change before pushing. I'm not sure why it fails to pip install on my machine, but if there is no preference for having a space then I think it's probably safer to use a dash.

tuner/tuner/candidate_gen.py
Max191: Yeah, there are a number of additional things that could probably use refactoring, but I left that out of the scope of this PR, since it is already very big. I think it makes sense to refactor this when we move the TD generation to use purely python bindings.

tuner/pyproject.toml
kuhar: Oh I've never tried installing it, I don't know if it actually works. I mostly copy-pasted from other project file in this repo

PR #579, opened 2024-11-21
[sharktank] Add Perplexity pre-submit test

archana-ramalingam:
Add a perplexity pre-submit test for llama3.1 8b fp16 with 5 prompts


sharktank/tests/evaluate/perplexity_iree_test.py
```diff
@@ -44,25 +50,26 @@ def test_llama3_8B_f16_decomposed(self):
...
                 f"--attention-kernel=decomposed",
+                f"--num-prompts={self.batch_size}",
             ]
         )
 
-        perplexity_difference = (
-            current_perplexity["mean_perplexity"]
-            - baseline_perplexity["mean_perplexity"]
+        baseline_mean_perplexity = round(
+            np.mean(baseline_perplexity["perplexities"][0 : self.batch_size]), 6
```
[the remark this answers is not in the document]
archana-ramalingam: Here I am rounding the `baseline_mean_perplexity` to 6 decimal places.

sharktank/tests/evaluate/perplexity_iree_test.py
```diff
@@ -7,18 +7,25 @@
 import unittest
 import pytest
 import json
+import numpy as np
 
 from sharktank.evaluate import perplexity_iree
 
-longrun = pytest.mark.skipif("not config.getoption('longrun')")
+is_mi300x = pytest.mark.skipif("config.getoption('iree_hip_target') != 'gfx942'")
+skipif_run_quick_llama_test = pytest.mark.skipif(
```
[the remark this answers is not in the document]
archana-ramalingam: To confirm, by default we want the shorter tests to run by default but longer tests only if a flag is passed?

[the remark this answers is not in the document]
IanNod: yes

.github/workflows/ci_eval_short.yaml
```diff
@@ -74,4 +74,4 @@ jobs:
             iree-base-runtime
 
       - name: Run perplexity test with vmfb
-        run:  pytest -n 8 -v -s sharktank/tests/evaluate/perplexity_iree_test.py --run-quick-llama-test --bs=5 --iree-device='hip://6' --iree-hip-target=gfx942 --iree-hal-target-backends=rocm --llama3-8b-f16-model-path=/data/llama3.1/8b/llama8b_f16.irpa --llama3-8b-tokenizer-path=/data/llama3.1/8b/tokenizer_config.json
```
IanNod: Is there a reason we are targeting only device hip://6? Should we have the runners available decide the device used?

archana-ramalingam: Ideally, but the specific `iree_device` needs to be passed to the `vmfbRunner`. We might require another script/flag to determine a free device and pass that info dynamically. Let me know if it's worth looking into.

IanNod: @saienduri have we setup ways to target specific devices on runners outside hardcoding like this?

IanNod (approved):

.github/workflows/ci_eval_short.yaml
saienduri: Yeah we do. We can specify ROCR_VISIBLE devices so that runners initialize with only certain gpus to avoid conflicts. I can make sure the seperation is in place to avoid conflicts on Monday.

Issue #453, opened 2024-11-08
[tuner] Reduce maintenance burden and prepare for more codegen pipelines


kuhar:
This is an uber-issue for making the tuner easier to maintain. The current implementation has a few issues that make the tuner library fragile and prone to getting out of sync with the IREE compiler. Specifically, we identified the following issues:

1. There are two ways to (re-)configure executable sources:
  a. By updating the lowering config and translation info in-situ. This is used when producing candidate dispatches using executable benchmarks as the source-of-truth.
  b. By using the transform dialect library script to match root ops and apply compilation info attributes to them. This is used during the model candidate compilation and benchmarking stage.
  
   As a result, we have duplicate logic to apply configurations found by the constraint solver. The fix is to write a pass that strips existing configuration from executable sources, and then use transform dialect to re-configure them. This can be done as a separate invocation of `iree-opt`.
  
2. The MLIR processing is mostly string-based. While this allowed us to quickly prototype, it makes the code prone to getting out of sync with the IREE compiler. The lowering configs and translation info attributes are considered compiler-internals and there's no stability guarantee as to the exact structure and format of these attributes. As a result, every time the format changes, we have to update the parsing and printing logic in the tuner to match the new format in the compiler.

   Here, the proposed solution is to expose these key attributes (translation info, compilation info, and MFMA intrinsic info) through python bindings. We already have it for the GPU pipeline options that can be used as a template for future bindings: https://github.com/iree-org/iree/pull/18840.
   
3. Make it easier to identify 'root ops'. We can make the IREE compiler annotate the root lingalg ops with a new attribute that the tuner can use to recognize them, without having to duplicate the compiler logic.
   
4. The `Configuration` representation is modeled after the requirements of the `LLVMGPUVectorDistribute` pipeline. This made it so that the surrounding code makes implicit assumptions about the problem representation. Instead, we should define an interface that allows us to support multiple compilation pipelines, such that the generated SMT constraints are specific to **both** the pipeline and the dispatch kind. Further, the constraint generation code should be decoupled from the parsing/printing code, such that projects like TKW can use just the constraint generation and benchmarking infra.

5. Move from two stages of compile-and-benchmark to just one. This made sense for SDXL where the best isolated dispatch does not necessarily perform best across the whole model, but it may not be necessary or even sufficiently general for other applications. This is related to the `libtuner.TuningClient` class; clients should be able to define their own tuning stages with libtuner providing the interface to specify the compilation and benchmarking commands.

[...]

PR #586, opened 2024-11-21
[tuner]: use python binding to select mma intrinsics

bangtianliu:
This PR is relevant to the task in https://github.com/nod-ai/shark-ai/issues/453: " Use IREE attributes for MFMA intrinsics in the tuner".

### DIFF
--- a/tuner/tuner/common.py
+++ b/tuner/tuner/common.py
@@ -12,6 +12,8 @@ from typing import Optional
 
 from iree.compiler import ir  # type: ignore
 
+from iree.compiler.dialects import iree_gpu
+
 
 class CommonTypes:
     def __init__(self, ctx: ir.Context):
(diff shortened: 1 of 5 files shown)


kuhar (changes requested):
The CI doesn't pass -- seems like the code doesn't type check

tuner/tuner/common.py
```diff
@@ -139,6 +146,10 @@ def is_compatible(intrinsic: MfmaIntrinsic) -> bool:
                 return False
             if problem_size.rhs_type.element_type != intrinsic.input_type:
                 return False
+
+        if available_mma_intrinsics and str(intrinsic) not in available_mma_intrinsics:
```
kuhar, line 150: I don't think we need to check that `available_mma_intrinsics` is non-empty?

tuner/tuner/common_test.py
```diff
@@ -109,7 +109,8 @@ def test_get_compatible_mfma_intrinsics(tuner_ctx: common.TunerContext) -> None:
             common.ShapedType([1280, 1280], tuner_ctx.type.f16),
             common.ShapedType([2048, 1280], tuner_ctx.type.f32),
             common.DispatchKind.mmt,
-        )
+        ),
+        [],
```
kuhar, line 113: We should make sure the new logic is covered by tests. Please add/modify existing tests to pass in the list of available intrinsics.

tuner/tuner/dispatch_constraints_test.py
```diff
@@ -37,7 +38,9 @@ def test_generate_solutions(tuner_ctx: common.TunerContext) -> None:
     problem_size = common.ProblemSize(
         matmul_size, lhs_type, rhs_type, res_type, common.DispatchKind.mmt
     )
-    configs = dispatch_constraints.generate_solutions(tuner_ctx.logger, problem_size, 4)
+    configs = dispatch_constraints.generate_solutions(
+        tuner_ctx.logger, problem_size, 4, []
```
kuhar, line 42: Doesn't this change the reason for no solutions? Now it can't produce any because there are no MFMAs to choose from.

kuhar (commented):
We should add a test that shows that filtering based on available intrinsics works in `get_compatible_mfma_intrinsics`

PR #587, opened 2024-11-21
(shortfin-sd) Interleave workers and their fibers by device.

monorimet:
This enables proper "filling" of multi-device topologies that populate each device evenly, rather than block allocating work to one device at a time.


nithinsubbiah:
Thanks for this patch. So with this can we run SDXL on CPX mode leveraging all the nodes? It would be helpful to add what this addresses in PR description

pdhirajkumarprasad:
verified and example is working but no

IanNod (approved):

Issue #584, opened 2024-11-21
Rework GitHub Actions workflows to build packages --> test packages


ScottTodd:
These workflows all currently build shortfin from source, duplicating all the boilerplate to fetch dependencies in some carefully balanced order:

- https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/ci-llama-large-tests.yaml
- https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/ci-llama-quick-tests.yaml
- https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/ci-sdxl.yaml
- https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/ci-sglang-benchmark.yml
- https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/ci-sglang-integration-tests.yml
- https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/ci-shark-ai.yml

For workflows that run on `pull_request` and `push` triggers, we can add a `build_dev_packages` job similar to https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/build_packages.yml that builds the packages and then have those workflows install artifacts from that job. For workflows that run on `schedule`, we can either do the same thing, or we can use the already built nightly packages (docs: https://github.com/nod-ai/shark-ai/blob/main/docs/nightly_releases.md).

In both cases, the complexity of package building will be isolated to a few package-oriented workflows and we'll gain confidence that the test jobs are compatible with our releases, so users will be able to use them without needing to build from source either.

Once we have something working, we can optimize the package build to improve CI turnaround times:

* cache pip dependencies
* cache CMake builds (or the entire build - see what IREE does)
* cache Dockerfiles
* skip the tracy build variant: https://github.com/nod-ai/shark-ai/blob/06599e9aa1eea6b96b2c9c8f1d21fe26b2c73a35/shortfin/build_tools/build_linux_package.sh#L94-L97 https://github.com/nod-ai/shark-ai/blob/06599e9aa1eea6b96b2c9c8f1d21fe26b2c73a35/shortfin/setup.py#L78 https://github.com/nod-ai/shark-ai/blob/06599e9aa1eea6b96b2c9c8f1d21fe26b2c73a35/shortfin/setup.py#L260-L263

See https://github.com/iree-org/iree/blob/main/.github/workflows/pkgci.yml for the shape of this sort of setup in IREE.

ScottTodd:
[...]

PR #589, opened 2024-11-22
Create pkgci.yml and pkgci_build_packages.yml.

ScottTodd:
Progress on https://github.com/nod-ai/shark-ai/issues/584.

## Summary

* Added `.github/workflows/pkgci_build_packages.yml` that builds `sharktank`, `shortfin`, and `shark-ai` dev packages and upload them to GitHub artifacts.
  * With ccache configured and tracing disabled, this job takes around 2 minutes to run.
* Added `.github/workflows/pkgci.yml` that just runs `pkgci_build_packages.yml` for now. Other jobs can be migrated to depend on that job and use/test the packages.

## Other details

* Experimented a bit with ccache. I'm seeing a ~30% cache hit rate but that still gets the shortfin build from 2m30s ([test logs here](https://github.com/ScottTodd/shark-ai/actions/runs/11964182118/job/33356053458#step:8:817)) down to 1m30s on standard GitHub-hosted `ubuntu-24.04` runners ([test logs here](https://github.com/ScottTodd/shark-ai/actions/runs/11964237192/job/33356209145#step:8:820)).
* Plumbed the `SHORTFIN_ENABLE_TRACING` setting through scripts / Docker. For dev packages we can keep tracing disabled (unless there is a clear reason to add it). If the cache hit rate improves then we might be able to enable tracing for low cost.
* Dropped `cache: "pip"` from `build_packages.yml` since it is counterproductive for a job that only installs `packaging`. Multiple workflows seem to be writing to the same cache and I see no way to customize the cache key. That, or the cache is unnecessarily large and we just need to prune it manually.


.github/workflows/pkgci_build_packages.yml
```diff
@@ -0,0 +1,71 @@
...
+      with:
+        python-version: 3.12
+
+    # Setup metadata.
+    - name: Install Python packages
+      run: pip install packaging
+    - name: Generate dev package versions
+      id: version_rc
+      run: |
+        version_suffix=".dev+${{ github.sha }}"
```
marbre, line 36:
```suggestion
        version_suffix=".dev0+${{ github.sha }}"
```

The version is normalized when building the packages and we might want to be explicit here:
```
shark_ai-3.1.0.dev0+20f1ecb69a24049675377cb4b41d53d599cda5dd-py3-none-any.whl
sharktank-3.1.0.dev0+20f1ecb69a24049675377cb4b41d53d599cda5dd-py3-none-any.whl
shortfin-3.1.0.dev0+20f1ecb69a24049675377cb4b41d53d599cda5dd-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
```

shortfin/build_tools/build_linux_package.sh
```diff
@@ -85,14 +116,44 @@ function run_in_docker() {
...
+
+    tar xf ccache.tar.xz
+    cp ccache-${CCACHE_VERSION}-linux-${ARCH}/ccache /usr/local/bin
+  elif [[ "${ARCH}" == "aarch64" ]]; then
+    # Latest version of ccache is not released for arm64, built it
+    git clone --depth 1 --branch "v${CCACHE_VERSION}" https://github.com/ccache/ccache.git
+    mkdir -p ccache/build && cd "$_"
+    cmake -G "Ninja" -DCMAKE_BUILD_TYPE=Release ..
+    ninja
+    cp ccache /usr/bin/
```
marbre, line 150: Do we need `aarch64` support? We probably can just install ccache for `x86_64`.

marbre (commented):
Nice! Just two minor things.

shortfin/build_tools/build_linux_package.sh
ScottTodd: We might at some point. I've written this code a few times before, most recently at https://github.com/iree-org/base-docker-images/blob/main/build_tools/install_ccache.sh, and supporting both architectures isn't much extra code. Better to write cross-platform/architecture code whenever possible instead of artificially limiting ourselves.

I plan on updating https://github.com/nod-ai/base-docker-images/blob/main/dockerfiles/manylinux_x86_64.Dockerfile to more closely match https://github.com/iree-org/base-docker-images/blob/main/dockerfiles/manylinux_x86_64.Dockerfile, as part of upgrading from manylinux2014 to manylinux_2_28: https://github.com/nod-ai/shark-ai/blob/779adc3d5add624cb80d5bcb579705dd42956f4b/shortfin/build_tools/build_linux_package.sh#L39-L43

ScottTodd:
Next steps with this:

* Refactor at least one workflow to use the package artifacts
* Put some more thought into how workflows can get different versions of the IREE / iree-turbine / etc. dependencies. Should they keep doing this? https://github.com/nod-ai/shark-ai/blob/e37b934a7eb3841f2a609cf94fa6b684838729ba/.github/workflows/ci-shark-ai.yml#L63-L72

ScottTodd:
Picking this back up now.

shortfin/build_tools/build_linux_package.sh
```diff
@@ -75,6 +89,23 @@ function run_in_docker() {
   echo "Using python versions: ${PYTHON_VERSIONS}"
   local orig_path="${PATH}"
 
+  # Configure caching.
```
renxida, line 92: YES!

renxida (approved):
Extremely excited for every bit of CI time reduction.

PR #592, opened 2024-11-22
Implement MMDIT block that is necessary for flux

KyleHerndon:


sharktank/sharktank/ops/default_impls.py
```diff
@@ -304,16 +304,26 @@ def interpolate_default(
...
 
-@layer_norm.override(Tensor, Tensor, Tensor)
 def layer_norm_default(input, weight, bias, *, eps):
     input = unbox_tensor(input)
-    weight = unbox_tensor(weight)
-    bias = unbox_tensor(bias)
+    if weight is not None:
+        weight = unbox_tensor(weight)
+    else:
+        weight = torch.ones(input.shape)
```
sogartar, line 312: Does this mean that we normalize over all values as one layer? I am curious if this is the natural default behavior. I guess the other extreme is each element on its own, but this does not seem useful.

sharktank/sharktank/layers/testing.py
```diff
@@ -43,3 +43,82 @@ def make_llama_attention_block_theta(
             ),
         }
     )
+
+
+def make_mmdit_block_theta(dtype: torch.dtype | None = None) -> Theta:
+    return Theta(
+        {
+            "img_attn.norm.key_norm.weight": DefaultPrimitiveTensor(  #
+                data=make_rand_torch((128,), dtype=dtype)
```
sogartar, line 52: I would move the numbers out into variables.
If not too difficult to construct them from the minimal amount of independent values. I would assume there is some dependence between all these values.

sharktank/sharktank/layers/mmdit_double.py
```diff
@@ -0,0 +1,133 @@
...
+
+        # calculate the img bloks
+        img = img + img_mod1.gate * self.img_attn_proj(img_attn)
+        img_mlp_in = (1 + img_mod2.scale) * ops.layer_norm(
+            img, None, None, eps=1e-6
+        ) + img_mod2.shift
+        img_mlp_out1 = self.img_mlp1(img_mlp_in)
+        img_mlp_out2 = ops.elementwise(F.gelu, img_mlp_out1)
+        img_mlp_out3 = self.img_mlp2(img_mlp_out2)
+        img = img + img_mod2.gate * img_mlp_out3
```
sogartar, line 121: In the block [picture](https://stability.ai/news/stable-diffusion-3-research-paper) I see a modulation layer here. Why can't we reuse the `ModulationLayer`?

sharktank/sharktank/layers/mmdit_double.py
```diff
@@ -0,0 +1,133 @@
...
+        txt_modulated = (1 + txt_mod1.scale) * txt_modulated + txt_mod1.shift
+        txt_qkv = self.txt_attn_qkv(txt_modulated)
+        txt_qkv_2 = txt_qkv.view(
+            txt_qkv.shape[0], txt_qkv.shape[1], 3, self.num_heads, -1
+        )  #
+        txt_qkv_3 = ops.permute(txt_qkv_2, (2, 0, 3, 1, 4))
+        txt_q, txt_k, txt_v = txt_qkv_3
+        txt_q, txt_k = qk_norm(
+            txt_q, txt_k, txt_v, self.txt_attn_norm_q, self.txt_attn_norm_k
+        )
```
sogartar, line 103: There seems to be potential for code reuse here. These seem like the same operations but one for the image and the other for the text.

The same hold after the attention where there is again repetition.

Since we don't have accuracy tests in this PR such refactoring could be left for a subsequent PR, where we can be more confident that not errors would occur.

sharktank/sharktank/layers/modulation.py
```diff
@@ -0,0 +1,32 @@
+import torch
```
sogartar, line 1: License.

sogartar (approved):
Looks good. I have a few questions. And a few suggestions that can be addressed now or later.

sharktank/sharktank/layers/testing.py
archana-ramalingam: Yeah, something like [this](https://github.com/nod-ai/shark-ai/blob/main/sharktank/sharktank/models/llama/testing.py). Also, should this+T5+CLIP dummy theta weights sit in `tests/layers/flux_testing.py` ?

sharktank/sharktank/ops/default_impls.py
sogartar: `torch.ones` would create a tensor with a default dtype. It is desirable to deduce a dtype either from the bias or from the incoming input.

sharktank/sharktank/layers/mmdit_double.py
```diff
@@ -0,0 +1,133 @@
...
+    x = ops.scaled_dot_product_attention(
+        q=q, k=k, v=v, a=None, is_causal=True, scale=None
+    )
+    x = ops.permute(x, (0, 2, 1, 3))
+    x = x.view(x.shape[0], x.shape[1], -1)
+
+    return x
+
+
+class MMDITDoubleBlock(ThetaLayer):
```
sogartar, line 38: Could add a reference to the source implementation if you based it on an existing one?

sharktank/sharktank/layers/mmdit_double.py
```diff
@@ -0,0 +1,133 @@
+import torch.nn.functional as F
```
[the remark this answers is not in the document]
KyleHerndon: Added. Done.

sharktank/sharktank/layers/mmdit_double.py
```diff
@@ -0,0 +1,133 @@
...
+from .modulation import ModulationLayer
+from .norm import RMSNormLayer
+from .paged_llama_attention_block import PagedLlamaAttentionBlock
+
+
+def qk_norm(q, k, v, rms_q, rms_k):
+    return rms_q(q).to(v), rms_k(k).to(v)
+
+
+def apply_rope(xq: Tensor, xk: Tensor, freqs_cis: Tensor) -> tuple[Tensor, Tensor]:
```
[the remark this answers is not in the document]
KyleHerndon: Added a TODO, I am pretty sure it can be, but it was sufficiently differently defined in the reference code that it is pretty non-trivial to do so.

sharktank/tests/layers/mmdit_test.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class MMDITTest(unittest.TestCase):
+    def setUp(self):
+        torch.manual_seed(12345)
+        self.hidden_size = 3072
+        self.num_heads = 24
+        self.batch_size = 3
+
+    def testExport(self):
```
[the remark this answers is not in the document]
KyleHerndon: Yeah, I have plans for numerical checks upcoming. I didn't need to adapt very significantly. Mostly replacing torch ops with our sharktank ops. I was able to follow the reference quite closely.

sharktank/sharktank/ops/default_impls.py
KyleHerndon: Nice catch on dtype. I don't really understand why this is used, but it is used in the reference implementation.

sharktank/sharktank/layers/testing.py
KyleHerndon: Done with variable refactor. I agree the new file location is a better idea, but it isn't directly related to the patch, so I'll send a follow up that changes this or approve one if it comes to me.

sharktank/sharktank/layers/mmdit_double.py
KyleHerndon: I added a refactoring TODO. When we have a numerics test in place, I'll refactor this, but I was following the code closely so I didn't want to make large changes. I think the only reason it isn't using the Modulation layer code is because they decided to use a different activation function.

sharktank/sharktank/layers/mmdit_double.py
KyleHerndon: Added a TODO for after numerics test is in.

sharktank/sharktank/layers/mmdit_double.py
KyleHerndon: Added. Done.

sharktank/sharktank/layers/modulation.py
KyleHerndon: Added. Done.

PR #595, opened 2024-11-22
[shortfin] Pin pydantic-core to 2.27.0

marbre:
Pins pydantic-core to 2.27.0, as building 2.27.1 fails to build for Python 3.13t.

### DIFF
--- a/shortfin/requirements-tests-nogil.txt
+++ b/shortfin/requirements-tests-nogil.txt
@@ -1,4 +1,5 @@
 pytest
 requests
 fastapi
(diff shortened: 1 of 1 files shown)


marbre:
Actually, pydantic-core 2.27.1 cannot be build with free-threaded Python as PyO3 was bumped to 0.22.6 to prevent accidental installs on free-threaded Python, see https://github.com/pydantic/pydantic-core/pull/1550.

Therefore, we should re-evaluate the deps in the requirements file: pydantic-core is required by pydantic which is required by fastapi.

PR #602, opened 2024-11-25
Add Baseline for SGLang Benchmark Test

stbaione:
# Description

The SGLang Benchmark Test has been running for awhile, but only benchmarks the shortfin server itself. In order to get a baseline metric and enable long-term convergence in-terms of performance, we need to be able to track metrics of the SGLang server using the same benchmark method.

This adds an `sglang_benchmark_test` to complement the `shortfin_benchmark_test`. Also restructures `app_tests/benchmark_tests/llm` -> `app_tests/benchmark_tests/llm/sglang_benchmarks`. This keeps the benchmark tests organized and allows for the folder to be extended with other types of benchmarks in the future.

# Why are we using docker to start the SGLang server?

Currently, the pyprompt.toml file inside of SGLang requires `vllm==0.6.3.dev13` to run on ROCm. I looked into potentially building vLLM from source for this test, but couldn't find a branch, tag, or release that matched that signature. From their own comments inside of `pyproject.toml`, it appears to only be available inside of a `ROCm` base image:

```toml
# HIP (Heterogeneous-computing Interface for Portability) for AMD
# => base docker rocm/vllm-dev:20241022, not from public vllm whl
srt_hip = ["sglang[runtime_common]", "torch", "vllm==0.6.3.dev13"]
```

Their [instructions](https://sgl-project.github.io/start/install.html#method-3-using-docker) on installing SGLang and running for ROCm also appear to suggest the docker method:

## Instructions from their docs for running with ROCm

```
docker build --build-arg SGL_BRANCH=v0.3.5.post2 -t v0.3.5.post2-rocm620 -f Dockerfile.rocm .

alias drun='docker run -it --rm --network=host --device=/dev/kfd --device=/dev/dri --ipc=host \
    --shm-size 16G --group-add video --cap-add=SYS_PTRACE --security-opt seccomp=unconfined \
    -v $HOME/dockerx:/dockerx -v /data:/data'

drun -p 30000:30000 \
    -v ~/.cache/huggingface:/root/.cache/huggingface \
    --env "HF_TOKEN=<secret>" \
    v0.3.5.post2-rocm620 \
    python3 -m sglang.launch_server --model-path meta-llama/Llama-3.1-8B-Instruct --host 0.0.0.0 --port 30000
```

The workflow file handles starting the container and cleaning up once the workflow is done. I set the timeout for waiting for the server to start to `10 minutes` to give the SGLang server enough time to load necessary model weights and startup.


renxida:
Link to successful run:

With SRT (Sglang RunTime):
https://github.com/nod-ai/shark-ai/actions/runs/12126822859/job/33812266791

With Shortfin:
https://github.com/nod-ai/shark-ai/actions/runs/12126822859/job/33812266454

.github/workflows/ci-sglang-benchmark.yml
```diff
@@ -69,21 +69,135 @@ jobs:
...
+    runs-on: mi300x-4
+    defaults:
+      run:
+        shell: bash
+    env:
+      PIP_CACHE_DIR: "${{ github.workspace }}/.pip-cache"
+    steps:
+      - name: Get Current Date
+        id: date
+        run: echo "::set-output name=date::$(date +'%Y-%m-%d')"
```
ScottTodd, line 108: Where is this used? Can you add a comment? (This looks like leftover debugging code)

.github/workflows/ci-sglang-benchmark.yml
```diff
@@ -69,21 +69,135 @@ jobs:
...
+
+      - name: "Checkout Code"
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Cache Pip Packages
+        uses: actions/cache@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
+        id: cache-pip
+        with:
+          path: ${{ env.PIP_CACHE_DIR }}
+          key: pip-${{ steps.setup_python.outputs.python-version }}-${{ hashFiles('*requirements*.txt','shortfin/requirements*.txt','sharktank/requirements*.txt') }}
```
ScottTodd, line 124: Does this really need `steps.setup_python.outputs.python-version` instead of `matrix.version`? I'm expecting some of the workflow refactoring to reduce the reliance on this `Cache Pip Packages` step anyways though...

ScottTodd (changes requested):

.github/workflows/ci-sglang-benchmark.yml
```diff
@@ -69,21 +69,135 @@ jobs:
...
+        uses: peaceiris/actions-gh-pages@4f9cc6602d3f66b9c108549d475ec49e8ef4d45e # v4.0.0
+        with:
+          github_token: ${{ secrets.SHARK_PLATFORM_GH_TOKEN }}
+          publish_dir: ./out/llm/shortfin
+          destination_dir: ./llm/sgl_benchmark/shortfin
+          keep_files: true
+
+  benchmark_sglang:
+    if: ${{ github.repository_owner == 'nod-ai' || github.event_name != 'schedule' }}
+    name: "SGLang Serving Benchmark With SGLang"
```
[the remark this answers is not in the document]
stbaione: We could theoretically move the SGLang benchmarking part to https://github.com/nod-ai/sglang.

The downside of that would be that it would fragment the benchmark results across repos. Would need to check actions across different repos to be able to update the benchmark numbers daily. Would also make it so that the SGLang benchmark results wouldn't be included in our pytest dashboard. We have plans to upload the stats to Grafana, which we can start working on once this half of the benchmark is in. Adds some complexity maintaining across different repos, but shouldn't be a crazy amount of complexity. I'm not sure what we would do if we move our `Shortfin` backend from our forked repo to the origin SGLang repo. I assume in that case, we wouldn't be able to carry our own benchmarking & upload process.

We could also make another repo for it. I would need some help getting that repo created and runner access set up for it, since I don't have that level of permission. That would have some of the fragmentation issues as well, not being included in the shark-ai pytest dashboard. Would be easier to keep the `grafana` stuff organized though. Maybe we could just add a link to that dashboard in our top-level `index.html` file in this repo, which would mitigate the fragmentation issue for both approaches in regards to pytest dashboard.

Ultimately, I was thinking that having the code here follows the "pattern" of other benchmark tests more closely and would keep things centralized. SGLang code doesn't rely on `shortfin`, but the benchmark itself is a target for shortfin. I `think` that having the benchmarks as a nightly run and using mi300x-4 for only the sglang integration and benchmark tests would mitigate strain on infrastructure.

Which option do you like the best, given above statements?

.github/workflows/ci-sglang-benchmark.yml
```diff
@@ -69,21 +69,135 @@ jobs:
...
+            -v ~/.cache/huggingface:/root/.cache/huggingface \
+            --env HF_TOKEN=${{ secrets.HF_TOKEN }} \
+            lmsysorg/sglang:v0.3.5.post1-rocm620 \
+            python3 -m sglang.launch_server \
+            --model-path meta-llama/Llama-3.1-8B-Instruct \
+            --host 0.0.0.0 \
+            --port 30000 \
+            --tp 1 \
+            --dtype float16 \
+            --disable-cuda-graph
```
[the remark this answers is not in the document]
stbaione: I was receiving OOM errors when attempting to run with `cuda-graph` enabled on the `mi300x-4` runner. 

I found [this comment](https://github.com/sgl-project/sglang/pull/1420#issuecomment-2423931906), which didn't seem to have a follow up. Trying it out locally caused the test to run successfully, so I ran with it, but admittedly, I didn't dig too far into an alternative for fixing the issue with cuda-graph enabled

.github/workflows/ci-sglang-benchmark.yml
stbaione: I saw it used in the other llama benchmarking tests `ci-llama-large-tests` and `ci-llama-quick-tests`. Used that as a template to setup this workflow and carried it over to be consistent. Will remove it though, I don't think it really adds anything.

.github/workflows/ci-sglang-benchmark.yml
```diff
@@ -69,21 +69,135 @@ jobs:
...
+
+      - name: Run SGLang Benchmark Tests
+        run: |
+          pytest -v app_tests/benchmark_tests/llm/sglang_benchmarks/sglang_benchmark_test.py --port 30000 --log-cli-level=INFO --html=out/llm/sglang/index.html
+
+      - name: Stop sglang-server
+        run: docker stop sglang-server || true # Stop container if it's running
+
+      - name: Cleanup SGLang Image
+        run: docker image rm lmsysorg/sglang:v0.3.5.post1-rocm620
```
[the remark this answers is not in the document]
stbaione: The image is `83.1GB`

.github/workflows/ci-sglang-benchmark.yml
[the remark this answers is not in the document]
ScottTodd: We don't need to tightly couple

* the report data generation (project code, test code, workflow files)
* the report data itself (structured json output or standalone .html report files)
* the dashboard presentation layer (currently gh-pages)

I consider the current dashboard existing at https://nod-ai.github.io/shark-ai/ to be a v0 prototype and I don't want to bend the repository structure by treating it as a final implementation. We could be reaching the point here where we need a more complete design (including a review process).

Some random notes from looking there:

* We could also use something like https://pypi.org/project/pytest-html-report-merger/ or https://pypi.org/project/pytest-html-merger/ to group test reports on that dashboard
* It looks like the current process overwrites previous results, e.g. https://github.com/nod-ai/shark-ai/commit/f7d53452dbd57debdcbdaa125da221dd3a4ed6f4 . Historical result tracking would be useful too: a dropdown to view over results, metrics like number of tests passing / xfailed over time, etc.

Maybe we can keep this complexity local to https://github.com/nod-ai/shark-ai/tree/main/app_tests/ and a few appropriately named github workflows (as you have it here, but with a clearer definition of the "core project" and "external integration tests"). That seems to be a reasonable starting point / compromise to me, as long as we draw a hard line there and stick to it.

.github/workflows/ci-sglang-benchmark.yml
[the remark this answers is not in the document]
ScottTodd: 💀 omg that's huge. Eeeeeeh. Might be worth the explicit cleanup step then, with a comment on why.

.github/workflows/ci-sglang-benchmark.yml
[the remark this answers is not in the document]
stbaione: Yeah, this makes a ton of sense and clears up the flow for me. What I'm thinking is:

## Reaching the boundaries of using the pytest dashboard
I think `Grafana` will be a good solution for this. We can get it to a point where the dashboard (if it sticks around) is more just to quickly see the status of a test, where Grafana would be used for actually tracking metrics. When I work on that (as next task), I can write it in a general way to make it easy for other benchmark tests to transition.

## Merging pytest reports
This is useful, I'll test locally and see which one fits the bill

## Isolating complexity
Yeah, I agree, what you described is a good starting point, but not the end-all-be-all. 

It makes sense to have a dedicated benchmarking repo as a more over-arching project, especially since we've been expanding benchmarks a good bit lately, and #609 transitions them to using packages, so they don't all NEED to run in `shark-ai` itself. I'll add a description at the top of this CI file adding more detail and explaining that this is an `external test`, which should not be copied in most cases or any case outside of `app_tests`. We should continue discussions on a dedicated repo though. It hasn't been formally discussed, but it would be a logical continuation to add benchmarking against more external projects in the future, like [TensorRT](https://developer.nvidia.com/tensorrt), which was mentioned in lead-sync today as faster than SGLang

.github/workflows/ci-sglang-benchmark.yml
```diff
@@ -69,21 +69,135 @@ jobs:
...
+            -e "git+https://github.com/iree-org/iree-turbine.git#egg=iree-turbine"
+          pip install --no-compile -r requirements.txt -e sharktank/ shortfin/
+
+          # Try with the latest nightly releases, not what iree-turbine pins.
+          # We could also pin to a known working or stable version.
+          # This should eventually stabilize. Do the best we can for now.
+          pip install -f https://iree.dev/pip-release-links.html --upgrade --pre \
+            iree-base-compiler \
+            iree-base-runtime \
+            "numpy<2.0"
```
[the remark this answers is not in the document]
stbaione: Reduced the dependencies in this block to only include `sharktank`. I use the `hf_datasets` script for downloading external model weights and ensuring both benchmark runs are using the same artifacts

.github/workflows/ci-sglang-benchmark.yml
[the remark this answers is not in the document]
stbaione: Yeah, it's pretty wack. The bulk comes from the base rocm image they use and there don't appear to be any extra steps taken to reduce layers/overall image size: [Dockerfile.rocm](https://github.com/sgl-project/sglang/blob/main/docker/Dockerfile.rocm).

Added back the step to remove the image, and included a comment on why it's being removed

.github/workflows/ci-sglang-benchmark.yml
stbaione: Changed to use `matrix.version`, good catch

.github/workflows/ci-sglang-benchmark.yml
stbaione: Has been removed in both jobs

ScottTodd: Ah, `date` was used in other jobs like here: https://github.com/nod-ai/shark-ai/blob/8cd3f850f0a7864d9e004483fcf4bbea08d3e21a/.github/workflows/ci-llama-large-tests.yaml#L89-L93

.github/workflows/ci-sglang-benchmark.yml
```diff
@@ -69,21 +69,135 @@ jobs:
...
+      # Instruction for SGLang image sourced from here:
+      #   https://sgl-project.github.io/start/install.html#method-3-using-docker
+      # We have to run in a docker container due to their vLLM dependency.
+      # From their pyproject.toml:
+      #   HIP (Heterogeneous-computing Interface for Portability) for AMD
+      #   => base docker rocm/vllm-dev:20241022, not from public vllm whl
+      #   srt_hip = ["sglang[runtime_common]", "torch", "vllm==0.6.3.dev13"]
+      - name: Pull SGLang Image (Had issues with sglang:v0.3.5.post2-rocm620)
+        run: |
+          docker pull lmsysorg/sglang:v0.3.5.post1-rocm620
```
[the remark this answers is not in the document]
stbaione: Resolving based on our conversations [here](https://github.com/nod-ai/shark-ai/pull/602#discussion_r1866670807), but agree that this would be useful transition towards having this in a different repo altogether.

That is kind of a nice selling point though, haha, can run shortfin on AMD GPUs with standard vLLM, and don't need to install an 83GB base image to make it work xD

.github/workflows/ci-sglang-benchmark.yml
stbaione: Oooooohhh that makes much more sense hahaha

.github/workflows/ci-sglang-benchmark.yml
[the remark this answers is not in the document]
stbaione: Added a `README` description to top of CI file, explaining the details of why we're including it here, for now. Leaving conversation unresolved, in case you have further suggestions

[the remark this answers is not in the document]
stbaione: Added steps to merge reports and upload as one artifact to pytest dashboard

.github/workflows/ci-sglang-benchmark.yml
```diff
@@ -186,18 +176,65 @@ jobs:
...
+      - name: Upload pytest report
+        uses: actions/upload-artifact@b4b15b8c7c6ac21ea08fcf65892d2ee8f75cf882
+        with:
+          name: sglang_benchmark
+          path: sglang_index.html
+
+  merge_and_upload_reports:
+    if: ${{ github.repository_owner == 'nod-ai' || github.event_name != 'schedule' }}
+    name: "Merge and upload benchmark reports"
+    needs: [benchmark_shortfin, benchmark_sglang]
```
ScottTodd, line 197: Is the duplicated `if` condition needed on this step here? I would expect the conditions in the `needs` steps to make this be skipped

https://stackoverflow.com/questions/65844033/conditional-needs-in-github-actions

One wrinkle to consider if how you want to handle one of the jobs succeeding and the other failing. Should that upload the results for the job that succeeded, or should it stop the combine step and the uploading?

.github/workflows/ci-sglang-benchmark.yml
```diff
@@ -186,18 +176,65 @@ jobs:
...
+    strategy:
+      matrix:
+        version: [3.11]
+      fail-fast: false
```
ScottTodd, line 201: `merge_and_upload_reports` shouldn't vary across python versions or need to run in a matrix. I'd simplify this step to just always use Python 3.11

.github/workflows/ci-sglang-benchmark.yml
```diff
@@ -186,18 +176,65 @@ jobs:
...
+        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16
+        with:
+          name: sglang_benchmark
+          path: reports
+
+      - name: Create merged report directory
+        run: mkdir merged_reports
+
+      - name: Merge html reports
+        run: pytest_html_merger -i reports -o merged_reports/index.html
```
ScottTodd, line 232: nit: I'd personally merge these two steps. I don't see much value in having a step for just `mkdir`

```suggestion
      - name: Merge html reports
        run: |
          mkdir merged_reports
          pytest_html_merger -i reports -o merged_reports/index.html
```

.github/workflows/ci-sglang-benchmark.yml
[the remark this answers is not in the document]
ScottTodd: Which script is that? Can the job use `huggingface-cli` or a script that only depends on the `huggingface` package instead of the `sharktank` package?

https://huggingface.co/docs/huggingface_hub/en/guides/cli

.github/workflows/ci-sglang-benchmark.yml
stbaione: Updated

.github/workflows/ci-sglang-benchmark.yml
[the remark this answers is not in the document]
stbaione: Removed `sharktank` dep and download tokenizer with `huggingface-cli`

.github/workflows/ci-sglang-benchmark.yml
stbaione: Duplicated `ifs` have been removed. 

That's a good point, really I just didn't want them to run at the same time. They're pretty resource heavy and didn't want that to cause the benchmark numbers to be off. However, if one fails, the other should still run. Updated test to handle that

.github/workflows/ci-sglang-benchmark.yml
stbaione: Hardcoded to `3.11`

.github/workflows/ci-sglang-benchmark.yml
ScottTodd: I would only use the setup you have if the workflows logically depend on one another:

```yml
  benchmark_sglang:
    name: "SGLang Serving Benchmark With SGLang"
    needs: benchmark_shortfin
    if: always()
```

Resource isolation and benchmark stability is the responsibility of the runner control plane and runner machines/images/VMs, not a single workflow. You have no control here over what other jobs might be running on the same physical machine. All your workflow should specify is what runner label it wants. There could be multiple physical machines with that label, or multiple runners on the same physical machine. Either way, the runners shouldn't be competing for resources.

stbaione: Gotcha, yeah that makes sense. I removed the dependency on `benchmark_shortfin` in `benchmark_sglang`. I then added an `if` statement for `merge_and_upload_reports` that waits for both jobs to finish, but still runs if either job completed successfully

ScottTodd (approved):

PR #610, opened 2024-11-26
[shortfin] Add C++ tokenizer wrapper library.

stellaraccident:
* This is gated by SHORTFIN_ENABLE_TOKENIZERS (presently off).
* I'd like to either take over the wrapper or get https://github.com/mlc-ai/tokenizers-cpp/issues/50 before putting much weight on this.
* There is no great C++ option for this component, so we go to the trouble of integrating a Rust component. We will need to do a bit of prep on our CI systems to enable this by default.
* Python API will be added in a subsequent commit. This should be more efficient than the tokenizers Python API since we will allow direct access to the tokens vs doing a lot of conversions.
* Size analysis: Prior to this patch, libshortfin was 1.8MB, which gave us an entire GPU and CPU runtime stack. After this patch (stripped) it is 8.4MB. Given how important the use case is, I'm willing to tolerate this for the moment. It seems like there is room for something better here, which is why I did not expose the underlying vendor'd API directly (edit: by switching to a nightly rust and activating a bunch of options from https://github.com/johnthagen/min-sized-rust, I was able to produce a binary that is 4.2MB, which is more reasonable).

### DIFF
--- a/shortfin/CMakeLists.txt
+++ b/shortfin/CMakeLists.txt
@@ -48,6 +48,7 @@ option(SHORTFIN_BUILD_TESTS "Builds C++ tests" ON)
 option(SHORTFIN_BUNDLE_DEPS "Download dependencies instead of using system libraries" ON)
 option(SHORTFIN_ENABLE_TRACING "Enable runtime tracing for iree and shortfin" OFF)
 option(SHORTFIN_ENABLE_LTO "Enables LTO if supported" ON)
+option(SHORTFIN_ENABLE_TOKENIZERS "Enables integration of native tokenizers library" OFF)
 
 set(SHORTFIN_IREE_SOURCE_DIR "" CACHE FILEPATH "Path to IREE source")
 
@@ -80,6 +81,7 @@ list(APPEND CMAKE_MODULE_PATH
   ${CMAKE_CURRENT_LIST_DIR}/build_tools/cmake/
 )
 include(shortfin_library)
+include(shortfin_testing)
 include(CheckCXXCompilerFlag)
 include(FetchContent)
 
@@ -90,7 +92,9 @@ include(FetchContent)
 if(SHORTFIN_ENABLE_LTO)
   include(CheckIPOSupported)
   check_ipo_supported(RESULT SHORTFIN_LTO_SUPPORTED OUTPUT SHORTFIN_LTO_ERROR)
-  if(SHORTFIN_LTO_SUPPORTED)
+  if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+    message(STATUS "Not enabling LTO for debug build")
+  elseif(SHORTFIN_LTO_SUPPORTED)
     message(STATUS "Shortfin LTO Enabled")
     set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
   else()
@@ -126,7 +130,9 @@ endif()
 message(STATUS "  - Host")
 
 ################################################################################
-# Dependencies
+# Bundled Dependencies
+# These dependencies are either bundled or used via installed packages based
+# on the SHORTFIN_BUNDLE_DEPS option.
 ################################################################################
 
 if(SHORTFIN_BUNDLE_DEPS)
@@ -164,6 +170,7 @@ if(SHORTFIN_BUNDLE_DEPS)
   shortfin_push_bundled_lib_options()
   # Enable spdlog shared library options so we can export it.
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DSPDLOG_SHARED_LIB -Dspdlog_EXPORTS")
+  message(STATUS "Fetching bundled projects")
   FetchContent_MakeAvailable(fmt spdlog xtl xtensor)
   shortfin_pop_bundled_lib_options()
 else()
@@ -172,7 +179,8 @@ else()
 endif()
 
 ################################################################################
-# IREE
+# IREE Dependency
+# This is always a source dependency on the IREE runtime.
 ################################################################################
 
 # Set IREE build flags.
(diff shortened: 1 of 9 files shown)


shortfin/CMakeLists.txt
```diff
@@ -237,6 +245,61 @@ else()
...
+function(shortfin_check_tokenizers)
+  # Make sure that rust/cargo is installed and usable.
+  find_program(SHORTFIN_CARGO_PATH NAMES cargo NO_CACHE)
+  if(NOT SHORTFIN_CARGO_PATH)
+    message(SEND_ERROR
+      "Building with -DSHORTFIN_ENABLE_TOKENIZERS=ON requires cargo (Rust's build tool). "
+      "Please follow Rust documentation to install. On Ubuntu, this can typically be accomplished with:\n"
+      "  sudo apt install rustup && rustup default stable"
+    )
+  endif()
```
ScottTodd, line 261: These docs? https://www.rust-lang.org/tools/install

Can you include that link in the message?

shortfin/CMakeLists.txt
```diff
@@ -256,7 +319,7 @@ if(SHORTFIN_BUILD_TESTS)
   enable_testing()
 endif()
 
-
+add_custom_target(shortfin_testdata_deps)
```
ScottTodd, line 322: Should this be moved under the `if(SHORTFIN_BUILD_TESTS)` above?

shortfin/build_tools/cmake/shortfin_testing.cmake
```diff
@@ -0,0 +1,47 @@
...
+    else()
+      message(SEND_ERROR "Error downloading file ${_RULE_URL} -> ${_RULE_OUTPUT_FILE}")
+    endif()
+  endif()
+
+  # Make clean remove it.
+  set_property(
+    TARGET shortfin_testdata_deps
+    APPEND PROPERTY ADDITIONAL_CLEAN_FILES
+      "${CMAKE_CURRENT_BINARY_DIR}/tokenizer.json"
```
ScottTodd, line 39: File name `tokenizer.json` is hardcoded here? Should this be computed from the arguments?

shortfin/src/shortfin/components/tokenizers/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+# Copyright 2024 Advanced Micro Devices, Inc.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+if(SHORTFIN_ENABLE_TOKENIZERS)
```
ScottTodd, line 7: nit: I'd invert this check to save on indentation

```cmake
if(NOT SHORTFIN_ENABLE_TOKENIZERS)
  return()
endif()

shortfin_cc_component(
...

File: shortfin/src/shortfin/components/tokenizers/CMakeLists.txt [hunk 5]
```diff
@@ -0,0 +1,39 @@
...
+    set_property(GLOBAL APPEND
+      PROPERTY SHORTFIN_LIB_OPTIONAL_COMPONENTS
+      shortfin_tokenizers)
+    target_compile_definitions(shortfin_public_defs INTERFACE SHORTFIN_HAVE_TOKENIZERS)
+
+    # Download test data.
+    shortfin_download_test_data(
+      URL "https://huggingface.co/google-bert/bert-base-cased/resolve/main/tokenizer.json"
+      OUTPUT_FILE "${CMAKE_CURRENT_BINARY_DIR}/tokenizer.json"
+    )
```
ScottTodd, line 31: I assume this file doesn't need authentication to download?

Looks like the file is 426KB. Might be small enough to check in instead of fetch from the network...

also nit: could pin to a specific commit hash to stabilize the supply chain a bit:
```suggestion
    # Download test data.
    shortfin_download_test_data(
      URL "https://huggingface.co/google-bert/bert-base-cased/resolve/cd5ef92a9fb2f889e972770a36d4ed042daf221e/tokenizer.json"
      OUTPUT_FILE "${CMAKE_CURRENT_BINARY_DIR}/tokenizer.json"
    )
```

marbre (commented):
I think Scott has already pointed out everything I might have said. No additional comments.

shortfin/CMakeLists.txt
stellaraccident: No idea - I didn't read any docs. Just followed the breadcrumbs of things that the tools said when a step hadn't been done.

shortfin/CMakeLists.txt
stellaraccident: No - various things expect the target to exist even if we don't actually enable testing.

shortfin/CMakeLists.txt
```diff
@@ -90,7 +92,9 @@ include(FetchContent)
 if(SHORTFIN_ENABLE_LTO)
   include(CheckIPOSupported)
   check_ipo_supported(RESULT SHORTFIN_LTO_SUPPORTED OUTPUT SHORTFIN_LTO_ERROR)
-  if(SHORTFIN_LTO_SUPPORTED)
+  if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+    message(STATUS "Not enabling LTO for debug build")
```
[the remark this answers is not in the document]
stellaraccident: I think I'll pass on this. It is a minor ergonomic thing and I don't want to further pollute the top level namespace.

shortfin/CMakeLists.txt
```diff
@@ -237,6 +245,61 @@ else()
 endif()
 shortfin_pop_bundled_lib_options()
 
+################################################################################
+# Tokenizer Library
+################################################################################
+
+function(shortfin_check_tokenizers)
+  # Make sure that rust/cargo is installed and usable.
+  find_program(SHORTFIN_CARGO_PATH NAMES cargo NO_CACHE)
```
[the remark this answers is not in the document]
stellaraccident: Commented:

```
  # Consider switching this to a cached variable once the tokenizers_cpp project
  # will accept an override vs running whatever is on the path. For now, just
  # verify the path is sane as that is what will get used.
```

shortfin/CMakeLists.txt
stellaraccident: Fixed.

shortfin/build_tools/cmake/shortfin_testing.cmake
stellaraccident: Oops. Copy pasta from before making it into its own function. Thanks.

shortfin/src/shortfin/components/tokenizers/CMakeLists.txt
stellaraccident: Yeah, it's a purely public one that others seem to use a lot. But I couldn't find anyone else who wasn't downloading it... decided to go with the crowd.

stellaraccident: Pinned it as suggested.

stellaraccident:
PTAL

ScottTodd (approved):

PR #613, opened 2024-11-26
Fix linear

dan-garvey:
Fixes punet regression from linear.py
I went a little crazy with the fake-quant arg and the logic was
ugly. Now the fake_quant arg only affects q_input. The purpose in having
it at all is to allow exporting and fake eager using the same irpa file


nithinsubbiah:
Looks like sharktank integration test is still failing

.github/workflows/ci-sharktank.yml
```diff
@@ -22,6 +22,31 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  test_punet:
```
nithinsubbiah, line 25: We should also change model version to this (https://github.com/nod-ai/shark-ai/pull/569/files) to reflect the production model

dan-garvey: ah my bad good catch

nithinsubbiah (approved):
Thanks for fixing the export. I don't know if the llama data dependent test failure is expected, contingent on that good to merge.

dan-garvey:
I'm not sure this is okay to merge as is, it takes over an hour to run the punet

PR #614, opened 2024-11-26
Add T5 encoder bfloat16 support

sogartar:
Adds various tests for verifying bfloat16 execution.

The bfloat16 eager execution matches Flux's output. One test verifies this against golden values from the Flux pipeline.

I would say that the eager numerical error is still quite high. My initial idea was to compare the numerical error in IREE bfloat16 vs eager float32. It should match the error profile as the one we get between eager bfloat16 compared against eager float32. The problem is that the bfloat16 eager has a high error that needs further investigation.
Due to this some tests are marked as xfail as we don't have a good metric to evaluate the IREE results.

### DIFF
--- a/sharktank/conftest.py
+++ b/sharktank/conftest.py
@@ -153,16 +153,28 @@ def pytest_addoption(parser):
     #     --outtype=f32 \
     #     t5-v1_1-small
     parser.addoption(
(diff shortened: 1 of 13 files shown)


sogartar:
The compilation errors is a known [issue](https://github.com/iree-org/iree/issues/19230).

PR #617, opened 2024-11-27
Update tooling to export toy LLaMa model

rsuderman:
We can export a toy version of the llama model with deterministic coefficients and structure. This allows simple compilation testing from sharktank through IREE.


stellaraccident (approved):
Thanks - figured it should be easy like this.

PR #619, opened 2024-11-27
[shortfin] Enable building Rust deps on Linux

marbre:
Sets up Rust to build shortfin with `SHORTFIN_ENABLE_TOKENIZERS` set to
`ON` on Linux. At the moment, building with Rust is limited to Linux as
there are build failures on Windows (see issue #620).

### DIFF
--- a/.github/workflows/ci-libshortfin.yml
+++ b/.github/workflows/ci-libshortfin.yml
@@ -50,7 +50,7 @@ jobs:
           - name: Ubuntu (Clang)(full)
             runs-on: ubuntu-24.04
             cmake-options:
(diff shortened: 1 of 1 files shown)


stellaraccident (approved):

ScottTodd (approved):
Some CMake warnings: https://github.com/nod-ai/shark-ai/actions/runs/12051730604/job/33603422869?pr=619#step:10:227

marbre:
> Some CMake warnings: https://github.com/nod-ai/shark-ai/actions/runs/12051730604/job/33603422869?pr=619#step:10:227

`msgpack` and `sentencepiece` (used in `tokenizers-cpp`) both set the CMake minimum to 3.1, which is something that probably would need to be adjusted upstream. There are some more warning which are worth a look but it seems to me that these are also issues that must be fixed upstream rather than downstream.

PR #622, opened 2024-11-27
Add `--durations` to some `pytest` commands in CI workflows.

ScottTodd:
This will help triage slow tests.

Docs:

* https://docs.pytest.org/en/stable/example/simple.html#profiling-test-duration
* https://docs.pytest.org/en/stable/how-to/usage.html#profiling-test-execution-duration

PR #621, opened 2024-11-27
Remove test_punet job from ci-sharktank.yml.

ScottTodd:
Partial revert of https://github.com/nod-ai/shark-ai/pull/613.

This job takes over an hour to run (specifically `test_punet_eager_fp16_validation `) and is not suitable for presubmit.

This is likely also not the right file for such tests, and if it was then the specialized tests should go under the generic `test` job, not at the top of the file.

PR #623, opened 2024-11-27
[CI] Add punet export test

nithinsubbiah:
Adds back punet export test to the CI to run on pre-commit. Takes about ~15 minutes for the tests to run

### DIFF
--- a/.github/workflows/ci-sharktank.yml
+++ b/.github/workflows/ci-sharktank.yml
@@ -22,43 +22,6 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  test_punet:
-    name: "Integration Tests - punet"
-    runs-on: nodai-amdgpu-mi250-x86-64
-    env:
-      PIP_CACHE_DIR: "${{ github.workspace }}/.pip-cache"
-    steps:
-      - name: "Setting up Python"
-        id: setup_python
-        uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.0
-        with:
-          python-version: 3.11
-
-      - name: "Checkout Code"
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-
-      - name: Cache Pip Packages
-        uses: actions/cache@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
-        id: cache-pip
-        with:
-          path: ${{ env.PIP_CACHE_DIR }}
-          key: pip-${{ steps.setup_python.outputs.python-version }}-${{ hashFiles('*requirements*.txt','sharktank/requirements*.txt') }}
-
-      - name: Install pip deps
-        run: |
-          python -m pip install --no-compile --upgrade pip
-          # Note: We install in three steps in order to satisfy requirements
-          # from non default locations first. Installing the PyTorch CPU
-          # wheels saves multiple minutes and a lot of bandwidth on runner setup.
-          pip install --no-compile -r pytorch-cpu-requirements.txt
-          pip install --no-compile -r requirements.txt -r sharktank/requirements-tests.txt -e sharktank/
-          # Update to the latest iree packages.
-          pip install -f https://iree.dev/pip-release-links.html --upgrade --pre \
-            iree-base-compiler iree-base-runtime --src deps \
-            -e "git+https://github.com/iree-org/iree-turbine.git#egg=iree-turbine"
-      - name: Run punet tests
-        run: |
-          pytest -v sharktank/ -m "model_punet and export"
   test:
     name: "Unit Tests and Type Checking"
     strategy:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-sharktank.yml
```diff
@@ -22,6 +22,43 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  test_punet:
+    name: "Integration Tests - punet"
+    runs-on: nodai-amdgpu-mi250-x86-64
```
marbre, line 27: Would a standard GH runner also suitable for this? See https://github.com/nod-ai/shark-ai/pull/621#discussion_r1861038318.

saienduri: Yeah we can just use `ubuntu-latest`

marbre: I would rather pin `ubuntu-24.04` and be more specific with the version :) Better to update manually than to see it breaking just because of an update some day.

ScottTodd: Yes, please use a specific version. `ubuntu-latest` is also going to change to `ubuntu 24` soon: https://github.blog/changelog/2024-11-05-notice-of-breaking-changes-for-github-actions/

.github/workflows/ci-sharktank.yml
```diff
@@ -22,6 +22,43 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  test_punet:
```
ScottTodd, line 25: Please order workflow jobs from simplest to most complicated. New jobs testing specific subsets of tests should go at the _bottom_ of the file, not the top. This affects both code structure and the display format on GitHub: 
image

.github/workflows/ci-sharktank.yml
```diff
@@ -22,6 +22,43 @@ concurrency:
...
+          # wheels saves multiple minutes and a lot of bandwidth on runner setup.
+          pip install --no-compile -r pytorch-cpu-requirements.txt
+          pip install --no-compile -r requirements.txt -r sharktank/requirements-tests.txt -e sharktank/
+          # Update to the latest iree packages.
+          pip install -f https://iree.dev/pip-release-links.html --upgrade --pre \
+            iree-base-compiler iree-base-runtime --src deps \
+            -e "git+https://github.com/iree-org/iree-turbine.git#egg=iree-turbine"
+      - name: Run punet tests
+        run: |
+          pytest -v sharktank/ -m "model_punet and export"
```
ScottTodd, line 61: Adding 40 lines of boilerplate to run tests matching a combination of marks for one particular model is a pretty bad code smell. Workflow changes generally shouldn't be required to add test coverage - tests should just get picked up by existing workflows automatically.

The work in https://github.com/nod-ai/shark-ai/issues/584 will simplify the pip install steps, but the test setup still needs more refactoring. Why are these tests not running as part of `pytest -n 4 sharktank/` below? Maybe we could add a "integration test" mark or directory path and run those in a job, instead of specializing this to `punet`? (Remember - we're going to be supporting 10s-100s of models, and we can't scale this sort of test or workflow code as currently written)

ScottTodd, line 61: Consider adding `--durations` like on https://github.com/nod-ai/shark-ai/pull/622 to new `pytest` commands (set to `0` to show all, or set to some medium-sized number like `10` if there are dozens of tests and you only want to see the durations for the N slowest)

saienduri: Yeah Nithin is going to add an integration test mark for now. We can also have it all under `pytest -n 4 sharktank/` but maybe good to have a bit of separation, so it is clear what jobs are testing going forward when bringing in a lot more models

.github/workflows/ci-sharktank.yml
nithinsubbiah: Changed to `ubuntu-24.04`

.github/workflows/ci-sharktank.yml
nithinsubbiah: Added this test as the first of many integration tests, I think its better to keep it as a separate tab integration test as Sai suggested

nithinsubbiah: Done

marbre:
Seems running `punet_quick` takes ~ 15 minutes on pre-summit when using the GH-hosted runners. Might be a little to long for pre-summit still?

saienduri (approved):

saienduri:
15 minutes should be fine. If that's too long, we can switch to the mi250 runners

.github/workflows/ci-sharktank.yml
ScottTodd: As part of https://github.com/nod-ai/shark-ai/issues/584, I think we should keep integration tests in separate workflow files that use package builds while we keep unit tests building from source.

This should merge with https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/ci-llama-quick-tests.yaml and https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/ci_eval_short.yaml as just "short integration tests" (20-30 minute target)

It can start here for now, but it will need to move soon IMO.

ScottTodd (approved):
Please add a PR description explaining what new test coverage this adds and how much CI time it takes.

Issue #130, opened 2024-08-20
libshortfin CI and Releases


stellaraccident:
# Bringing up libshortfin CI and releases

## Overall Description

[libshortfin](https://github.com/nod-ai/sharktank/tree/main/libshortfin) is a 
C++ project providing serving oriented APIs atop the IREE runtime. 
It aims to supercede prior pure-Python systems that used IREE via
its low-level synchronous API. From the get-go, libshortfin is:

* Multi-device
* Parallel (designed to scale in a single process)
* Exposing a high level Python API that directly mirrors the C++ API
* Async (integrating with both Python asyncio and C++ coroutines)

Internally, it consists of three primary components:

* `libshortfin` C++ library (built as both a static and a dynamic library)
* `_shortfin` native Python module
* `shortfin` pure Python module (re-exports from the native module and adds
  higher level Python-only features)

#### Native deps

`libshortfin` currently depends purely on the 
[IREE `runtime` C APIs](https://github.com/iree-org/iree/tree/main/runtime). 
In the future, it will also expose an interface to the 
[IREE compiler C API](https://github.com/iree-org/iree/tree/main/compiler/bindings/c)
via a delay loaded stub, allowing the compiler to be dynamically loaded
(same as some other integrations). The compiler will remain an optional dep
and pure-inference or deployed systems can leave it off so long as whatever
is built on `libshortfin` does not need it (i.e. if pipelines are pre-compiled).

`libshortfin` currently has the following additional native dependencies:

* spdlog
* xtensor
* gtest (if testing is enabled)

#### Python deps

`shortfin` does not have any Python deps for minimal functionality, but it
is expected that a number of deps will be useful for various applications of it,
and it may provide features based on these deps in an optional fashion that
people can use if helpful/available:

* pytest (for testing)
* numpy
* torch
* uvicorn (or other ASGI web server)
* iree-turbine (for torch compilation)
  * Transitive deps on iree-compiler and iree-runtime
* onnx (for onnx compilation)
* etc

#### Variants

The native `_shortfin.lib` Python package is provided, similar to `iree.runtime`,
by redirecting at runtime to a concrete native library module based on env
flags, allowing the selection of multiple packages runtimes such as "default",
"tracy", "assert", etc. As with the `iree.runtime`, we expect that our normal
wheels will bundle the `default` and `tracy` versions so that users can always
use an instrumented build.

### Dependent Projects

It is intended that both C++ and pure/native Python extensions will be built
[...]

PR #627, opened 2024-11-27
[shortfin] Upgrade package build dockerfile to manylinux_2_28.

ScottTodd:
Progress on https://github.com/nod-ai/shark-ai/issues/130.

The manylinux2014 image includes gcc 10.2.1 by default while manylinux_2_28 includes gcc 12.2.1. At one point we had warnings/errors building on the newer gcc version, but that is no longer the case.

With the new Rust dependency coming from https://github.com/nod-ai/shark-ai/pull/610, we will likely want to revive https://github.com/nod-ai/base-docker-images/blob/main/dockerfiles/manylinux_x86_64.Dockerfile, add more dependencies there, then switch from the upstream `quay.io/...` image to that `ghcr.io/nod-ai/...` image.

Tested locally with `OUTPUT_DIR="/tmp/wheelhouse" sudo -E ./build_tools/build_linux_package.sh`. If the nightly package build fails for some reason we can easily revert this.

### DIFF
--- a/shortfin/build_tools/build_linux_package.sh
+++ b/shortfin/build_tools/build_linux_package.sh
@@ -36,11 +36,8 @@ REPO_ROOT="$(cd "$THIS_DIR"/../../ && pwd)"
 SCRIPT_NAME="$(basename $0)"
 ARCH="$(uname -m)"
 
-# TODO(#130): Update to manylinux_2_28, upstream or a fork
-#   * upstream uses a version of gcc that has build warnings/errors
-#   * https://github.com/nod-ai/base-docker-images is a bit out of date but can include a recent clang
-# MANYLINUX_DOCKER_IMAGE="${MANYLINUX_DOCKER_IMAGE:-quay.io/pypa/manylinux_2_28_${ARCH}:latest}"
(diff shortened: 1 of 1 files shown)


ScottTodd:
> With the new Rust dependency coming from #610, we will likely want to revive https://github.com/nod-ai/base-docker-images/blob/main/dockerfiles/manylinux_x86_64.Dockerfile, add more dependencies there, then switch from the upstream `quay.io/...` image to that `ghcr.io/nod-ai/...` image.

As expected, flipping `SHORTFIN_ENABLE_TOKENIZERS=ON` using the stock manylinux image results in:
```
  CMake Error at CMakeLists.txt:261 (message):
    Building with -DSHORTFIN_ENABLE_TOKENIZERS=ON requires cargo (Rust's build
    tool).  Please follow Rust documentation to install.  On Ubuntu, this can
    typically be accomplished with:

      sudo apt install rustup && rustup default stable

    See https://www.rust-lang.org/tools/install
  Call Stack (most recent call first):
    CMakeLists.txt:290 (shortfin_check_tokenizers)
```

marbre (approved):

PR #608, opened 2024-11-26
Implement PageAllocation as a handle into a PagedAttentionCache, allowing publishing and releasing an allocation via handle rather than cache

renxida:
Deinitialization looks wonky for now. Will test extensively to get deinit right once I merge #600

Closes #607

PR #628, opened 2024-11-28
TriePagedAttentionCache - 2

renxida:
See an earlier draft of this PR at #600 

Instead of using a radix tree, a block trie is more appropriate for our paged attention. We can always add back the one feature that distinguishes a radix tree from this implementation (merging tree nodes on long un-branching paths)

This PR depends on #608 and #565 and should not be reviewed until they are merged.


renxida:
Replaced by #632

PR #629, opened 2024-11-28
[tuner]: use lowering config binding

bangtianliu:
This PR is relevant to the task in https://github.com/nod-ai/shark-ai/issues/453 : use IREE bindings for compilation info (incl., lowering_config and translation_info).  

Remove data class `ReorderWorkgroupsStrategy`, and use lowering_config binding.

### DIFF
--- a/tuner/tuner/candidate_gen.py
+++ b/tuner/tuner/candidate_gen.py
@@ -55,14 +55,14 @@ def apply_configuration(
     expr1 = re.compile(
         r"LLVMGPUVectorDistribute workgroup_size = \[.+\] subgroup_size = ([0-9]+),"
     )
-    expr2 = re.compile(r"workgroup = \[\[([0-9]+)(, ([0-9]+))+\]\]")
-    expr3 = re.compile(r"reduction = \[\[([0-9]+)(, ([0-9]+))+\]\]")
+    expr2 = re.compile(r"workgroup = \[([0-9]+)(, ([0-9]+))+\]")
+    expr3 = re.compile(r"reduction = \[([0-9]+)(, ([0-9]+))+\]")
     expr4 = re.compile(r"gpu_pipeline_options = #iree_gpu\.pipeline_options<([^>]*)>")
     expr5 = re.compile(r"\"amdgpu-waves-per-eu\" = \"([0-9])\"")
     repl0 = f"<intrinsic = {intrinsic}, subgroup_m_count = {subgroup_m_count}, subgroup_n_count = {subgroup_n_count}>"
     repl1 = f'LLVMGPUVectorDistribute workgroup_size = [{", ".join(map(str, configuration.workgroup_size))}] subgroup_size = {configuration.subgroup_size},'
-    repl2 = f'workgroup = [[{", ".join(map(str, workgroup_sizes))}]]'
-    repl3 = f'reduction = [[{", ".join(map(str, reduction_sizes))}]]'
+    repl2 = f'workgroup = [{", ".join(map(str, workgroup_sizes))}]'
+    repl3 = f'reduction = [{", ".join(map(str, reduction_sizes))}]'
     repl4 = f"gpu_pipeline_options = {configuration.gpu_pipeline_options}"
     repl5 = f'"amdgpu-waves-per-eu" = "{configuration.waves_per_eu}"'
 
@@ -125,8 +125,6 @@ class MmtTuner(DispatchTuner, MmtParser):
     def get_transform_function_mmt(
         self, problem_size: ProblemSize, functionName: str, configuration: Configuration
     ) -> str:
-        workgroup_sizes = ", ".join(map(str, get_mmt_workgroup_sizes(configuration)))
-        reduction_sizes = ", ".join(map(str, get_mmt_reduction_sizes(configuration)))
         intrinsic = configuration.intrinsic()
         subgroup_m_count = configuration.subgroup_m_count()
         subgroup_n_count = configuration.subgroup_n_count()
@@ -141,7 +139,7 @@ class MmtTuner(DispatchTuner, MmtParser):
     transform.iree.match.cast_compatible_type %lhs = tensor<{problem_size.lhs_type}> : !transform.any_value
     transform.iree.match.cast_compatible_type %rhs = tensor<{problem_size.rhs_type}> : !transform.any_value
     %config = transform.param.constant #iree_codegen.compilation_info<
-        lowering_config = #iree_codegen.lowering_config<workgroup = [[{workgroup_sizes}]], reduction = [[{reduction_sizes}]]>,
+        lowering_config = {configuration.lowering_config}>,
         translation_info = #iree_codegen.translation_info<LLVMGPUVectorDistribute
         workgroup_size = [{wg_x}, {wg_y}, {wg_z}] subgroup_size = {configuration.subgroup_size},
         {{mma_schedule = #iree_gpu.mma_schedule<
@@ -218,7 +216,7 @@ class ConvTuner(DispatchTuner, ConvParser):
         outs(%out : {output}) -> {output}
     }} : (!transform.any_op) -> (!transform.any_value, !transform.any_value)
         %config = transform.param.constant #iree_codegen.compilation_info<
-        lowering_config = #iree_codegen.lowering_config<workgroup = [[{workgroup_sizes}]], reduction = [[{reduction_sizes}]]>,
+        lowering_config = {configuration.lowering_config}>,
         translation_info = #iree_codegen.translation_info<LLVMGPUVectorDistribute
         workgroup_size = [{wg_x}, {wg_y}, {wg_z}] subgroup_size = {configuration.subgroup_size},
             {{mma_schedule = #iree_gpu.mma_schedule<
@@ -290,7 +288,7 @@ transform.named_sequence @{functionName}(%generic: !transform.any_op {{transform
 transform.iree.match.cast_compatible_type %lhs = tensor<{lhs_dynamic_batch}> : !transform.any_value
 transform.iree.match.cast_compatible_type %rhs = tensor<{problem_size.rhs_type}> : !transform.any_value
 %config = transform.param.constant #iree_codegen.compilation_info<
-    lowering_config = #iree_codegen.lowering_config<workgroup = [[{workgroup_sizes}]], reduction = [[{reduction_sizes}]]>,
+    lowering_config = {configuration.lowering_config}>,
     translation_info = #iree_codegen.translation_info<LLVMGPUVectorDistribute
     workgroup_size = [{wg_x}, {wg_y}, {wg_z}] subgroup_size = {configuration.subgroup_size},
     {{mma_schedule = #iree_gpu.mma_schedule<
@@ -361,12 +359,6 @@ class BatchMmtTuner(DispatchTuner, BatchMmtParser):
         functionName: str,
         configuration: Configuration,
     ) -> str:
-        workgroup_sizes = ", ".join(
-            map(str, get_batch_mmt_workgroup_sizes(configuration))
-        )
-        reduction_sizes = ", ".join(
-            map(str, get_batch_mmt_reduction_sizes(configuration))
-        )
         intrinsic = configuration.intrinsic()
         subgroup_m_count = configuration.subgroup_m_count()
         subgroup_n_count = configuration.subgroup_n_count()
@@ -382,7 +374,7 @@ transform.named_sequence @{functionName}(%generic: !transform.any_op {{transform
 transform.iree.match.cast_compatible_type %lhs = tensor<{problem_size.lhs_type}> : !transform.any_value
 transform.iree.match.cast_compatible_type %rhs = tensor<{problem_size.rhs_type}> : !transform.any_value
 %config = transform.param.constant #iree_codegen.compilation_info<
-    lowering_config = #iree_codegen.lowering_config<workgroup = [[{workgroup_sizes}]], reduction = [[{reduction_sizes}]]>,
+    lowering_config = {configuration.lowering_config}>,
     translation_info = #iree_codegen.translation_info<LLVMGPUVectorDistribute
     workgroup_size = [{wg_x}, {wg_y}, {wg_z}] subgroup_size = {configuration.subgroup_size},
     {{mma_schedule = #iree_gpu.mma_schedule<
@@ -436,12 +428,6 @@ class BatchMatmulTuner(DispatchTuner, BatchMatmulParser):
         input1 = f"tensor<{problem_size.rhs_type}>"
         output = f"tensor<{problem_size.res_type}>"
 
-        workgroup_sizes = ", ".join(
-            map(str, get_contract_workgroup_sizes(configuration, tile_dims))
-        )
-        reduction_sizes = ", ".join(
-            map(str, get_contract_reduction_sizes(configuration, tile_dims))
-        )
         intrinsic = configuration.intrinsic()
         subgroup_m_count = configuration.subgroup_m_count()
         subgroup_n_count = configuration.subgroup_n_count()
@@ -459,7 +445,7 @@ class BatchMatmulTuner(DispatchTuner, BatchMatmulParser):
         outs(%out : {output}) -> {output}
     }} : (!transform.any_op) -> (!transform.any_value, !transform.any_value)
         %config = transform.param.constant #iree_codegen.compilation_info<
-        lowering_config = #iree_codegen.lowering_config<workgroup = [[{workgroup_sizes}]], reduction = [[{reduction_sizes}]]>,
+        lowering_config = {configuration.lowering_config}>,
         translation_info = #iree_codegen.translation_info<LLVMGPUPadAndVectorDistribute
         workgroup_size = [{wg_x}, {wg_y}, {wg_z}] subgroup_size = {configuration.subgroup_size},
             {{mma_schedule = #iree_gpu.mma_schedule<
--- a/tuner/tuner/candidate_gen.py
+++ b/tuner/tuner/candidate_gen.py
@@ -42,6 +42,9 @@ tune_logger = logging.getLogger("tune")
 def apply_configuration(
     template: list[str], configuration: Configuration, tile_sizes: list[int]
 ) -> str:
+    intrinsic = configuration.intrinsic
+    subgroup_m_count = configuration.subgroup_m_count
+    subgroup_n_count = configuration.subgroup_n_count
     tune_logger.info(f"Applying: {configuration}")
     expr0 = re.compile(
         r"<intrinsic = #iree_gpu\.mma_layout<(.+)>, subgroup_m_count = ([0-9]+), subgroup_n_count = ([0-9]+)>"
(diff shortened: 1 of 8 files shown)


tuner/tuner/candidate_gen_test.py
```diff
@@ -48,13 +48,25 @@ def test_apply_params_mmt(tuner_ctx: common.TunerContext) -> None:
 
     mma_intrinsic = iree_gpu.MMAIntrinsic.MFMA_F32_16x16x16_F16
     mma_attr = iree_gpu.MMAAttr.get(mma_intrinsic)
+    lowering_config_dict = {
+        "mma_kind": mma_attr,
+        "workgroup": ir.ArrayAttr.get(
+            [
+                ir.IntegerAttr.get(tuner_ctx.type.i32, 8),
+                ir.IntegerAttr.get(tuner_ctx.type.i32, 8),
+                ir.IntegerAttr.get(tuner_ctx.type.i32, 8),
```
kuhar, line 57: We can add a helper functions to `tuner_ctx.type` like `tuner_ctx.type.getI32(8)` etc. This should save a lot of typing in code like this.

tuner/tuner/candidate_gen_test.py
```diff
@@ -48,13 +48,25 @@ def test_apply_params_mmt(tuner_ctx: common.TunerContext) -> None:
...
+            ]
+        ),
+        "reduction": ir.ArrayAttr.get([]),
+        "subgroup_m_count": ir.IntegerAttr.get(tuner_ctx.type.i32, 16),
+        "subgroup_n_count": ir.IntegerAttr.get(tuner_ctx.type.i32, 16),
+    }
+
+    lowering_config_attrs = ir.DictAttr.get(lowering_config_dict)
```
kuhar, line 65: We can create a helper function that accepts python types and returns a lowering config attribute to avoid having to create these dictionaries by hand.

For example, this could be something like: `common.getLoweringConfig(mma_kind=mma_attr, workgroup=[8,8,0], reduction=[0, 0, 8], subgroup_m_count=16, subgroup_n_count=16)`.

tuner/tuner/candidate_gen_test.py
```diff
@@ -48,13 +48,25 @@ def test_apply_params_mmt(tuner_ctx: common.TunerContext) -> None:
...
+    lowering_config_dict = {
+        "mma_kind": mma_attr,
+        "workgroup": ir.ArrayAttr.get(
+            [
+                ir.IntegerAttr.get(tuner_ctx.type.i32, 8),
+                ir.IntegerAttr.get(tuner_ctx.type.i32, 8),
+                ir.IntegerAttr.get(tuner_ctx.type.i32, 8),
+            ]
+        ),
+        "reduction": ir.ArrayAttr.get([]),
```
kuhar, line 60: The tile sizes are incorrect here: `[8, 8, 8]` in the single-array format should be `workgroup=[8, 8, 0], reduction = [0, 0, 8]`

tuner/tuner/common.py
```diff
@@ -105,26 +105,34 @@ def is_comptible(mma_intrinsic: iree_gpu.MMAIntrinsic) -> bool:
...
-    tile_sizes: list[int]
-    subgroup_m_count: int
-    subgroup_n_count: int
+    lowering_config: iree_gpu.LoweringConfigAttr
     gpu_pipeline_options: iree_gpu.PipelineOptionsAttr
     waves_per_eu: int
 
+    @property
+    def intrinsic(self) -> iree_gpu.MMAAttr:
+        return self.lowering_config.attributes["mma_kind"]
```
kuhar, line 118: I wouldn't make these properties because this can fail when the lowering config doesn't contain that attribute

tuner/tuner/common.py
```diff
@@ -105,26 +105,34 @@ def is_comptible(mma_intrinsic: iree_gpu.MMAIntrinsic) -> bool:
...
+    def tilesize_reduction(self) -> list[int]:
+        return [attr.value for attr in self.lowering_config.attributes["reduction"]]
+
+    @property
+    def subgroup_m_count(self) -> int:
+        return self.lowering_config.attributes["subgroup_m_count"].value
+
+    @property
+    def subgroup_n_count(self) -> int:
+        return self.lowering_config.attributes["subgroup_n_count"].value
```
kuhar, line 134: Adding these is a good idea, but we should make sure we have tests that exercise this code

kuhar (changes requested):

tuner/tuner/candidate_gen_test.py
kuhar: This could also be new method defined in python bindings, but I think it will be easier to define this in python as a new helper.

tuner/tuner/candidate_gen.py
```diff
@@ -127,12 +141,12 @@ def get_transform_function_mmt(
     transform.iree.match.cast_compatible_type %lhs = tensor<{problem_size.lhs_type}> : !transform.any_value
     transform.iree.match.cast_compatible_type %rhs = tensor<{problem_size.rhs_type}> : !transform.any_value
     %config = transform.param.constant #iree_codegen.compilation_info<
-        lowering_config = #iree_codegen.lowering_config<tile_sizes = [[{tile_sizes}]]>,
+        lowering_config = #iree_codegen.lowering_config<workgroup = [[{workgroup_sizes}]], reduction = [[{reduction_sizes}]]>,
```
kuhar, line 144: We should print the lowering config attribute here instead of extracting stuff out of it.

tuner/tuner/common.py
```diff
@@ -105,26 +110,63 @@ def is_comptible(mma_intrinsic: iree_gpu.MMAIntrinsic) -> bool:
...
-    tile_sizes: list[int]
-    subgroup_m_count: int
-    subgroup_n_count: int
+    lowering_config: iree_gpu.LoweringConfigAttr
     gpu_pipeline_options: iree_gpu.PipelineOptionsAttr
     waves_per_eu: int
 
+    def intrinsic(self) -> Optional[iree_gpu.MMAAttr]:
+        if self.lowering_config.attributes.__contains__("mma_kind"):
+            return self.lowering_config.attributes.__getitem__("mma_kind")
```
kuhar, line 123: Don't use private methods (that start with an underscore). For Dictattr, I think you can use the `in` and `[...]` operators

kuhar: Also, I think this pattern would be supported as `my_dict.get(key, default)`, but I'm not sure if the python bindings support that

tuner/tuner/common.py
```diff
@@ -105,26 +110,63 @@ def is_comptible(mma_intrinsic: iree_gpu.MMAIntrinsic) -> bool:
...
+    for key, value in kwargs.items():
+        if isinstance(value, list):
+            lowering_config_dict[key] = ir.ArrayAttr.get(
+                [tuner_ctx.type.getI64(x) for x in value]
+            )
+        elif isinstance(value, int):
+            lowering_config_dict[key] = tuner_ctx.type.getI64(value)
+        elif isinstance(value, iree_gpu.MMAAttr):
+            lowering_config_dict[key] = value
+        else:
```
kuhar, line 165: I think it would be better to do the conversion based on known keys instead of just values. We can check if the passes value is an attribute (nothing to do then), or dispatch based on the type otherwise (I think you can use a `match` statement for that ([some examples](https://benhoyt.com/writings/python-pattern-matching/)).

tuner/tuner/common.py
```diff
@@ -105,26 +110,63 @@ def is_comptible(mma_intrinsic: iree_gpu.MMAIntrinsic) -> bool:
...
+        return None
+
+    def subgroup_n_count(self) -> Optional[int]:
+        if self.lowering_config.attributes.__contains__("subgroup_n_count"):
+            attr = self.lowering_config.attributes.__getitem__("subgroup_n_count")
+            return attr.value
+        return None
+
+
+def get_lowering_config(
```
kuhar, line 151: Please add tests for this and make sure the returned attributes roundtrip with the helpers from `Configuration`

tuner/tuner/dispatch_parser.py
```diff
@@ -140,18 +164,21 @@ class ConvParser(DispatchParser):
...
         fh = 1
         fw = 1
 
         oh = 1
 
-        oc = n
-        ow = m
-        ic = k
-        return [batch, oh, ow, oc, fh, fw, ic]
+        ow, oc, _ = configuration.tilesize_workgroup()
```
kuhar, line 174:
```suggestion
        ow, oc, _ic = configuration.tilesize_workgroup()
```

tuner/tuner/dispatch_constraints.py
```diff
@@ -217,25 +219,46 @@ def generate_solutions(
...
+                    ir.IntegerAttr.get(int_type, 0),
+                ]
+            ),
+            "reduction": ir.ArrayAttr.get(
+                [
+                    ir.IntegerAttr.get(int_type, 0),
+                    ir.IntegerAttr.get(int_type, 0),
+                    ir.IntegerAttr.get(int_type, lookup(k)),
+                ]
+            ),  # placeholder now to be consistent with iree
```
kuhar, line 251: Use proper casing and punctuation.

tuner/tuner/dispatch_parser_test.py
```diff
@@ -42,60 +42,93 @@ def test_parse_tensor_type(tuner_ctx: common.TunerContext) -> None:
...
+    assert dispatch_parser.get_contract_reduction_sizes(config, "mnk") == [0, 0, 16]
+    assert dispatch_parser.get_contract_workgroup_sizes(config, "nmk") == [8, 4, 0]
+    assert dispatch_parser.get_contract_reduction_sizes(config, "nmk") == [0, 0, 16]
+    assert dispatch_parser.get_contract_workgroup_sizes(config, "knm") == [0, 8, 4]
+    assert dispatch_parser.get_contract_reduction_sizes(config, "knm") == [16, 0, 0]
+    assert dispatch_parser.get_contract_workgroup_sizes(config, "kkk") == [
+        0,
+        0,
+        0,
+    ]
```
kuhar, line 130: This formatting looks weird -- I'd expect the `0, 0, 0` to fit on the same line since `16, 0, 0` is longer and fits

tuner/tuner/common.py
```diff
@@ -105,26 +110,63 @@ def is_comptible(mma_intrinsic: iree_gpu.MMAIntrinsic) -> bool:
...
     workgroup_size: list[int]
-    intrinsic: iree_gpu.MMAAttr
-    tile_sizes: list[int]
-    subgroup_m_count: int
-    subgroup_n_count: int
+    lowering_config: iree_gpu.LoweringConfigAttr
     gpu_pipeline_options: iree_gpu.PipelineOptionsAttr
     waves_per_eu: int
 
+    def intrinsic(self) -> Optional[iree_gpu.MMAAttr]:
```
kuhar, line 121: If these helpers are useful enough, we could also add them to the iree python bindings. I think these helpers are also fine, but I'd rather see them as free functions instead of methods in `Configuration`. They don't need to know about `subgroup_size` and workgroup_size` after all -- the only input should the the lowering config attr.

bangtianliu: Good point, I will send another IREE PR to support this.

tuner/tuner/common.py
```diff
@@ -105,26 +110,83 @@ def is_comptible(mma_intrinsic: iree_gpu.MMAIntrinsic) -> bool:
...
+        if "mma_kind" in self.lowering_config.attributes:
+            return self.lowering_config.attributes["mma_kind"]
+        return None
+
+    def tilesize_workgroup(self) -> list[int]:
```
kuhar, line 126: There shouldn't be methods of `Configuration`. We can have them as free functions instead.

tuner/tuner/candidate_gen.py
```diff
@@ -40,36 +40,46 @@
...
+    expr3 = re.compile(r"reduction = \[([0-9]+)(, ([0-9]+))+\]")
+    expr4 = re.compile(r"gpu_pipeline_options = #iree_gpu\.pipeline_options<([^>]*)>")
+    expr5 = re.compile(r"\"amdgpu-waves-per-eu\" = \"([0-9])\"")
+    repl0 = f"<intrinsic = {intrinsic}, subgroup_m_count = {subgroup_m_count}, subgroup_n_count = {subgroup_n_count}>"
     repl1 = f'LLVMGPUVectorDistribute workgroup_size = [{", ".join(map(str, configuration.workgroup_size))}] subgroup_size = {configuration.subgroup_size},'
-    repl2 = f'tile_sizes = [[{", ".join(map(str, tile_sizes))}]]'
-    repl3 = f"gpu_pipeline_options = {configuration.gpu_pipeline_options}"
-    repl4 = f'"amdgpu-waves-per-eu" = "{configuration.waves_per_eu}"'
+    repl2 = f'workgroup = [{", ".join(map(str, workgroup_sizes))}]'
+    repl3 = f'reduction = [{", ".join(map(str, reduction_sizes))}]'
```
kuhar, line 65: Can we print the attributes directly? `config.lowering_config.attributes["workgroup"]`

tuner/tuner/common.py
```diff
@@ -105,27 +110,84 @@ def is_comptible(mma_intrinsic: iree_gpu.MMAIntrinsic) -> bool:
...
     waves_per_eu: int
 
 
+def get_intrinsic(config: Configuration) -> Optional[iree_gpu.MMAAttr]:
+    if "mma_kind" in config.lowering_config.attributes:
+        return config.lowering_config.attributes["mma_kind"]
+    return None
+
+
+def get_tilesize_workgroup(config: Configuration) -> list[int]:
```
kuhar, line 128:
```suggestion
def get_workgroup_tile_sizes(config: Configuration) -> list[int]:
```

tuner/tuner/common.py
```diff
@@ -105,27 +110,84 @@ def is_comptible(mma_intrinsic: iree_gpu.MMAIntrinsic) -> bool:
...
+    if "workgroup" in config.lowering_config.attributes:
+        workgroup_attrs = config.lowering_config.attributes["workgroup"]
+        return [attr.value for attr in workgroup_attrs]
+    return []
+
+
+def get_tilesize_reduction(config: Configuration) -> list[int]:
```
kuhar, line 135:
```suggestion
def get_reduction_tile_sizes(config: Configuration) -> list[int]:
```

tuner/tuner/common.py
```diff
@@ -105,27 +110,84 @@ def is_comptible(mma_intrinsic: iree_gpu.MMAIntrinsic) -> bool:
...
+                ), f"Unsupported type for key '{key}': {type(value).__name__}"
+                if isinstance(value, int):
+                    promoted_value = tuner_ctx.type.getI64(value)
+            case "mma_kind":
+                assert isinstance(
+                    value, iree_gpu.MMAAttr
+                ), f"Unsupported type for key '{key}': {type(value).__name__}"
+            case _:
+                raise KeyError(f"Unhandled key in lowering configuration: {key}")
+        # Single assignment after the match.
```
kuhar, line 185: This is pretty obvious, I don't think we need this comment but could use an empty line

tuner/tuner/common.py
```diff
@@ -105,27 +110,84 @@ def is_comptible(mma_intrinsic: iree_gpu.MMAIntrinsic) -> bool:
...
+) -> iree_gpu.LoweringConfigAttr:
+    lowering_config_dict: dict[str, Any] = {}
+    for key, value in kwargs.items():
+        # A local variable to hold the transformed value.
+        promoted_value = value
+        match key:
+            case "workgroup" | "reduction":
+                assert isinstance(
+                    value, (list, ir.ArrayAttr)
+                ), f"Unsupported type for key '{key}': {type(value).__name__}"
```
kuhar, line 168: Don't check the type in the assert -- if the previous if doesn't match, `assert False`

tuner/tuner/candidate_gen.py
```diff
@@ -40,36 +40,46 @@
 
 
 def apply_configuration(
-    template: list[str], configuration: Configuration, tile_sizes: list[int]
+    template: list[str],
+    configuration: Configuration,
+    workgroup_sizes: list[int],
+    reduction_sizes: list[int],
```
kuhar, line 46: Could we extract workgroup and tile sizes from the `configuration` directly and remove these two function arguments?

bangtianliu: yeah, good catch!

bangtianliu: Oh, actually no. The tile sizes from the configuration are not used directly https://github.com/nod-ai/shark-ai/blob/main/tuner/tuner/dispatch_parser.py#L41, so these two function arguments cannot be removed.

tuner/tuner/candidate_gen.py
bangtianliu: Similar to the reply below, the tile sizes are generated by functions that vary based on the operators, as shown here: https://github.com/nod-ai/shark-ai/blob/main/tuner/tuner/dispatch_parser.py#L23toL40.

tuner/tuner/candidate_gen.py
kuhar: This definitely should be removed eventually, but we don't have to put everything in the same PR. Could you add a TODO so that we don't forget about this?

tuner/tuner/dispatch_constraints.py
```diff
@@ -217,25 +219,46 @@ def generate_solutions(
...
+                    ir.IntegerAttr.get(int_type, 0),
+                    ir.IntegerAttr.get(int_type, lookup(k)),
+                ]
+            ),  # Placeholder now to be consistent with iree.
+            "subgroup_m_count": ir.IntegerAttr.get(int_type, lookup(sg_m_cnt)),
+            "subgroup_n_count": ir.IntegerAttr.get(int_type, lookup(sg_n_cnt)),
+        }
+
+        lowering_config_attrs = ir.DictAttr.get(lowering_config_dict)
+        lowering_config = iree_gpu.LoweringConfigAttr.get(lowering_config_attrs)
```
kuhar, line 257: Why aren't we using the new `get_lowering_config` helper here?

tuner/tuner/candidate_gen.py
bangtianliu: Sure, will do.

tuner/tuner/dispatch_constraints.py
```diff
@@ -217,25 +219,39 @@ def generate_solutions(
     )
     solver.add(z3.simplify(z3.And(constraints)))
     logger.debug(f"Initial constraints: {solver}")
+
+    int_type = ir.IntegerType.get_signless(64)
+
+    tuner_ctx = TunerContext(ir.Context(), logger)
```
kuhar, line 225: We shouldn't be creating a tuner context here. If it's needed, it should be passed as a function argument. This is likely to break object comparisons if we accidentally end up with two contexts at once.

tuner/tuner/dispatch_constraints.py
```diff
@@ -217,25 +219,39 @@ def generate_solutions(
     )
     solver.add(z3.simplify(z3.And(constraints)))
     logger.debug(f"Initial constraints: {solver}")
+
+    int_type = ir.IntegerType.get_signless(64)
```
kuhar, line 223: use the int type from the context

PR #631, opened 2024-12-02
[sharktank] Fix flaxy quantizer test 

dan-garvey:
Forcibly assigns type to avoid test flake when it selects bfloat16 for unknown reasons.


dan-garvey:
I wasn't able to reproduce locally, but based on the error we hit on CI this should avoid it. 🤞

ScottTodd (approved):

Issue #638, opened 2024-12-03
Shortfin LLM Regression from IREE version `3.1.0rc20241203`


stbaione:
# Description

It appears that there was a regression for Shortfin LLM introduced by: [0f21989438e5c422287882cb67a315329b7b744e](https://github.com/iree-org/iree/commit/0f21989438e5c422287882cb67a315329b7b744e)

Have already filed an [IREE Issue](https://github.com/iree-org/iree/issues/19360)

```Memory access fault by GPU node-8 (Agent handle: 0x5610e4b7e750) on address 0x7eff79e03000. Reason: Unknown.```

Can be seen in this [CI - sharktank perplexity short](https://github.com/nod-ai/shark-ai/actions/runs/12144449804/job/33863814189#step:6:42) run, where the iree-base-compiler/iree-base-runtime version was upgraded from 3.1.0rc20241202 to 3.1.0rc20241203.

I saw the same error when validating my own CI in a PR, and was able to reproduce it locally with the shortfin server.

# Repro Steps
I was able to bisect and verify the issue by exporting/compiling fresh model artifacts and sending a request to the shortfin LLM server. Repro steps outlined below:

1. Create new directory
2. Download Llama3_8b_fp16 w/ sharktank.utils.hf_datasets:
```bash
python -m sharktank.utils.hf_datasets llama3_8B_fp16 --local-dir ./artifacts

```
3. Export to MLIR:

```bash
python -m sharktank.examples.export_paged_llm_v1 --gguf-file=./artifacts/meta-llama-3.1-8b-instruct.f16.gguf --output-mlir=./artifacts/model.mlir --output-config=./artifacts/config.json --bs=1,4

```
5. Compile model:

```
iree-compile ./artifacts/model.mlir -o ./artifacts/model.vmfb

```
7. Save the following json data to ./artifacts/config.json:
```json
{
        "module_name": "module",
        "module_abi_version": 1,
        "max_seq_len": 131072,
        "attn_head_count": 8,
        "attn_head_dim": 128,
        "prefill_batch_sizes": [1, 4],
        "decode_batch_sizes": [1, 4],
        "transformer_block_count": 32,
        "paged_kv_cache": {"block_seq_stride": 16, "device_block_count": 256},
    }

```

9. Start Shortfin LLM Server:

```bash
[...]

PR #639, opened 2024-12-03
Workflow cleanup: move checkout to start, prune unused steps+vars.

ScottTodd:
Cutting down on boilerplate and standardizing some workflow syntax before deeper refactoring work.

* Move `users: actions/checkout` to the start of each job (for consistency, since all workflows use the source code but not all workflows need dependencies or python setup, etc.)
* Per https://github.com/actions/checkout, `submodules: false` is the default
* Step names can be omitted with `uses` (docs: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_iduses) . Our [`pre-commit.yml`](https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/pre-commit.yaml) workflow does this already: https://github.com/nod-ai/shark-ai/blob/c0ca2e2788ea09a890ffe347dfbffe57606670b8/.github/workflows/pre-commit.yaml#L11-L14
image
* Remove copy/pasted `Get Current Date` step from some workflows
* Remove unused `SHARK_PLATFORM_REPO_ROOT` env var from some workflows (unused since https://github.com/nod-ai/shark-ai/pull/305, copy/pasted into a few places)

### DIFF
--- a/.github/workflows/ci-sharktank.yml
+++ b/.github/workflows/ci-sharktank.yml
@@ -36,14 +36,13 @@ jobs:
     env:
       PIP_CACHE_DIR: "${{ github.workspace }}/.pip-cache"
     steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
       - name: "Setting up Python"
         id: setup_python
         uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.0
         with:
           python-version: ${{matrix.version}}
-
-      - name: "Checkout Code"
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
       - name: Cache Pip Packages
         uses: actions/cache@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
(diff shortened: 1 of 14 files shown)


stbaione (approved):
LGTM, test failures for `perplexity short` and `Llama Benchmarking` are due to #638, not anything with this PR

.github/workflows/ci_eval.yaml
```diff
@@ -36,17 +36,15 @@ jobs:
         shell: bash
     env:
       PIP_CACHE_DIR: "${{ github.workspace }}/.pip-cache"
-      SHARK_PLATFORM_REPO_ROOT: ${{ github.workspace }}
     steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
archana-ramalingam, line 40: Just seeing this now, there is another job in this workflow that needs to be updated as well. Maybe in another PR.

.github/workflows/ci-sharktank.yml
```diff
@@ -36,15 +36,14 @@ jobs:
     env:
       PIP_CACHE_DIR: "${{ github.workspace }}/.pip-cache"
     steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
```
archana-ramalingam, line 40: Same here, there is another job in this workflow that needs to be updated as well.

ScottTodd:
Ah. Thanks for noticing. Sent https://github.com/nod-ai/shark-ai/pull/642

marbre (commented):
Thanks for the cleanup.

PR #625, opened 2024-11-27
Convert `ci-shark-ai.yml` to `pkgci_shark_ai.yml`.

ScottTodd:
This builds on https://github.com/nod-ai/shark-ai/pull/589 to make progress on https://github.com/nod-ai/shark-ai/issues/584.

On the surface, this adds complexity right now. The benefits are:

* Integration test workflows use dev packages, without needing to build them from source or use editable installs
* Users and developers can access the same dev packages to reproduce CI results
* We only need to build shortfin from source once for multiple jobs (potentially saving 2-3 minutes)
* Only one runner needs the build requirements (potentially including clang, ninja, CMake, Rust, etc.), other runners only need Python

This also switches to using Python venvs, tightening configuration control in workflows and ensuring that only the expected packages are installed and used for tests, without leaking installed packages across workflow runs. That does increase the workflow time though.

---

## Metrics using `pip`

Logs before: https://github.com/nod-ai/shark-ai/actions/runs/12058325322

* **Total: 7m6s**
* 3m5s on `Install pip deps` (no venv, reusing already installed packages and getting the wrong IREE versions)
* 3m14s on `Run LLM Integration Tests`

Logs after (cold cache): https://github.com/nod-ai/shark-ai/actions/runs/12059301876/attempts/1?pr=625

* **Total: 9m57s**
* 2m42s on `Build shortfin` (cold cache)
* 2m37s on `Setup venv` (from scratch, no deps reused)
* 16s on `Install nightly IREE packages`
* 3m25s on `Run LLM Integration Tests`

Logs after (warm cache): https://github.com/nod-ai/shark-ai/actions/runs/12059301876/attempts/2?pr=625

* **Total: 8m20s**
* 1m24s on `Build shortfin` (warm cache, 46% cache hit rate)
* 2m23s on `Setup venv` (from scratch, no deps reused)
* 17s on `Install nightly IREE packages`
* 3m23s on `Run LLM Integration Tests`

## Metrics using `uv`

`uv` (https://docs.astral.sh/uv/) can be used as an alternative to `pip`

Logs (cold cache): https://github.com/nod-ai/shark-ai/actions/runs/12147344797?pr=625

* **Total: 6m31s**
* 2m48s on `Build shortfin` (cold cache)
* 1m58s on `Setup venv` (from scratch, no deps reused)
* 24s on `Install nightly IREE packages`
* 24s on `Run LLM Integration Tests` (failed)

[...]

PR #640, opened 2024-12-03
Install python packages into venvs across workflows.

ScottTodd:
Many of these workflows are using persistent self-hosted runners, so it looks like they have been reusing the same system-wide Python environment between workflow runs (plus layer of caching on top). This switches to using venvs at `${{ github.workspace }}/.venv` that should be ephemeral, giving us more explicit control over which packages are installed.

More work is planned as part of https://github.com/nod-ai/shark-ai/issues/584 to refactor these workflows further - replacing the package installs code like `pip install --no-compile -r requirements.txt -r sharktank/requirements-tests.txt -e sharktank/` with a `setup_venv.py` script that uses dev/nightly/stable packages (from an appropriate source).

This also disables pip caching, since that is not directly compatible with using venvs. As a result, some workflows are slower now, but they are more predictable in what they install. Good reading for adding caching back:
* https://adamj.eu/tech/2023/11/02/github-actions-faster-python-virtual-environments/
* https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages

### DIFF
--- a/.github/workflows/ci-llama-quick-tests.yaml
+++ b/.github/workflows/ci-llama-quick-tests.yaml
@@ -33,7 +33,6 @@ jobs:
       run:
         shell: bash
     env:
-      PIP_CACHE_DIR: "${{ github.workspace }}/.pip-cache"
       VENV_DIR: ${{ github.workspace }}/.venv
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
--- a/.github/workflows/ci-llama-quick-tests.yaml
+++ b/.github/workflows/ci-llama-quick-tests.yaml
@@ -33,7 +33,6 @@ jobs:
       run:
         shell: bash
     env:
-      PIP_CACHE_DIR: "${{ github.workspace }}/.pip-cache"
       VENV_DIR: ${{ github.workspace }}/.venv
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
@@ -52,10 +51,9 @@ jobs:
 
       - name: Cache Pip Packages
         uses: actions/cache@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
-        id: cache-pip
         with:
-          path: ${{ env.PIP_CACHE_DIR }}
-          key: pip-${{ steps.setup_python.outputs.python-version }}-${{ hashFiles('*requirements.txt') }}
+          path: ${{ env.VENV_DIR }}
+          key: pip-venv-${{ steps.setup_python.outputs.python-version }}-${{ hashFiles('*requirements.txt') }}
 
       - name: Install pip deps
         run: |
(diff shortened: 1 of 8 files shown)


.github/workflows/ci-sharktank.yml
```diff
@@ -98,7 +97,6 @@ jobs:
```
archana-ramalingam, line 97: We could add missed changes from [here](https://github.com/nod-ai/shark-ai/pull/639#discussion_r1868485174).

ScottTodd: Keeping PRs small and focused. Sent https://github.com/nod-ai/shark-ai/pull/642 for that.

.github/workflows/ci-llama-quick-tests.yaml
```diff
@@ -47,6 +47,8 @@ jobs:
         uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.0
         with:
           python-version: ${{matrix.version}}
+      - name: Create Python venv
+        run: python -m venv ${VENV_DIR}
 
       - name: Cache Pip Packages
```
ScottTodd, line 53: Notes to self:

Need to be deliberate with pip caching. I think I want the downloads to be cached, but _not_ any installed packages.

Reference https://adamj.eu/tech/2023/11/02/github-actions-faster-python-virtual-environments/ and https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages.

Both of those approaches hash the requirements files as the cache key, but we have so many separate requirements files _and_ we install our own dev and nightly packages on top, which are not pinned in any requirement files to be cached.

Some of that will hopefully be less consequential, if/when we switch to the `setup_venv.py` + `uv` setup style that I'm prototyping on https://github.com/nod-ai/shark-ai/pull/625.

ScottTodd: Maybe try landing tomorrow without any pip/venv caching, then figure out how to add it back carefully.  Some extra CI time is worth the environment consistency.

.github/workflows/ci-sglang-benchmark.yml
```diff
@@ -46,6 +46,7 @@ jobs:
         shell: bash
     env:
       PIP_CACHE_DIR: "${{ github.workspace }}/.pip-cache"
```
stbaione, line 48: `PIP_CACHE_DIR` can be removed

.github/workflows/ci-sglang-integration-tests.yml
```diff
@@ -35,6 +35,7 @@ jobs:
         shell: bash
     env:
       PIP_CACHE_DIR: "${{ github.workspace }}/.pip-cache"
```
stbaione, line 37: `PIP_CACHE_DIR` can be removed

.github/workflows/ci-sglang-benchmark.yml
ScottTodd: Ah, thanks. Removed.

stbaione (approved):

marbre (approved):

PR #643, opened 2024-12-04
Add CLIP text model

sogartar:
Ports the CLIP text model from Hugging Face. This is the first iteration so not much is changed from the original model. Things like dropout and checkpointing are removed.
Add numeric verification tests for the various components of the stack when executing in eager mode. Verifications are made for float32 and bfloat16. There are tests for toy-sized components and the whole model as well as the Large pretrained variant.
These tests does not include testing with IREE.

Functionalities for mask creation are not yet ported.

### DIFF
--- a/sharktank/sharktank/layers/norm.py
+++ b/sharktank/sharktank/layers/norm.py
@@ -52,10 +52,9 @@ class LayerNorm(ThetaLayer):
     ):
         super().__init__(theta)
         self.weight = self.theta_tensor(weight_name)
+        self.bias = None
         if bias_name in self.theta.keys:
             self.bias = self.theta_tensor(bias_name)
-        else:
-            self.bias = None
         self.eps = eps
 
     def forward(self, x: torch.Tensor):
--- a/sharktank/sharktank/layers/norm.py
+++ b/sharktank/sharktank/layers/norm.py
@@ -39,3 +39,24 @@ class RMSNormLayer(ThetaLayer):
         # often in higher precision. Downcast back to expected.
         norm = ops.to(norm, orig_dtype)
         return norm
+
+
+class LayerNorm(ThetaLayer):
+    def __init__(
+        self,
+        theta: Theta,
+        *,
+        weight_name: str = "weight",
+        bias_name: str = "bias",
+        eps: float = 1e-05,
+    ):
+        super().__init__(theta)
+        self.weight = self.theta_tensor(weight_name)
+        if bias_name in self.theta.keys:
+            self.bias = self.theta_tensor(bias_name)
+        else:
+            self.bias = None
+        self.eps = eps
+
+    def forward(self, x: torch.Tensor):
+        return ops.layer_norm(x, weight=self.weight, bias=self.bias, eps=self.eps)
--- a/sharktank/sharktank/models/clip/clip.py
+++ b/sharktank/sharktank/models/clip/clip.py
@@ -427,14 +427,28 @@ class ClipTextTransformer(nn.Module):
         )
         last_hidden_state = self.final_layer_norm(last_hidden_state)
 
-        # We don't support this variant.
-        assert self.eos_token_id != 2
-
-        pooled_output = last_hidden_state[
-            torch.arange(last_hidden_state.shape[0]),
-            # We need to get the first position of `eos_token_id` value (`pad_token_ids` might equal to `eos_token_id`)
-            (input_ids == self.eos_token_id).int().argmax(dim=-1),
-        ]
+        if self.eos_token_id == 2:
+            # The `eos_token_id` was incorrect before PR
+            # https://github.com/huggingface/transformers/pull/24773
+            # Let's keep what have been done here.
+            # A CLIP model with such `eos_token_id` in the config can't work correctly with extra new tokens added
+            # ------------------------------------------------------------
+            # text_embeds.shape = [batch_size, sequence_length, transformer.width]
+            # take features from the eot embedding (eot_token is the highest number in each sequence)
+            # casting to torch.int for onnx compatibility: argmax doesn't support int64 inputs with opset 14
+            pooled_output = last_hidden_state[
+                torch.arange(last_hidden_state.shape[0]),
+                input_ids.argmax(dim=-1),
+            ]
+        else:
+            # The config gets updated `eos_token_id` from PR
+            # https://github.com/huggingface/transformers/pull/24773
+            # (so the use of exta new tokens is possible)
+            pooled_output = last_hidden_state[
+                torch.arange(last_hidden_state.shape[0]),
+                # We need to get the first position of `eos_token_id` value (`pad_token_ids` might equal to `eos_token_id`)
+                (input_ids == self.eos_token_id).int().argmax(dim=-1),
+            ]
 
         if not return_dict:
             return (last_hidden_state, pooled_output) + encoder_outputs[1:]
--- a/sharktank/sharktank/models/clip/clip.py
+++ b/sharktank/sharktank/models/clip/clip.py
@@ -5,6 +5,10 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+# CLIP text model
+# It is based on
+# https://github.com/huggingface/transformers/blob/v4.40-release/src/transformers/models/clip/modeling_clip.py
+
 from typing import Any, Optional, Tuple, Union
 import torch
 from torch import nn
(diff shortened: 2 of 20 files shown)


sharktank/sharktank/layers/norm.py
```diff
@@ -39,3 +39,24 @@ def forward(self, x: torch.Tensor):
...
+        weight_name: str = "weight",
+        bias_name: str = "bias",
+        eps: float = 1e-05,
+    ):
+        super().__init__(theta)
+        self.weight = self.theta_tensor(weight_name)
+        if bias_name in self.theta.keys:
+            self.bias = self.theta_tensor(bias_name)
+        else:
+            self.bias = None
```
rsuderman, line 58: Don't bother if-else ing. Just set `self.bias = None` prior to the if statement

sharktank/sharktank/models/clip/clip.py
```diff
@@ -0,0 +1,502 @@
...
+            theta=theta("token_embedding"), dtype=None
+        )
+        self.position_embedding = TokenEmbeddingLayer(
+            theta=theta("position_embedding"), dtype=None
+        )
+
+        # position_ids (1, len position emb) is contiguous in memory and exported when serialized
+        self.register_buffer(
+            "position_ids",
+            torch.arange(config.max_position_embeddings).expand((1, -1)),
```
rsuderman, line 41: You should use `unsqueeze` instead of `expand`

sharktank/sharktank/models/clip/clip.py
```diff
@@ -0,0 +1,502 @@
...
+        super().__init__()
+        self.embed_dim = config.hidden_size
+        self.num_heads = config.num_attention_heads
+        self.head_dim = self.embed_dim // self.num_heads
+        if self.head_dim * self.num_heads != self.embed_dim:
+            raise ValueError(
+                f"embed_dim must be divisible by num_heads (got `embed_dim`: {self.embed_dim} and `num_heads`:"
+                f" {self.num_heads})."
+            )
+        self.scale = self.head_dim**-0.5
```
rsuderman, line 80: Avoid using `**` for inverse powers. Its better to do `1.0 / math.sqrt(self.head_dim)`. The numerical precision on `pow` operators is usually significantly worse.

sharktank/sharktank/models/clip/clip.py
```diff
@@ -0,0 +1,502 @@
...
+        attn_weights = ops.matmul(query_states, key_states.transpose(1, 2))
+
+        if attn_weights.size() != (bsz * self.num_heads, tgt_len, src_len):
+            raise ValueError(
+                f"Attention weights should be of size {(bsz * self.num_heads, tgt_len, src_len)}, but is"
+                f" {attn_weights.size()}"
+            )
+
+        # apply the causal_attention_mask first
+        if causal_attention_mask is not None:
```
rsuderman, line 125: Why do you separate the causal attention mask from the regular attention mask? They should just occur together. Even the causal attention mask should really just be a bool.

sharktank/sharktank/models/clip/clip.py
```diff
@@ -0,0 +1,502 @@
...
+                raise ValueError(
+                    f"Attention mask should be of size {(bsz, 1, tgt_len, src_len)}, but is {attention_mask.size()}"
+                )
+            attn_weights = (
+                attn_weights.view(bsz, self.num_heads, tgt_len, src_len)
+                + attention_mask
+            )
+            attn_weights = attn_weights.view(bsz * self.num_heads, tgt_len, src_len)
+
+        attn_weights = ops.softmax(attn_weights, dim=-1)
```
rsuderman, line 148: Rather than decomposing we should use the `ops.scaled_dot_product_attention` operation. Attention is attention, so we should avoid replicating the decomposed version everywhere.

sharktank/sharktank/models/clip/clip.py
```diff
@@ -0,0 +1,502 @@
...
+        output_hidden_states = (
+            output_hidden_states
+            if output_hidden_states is not None
+            else self.config.output_hidden_states
+        )
+        return_dict = (
+            return_dict if return_dict is not None else self.config.use_return_dict
+        )
+
+        encoder_states = () if output_hidden_states else None
```
rsuderman, line 316: This logic of empty tuple vs `none` and adding feels unclear what it actually is attempting to do. Relying on `none + tuple` feels weird

sharktank/sharktank/types/tensors.py
```diff
@@ -378,6 +378,12 @@ def reshape(self, *args: Union[List[List[int]], List[int]]) -> "AnyTensor":
             shape = args[0]
         return reshape(self, shape)
 
+    def size(self, dim: Optional[int] = None) -> tuple[int]:
+        if dim is None:
+            return tuple(self.shape)
+        else:
```
rsuderman, line 384: No `else` condition required. Just include the return when the condition is not taken.

rsuderman (changes requested):

sharktank/sharktank/models/clip/clip.py
sogartar: I forgot to mention in the PR description that this is the initial port of the model and I did not attempt to optimize anything. The main goal was to put it under test. I would rather do these modifications later as they would be tracked in different commits and will be clear what changed compared to the original.

sharktank/sharktank/layers/norm.py
sogartar: Done.

sharktank/sharktank/types/tensors.py
sogartar: Done.

sharktank/sharktank/models/clip/clip.py
```diff
@@ -0,0 +1,502 @@
...
+        if inputs_embeds is None:
+            inputs_embeds = self.token_embedding(input_ids)
+
+        position_embeddings = self.position_embedding(position_ids)
+        embeddings = inputs_embeds + position_embeddings
+
+        return embeddings
+
+
+class ClipAttention(BaseLayer):
```
IanNod, line 67: At a quick glance is seems much of this is reused from the punet Attention (outside the decomposed SDPA which we should try avoiding) from here https://github.com/nod-ai/shark-ai/blob/5f98e7321564eeec6d0901250699938c7307ef65/sharktank/sharktank/models/punet/layers.py#L390. Can we just extend and reuse that implementation?

sharktank/sharktank/models/clip/clip.py
```diff
@@ -0,0 +1,502 @@
...
+            return_dict=return_dict,
+        )
+
+        last_hidden_state = (
+            encoder_outputs["last_hidden_state"] if return_dict else encoder_outputs[0]
+        )
+        last_hidden_state = self.final_layer_norm(last_hidden_state)
+
+        if self.eos_token_id == 2:
+            # The `eos_token_id` was incorrect before PR #24773: Let's keep what have been done here.
```
IanNod, line 427: This PR I assume is reference to diffusers? Confusing to have in our repo

sharktank/tests/models/clip/clip_test.py
```diff
@@ -0,0 +1,449 @@
...
+@pytest.mark.usefixtures("path_prefix")
+class ClipExportTest(TempDirTestBase):
+    def setUp(self):
+        super().setUp()
+        if self.path_prefix is None:
+            self.path_prefix = f"{self._temp_dir}/"
+
+    @with_clip_data
+    def testSmokeExportLargeF32FromHuggingFace(self):
+        repo_id = "openai/clip-vit-large-patch14"
```
IanNod, line 60: We might want to avoid downloading full models for developers running locally. Should we implement a toy model to accommodate? Thoughts @rsuderman?

sharktank/sharktank/models/clip/clip.py
KyleHerndon: If you're directly referencing the huggingface code for implementation, do you want to link it in a comment?

sharktank/tests/models/clip/clip_test.py
sogartar: Because it downloads data I marked it with `with_clip_data`.

sharktank/sharktank/models/clip/clip.py
sogartar: I added a comment at the top of the file.

sharktank/sharktank/models/clip/clip.py
sogartar: I removed this EOS variant.

sharktank/sharktank/models/clip/clip.py
sogartar: I would like to iterate on improving code reuse and performance in subsequent PRs.

sharktank/sharktank/models/clip/clip.py
sogartar: We actually go through this case in the actual model so I put it back in, but changed the PR reference to a full link.

KyleHerndon (approved):

sharktank/sharktank/models/clip/clip.py
sogartar: I will address this in a later PR.

sharktank/sharktank/models/clip/clip.py
sogartar: This is coming from the original model in Hugging Face. I will address this in a later PR.

sharktank/sharktank/models/clip/clip.py
sogartar: This is coming from the original model in Hugging Face. I will address this in a later PR as the whole Layer should get merged with the one from Punet.

sharktank/sharktank/models/clip/clip.py
sogartar: I would probably substitute this mechanism with tensor tracing in the future.

PR #632, opened 2024-12-02
TriePagedAttentionCache

renxida:
feat: Add TriePagedAttentionCache with initial implementation

Added TriePagedAttentionCache as an optional prefix sharing algorithm, selectable via:
`config["paged_kv_cache"]["prefix_sharing_algorithm"] = "trie"`

Current Status:
- Basic implementation and unit tests complete
- Integration test cases for both Base and Trie implementations, with trie implementation xfailed due to pending cache allocation improvements
- BasePagedAttentionCache remains the default

Next Steps:
To achieve full functionality, we need to support cache re-allocations to extend the associated tokens & pages.

PR #645, opened 2024-12-04
Fix/skip trie_attention_cache_test on Windows.

ScottTodd:
Context: https://github.com/nod-ai/shark-ai/pull/632#discussion_r1869981637

### DIFF
--- a/shortfin/python/shortfin_apps/llm/components/kvcache/trie_attention_cache.py
+++ b/shortfin/python/shortfin_apps/llm/components/kvcache/trie_attention_cache.py
@@ -90,6 +90,10 @@ class TrieNode:
         """Nodes are equal only if they are the same object."""
         return self is other
 
+    def __lt__(self, other):
+        """Sort nodes by their memory address."""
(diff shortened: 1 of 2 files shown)


shortfin/python/shortfin_apps/llm/components/kvcache/trie_attention_cache.py
```diff
@@ -90,6 +90,10 @@ def __eq__(self, other: object) -> bool:
         """Nodes are equal only if they are the same object."""
         return self is other
 
+    def __lt__(self, other):
+        """Sort nodes by their memory address."""
+        return id(self) < id(other)
```
ScottTodd, line 95: If the tests passed on Linux without this and they still fail on Windows with it... maybe this isn't the right implementation 🤔 

Could leave this off and just keep the test skipped.

ScottTodd:
Ping?

renxida (approved):
Sry! missed this. LGTM, merging.

PR #649, opened 2024-12-05
Use correct args for input len 2048 llama test

aviator19941:
Uses args for input len 2048 llama test instead of args for input len 128 llama test.


saienduri (approved):
Remove `pull_request` before merging :)

PR #653, opened 2024-12-06
Adds sdxl's VAE decoder implementation

IanNod:


sharktank/sharktank/models/vae/layers.py
```diff
@@ -0,0 +1,267 @@
...
+                )
+            else:
+                attentions.append(None)
+
+        if resnet_time_scale_shift == "spatial":
+            # TODO
+            raise AssertionError(
+                f"ResnetBlock2D spatial time scale shift not yet implemented"
+            )
+        else:
```
KyleHerndon, line 242: same here

sharktank/sharktank/models/vae/tools/run_vae.py
```diff
@@ -0,0 +1,137 @@
...
+
+    mdl = VaeDecoderModel.from_dataset(ds)
+
+    # Run a step for debugging.
+    if args.inputs:
+        inputs = load_inputs(args.inputs, dtype=dtype, device=device, bs=args.bs)
+    else:
+        inputs = get_random_inputs(dtype=dtype, device=device, bs=args.bs)
+
+    if args.export:
```
KyleHerndon, line 105: When this branch is reached, it looks like VAE is not run, only exported. I would not expect such a thing of a file named `run_vae.py`. Could you at least add a TODO to move the export code to another file or put a print statement "VAE exported. Skipping execution" or something like that?

sharktank/sharktank/types/tensors.py
```diff
@@ -393,6 +393,11 @@ def unsqueeze(self, dim: int) -> "AnyTensor":
 
         return unsqueeze(self, dim)
 
+    def squeeze(self, dim: Optional[int] = None) -> "AnyTensor":
```
KyleHerndon, line 396: Likewise lets keep this alphabetized.

sharktank/sharktank/models/vae/tools/diffuser_ref.py
```diff
@@ -0,0 +1,55 @@
...
+                    custom_vae,
+                )
+            except:
+                # some larger repo with vae subfolder
+                self.vae = AutoencoderKL.from_pretrained(
+                    custom_vae,
+                    subfolder="vae",
+                )
+
+    def decode(self, inp):
```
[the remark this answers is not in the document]
IanNod: This is the diffusers reference code so is not using our HParams at all. They also do not do the scaling inside vae decode but in their pipeline as input here https://github.com/huggingface/diffusers/blob/18f9b990884883533491fc87f303e7305dc27d75/src/diffusers/pipelines/stable_diffusion/pipeline_stable_diffusion.py#L591. I added it to our implementation as forcing shortfin to handle it in the pipeline would introduce extra host/device roundtrips which would impact performance.

I will add comments to clarify here

sharktank/sharktank/models/vae/tools/sample_data.py
```diff
@@ -0,0 +1,18 @@
...
+
+"""Various utilities for deriving stable sample data for the model."""
+
+from pathlib import Path
+
+import torch
+
+
+def get_random_inputs(dtype, device, bs: int = 2):
+    torch.random.manual_seed(42)
```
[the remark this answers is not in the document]
IanNod: good catch. Left over from some debugging awhile ago. Will remove

sharktank/sharktank/models/vae/layers.py
```diff
@@ -0,0 +1,267 @@
...
+        attn_groups: Optional[int] = None,
+    ):
+        super().__init__(theta)
+        attentions = []
+
+        resnet_groups = resnet_groups if resnet_time_scale_shift == "default" else None
+
+        # there is always at least one resnet
+        if resnet_time_scale_shift == "spatial":
+            # TODO
```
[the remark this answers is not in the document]
IanNod: Each layer always has a resnet and specifically either ResnetBlockCondNorm2d for spatial time shift or ResnetBlock2D otherwise. No else needed

sharktank/sharktank/models/vae/tools/diffuser_ref.py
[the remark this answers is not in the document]
monorimet: You can probably replace `img = 1 / 0.13025 * inp` with `img = img / self.vae.config.scaling_factor` [source](https://github.com/huggingface/diffusers/blob/main/src/diffusers/pipelines/stable_diffusion_xl/pipeline_stable_diffusion_xl.py#L1289C27-L1289C67)

this is done differently in flux/sd3 where we have a shift factor:
```
img = img / self.scale_factor + self.shift_factor
```

sharktank/sharktank/models/vae/layers.py
```diff
@@ -0,0 +1,261 @@
+# Copyright 2024 Advanced Micro Devices, Inc.
```
KyleHerndon, line 1: Just realized. This should probably be broken up and put in `sharktank/sharktank/layers/`

IanNod: Yeah, it should. I kept it separate following punet layers and was planning to clean up both in a follow up PR

PR #655, opened 2024-12-06
Concurrent CPU Integration Tests + Reuse Model Artifacts

stbaione:
# Description

There are two things that get added in this PR:
1. Reuse model artifacts when multiple tests in the same module request export/compilation of the same artifacts
2. Add concurrency tests for `2, 4, and 8` requests at the same time.

# Reusing model aritfacts

Currently, our `cpu_llm_server_integration_tests` generate new model artifacts for each tests, even when they are requesting the exact same artifacts. This causes the tests to take much longer to run than they should, and makes it harder to add more tests without drastically increasing overall test time.

We add a static `MODEL_DIR_CACHE`, which is just a hashmap that stores `{ request.params_hash: temporary_dir }`. If a test requests the same artifacts as a previous test, we reuse the already existing artifacts, instead of generating new ones.

# Adding concurrency tests

We recenly found a bug in concurrency with the Shortfin LLM Server. When sending multiple requests at the same time, we end up with responses that have incorrect tokens. 

This adds basic concurrent integration tests for 2, 4, and 8 requests sent in parallel. Currently, they are xfailed, but we will be able to use these to validate our fix, when we get there, and ensure that we don't have a regression in concurrency in the future. Will extend the `periodic SGLang Integration tests` to further test concurrency on GPU, with more complex prompts, but for a PR triggered test, this should serve as a good guard.

### DIFF
--- a/app_tests/integration_tests/llm/shortfin/conftest.py
+++ b/app_tests/integration_tests/llm/shortfin/conftest.py
@@ -4,12 +4,10 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+import hashlib
 import json
 import logging
-import os
-from pathlib import Path
 import pytest
-import shutil
 
 pytest.importorskip("transformers")
 from ..utils import (
@@ -24,6 +22,8 @@ from ..utils import (
 )
 
 logger = logging.getLogger(__name__)
+
+MODEL_DIR_CACHE = {}
 
 
 @pytest.fixture(scope="module")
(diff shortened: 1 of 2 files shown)


app_tests/integration_tests/llm/shortfin/conftest.py
```diff
@@ -46,58 +46,73 @@ def model_test_dir(request, tmp_path_factory):
         "Preparing model artifacts..." + start_log_group("Preparing model artifacts")
     )
 
+    param_key = hashlib.md5(str(request.param).encode()).hexdigest()
```
renxida, line 49: nice fast hash : )

renxida (approved):
lgtm. once tested feel free to merge!

PR #657, opened 2024-12-06
Switch workflows to use new nightly iree-turbine packages.

ScottTodd:
This simplification will help with https://github.com/nod-ai/shark-ai/issues/584.

Nightly releases of iree-turbine are now being built thanks to https://github.com/iree-org/iree-turbine/pull/314 and published at the same index as the other IREE packages thanks to https://github.com/iree-org/iree/pull/19391.

### DIFF
--- a/.github/workflows/ci-llama-large-tests.yaml
+++ b/.github/workflows/ci-llama-large-tests.yaml
@@ -61,7 +61,7 @@ jobs:
 
           # Install nightly IREE packages.
           # We could also pin to a known working or stable version.
-          pip install -f https://iree.dev/pip-release-links.html --pre \
+          pip install -f https://iree.dev/pip-release-links.html --pre --upgrade \
             iree-base-compiler \
             iree-base-runtime \
             iree-turbine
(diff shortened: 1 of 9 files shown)


.github/workflows/ci-llama-large-tests.yaml
```diff
@@ -59,14 +59,12 @@ jobs:
...
-
-          # Test with nightly releases, not what iree-turbine uses.
-          pip install -f https://iree.dev/pip-release-links.html --upgrade --pre \
+          # Install nightly IREE packages.
+          # We could also pin to a known working or stable version.
+          pip install -f https://iree.dev/pip-release-links.html --pre --upgrade \
             iree-base-compiler \
-            iree-base-runtime
+            iree-base-runtime \
+            iree-turbine
```
archana-ramalingam, line 67: QQ: Doesn't installing `iree-turbine` get us the latest/relevant version of compiler and runtime. Do we have to explicitly install/update them too?

marbre (approved):
Nice! I also prefer being explicit and not letting `iree-turbine` pulling in the IREE base packages.

stbaione (approved):

PR #663, opened 2024-12-09
A basic working version of the flux model

KyleHerndon:
This version of the flux model should work, as it directly modifies the reference implementation, but could really use some refactoring, especially to reduce code duplication


sogartar (approved):
Needs some refactoring of tests to reduce the model size. They don't cause memory exhaustion now, but still they can be much smaller.

PR #664, opened 2024-12-09
Add exporting and numerics verification for CLIP Large text model with IREE

sogartar:
Add exporting to MLIR and IRRE parameters. We don't make the context length dynamic since the maximum is only 77 anyway, so the token sequences are padded to 77. We could explore later making this dynamic.

This adds comparison of IREE execution of float32, bfloat16 model variants against float32 torch eager. For bfloat16 results are close up to 1.43e-2 using cosine similarity.

Toy-sized model comparison for float32 and bfloat16 is also provided.


sharktank/sharktank/models/clip/clip.py
```diff
@@ -487,6 +487,19 @@ def get_input_embeddings(self) -> nn.Module:
     def set_input_embeddings(self, value):
         self.text_model.embeddings.token_embedding = value
 
+    def sample_inputs(self, batch_size: int) -> OrderedDict[str, AnyTensor]:
+        return OrderedDict(
+            [
+                (
+                    "input_ids",
```
KyleHerndon, line 494: Is empty better than some random values here? Probably doesn't matter either way, but all zeros causes bigger differences on some models iirc.

KyleHerndon (approved):

sharktank/tests/models/clip/clip_test.py
```diff
@@ -48,29 +72,278 @@
 with_clip_data = pytest.mark.skipif("not config.getoption('with_clip_data')")
 
 
-@pytest.mark.usefixtures("path_prefix")
-class ClipExportTest(TempDirTestBase):
+def assert_last_hidden_states_close(
```
KyleHerndon, line 75: Can we put this in a `utils` file?

sogartar: I moved it.

PR #666, opened 2024-12-09
Fix version_suffix for nighly shark-ai packages in build_packages.yml.

ScottTodd:
Tested here: https://github.com/ScottTodd/shark-ai/actions/runs/12244417956 and confirmed that the `METADATA` embedded within `shark_ai.*.whl` has `Requires-Dist: shortfin==3.1.0rc20241209` as expected, compared to the baseline of `Requires-Dist: shortfin==3.1.0`.

I then also ran
```bash
pip install --pre --find-links https://github.com/ScottTodd/shark-ai/releases/expanded_assets/dev-wheels shark-ai
# Looking in links: https://github.com/ScottTodd/shark-ai/releases/expanded_assets/dev-wheels
# Collecting shark-ai
#   Downloading https://github.com/ScottTodd/shark-ai/releases/download/dev-wheels/shark_ai-3.1.0rc20241209-py3-none-any.whl (1.4 kB)
# Collecting iree-base-compiler (from shark-ai)
#   Using cached iree_base_compiler-3.0.0-cp311-cp311-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl.metadata (1.0 kB)
# Collecting iree-base-runtime (from shark-ai)
#   Using cached iree_base_runtime-3.0.0-cp311-cp311-manylinux_2_28_x86_64.whl.metadata (1.0 kB)
# Collecting iree-turbine (from shark-ai)
#   Using cached iree_turbine-3.0.0-py3-none-any.whl.metadata (5.8 kB)
# Collecting shortfin==3.1.0rc20241209 (from shark-ai)
#   Downloading https://github.com/ScottTodd/shark-ai/releases/download/dev-wheels/shortfin-3.1.0rc20241209-cp311-cp311-manylinux_2_28_x86_64.whl (2.7 MB)

pip freeze | grep sh
# shark-ai==3.1.0rc20241209
# shortfin==3.1.0rc20241209
```

Compared to the baseline of
```bash
pip install --pre --find-links https://github.com/nod-ai/shark-ai/releases/expanded_assets/dev-wheels shark-ai
# Looking in links: https://github.com/nod-ai/shark-ai/releases/expanded_assets/dev-wheels
# Collecting shark-ai
#   Downloading https://github.com/nod-ai/shark-ai/releases/download/dev-wheels/shark_ai-3.1.0rc20241209-py3-none-any.whl (1.4 kB)
# Collecting iree-base-compiler (from shark-ai)
[...]

PR #667, opened 2024-12-09
Add dev package support to build_packages.yml.

ScottTodd:
Progress on https://github.com/nod-ai/shark-ai/issues/584. ~~Depends on https://github.com/nod-ai/shark-ai/pull/666 (the first commit).~~

This is refactors the `build_packages.yml` workflow so it can be used via `workflow_call` as part of a "pkgci" setup, as an alternative to creating a new `pkgci_build_packages.yml` workflow as originally proposed in https://github.com/nod-ai/shark-ai/pull/589. This lets us reuse the same workflow for building stable, nightly, and dev packages, all across the same matrix of Python versions and operating systems. Package builds take about 2 minutes (wall time) across the full matrix, so we might as well build them all, instead of artificially constraining ourselves to a subset like only Linux on Python 3.11.

Triggers for the workflow are now this:

Trigger | Scenario | Build type(s)
-- | -- | --
`schedule` | Nightly pre-release build | `rc`
`workflow_dispatch` | Workflow testing, manual releasing | `rc` default, `stable` and `dev` possible
`workflow_call` | Pull request or push "pkgci" dev builds | `dev` default, `stable` and `rc` possible

With this workflow behavior:

Build type | Version suffix | Cache enabled? | Tracing enabled? | Pushes to release?
-- | -- | -- | -- | --
`stable` | None | No | Yes | No
`rc` | `rcYYYYMMDD` | No | Yes | Yes
`dev` | `.dev0+${{ github.sha }}` | Yes | No | No

Tested over at https://github.com/ScottTodd/shark-ai/actions/workflows/build_packages.yml. Example run: https://github.com/ScottTodd/shark-ai/actions/runs/12245900071 (warm cache)

### DIFF
--- a/.github/workflows/build_packages.yml
+++ b/.github/workflows/build_packages.yml
@@ -7,7 +7,21 @@
 name: Build packages
 
 on:
+  # Trigger from another workflow (typically to build dev packages and then test them)
+  workflow_call:
+    inputs:
+      build_type:
+        description: The type of build version to produce ("stable", "rc", or "dev")
+        type: string
+        default: "dev"
+  # Trigger manually (typically to test the workflow or manually build a release [candidate])
   workflow_dispatch:
+    inputs:
+      build_type:
+        description: The type of build version to produce ("stable", "rc", or "dev")
+        type: string
+        default: "rc"
+  # Trigger on a schedule to build nightly release candidates.
   schedule:
     # Runs at 11:00 AM UTC, which is 3:00 AM PST (UTC-8)
     - cron: '0 11 * * *'
@@ -16,9 +30,7 @@ permissions:
   contents: read
 
 jobs:
-  # Note: metadata generation could happen in a separate trigger/schedule
-  # workflow. For cross platform builds, it's useful to just generate the
-  # metadata on Linux and pass that to later jobs using artifacts.
+  # Generate metadata on Linux and pass to later jobs.
   setup_metadata:
     if: ${{ github.repository_owner == 'nod-ai' || github.event_name != 'schedule' }}
     runs-on: ubuntu-24.04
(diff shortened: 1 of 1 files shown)


.github/workflows/build_packages.yml
```diff
@@ -34,18 +43,35 @@ jobs:
...
+      - name: Compute rc version suffix
+        if: ${{ inputs.build_type == 'rc' || inputs.build_type == '' }}
+        run: |
+          version_suffix="$(printf 'rc%(%Y%m%d)T')"
+          echo "version_suffix=${version_suffix}" >> $GITHUB_ENV
+      - name: Compute dev version suffix
+        if: ${{ inputs.build_type == 'dev' }}
+        run: |
+          version_suffix=".dev0+${{ github.sha }}"
+          echo "version_suffix=${version_suffix}" >> $GITHUB_ENV
```
ScottTodd, line 61: We could also move this logic into `compute_local_version.py`, as it is in IREE: https://github.com/iree-org/iree/blob/ab3c9bb031f82f8481da7d867fab03db9180646a/build_tools/python_deploy/compute_local_version.py#L26-L30

We'd then need to pass the selection from the workflow inputs through to the script somehow:

```yml
      - name: Generate release candidate versions
        id: version_rc
        run: |
          python3 build_tools/python_deploy/compute_local_version.py --release-type ${{ inputs.build_type }} sharktank
          # extract version_suffix from the local version to forward to the job output?
          # that is currently passed to `write_requirements.py` down below as part of building the shark-ai package
```

```python
release_type = parser.add_mutually_exclusive_group()
release_type.add_argument("-stable", "--stable-release", action="store_true")  # default
release_type.add_argument("-rc", "--nightly-release", action="store_true")
release_type.add_argument("-dev", "--development-release", action="store_true")
release_type.add_argument("--custom-string", action="store", type=str)
release_type.add_argument("--build-type", action="store", type=str)

args = parser.parse_args()
if args.build_type == "dev":
  args.development_release = True
elif args.build_type == "stable
  ...
```

I generally prefer to keep workflow files simpler and put logic in scripts that can be run outside of the CI. Feeling lazy about this though, since it is already working as-is :P. Opinions?

.github/workflows/build_packages.yml
```diff
@@ -7,7 +7,21 @@
 name: Build packages
 
 on:
+  # Trigger from another workflow (typically to build dev packages and then test them)
+  workflow_call:
+    inputs:
+      build_type:
+        description: The type of build version to produce ("stable", "rc", or "dev")
+        type: string
+        default: "dev"
```
ScottTodd, line 16: > This is refactors the `build_packages.yml` workflow so it can be used via `workflow_call` as part of a "pkgci" setup, as an alternative to creating a new `pkgci_build_packages.yml` workflow as originally proposed in #589. This lets us reuse the same workflow for building stable, nightly, and dev packages, all across the same matrix of Python versions and operating systems. Package builds take about 2 minutes (wall time) across the full matrix, so we might as well build them all, instead of artificially constraining ourselves to a subset like only Linux on Python 3.11.

Sample workflow run with that setup:

https://github.com/ScottTodd/shark-ai/actions/runs/12246364049

image

.github/workflows/build_packages.yml
```diff
@@ -16,37 +30,51 @@ permissions:
...
+          version_suffix=".dev0+${{ github.sha }}"
+          echo "version_suffix=${version_suffix}" >> $GITHUB_ENV
+
+      - name: Generate release candidate versions
+        id: version_rc
+        run: |
+          echo "version_suffix=${version_suffix}" >> $GITHUB_OUTPUT
+          python3 build_tools/python_deploy/compute_local_version.py --version-suffix=${version_suffix} sharktank
+          python3 build_tools/python_deploy/compute_local_version.py --version-suffix=${version_suffix} shortfin
+          python3 build_tools/python_deploy/compute_common_version.py -rc --version-suffix=${version_suffix} --write-json
```
marbre, line 72: I think this works (as the version suffix is empty for stable releases and overwrites for dev releases) but we should refactor in a follow up (happy to take this) as it rather confusing that `-rc` is passed for all builds. I needed to look twice here to understand why it is probably fine.

.github/workflows/build_packages.yml
marbre: Well I would like to harmonize what we have here and what I've done for IREE in general. However, I am fine with going what you have right now (assuming that one of us fills an issue, see my review comment about refactoring to reduce confusion).

marbre (approved):

.github/workflows/build_packages.yml
ScottTodd: Oh, I missed the `-rc` in the middle there. Yeah, it still works but we should refactor :P

ScottTodd:
Rebased + applied suggestions.

.github/workflows/build_packages.yml
marbre: This is addressed with #674.

PR #668, opened 2024-12-09
Shortfin LLM Debug Ergonomics: one flag to dump them all

renxida:
Thanks @stbaione for originally writing a lot of the debug code in here. I organized  his debug logging & dumping into a separate file & made the output easier to deal with.


shortfin/python/shortfin_apps/llm/components/service.py
```diff
@@ -433,7 +433,36 @@ async def run(self):
                 fn,
                 "".join([f"\n  {i}: {ary.shape}" for i, ary in enumerate(args)]),
             )
-            # Invoke. Logits are of shape [bs, bsl, d].
+
+            # pre-invocation args dump
+            try:
```
stbaione, line 438: Kind of a nit, but I feel like a developer would expect `service.py` to handle checking the `SHORTFIN_DEBUG_LLM_SERVICE` var and invoking `pre_invocation_debug_dump` if True, instead of `debug_service.py` having control over that.

I think this may be a little cleaner if we made it a private `async def _pre_invocation_debug_dump` on `InferenceExecutorProcess`, initialized a constant `SHORTFIN_DEBUG_LLM_SERVICE` at the top of `service.py`, and then called:

```python
if SHORTFIN_DEBUG_LLM_SERVICE:
    await self._pre_invocation_debug_dump(...)
```

And moved the import & error handling logic to that private function, instead of `debug_service.py` reading the env variable and controlling whether or not it gets dumped with:

```python
if not SHORTFIN_DEBUG_LLM_SERVICE:
        return
```

Makes `run` a little less noisy too

ScottTodd: Yeah this is a bit strange, but I can see how injecting in this way is convenient.

This could also be configured via a new argument to the server, here: https://github.com/nod-ai/shark-ai/blob/main/shortfin/python/shortfin_apps/llm/server.py. That would be more discoverable than an environment variable, since it would appear in `--help` output.

shortfin/python/shortfin_apps/llm/components/debug_service.py
```diff
@@ -0,0 +1,219 @@
...
+
+# Get environment variable, default to False if not set
+SHORTFIN_DEBUG_LLM_SERVICE = os.getenv(
+    "SHORTFIN_DEBUG_LLM_SERVICE", "False"
+).lower() in ("true", "yes", "1", "y")
+if SHORTFIN_DEBUG_LLM_SERVICE:
+    logger.info("DEBUG_LLM_SERVICE=True")
+    dump_id = 0
+    boot_timestamp = datetime.now().isoformat()
+    DEBUG_DATA_DIR = Path.home() / "sfdebug"
```
ScottTodd, line 33: This could be configurable. We might want to use a hidden directory (prefixed with `.`) or namespace similar outputs across this project to a `shark-ai` or `shortfin` (or this `sfdebug`) folder.

shortfin/python/shortfin_apps/llm/components/service.py
renxida: the intention is to make this attach to service.py at as few points as possible, so that if we want to make a more unified debug interface it's easy to do so. I do agree that this is a little cumbersome though.

Currently working on some simplifications, where instead of making the debug dump function a part of InferenceExecProcess, we pass self and all local variables to the dump function.

shortfin/python/shortfin_apps/llm/components/debug_service.py
renxida: I prefer using an env var & using a hardcoded debug data dir for simplicity, but i like the hidden dir idea. Might do ~/.shortfin/debug

stbaione (approved):

shortfin/python/shortfin_apps/llm/components/service.py
```diff
@@ -438,7 +446,12 @@ async def run(self):
...
                 "".join([f"\n  {i}: {ary.shape}" for i, ary in enumerate(args)]),
             )
-            # Invoke. Logits are of shape [bs, bsl, d].
+
+            # pre-invocation args dump
+            if SHORTFIN_DEBUG_LLM_SERVICE:
+                await pre_invocation_debug_dump(executor=self, local_vars=locals())
+
+            # invoke VMFB
             (logits,) = await fn(*args, fiber=self.fiber)
```
ScottTodd, line 455: Keep the full `# Invoke. Logits are of shape [bs, bsl, d].` comment?

shortfin/python/shortfin_apps/llm/components/debug_service.py
```diff
@@ -0,0 +1,221 @@
...
+    dump_id = 0
+    boot_timestamp = datetime.now().isoformat()
+    DEBUG_DATA_DIR = Path.home() / ".shortfin/debug/"
+    DUMP_DIR_THIS_SESSION = (
+        DEBUG_DATA_DIR / "llm_service_invocation_dumps" / "{boot_timestamp}"
+    )
+    DUMP_DIR_THIS_SESSION.mkdir(parents=True, exist_ok=False)
+    logger.info(
+        f"[debug_service.py] Please find debug dumps for service.py in {DUMP_DIR_THIS_SESSION}"
+    )
```
ScottTodd, line 40: Can you move this to an `init()` function so it doesn't always run at import time? This debug file could provide a class instead of a loose function that uses global variables. Then there would be no need to duplicate this `os.getenv` code across both files or add the `Importing this should do nothing` comment

```python
# debug_service.py

class ShortfinLLMDebugService:
  def __init__(self):
    dump_id = 0
    dump_dir = ""
    dump_dir.mkdir(...)

  async def pre_invocation_debug_dump(...):
    is_decode = ...

# service.py

class GenerateService:
    def __init__(self, ...):
      ...

      SHORTFIN_DEBUG_LLM_SERVICE = os.getenv(
          "SHORTFIN_DEBUG_LLM_SERVICE", "False"
      ).lower() in ("true", "yes", "1", "y")
      if SHORTFIN_DEBUG_LLM_SERVICE:
        from .debug_service import ShortfinLLMDebugService
        llm_debug_service = ShortfinLLMDebugService

...

    async def run(self):
      ...
      if self.llm_debug_service:
        llm_debug_service.pre_invocation_debug_dump(...)
```

PR #672, opened 2024-12-10
Add CLI script exporting CLIP Toy model IREE test data

sogartar:
This is required to have an easy way of exporting test data that will be
used in IREE to guard against regressions.

E.g.
```
python -m sharktank.models.clip.export_toy_text_model_iree_test_data \
  --output-dir=clip_toy_text_model
```

Refactor some of the existing tests to reuse the new export logic.

### DIFF
--- a/sharktank/sharktank/models/clip/export.py
+++ b/sharktank/sharktank/models/clip/export.py
@@ -11,8 +11,8 @@ from transformers.models.clip.modeling_clip import (
     CLIPEncoderLayer as HfCLIPEncoderLayer,
     CLIPEncoder as HfCLIPEncoder,
 )
-from os import PathLike
 import torch
+from os import PathLike
 
 from ...types.theta import Theta, Dataset, torch_module_to_theta
 from ...layers.configs import ClipTextConfig
@@ -50,9 +50,14 @@ def clip_text_model_to_dataset(model: ClipTextModel) -> Dataset:
     return Dataset(properties=model.config.to_properties(), root_theta=model.theta)
 
 
+def export_clip_text_model_iree_parameters(model: ClipTextModel, output_path: PathLike):
+    dataset = clip_text_model_to_dataset(model)
+    dataset.save(output_path)
+
+
 def export_clip_text_model_dataset_from_hugging_face(
-    model_or_name_or_path: Union[str, PathLike, transformers.CLIPTextModel],
-    output_path: Union[str, PathLike],
+    model_or_name_or_path: Union[PathLike, transformers.CLIPTextModel],
+    output_path: PathLike,
     dtype: Optional[torch.dtype] = None,
 ):
     if isinstance(model_or_name_or_path, transformers.CLIPTextModel):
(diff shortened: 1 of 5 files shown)


sogartar:
This PR requires that #664 is merged first.

sharktank/sharktank/models/clip/export.py
```diff
@@ -67,7 +72,7 @@ def export_clip_text_model_dataset_from_hugging_face(
 
 
 def export_clip_text_model_mlir(
-    model: Union[ClipTextModel, PathLike],
+    model: Union[ClipTextModel, AnyPath],
```
rsuderman, line 75: Leave this as `PathLike` and remove the `AnyPath`. Users should just make sure to convert the string before invoking. (Same as above)

sharktank/sharktank/models/clip/testing.py
```diff
@@ -4,14 +4,161 @@
...
+            strict=True,
+        )
+    ):
+        export_clip_text_model_iree_test_data(
+            reference_model=reference_model,
+            target_dtype=target_dtype,
+            input_ids=input_ids,
+            target_iree_parameters_output_path=target_iree_parameters_output_path,
+            target_mlir_output_path=target_mlir_output_path,
+            input_ids_output_path=input_ids_output_path if i == 0 else None,
```
rsuderman, line 113: Split this across lines or put in an `if` conditional. The `,` between make the behavior extremely ambiguous.

sharktank/sharktank/models/clip/testing.py
```diff
@@ -4,14 +4,161 @@
...
+            expected_last_hidden_state_output_path=expected_last_hidden_state_output_path
```
rsuderman, line 114: Same comment as above. This looks like it should just be in an `if` condition rather than a oneliner

sharktank/tests/models/clip/clip_test.py
```diff
@@ -96,6 +97,11 @@ def testSmokeExportLargeF32FromHuggingFace(self):
             huggingface_repo_id, output_path
         )
 
+    def testSmokeExportToyIreeTestData(self):
+        from sharktank.models.clip.export_toy_text_model_iree_test_data import main
+
+        main([f"--output-path-prefix={self.path_prefix}clip_toy_text_model"])
```
rsuderman, line 103: Use pythons path library to operate on this instead of string manipulations.

sharktank/tests/models/clip/clip_test.py
```diff
@@ -147,30 +153,24 @@ def runTestCompareIreeAgainstTorchEagerWithInputTokens(
...
-
         mlir_path = f"{target_model_path_prefix}.mlir"
-        if not self.caching or not os.path.exists(mlir_path):
-            export_clip_text_model_mlir(
-                parameters_path, batch_sizes=[batch_size], mlir_output_path=mlir_path
+
+        if (
+            not self.caching
+            or not os.path.exists(mlir_path)
+            or not os.path.exists(parameters_path)
```
rsuderman, line 164: This conditional feels weird / wrong. Is it checking for defaults? Also it feels particularly weird for a test. Reconsider why this is needed.

rsuderman (changes requested):

sharktank/sharktank/models/clip/export.py
sogartar: I removed AnyPath.

sharktank/sharktank/models/clip/testing.py
sogartar: Done.

sharktank/tests/models/clip/clip_test.py
sogartar: Done. I changed it to use a directory as a prefix.

sharktank/tests/models/clip/clip_test.py
sogartar: I removed the caching. It is useful to speedup iteration, but it does add clutter.

rsuderman (approved):

sogartar:
Here is an IREE [fix](https://github.com/iree-org/iree/pull/19582) for that failing data-dependent tests.

PR #673, opened 2024-12-11
Pin OS to specific versions

marbre:
With this, runner images are pinned to specific versions instead of referring to `latest`. This allows to have a controlled upgrade path instead instead of just waiting for when a change in the reference affects the workflows.

### DIFF
--- a/.github/workflows/ci-sharktank.yml
+++ b/.github/workflows/ci-sharktank.yml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       matrix:
         version: [3.11]
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-24.04, windows-2022]
       fail-fast: false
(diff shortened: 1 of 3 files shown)


marbre:
`Unit Tests and Type Checking (3.11, ubuntu-latest)` cannot report its status as it now runs on `ubuntu-24.04` instead of `ubuntu-latest`.

ScottTodd:
> `Unit Tests and Type Checking (3.11, ubuntu-latest)` cannot report its status as it now runs on `ubuntu-24.04` instead of `ubuntu-latest`.

Unavoidable with how GitHub handles required checks. You can either merge through it then swap which check is required, or remove the current check first. Grouping workflows via a "summary" step like how we have iree (and now iree-turbine) workflows set up does let you provide a reasonably stable check name to set as required.

PR #677, opened 2024-12-11
loop over cache_partitions to enable fusion

dan-garvey:


rsuderman (approved):
LGTM assuming it passes CI

PR #676, opened 2024-12-11
Update user docs for running `llm server` + upgrade `gguf` to `0.11.0`

stbaione:
# Description

Did a pass through and made updates + fixes to the user docs for `e2e_llama8b_mi300x.md`.

1. Update install instructions for `shark-ai`
2. Update nightly install instructions for `shortfin` and `sharktank`
3. Update paths for model artifacts to ensure they work with `llama3.1-8b-fp16-instruct`
4. Remove steps to `write edited config`. No longer needed after #487 

Added back `sentencepiece` as a requirement for `sharktank`. Not having it caused `export_paged_llm_v1` to break when installing nightly:

```text
ModuleNotFoundError: No module named 'sentencepiece'
```

This was obfuscated when building from source, because `shortfin` includes `sentencepiece` in `requirements-tests.txt`.

PR #687, opened 2024-12-12
[sharktank] Depend on `sentencepiece`

marbre:
While `gguf`'s `pyproject.toml` lists `sentencepiece` as a dependency (https://github.com/ggerganov/llama.cpp/blob/a07c32ea54850c989f0ef6989da5b955b77b7172/pyproject.toml#L20), the package does not get pulled in when installing gguf via `uv pip` or `pip`. However, without the package the sanity test `python -c "from sharktank import ops; print('Sanity check passed')"`, documented in https://github.com/nod-ai/shark-ai/blob/main/docs/nightly_releases.md#quickstart---sharktank, fails.

### DIFF
--- a/sharktank/requirements.txt
+++ b/sharktank/requirements.txt
@@ -5,7 +5,7 @@ gguf==0.10.0
 numpy<2.0
 
 # Needed for newer gguf versions (TODO: remove when gguf package includes this)
-# sentencepiece>=0.1.98,<=0.2.0
+sentencepiece>=0.1.98,<=0.2.0
 
 # Model deps.
 huggingface-hub==0.22.2


sharktank/requirements.txt
```diff
@@ -5,7 +5,7 @@ gguf==0.10.0
 numpy<2.0
 
 # Needed for newer gguf versions (TODO: remove when gguf package includes this)
-# sentencepiece>=0.1.98,<=0.2.0
+sentencepiece>=0.1.98,<=0.2.0
```
ScottTodd, line 8: See also https://github.com/nod-ai/shark-ai/pull/676#discussion_r1880805172

stbaione: See response https://github.com/nod-ai/shark-ai/pull/676#discussion_r18826010700. v0.11.0 released an hour ago, which has `sentecepiece` included as a dep

stbaione: Upgrading `gguf` to `0.11.0` and removing `sentencepiece` dep in #676

marbre: Ah, I saw the tag for `0.11.0` but there was no whl on PyPI and the `project.toml` looked identical. Thanks for updating!

PR #690, opened 2024-12-12
Enable flash attention by default

rsuderman:


aviator19941 (approved):

aviator19941:
@rsuderman Can you also add the block_seq_stride flag defaulted to 32 for `export_paged_llm_v1.py` like in https://github.com/rsuderman/sharktank/tree/configure_blocking?

Issue #357, opened 2024-10-29
Organize github actions workflows and expand build/test matrices


ScottTodd:
The files in https://github.com/nod-ai/SHARK-Platform/tree/main/.github/workflows are proliferating with overlapping workflows while missing some coverage we do care about.

GitHub has tools to manage this scaling:
* https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow
* https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions

We generally want to build/test/release across:
* Python versions: 3.10, 3.11, 3.12, 3.13, 3.13t
* Operating systems: Linux, Windows (macOS too?)
* Accelerators/devices: CPU, GPU (MI300, others)
* Other configurations: address sanitizer (ASan), instrumented (Tracy), bundled or bring-your-own deps like IREE, etc.
* (Stretch) architectures: x86, aarch64

Each subproject (sharktank, shortfin, tuner) is currently self-contained, so we can keep at least one workflow per subproject. Common tasks like running pre-commit and mypy type checking could stay in https://github.com/nod-ai/SHARK-Platform/blob/main/.github/workflows/pre-commit.yaml (note that iree-turbine puts MyPy Type Checking as the last step in multiple build/test workflows: https://github.com/iree-org/iree-turbine/blob/2b45c0fdec21f69b9cc088ec9852e98f5219c37c/.github/workflows/ci.yaml#L65-L68)

ScottTodd:
As part of this we could have these workflows test both nightly IREE and iree-turbine packages _and_ pinned versions. I think blocking CI should use pinned versions, while the currently mode of testing with the latest nightly/source packages can drop to being allowed to fail.

PR #693, opened 2024-12-13
[sharktank] Extend to test with Python 3.12

marbre:
Progress on https://github.com/nod-ai/shark-ai/issues/357

Extends to test with Python 3.12 and newer `torch` versions on Linux.

For `torch.compile` support on Python 3.12, torch ge 2.4.0 is required. Testing with torch 2.3.0 is therefore skipped when running Python 3.12. To limit the number of jobs, tests are currently only expanded for Linux, whereas Windows (and unpinning NumPy) should be handled in a follow up.

The error logged for `testExportNondecomposed` when failing with torch 2.5.1 slows down the CI by more than 20 minutes, therefore the test is now skipped instead of marked xfail.

### DIFF
--- a/.github/workflows/ci-sharktank.yml
+++ b/.github/workflows/ci-sharktank.yml
@@ -27,16 +27,16 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.11", "3.12"]
-        torch-version: ["2.3.0+cpu", "2.4.1+cpu", "2.5.1+cpu"]
+        torch-version: ["2.3.0", "2.4.1", "2.5.1"]
         os: [ubuntu-24.04]
         include:
           - os: windows-2022
             python-version: "3.11"
-            torch-version: "2.3.0+cpu"
+            torch-version: "2.3.0"
         exclude:
           - python-version: "3.12"
              # `torch.compile` requires torch>=2.4.0 for Python 3.12+
-            torch-version: "2.3.0+cpu"
+            torch-version: "2.3.0"
       fail-fast: false
     runs-on: ${{matrix.os}}
     defaults:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-sharktank.yml
```diff
@@ -26,8 +26,17 @@ jobs:
     name: "Unit Tests and Type Checking"
```
ScottTodd, line 26: Maybe rename to just "Unit Tests" or "Test" so more of the matrix shows up in the UI

image

Could also qualify "torch {{ matrix.torch-version }}"

Here's a mock-up to show where the text gets cut off: 
image

.github/workflows/ci-sharktank.yml
```diff
@@ -26,8 +26,17 @@ jobs:
...
     strategy:
       matrix:
-        version: [3.11]
-        os: [ubuntu-24.04, windows-2022]
+        python-version: ["3.11", "3.12"]
+        torch-version: ["2.3.0+cpu", "2.4.1+cpu", "2.5.1+cpu"]
```
ScottTodd, line 30: You could also omit the `+cpu` from the matrix and only add it back in
```
pip install ... torch==${{matrix.torch-version}}+cpu
```

4 more characters to use in the job names then :P

ScottTodd (approved):
Nice! I made some small edits to the PR description and left some suggestions for abbreviating the job names

.github/workflows/ci-sharktank.yml
marbre: I renamed to "Unit Tests" for now but we probably want to harmonize this more across different workflows in the future.

ScottTodd: Nice. Definitely better now. Could still include "torch" next to the torch version but it might be obvious enough to some people.

image

.github/workflows/ci-sharktank.yml
marbre: Done.

.github/workflows/ci-sharktank.yml
marbre: Renamed to `"Unit Tests :: ${{ matrix.os }}, ${{ matrix.python-version }}, ${{ matrix.torch-version }}"` as for Windows it was `(windows-2022, 3.11, 2.3.0)` and now have the same order for all jobs. WDYT?

ScottTodd: Ah, the ordering helps too. I think having two versions right next to each other can be confusing unless you are very familiar with Python _and_ PyTorch versions. I personally have to unpack `3.11, 2.4.1` and think for a second, compared to `3.11, torch 2.4.1`. We don't have many characters to spare in the UI though.

marbre: Save another character but we should aim to harmonize the naming in a follow up.

ScottTodd (approved):

PR #695, opened 2024-12-13
Add unit test summary jobs to sharktank and shortfin workflows.

ScottTodd:
Progress on https://github.com/nod-ai/shark-ai/issues/357.

This gives us a single job in each unit test workflow to set as a "required check" that will block merging, replacing the current single `Unit Tests (3.11, 2.3.0, ubuntu-24.04)` check: 
image

I'm also switching ci-libshortfin.yml to run on all changes, which is a requirement for making a check required. We could still conditionally skip jobs based on paths modified... just not using GitHub's built in path filtering.

### DIFF
--- a/.github/workflows/ci-libshortfin.yml
+++ b/.github/workflows/ci-libshortfin.yml
@@ -9,15 +9,9 @@ name: CI - shortfin
 on:
   workflow_dispatch:
   pull_request:
-    paths:
-      - '.github/workflows/ci-libshortfin.yml'
-      - 'shortfin/**'
   push:
     branches:
       - main
-    paths:
-      - '.github/workflows/ci-libshortfin.yml'
-      - 'shortfin/**'
 
 permissions:
   contents: read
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-libshortfin.yml
```diff
@@ -152,3 +146,23 @@ jobs:
...
+          echo '${{ toJson(needs) }}'
+          FAILED_JOBS="$(echo '${{ toJson(needs) }}' \
+            | jq --raw-output \
+            'map_values(select(.result!="success" and .result!="skipped")) | keys | join(",")' \
+          )"
+          echo "failed-jobs=${FAILED_JOBS}" >> $GITHUB_OUTPUT
+          if [[ "${FAILED_JOBS}" != "" ]]; then
+            echo "The following jobs failed: ${FAILED_JOBS}"
+            exit 1
+          fi
```
ScottTodd, line 168: This failed jobs list code is forked from IREE. We could simplify for cases where there is only one job or we aren't using the `GITHUB_OUTPUT` (e.g. to send an alert to an external service).

marbre: I don't know if there are plans to send it out via `GITHUB_OUTPUT` thus simplifying sounds good to me and if really needed we know where to fork the complexity from :smile:

ScottTodd: Okay, I'll simplify. Will need to test a bit :P

ScottTodd: Could just `echo "The job matrix failed"` I guess...

ScottTodd: Eh. Supporting a list of jobs is still useful for the future. I'll just remove the currently unused `echo "failed-jobs=${FAILED_JOBS}" >> $GITHUB_OUTPUT`.

marbre (approved):

PR #699, opened 2024-12-16
Explicitly install nightlies

marbre:
As `iree-turbine` is specified in `sharktank/requirements.txt`, the
dependency is already fulfilled and nightlies don't get installed. One
option would be to enforce to `--upgrade` but instead the packages get
not installed before stable releases get pulled in via sharktank's
requirements file. For workflow running on non-GH hosted runners,
`--upgrade` is passed additionaly to avoid using an eventually cached
version on the self-hosted runners.

### DIFF
--- a/.github/workflows/ci-llama-large-tests.yaml
+++ b/.github/workflows/ci-llama-large-tests.yaml
@@ -53,11 +53,11 @@ jobs:
         run: |
           source ${VENV_DIR}/bin/activate
           python -m pip install --no-compile --upgrade pip
+
           # Note: We install in three steps in order to satisfy requirements
           # from non default locations first. Installing the PyTorch CPU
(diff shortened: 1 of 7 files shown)


marbre:
Passing `--upgrade` in workflows on self-hosted runners might not be necessary but it shouldn't hurt to have it for the moment and we can check the logs if the packages get freshly installed or if a cached version is upgraded and eventually drop the flag later.

ScottTodd (approved):

ScottTodd:
> Passing `--upgrade` in workflows on self-hosted runners might not be necessary but it shouldn't hurt to have it for the moment and we can check the logs if the packages get freshly installed or if a cached version is upgraded and eventually drop the flag later.

Agreed. I think the venvs should be fresh between runs, even on self-hosted runners, but adding `--upgrade` is pretty harmless here.

PR #701, opened 2024-12-16
[shark-ai] Include sharktank in the meta package

marbre:


marbre:
The downside right now is that
https://github.com/nod-ai/shark-ai/blob/eb1d4c3f18e6e62412ab48bad001e4a699346e43/sharktank/requirements.txt#L14
is pulled in from PyPI which comes with a lot of dependencies.

ScottTodd:
> The downside right now is that
> 
> https://github.com/nod-ai/shark-ai/blob/eb1d4c3f18e6e62412ab48bad001e4a699346e43/sharktank/requirements.txt#L14
> 
> 
> is pulled in from PyPI which comes with a lot of dependencies and we cannot point to the e.g. CPU-only version if we want to publish the meta package on PyPI.

iree-turbine also has that pin: https://github.com/iree-org/iree-turbine/blob/91d8f598ca4ae5927bc0a735e17c4af8fbfaf90d/setup.py#L110. We could switch all packages to not require `torch` directly and instead check at import time if it is installed

ScottTodd (approved):
I think we should proceed with this and then trim the requirements (notably `torch`) in a follow-up (ideally before the next stable release).

We can also explore moving some deps into extras.

PR #606, opened 2024-11-25
[tuner] Add direct TD spec generation for candidates

Max191:
This PR adds direct transform dialect spec generation for candidate configurations. This is the first part of the large refactoring described in https://github.com/nod-ai/shark-ai/pull/577. The way TD specs are generated is by matching against certain types of operations, and then creating a named sequence with `transform.iree.match.cast_compatible_dag_from_root` based on the matched operation. This is done for each configuration found, and the specs are saved to the temporary tuning directory to be used later in tuning.

One main difference in the flow of candidate generation is that state is no longer tracked by saving files to a temporary directory. Instead, ir modules are passed to each function, and only at the very end of candidate generation are the transform dialect specs written to files. This makes things cleaner, since there no longer needs to be a coordination of file paths.

PR #702, opened 2024-12-16
[tuner] Add unified compilation step for models and dispatches

Max191:
This PR adds a new `compile()` function for compiling both models and dispatches during tuning. This will eventually replace the old path, which has a separate function for model compilation and dispatch compilation. The new compilation flow is now simpler:

1. Populate each CandidateTracker with the input and output filepaths. The input filepaths can be overridden by an optional function argument to the compile() function. This argument can be used for model tuning, passing the model filepath as the new input file.
2. For each candidate, strip the compilation info using iree-opt, and compile to a vmfb with the iree compiler python bindings. Set the candidate's TD spec file (generated during candidate generation), and add any additional iree-compile flags that came from the TuningClient. The extra flags are taken from a new abstract TuningClient function called `get_iree_compile_flags`.
3. For all successful compilations, save the vmfbs to the designated output path, and skip any failed compilation. For any failed compilation, a failure dump is saved instead of the vmfb.
4. Remove duplicate vmfbs, and return the ids of all unique candidates.

### DIFF
--- a/tuner/examples/dispatch/dispatch_tuner.py
+++ b/tuner/examples/dispatch/dispatch_tuner.py
@@ -79,6 +79,9 @@ class DispatchTuner(libtuner.TuningClient):
     ) -> list[str]:
         return []
 
+    def get_iree_compile_flags(self) -> list[str]:
+        return []
+
 
 def main():
     args = libtuner.parse_arguments()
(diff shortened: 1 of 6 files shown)


Max191:
This PR is based on https://github.com/nod-ai/shark-ai/pull/606. Please only review the last commit.

PR #704, opened 2024-12-16
[tuner] Add unified benchmarking and compilation for models + dispatches

Max191:
This PR adds `benchmark()` and `compile()` functions to the tuner that can be used for both model and dispatch tuning. The new functions will replace the split benchmark/compile_models and benchmark/compile_dispatches functions. The new benchmarking and compilation functions now use the iree_runtime/iree_compiler python bindings which makes much of the code simpler. Particularly, benchmark results are now mostly parsed by the bindings, and the parse_*_benchmark_results functions are no longer needed. The new compilation and benchmarking flows are described below.

### Compilation ###

1. Populate each CandidateTracker with the input and output filepaths. The input filepaths can be overridden by an optional function argument to the compile() function. This argument can be used for model tuning, passing the model filepath as the new input file.
2. For each candidate, strip the compilation info using iree-opt, and compile to a vmfb with the iree compiler python bindings. Set the candidate's TD spec file (generated during candidate generation), and add any additional iree-compile flags that came from the TuningClient. The extra flags are taken from a new abstract TuningClient function called get_iree_compile_flags.
3. For all successful compilations, save the vmfbs to the designated output path, and skip any failed compilation. For any failed compilation, a failure dump is saved instead of the vmfb.
4. Remove duplicate vmfbs, and return the ids of all unique candidates.

### Benchmarking ###

1. Create benchmark task structs for each candidate with its CandidateTracker and the TuningClient
2. Run the candidate benchmarks on the available devices. Each benchmark task will benchmark the vmfb from the CandidateTracker using the iree_runtime python bindings, and return a benchmark result containing the candidate_id, benchmark time, and device_id.
3. Then the same benchmarking is done on the untuned baseline configuration once for each available device.
4. The results from the candidate benchmarks are compared with the baseline benchmarks from the same device, and the fastest candidates are logged and returned. The number of candidates returned is determined by an optional argument to the benchmark function, and all candidates will be returned by default.

### DIFF
--- a/tuner/examples/test/README.md
+++ b/tuner/examples/test/README.md
@@ -35,5 +35,5 @@ python -m examples.test double_mmt.mlir mmt_benchmark.mlir \
 python -m examples.test <model_file_path> <benchmark_file_path> \
     --test_num_dispatch_candidates=<num_dispatch_candidates> \
     --test_num_model_candidates=<num_model_candidates> \
-    --num-candidates=<num_generated_candidates>
+    --test_hip_target=<hip_target> \ --num-candidates=<num_generated_candidates>
 ```
--- a/tuner/examples/test/README.md
+++ b/tuner/examples/test/README.md
@@ -1,0 +1,32 @@
+# Example Tuner Test
+
+Example of tuning a dispatch and full model.
+
+## Environments
+Follow instructions in [`/tuner/README.md`](../README.md)
+
+## Running the Tuner
+
+### Choose a model to tune
+This example uses the simple `double_mmt.mlir` file.
+
+### Generate a benchmark file
+Use the usual `iree-compile` command for your model and add
+`--iree-hal-dump-executable-files-to=dump --iree-config-add-tuner-attributes`. For example:
+```shell
+iree-compile double_mmt.mlir --iree-hal-target-backends=rocm --iree-hip-target=gfx942 --iree-hal-dump-executable-files-to=dump --iree-config-add-tuner-attributes -o /dev/null
+```
+
+Next, copy the `*_benchmark.mlir` file from the `dump` directory to some temporary directory of choice.
+This will be the input to the dispatch tuner along with the original `double_mmt.mlir` model file.
+
+### Recommended Trial Run
+For an initial trial to test the tuning loop, use:
+```shell
+python -m examples.test double_mmt.mlir 5 3 mmt_benchmark.mlir --num-candidates=30
+```
+
+### Basic Usage
+```shell
+python -m examples.test <model_file_path> <num_dispatch_candidates> <num_model_candidates> <benchmark_file_path>
+```
--- a/tuner/examples/test/README.md
+++ b/tuner/examples/test/README.md
@@ -11,22 +11,29 @@ Follow instructions in [`/tuner/README.md`](../README.md)
 This example uses the simple `double_mmt.mlir` file.
 
 ### Generate a benchmark file
-Use the usual `iree-compile` command for your model and add
-`--iree-hal-dump-executable-files-to=dump --iree-config-add-tuner-attributes`. For example:
+Use the usual `iree-compile` command for your model, add
+`--iree-hal-dump-executable-files-to=dump --iree-config-add-tuner-attributes`,
+and get the dispatch benchmark that you want to tune. For example:
 ```shell
-iree-compile double_mmt.mlir --iree-hal-target-backends=rocm --iree-hip-target=gfx942 --iree-hal-dump-executable-files-to=dump --iree-config-add-tuner-attributes -o /dev/null
-```
+iree-compile double_mmt.mlir --iree-hal-target-backends=rocm \
+    --iree-hip-target=gfx942 --iree-hal-dump-executable-files-to=dump \
+    --iree-config-add-tuner-attributes -o /dev/null
 
-Next, copy the `*_benchmark.mlir` file from the `dump` directory to some temporary directory of choice.
-This will be the input to the dispatch tuner along with the original `double_mmt.mlir` model file.
+cp dump/module_main_dispatch_0_rocm_hsaco_fb_benchmark.mlir mmt_benchmark.mlir
+```
 
 ### Recommended Trial Run
 For an initial trial to test the tuning loop, use:
 ```shell
-python -m examples.test double_mmt.mlir 5 3 mmt_benchmark.mlir --num-candidates=30
+python -m examples.test double_mmt.mlir mmt_benchmark.mlir \
+    --test_num_dispatch_candidates=5 --test_num_model_candidates=3 \
+    --num-candidates=30
 ```
 
 ### Basic Usage
 ```shell
-python -m examples.test <model_file_path> <num_dispatch_candidates> <num_model_candidates> <benchmark_file_path>
+python -m examples.test <model_file_path> <benchmark_file_path> \
+    --test_num_dispatch_candidates=<num_dispatch_candidates> \
+    --test_num_model_candidates=<num_model_candidates> \
+    --num-candidates=<num_generated_candidates>
 ```
--- a/tuner/examples/test/tuner_test.py
+++ b/tuner/examples/test/tuner_test.py
@@ -64,17 +64,19 @@ def main():
     parser = argparse.ArgumentParser(description="Autotune test script")
     test_args = parser.add_argument_group("Example Test Options")
     test_args.add_argument(
-        "test_model_file", type=Path, help="Path to the model file to benchmark (.mlir)"
+        "test_model_file", type=Path, help="Path to the model file to tune (.mlir)"
     )
     test_args.add_argument(
-        "test_num_dispatch_candidates",
+        "--test_num_dispatch_candidates",
         type=int,
+        default=None,
         help="Number of dispatch candidates to keep for model benchmarks.",
     )
     test_args.add_argument(
-        "test_num_model_candidates",
+        "--test_num_model_candidates",
         type=int,
-        help="Number of model candidates to keep for model benchmarks.",
+        default=None,
+        help="Number of model candidates to produce after tuning.",
     )
     # Remaining arguments come from libtuner
     args = libtuner.parse_arguments(parser)
--- a/tuner/examples/test/tuner_test.py
+++ b/tuner/examples/test/tuner_test.py
@@ -11,10 +11,7 @@ from tuner import libtuner
 
 class TestTuner(libtuner.TuningClient):
     def __init__(self):
-        self.compile_flags = [
-            "--iree-hip-target=gfx942",
-            "--compile-from=executable-sources",
-        ]
+        self.compile_flags = ["--compile-from=executable-sources"]
         self.benchmark_flags = ["--benchmark_repetitions=3", "--input=1"]
 
     def get_iree_compile_flags(self) -> list[str]:
@@ -96,10 +93,10 @@ def main():
     # TODO(Max191): Some bug seems to be causing OOM errors in benchmarking
     # when device validation happens, so this is commented for now. Uncomment
     # when the bug is fixed.
-    # if not args.dry_run:
-    #     print("Validating devices")
-    #     libtuner.validate_devices(args.devices)
-    #     print("Validation successful!\n")
+    if not args.dry_run:
+        print("Validating devices")
+        libtuner.validate_devices(args.devices)
+        print("Validation successful!\n")
 
     print("Generating candidates...")
     candidates = libtuner.generate_candidate_specs(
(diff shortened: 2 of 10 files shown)


Max191:
This PR is based on https://github.com/nod-ai/shark-ai/pull/702. Please only review the last commit.

tuner/examples/test/tuner_test.py
```diff
@@ -4,15 +4,88 @@
 # See https://llvm.org/LICENSE.txt for license information.
```
Max191, line 4: I will add a README for this tuner example that explains how to use it.

kuhar (commented):
Some initial comments, I need to do another scan over this code

tuner/tuner/libtuner.py
```diff
@@ -598,6 +641,158 @@ def run_command(run_pack: RunPack) -> RunResult:
...
+
+    # Benchmark the module.
+    try:
+        timeout = tuning_client.get_benchmark_timeout_s()
+        benchmark_results = ireert.benchmark.benchmark_module(
+            vm_module,
+            entry_function=func_name,
+            device=device,
+            timeout=timeout,
+            **extra_flags,
```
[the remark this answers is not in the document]
Max191: I reused most of the multiprocess stuff she wrote, but I just replaced the worker function that gets called. In my own testing it is using the correct devices.

Afaict, the iree_runtime.benchmark.benchmark_module function effectively does the same thing under the hood that the old path used to do (calling a subprocess with the appropriate command), but it handles all the iree specific flag and output parsing.

tuner/tuner/libtuner.py
```diff
@@ -1172,6 +1434,72 @@ def benchmark_dispatches(
...
+    num_candidates: Optional[int] = None,
+):
+    logging.debug("benchmark()")
+
+    task_list = [
+        BenchmarkPack(
+            tuning_client=tuning_client, candidate_tracker=candidate_trackers[i]
+        )
+        for i in compiled_candidates
+        if i != 0
```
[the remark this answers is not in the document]
Max191: Yes, it is filtering out the baseline candidate. The black formatted put that on a separate line.

tuner/tuner/spec_builder.py
```diff
@@ -24,12 +38,43 @@ def build_td_spec(
...
+    # Get the names ssa names of operands to make sure they match in the
+    # template after string formatting.
+    captured_values: set[ir.Value] = set()
     for operand in op.operands:
+        if operand in captured_values:
+            # TODO(Max191): Remove this warning when the transform for the
+            # `cast_compatible_dag_from_root` op fixes a bug in the matching
+            # logic that causes failure to match when the same operand is
+            # repeated. For now, still avoid adding duplicate SSA values to
+            # prevent parsing failure.
```
Max191, line 68: There is a bug in the transform op, which I think is a very simple fix, but we need to land it in IREE before this works properly for repeated operands.

tuner/tuner/libtuner.py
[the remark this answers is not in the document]
Max191: There were some unrelated issues when using the python bindings with multithreading, but I fixed it in my last commit. The multithreading is working properly now with multiple gpu devices.

tuner/examples/test/README.md
```diff
@@ -0,0 +1,32 @@
...
+## Running the Tuner
+
+### Choose a model to tune
+This example uses the simple `double_mmt.mlir` file.
+
+### Generate a benchmark file
+Use the usual `iree-compile` command for your model and add
+`--iree-hal-dump-executable-files-to=dump --iree-config-add-tuner-attributes`. For example:
+```shell
+iree-compile double_mmt.mlir --iree-hal-target-backends=rocm --iree-hip-target=gfx942 --iree-hal-dump-executable-files-to=dump --iree-config-add-tuner-attributes -o /dev/null
```
kuhar, line 17: nit: break this line to fit the column limit

tuner/examples/test/README.md
```diff
@@ -0,0 +1,32 @@
...
+```
+
+Next, copy the `*_benchmark.mlir` file from the `dump` directory to some temporary directory of choice.
```
kuhar, line 20: Add the copy command to the shell snippet above to reduce the amount of prose?

tuner/examples/test/README.md
```diff
@@ -0,0 +1,32 @@
...
+This will be the input to the dispatch tuner along with the original `double_mmt.mlir` model file.
+
+### Recommended Trial Run
+For an initial trial to test the tuning loop, use:
+```shell
+python -m examples.test double_mmt.mlir 5 3 mmt_benchmark.mlir --num-candidates=30
```
kuhar, line 26: Can we use named arguments? It's hard to say what 5 and 3 mean and if the order of the input program and input benchmark matter.

tuner/examples/test/README.md
```diff
@@ -0,0 +1,32 @@
...
+```
+
+### Basic Usage
+```shell
+python -m examples.test <model_file_path> <num_dispatch_candidates> <num_model_candidates> <benchmark_file_path>
```
kuhar, line 31: Num candidates is a flag below but here it's a positional argument

tuner/examples/test/tuner_test.py
```diff
@@ -4,22 +4,95 @@
...
 
+import argparse
+from pathlib import Path
 from tuner import libtuner
 
 
+class TestTuner(libtuner.TuningClient):
+    def __init__(self):
+        self.compile_flags = ["--compile-from=executable-sources"]
+        self.benchmark_flags = ["--benchmark_repetitions=3", "--input=1"]
```
kuhar, line 15: Do we need the input argument?

tuner/examples/test/tuner_test.py
```diff
@@ -33,6 +106,53 @@ def main():
...
+        candidate_trackers,
+        test_tuner,
+        args.test_num_dispatch_candidates,
+    )
+
+    print("Compiling models with top candidates...")
+    test_tuner.compile_flags = [
+        "--iree-hal-target-backends=rocm",
+        "--iree-hip-target=gfx942",
+    ]
```
kuhar, line 129: Since we know how to compile the model, could we generate the benchmark files for the user to skip manual steps? Doesn't have to be in this PR, but it would be nice to have a TODO.

tuner/examples/test/tuner_test.py
```diff
@@ -33,6 +106,53 @@ def main():
...
+        compiled_candidates,
+        candidate_trackers,
+        test_tuner,
+        args.test_num_dispatch_candidates,
+    )
+
+    print("Compiling models with top candidates...")
+    test_tuner.compile_flags = [
+        "--iree-hal-target-backends=rocm",
+        "--iree-hip-target=gfx942",
```
kuhar, line 128: It could be that the benchmark is compiled for a different target.

tuner/tuner/libtuner.py
```diff
@@ -70,6 +75,7 @@ class CandidateTracker:
     dispatch_config_path: Optional[Path] = None
     configuration: Optional[candidate_gen.iree_codegen.CompilationInfoAttr] = None
     compilation_successful: Optional[bool] = None
+    compiled_vmfb_path: Optional[Path] = None
```
kuhar, line 78: Is this vmfb of the model or the dispatch?

tuner/tuner/libtuner.py
```diff
@@ -598,6 +642,172 @@ def run_command(run_pack: RunPack) -> RunResult:
...
+            strip_command,
+            capture_output=True,
+            text=True,
+        )
+        stripped_mlir = result.stdout
+    except subprocess.CalledProcessError as e:
+        print(e.output)
+        cmd = " ".join(strip_command)
+        logging.error(f"Command '{cmd}' returned non-zero exit status {e.returncode}.")
+        logging.error(f"Command '{cmd}' failed with error: {e.stderr}")
```
kuhar, line 679: I think it'd be cleaner to define a helper function to launch external processes and have uniform error handling in the tuner

tuner/tuner/libtuner.py
```diff
@@ -598,6 +642,172 @@ def run_command(run_pack: RunPack) -> RunResult:
...
+        logging.warning(
+            f"Benchmark of candidate {candidate_id} timed out after {timeout} seconds."
+        )
+        return BenchmarkResult(
+            candidate_id=candidate_id,
+            time=math.inf,
+            device_id=str(device_id),
+        )
+    except KeyboardInterrupt:
+        print("Ctrl+C detected, terminating child processes...")
```
kuhar, line 783: What is child process in this context? The whole tuner?

tuner/tuner/libtuner.py
```diff
@@ -908,6 +1121,87 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+
+    # If `input_file` is not None, then replace the currently tracked template
+    # with the passed input mlir file.
+    if input_file is not None:
+        shutil.copy(input_file, path_config.template_mlir)
+
+    # Strip compilation info and root_op attribute from the source and save
+    # the stripped IR, since the TD specs do not expect these attributes.
+    stripped_mlir = strip_compilation_info(path_config.template_mlir)
+    with ir.Context():
```
kuhar, line 1146: Why are we creating a new context instead of using the one from the tuning client?

tuner/tuner/spec_builder.py
```diff
@@ -13,6 +13,20 @@
 from .common import *
 from .dispatch_constraints import *
 from .dispatch_parser import *
+from .op_matchers import ROOT_OP_ATTR_NAME
+
+
+def get_placeholder_spec(context: ir.Context) -> ir.Module:
+    spec_text = f"""
+        module attributes {{ transform.with_named_sequence }} {{
```
kuhar, line 21: nice 👍

tuner/examples/test/tuner_test.py
Max191: Benchmarks take a single input for the number of iterations to run afaict. I think we can add some bool option (i.e., is_benchmark or something) to the benchmark function that can add this input automatically, but I'll leave that as a followup.

tuner/examples/test/tuner_test.py
Max191: Yeah, that's a nice idea, and I think it's better as a followup.

tuner/tuner/libtuner.py
Max191: This is used for both. When you run the compile() function, it will populate this path. I think it is fine to overwrite things like this, because eventually I want to move the candidate_trackers to be private to some tuner object that can save state between tuning phases. Then we can ensure that the paths are filled as expected, and the tuning client doesn't have to touch these candidate trackers.

tuner/tuner/libtuner.py
Max191: I could do this, but it will only be used in this strip_compilation_info function right now. The compilation and benchmarking happens within the python bindings, and they raise their own exception types, so right now this is the only place something like that would be used. I'm fine with adding it anyway, though, if you think it is cleaner that way.

tuner/tuner/libtuner.py
Max191: Yeah, I need to update this message. It is left over from copy paste.

tuner/tuner/libtuner.py
Max191: The context isn't actually being used for anything other than to make the parser happy here. This is going from input_file->output_file, so the context is completely local to this one section of code. Also, afaict, there is no guarantee that the tuning client has a context?

tuner/tuner/libtuner.py
Max191: I used the existing run_command helper, which has some error handling.

kuhar (approved):
LGTM % context reuse (if we can easily do that)

tuner/tuner/libtuner.py
Max191: I see, then I can add a context to the TuningClient class and thread it through.

kuhar: Feel free to add a TODO, it's not super important

Max191:
> LGTM % context reuse (if we can easily do that)

This ended up needing a lot of changes, since context was not threaded through many of the util functions. You might want to take another look at the PR now.

tuner/tuner/candidate_gen_test.py
```diff
@@ -77,20 +77,21 @@ def test_get_td_spec_contraction(tuner_ctx: common.TunerContext) -> None:
...
     )
     pipeline_options = iree_gpu.PipelineOptionsAttr.get(prefetch_shared_memory=True)
     config_dict = common.get_translation_info_config(pipeline_options, waves_per_eu=8)
     translation_info = iree_codegen.TranslationInfoAttr.get(
-        pipeline_attr, None, [16, 16, 1], 16, config_dict
+        pipeline_attr, None, [16, 16, 1], 16, config_dict, ctx=tuner_ctx.mlir_ctx
     )
     compilation_info = iree_codegen.CompilationInfoAttr.get(
-        lowering_config, translation_info
+        lowering_config, translation_info, ctx=tuner_ctx.mlir_ctx
```
kuhar, line 89: I don't think we need this or similar elsewhere -- `None` context is always replaced with the first context in the context stack: https://mlir.llvm.org/docs/Bindings/Python/#context-management

PR #711, opened 2024-12-17
[shortfin] Bump fmt and spdlog and test with gcc 14

marbre:
Bumps libftm to 11.0.2 to mitigate a build error with occurring with
gcc 14. Bumping spdlog to 1.15.0 (which bundles libfmt 11.0.2)
accordingly to keep the libs in sync. Furthermore expands testing
with gcc 14.

### DIFF
--- a/.github/workflows/ci-libshortfin.yml
+++ b/.github/workflows/ci-libshortfin.yml
@@ -38,7 +38,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        name: ["Ubuntu (Clang)(full)", "Ubuntu (Clang)(host-only)", "Ubuntu (GCC)", "Windows (MSVC)"]
+        name: ["Ubuntu (Clang)(full)", "Ubuntu (Clang)(host-only)", "Windows (MSVC)"]
         python-version: ["3.10", "3.11", "3.12"]
         include:
           - name: Ubuntu (Clang)(full)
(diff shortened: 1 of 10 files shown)


shortfin/CMakeLists.txt
```diff
@@ -39,6 +39,10 @@ if(NOT WIN32)
   set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 endif()
 
+# For unicode support Windows libfmt requires compiling with /utf-8.
+add_compile_options("$<$<C_COMPILER_ID:MSVC>:/utf-8>")
+add_compile_options("$<$<CXX_COMPILER_ID:MSVC>:/utf-8>")
```
ScottTodd, line 44: I take it these can't be added here, since they need to apply to `fmt::fmt` too?
https://github.com/nod-ai/shark-ai/blob/d08b0e5dc4ecf4af364121933fbe0aa61031cc66/shortfin/build_tools/cmake/shortfin_library.cmake#L7-L17

.github/workflows/ci-libshortfin.yml
```diff
@@ -53,16 +53,19 @@ jobs:
             cmake-options:
               -DCMAKE_C_COMPILER=clang-18 -DCMAKE_CXX_COMPILER=clang++-18 -DCMAKE_LINKER_TYPE=LLD -DSHORTFIN_HAVE_AMDGPU=OFF -DSHORTFIN_BUILD_STATIC=ON -DSHORTFIN_BUILD_DYNAMIC=ON
             additional-packages: clang lld
-          - name: Ubuntu (GCC)
+          - name: Ubuntu (GCC 13)
             runs-on: ubuntu-24.04
+            # Only test with GCC 13 and Python 3.12
+            python-version: "3.12"
```
ScottTodd, line 59: Can you set the gcc version explicitly here too for consistency? If we change the ubuntu-24.04 image and the default compiler changes, this job might no longer test gcc 13.
```suggestion
          - name: Ubuntu (GCC 13)
            runs-on: ubuntu-24.04
            # Only test with GCC 13 and Python 3.12
            python-version: "3.12"
            cmake-options:
              -DCMAKE_C_COMPILER=gcc-13 -DCMAKE_CXX_COMPILER=g++-13
```

ScottTodd (approved):

shortfin/CMakeLists.txt
marbre: Sure, I give it a try.

ScottTodd: Didn't seem to work: https://github.com/nod-ai/shark-ai/actions/runs/12379929075/job/34555169307?pr=711#step:10:593. Keeping it at the top level is fine I guess. We try to keep options scoped as tightly as possible. At least https://cmake.org/cmake/help/latest/command/add_compile_options.html applies to the current directory and below, so downstream users shouldn't also need to set this... I think.

marbre: Yeah, I just moved it back to the top level.

marbre:
Test failures seem unrelated.

PR #712, opened 2024-12-18
[sharktank] Avoid torch pre-releases

marbre:
Drops passing `--pre` to avoid that pre-releases of torch get installed and switch to stable channel.

### DIFF
--- a/.github/workflows/ci-sharktank.yml
+++ b/.github/workflows/ci-sharktank.yml
@@ -74,7 +74,7 @@ jobs:
           # Note: We install in three steps in order to satisfy requirements
           # from non default locations first. Installing the PyTorch CPU
           # wheels saves multiple minutes and a lot of bandwidth on runner setup.
-          pip install --no-compile --pre --index-url https://download.pytorch.org/whl/test/cpu torch==${{matrix.torch-version}}+cpu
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-sharktank.yml
```diff
@@ -74,7 +74,7 @@ jobs:
           # Note: We install in three steps in order to satisfy requirements
           # from non default locations first. Installing the PyTorch CPU
           # wheels saves multiple minutes and a lot of bandwidth on runner setup.
-          pip install --no-compile --pre --index-url https://download.pytorch.org/whl/test/cpu torch==${{matrix.torch-version}}+cpu
+          pip install --no-compile --index-url https://download.pytorch.org/whl/test/cpu torch==${{matrix.torch-version}}+cpu
```
ScottTodd, line 77: Also switch off the `test` index:
```suggestion
          pip install --no-compile --index-url https://download.pytorch.org/whl/cpu torch==${{matrix.torch-version}}+cpu
```

Unless we also want to explicitly test 2.6.0 here, which we could do. Since the version is explicit with `==`, both `--pre` and the test index _should_ be safe.

marbre: I switched to the stable channel for now. We can switch back to `test` if we want to test 2.6.0 in the future.

ScottTodd (approved):

PR #713, opened 2024-12-18
Bump IREE to iree-3.1.0rc20241217

monorimet:
Includes a fix for shortfin's array storage subspan method: attach the storage's host allocator to iree_hal_buffer_subspan() invocation


marbre:
With #721, IREE was bumped to an even more recent version (3.1.0rc20241220).

PR #714, opened 2024-12-18
[sharktank/shortfin] Schedule sharktank CI after IREE nightly release

archana-ramalingam:
Schedule sharktank CI after IREE nightly release, which happens around 1.20 AM to 1.40 AM PST.

### DIFF
--- a/.github/workflows/ci-llama-large-tests.yaml
+++ b/.github/workflows/ci-llama-large-tests.yaml
@@ -9,8 +9,8 @@ name: Llama Benchmarking Tests
 on:
   workflow_dispatch:
   schedule:
-    # Weekdays at 09:50 AM UTC = 01:50 AM PST / 02:50 AM PDT
-    - cron: "50 9 * * 1-5"
+    # Weekdays at 11:00 AM UTC = 03:00 AM PST / 04:00 AM PDT
+    - cron: "0 11 * * 1-5"
 
 concurrency:
   # A PR number if a pull request and otherwise the commit hash. This cancels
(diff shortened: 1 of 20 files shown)


stbaione (approved):

aviator19941 (approved):

.github/workflows/ci-llama-large-tests.yaml
```diff
@@ -9,8 +9,8 @@ name: Llama Benchmarking Tests
 on:
   workflow_dispatch:
   schedule:
-    # Weekdays at 4:00 AM UTC = 9:00 PM PST.
-    - cron: "0 4 * * 1-5"
+    # Weekdays at 09:50 AM UTC = 01:50 AM PST / 02:50 AM PDT
+    - cron: "50 9 * * 1-5"
```
ScottTodd, line 13: You might want to give this a bit of extra buffer, in case the IREE release builds take longer over time as more code, dependencies, and configurations are added.

Looking through the history at https://github.com/iree-org/iree/actions/workflows/validate_and_publish_release.yml, publication time is pretty consistently between 1:20AM and 1:30AM but I think giving a 1 hour buffer would be safer.

ScottTodd: There are also ways to trigger on events like releases being published, but those get harder across organizations and repositories.

archana-ramalingam: Good point, I can make it 3 AM PST for now, till we have a way to trigger this automatically after every sharktank/shortfin/IREE nightly release.

PR #717, opened 2024-12-19
Add export of Flux-dev transformer and uploading to Azure

sogartar:
There was a slight difference in the Flux schnell and dev variants. Namely, dev has a guidance layer and schenll does not.

Fixed some tensor argument element value types as they were always passed as f32 while some of them should use the model's dtype.

Refactored a bit the Flux transformer export boilerplate.

Added a script that uploads models to Azure. Right now it uploads the Flux transformer models.
This can become a part of the CI jobs at some point.

### DIFF
--- a/sharktank/requirements.txt
+++ b/sharktank/requirements.txt
@@ -12,3 +12,6 @@ datasets
 # Serving deps.
 fastapi>=0.112.2
 uvicorn>=0.30.6
+
+azure-identity>=1.19
+azure-storage-blob>=12.24
--- a/sharktank/requirements.txt
+++ b/sharktank/requirements.txt
@@ -12,6 +12,3 @@ datasets
 # Serving deps.
 fastapi>=0.112.2
 uvicorn>=0.30.6
-
-azure-identity>=1.19
-azure-storage-blob>=12.24
--- a/sharktank/sharktank/tools/upload_all_models_to_azure.py
+++ b/sharktank/sharktank/tools/upload_all_models_to_azure.py
@@ -1,3 +1,9 @@
+# Copyright 2024 Advanced Micro Devices, Inc.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
 from ..utils.azure import upload_all_models
 
 import logging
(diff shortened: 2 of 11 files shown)


sharktank/requirements.txt
```diff
@@ -12,3 +12,6 @@ datasets
 # Serving deps.
 fastapi>=0.112.2
 uvicorn>=0.30.6
+
+azure-identity>=1.19
+azure-storage-blob>=12.24
```
ScottTodd, line 17: Let's put these new requirements in a dev requirements file or an "extras" so they don't spill into the packages we publish unconditionally.

sharktank/sharktank/tools/upload_all_models_to_azure.py
```diff
@@ -0,0 +1,51 @@
+from ..utils.azure import upload_all_models
```
ScottTodd, line 1: copyright headers at the tops of files please

```suggestion
# Copyright 2024 Advanced Micro Devices, Inc.
#
# Licensed under the Apache License v2.0 with LLVM Exceptions.
# See https://llvm.org/LICENSE.txt for license information.
# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception

from ..utils.azure import upload_all_models
```

sharktank/requirements.txt
ScottTodd: The sharktank package currently has two requirements files, one for the standard install and one for the `sharktank[testing]` extra: https://github.com/nod-ai/shark-ai/blob/5e3f5e645abafc58c69a7f871e4b18b318d6653d/sharktank/pyproject.toml#L33-L37

Maybe `testing` is a good fit for this? We can also add a new category as needed

sharktank/requirements.txt
```diff
@@ -12,3 +12,6 @@ datasets
 # Serving deps.
 fastapi>=0.112.2
 uvicorn>=0.30.6
+
+azure-identity>=1.19
```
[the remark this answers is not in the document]
IanNod: Just saw @ScottTodd's comment on this above. I agree that maybe we should add another dev or extras requirements file

sharktank/sharktank/tools/upload_all_models_to_azure.py
sogartar: Done

sharktank/requirements.txt
sogartar: This is not required by any tests right now. I added a dev package component. And put them in another file requirements file.

sharktank/requirements.txt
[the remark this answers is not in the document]
sogartar: I answered on the other thread.

IanNod (approved):
Looks good to me

Issue #691, opened 2024-12-12
Streamline LLM import/compile/serve user experience


ScottTodd:
Here is our current documentation for running llama models through shark-ai: https://github.com/nod-ai/shark-ai/blob/main/docs/shortfin/llm/user/e2e_llama8b_mi300x.md ([permalink](https://github.com/nod-ai/shark-ai/blob/d279afff48c56ab5e0e6e5ebe2717ccf0b26ee50/docs/shortfin/llm/user/e2e_llama8b_mi300x.md)).

Our current steps are convoluted compared to similar documentation in other projects.

## Comparison with other projects

vLLM

* https://docs.vllm.ai/en/latest/getting_started/quickstart.html
* https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html#example-of-usage-for-a-pair-of-a-string-and-a-list-of-texts

TensorRT-LLM

* https://nvidia.github.io/TensorRT-LLM/quick-start-guide.html
* https://nvidia.github.io/TensorRT-LLM/commands/trtllm-serve.html
* https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html 

MLC LLM

* https://llm.mlc.ai/docs/get_started/introduction.html
* https://llm.mlc.ai/docs/compilation/compile_models.html#clone-from-hf-and-convert-weight
* https://github.com/bentoml/BentoMLCLLM?tab=readme-ov-file#set-up-the-environment

Ollama

* https://github.com/ollama/ollama

TorchServe

* https://pytorch.org/serve/batch_inference_with_ts.html

## Feedback

I have some inlined comments on one file [in this commit on my fork](https://github.com/ScottTodd/shark-ai/commit/bfde15b423471b8d7d1db8d06686227af3899940). Here I'll try to summarize them as tasks:

- [ ] Only one line in that guide is actually device-specific: `--iree-hip-target=gfx942`. The file should be generalized.
- [ ] Any actions that create files (downloading, importing, compiling, etc.) should choose a consistent and structured default location. Similarly, tool arguments should have sensible default values, so the documentation does not need to specify 9+ environment variables
- [ ] The [`sharktank.utils.hf_datasets` module ](https://github.com/nod-ai/shark-ai/blob/main/sharktank/sharktank/utils/hf_datasets.py) is a tool for developer convenience and testing. Users should be directed to download from huggingface directly, using standard tools and APIs (e.g. the huggingface-cli tool or the hugging_face library)
[...]

PR #718, opened 2024-12-19
Iterate on llama user guide [2].

ScottTodd:
Progress on https://github.com/nod-ai/shark-ai/issues/691.

* Generalize guide to any llama model on any accelerator, mentioning specifics of platform/model support where it matters
* Restructure the introduction section with more context and an overview of the rest of the guide (more still to do here, explaining this tech stack)
* Add prerequisites section, modeled after the [user guide](https://github.com/nod-ai/shark-ai/blob/main/docs/user_guide.md)
* Add more "why" explanations for many steps (more still to do here)
* Start trimming environment variable instructions


stbaione (approved):

docs/shortfin/llm/user/llama_end_to_end.md
ScottTodd, line 1: Regarding the `docs/shortfin/llm/user/llama_end_to_end.md` path, I have a few remaining reservations:

* This workflow uses both sharktank and shortfin, so nesting under "shortfin" is potentially confusing. Might want to just move existing top level docs to a docs/developers and keep user docs un-nested for now.
* I don't like the phrase "end to end" generally, but this is about more than just the final "serving" step right now. Maybe just "llama.md" or something like
    
    ```
    docs/
      developers/
      guides/
        serving_llama.md
        serving_sdxl.md
      reference/
        shortfin.md
        sharktank.md
    ```

In any case, I think we might hit the limits of what GitHub-flavored markdown allows us to express pretty soon. Markdown is accessible but it isn't pretty in the pages themselves or the directory view in GitHub

PR #719, opened 2024-12-19
[llama] Added the fused rotary embedding kernel

rsuderman:
Reworked rotary embedding application to be performed via a custom kernel. This includes dropping `static_table` for the sake of maintenance (it was largely unused). It includes a simple numerical test however under the hood no numerical change should occur.

Existing baseline vs hugging face remained unchanged.

### DIFF
--- a/sharktank/sharktank/examples/export_paged_llm_v1.py
+++ b/sharktank/sharktank/examples/export_paged_llm_v1.py
@@ -218,22 +218,16 @@ def main():
             else:
                 cache_tensors = cs
 
-            sl = tokens.shape[1]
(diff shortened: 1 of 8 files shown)


aviator19941:
Is it expected that the exported IR becomes much bigger (for 8b f16 bs4 tp1 IR is about 34 GB)?

rsuderman:
> Is it expected that the exported IR becomes much bigger (for 8b f16 bs4 tp1 IR is about 34 GB)?

Updated how we do causal attention on prefill. It was baking out the attention matrix as a constant which was very beefy.

aviator19941 (approved):

Issue #729, opened 2024-12-30
Shortfin LLM CI Fails on export after IREE nightly 1227


renxida:
Shortfin LLM CI fails in `sharktank.examples.export_paged_llm_v1` with error:
```python
TypeError: get(): incompatible function arguments
Supported: get(value: str, context: mlir.ir.Context | None = None) -> iree.compiler._mlir_libs._mlir.ir.FlatSymbolRefAttr
```

when 

```
            iree-base-compiler==3.1.0rc20241227 \
            iree-base-runtime==3.1.0rc20241227 \
            iree-turbine==3.1.0rc20241227
```
@nithinsubbiah mentions in slack that this issue also seems to affect SDXL export functionality

To resolve this:

1. Pin IREE to version 1226 temporarily
2. Address IREE 1227 compatibility issue

This is likely due to iree-org/iree and not iree-turbine: iree-turbine has not been committed to between Dec 20 and Dec 30. IREE had a llvm version bump on Dec 26

renxida:
Slack thread: https://xilinx.slack.com/archives/C077FDT0002/p1735374494332789?thread_ts=1735374062.079439&cid=C077FDT0002

renxida:
https://github.com/nod-ai/shark-ai/actions/runs/12550021461?pr=730

Pinning the iree-base-compiler, iree-base-runtime, and iree-turbine verisons to 1226 fixes the export failure but the server still fails to start.

Also, almost all of the sharktank tests fails when i do this.

ScottTodd:
Is this fixed now? By https://github.com/nod-ai/shark-ai/pull/758 ?

https://github.com/nod-ai/shark-ai/actions/workflows/ci-shark-ai.yml?page=2&query=branch%3Amain
![image](https://github.com/user-attachments/assets/01e43543-0aba-4bfd-a8d3-15417e4815ec)

renxida:
Yup this one is fixed!

PR #730, opened 2024-12-30
pin iree compiler versions to temporarily mitigate #729 iree rc20241230 inconpatibility

renxida:


renxida:
Cherry-picked #731 to be able to actually see the remaining failures after version pin

PR #731, opened 2024-12-30
make app_tests log subprocess stdout and stderr upon failure

renxida:


renxida:
`pytest -s` correctly captures the stderr / stdout that was missing. Closed because incorporated into #758

Issue #402, opened 2024-10-31
Find More General and Easier to use Alternative For Compiling Models for Shortfin LLM Server


stbaione:
Discussion for this in #373 and #284.

The export script in sharktank was built specifically for llama 3.1 models and has some rough edges. Along with this, it requires users to chain together cli commands: `python -m sharktank.examples.export_paged_llm_v1.py [--options]`, then `iree-compile [--options]`.

It has some rough edges, is a bit cumbersome from a user perspective, and requires CI runs to invoke cli commands via `subprocess`, instead of having a programmatic in-memory alternative.

We should find a more general and easier to use solution to handle generating mlir for LLM models and compiling those models to `.vmfb` for shortfin server.

Below is a starting point recommendation provided by @ScottTodd:

"Users shouldn't need to chain together `python -m sharktank.examples`. and `iree-compile ...` commands. We can aim for something like https://docs.vllm.ai/en/latest/getting_started/quickstart.html#offline-batched-inference

```python
llm = LLM(model="facebook/opt-125m")
outputs = llm.generate(prompts, sampling_params)
```
(that's as minimal as it gets - we'll want to pass options like the compilation target though)"

stellaraccident:
Hang tight for a bit. More tooling is coming that will make this all one command. Building it out for sdxl first. 

https://github.com/iree-org/iree/pull/18630#pullrequestreview-2409072569

ScottTodd:
Ah yes, I was just going to connect those dots too.

For SDXL there are multiple submodels (VAE + UNet + CLIP), so having the build system manage all of them is especially helpful. Ideally we can standardize on a similar set of APIs for llama, SDXL, and future supported models.

stbaione:
Closing as something is already in the works for this

ScottTodd:
Well we still need code written. Fine to keep this as a tracking issue, blocked on the work happening for SDXL.

renxida:
Looks like iree.build is merged!

PR #733, opened 2025-01-01
Implement bash script to unify user-end commands

vinayakdsci:
Progress on https://github.com/nod-ai/shark-ai/issues/402 and https://github.com/nod-ai/shark-ai/issues/691.

Adds a bash script that will unify all the commands in [llama_serving.md](https://github.com/nod-ai/shark-ai/blob/main/docs/shortfin/llm/user/llama_serving.md), providing a cleaner and single interface for the user to interact with.

### DIFF
--- a/run_shark_ai.sh
+++ b/run_shark_ai.sh
@@ -19,185 +19,277 @@ shopt -s extglob
 #   6. -i -- path to the IREE build dir (for dev purposes).
 #   7. -v -- enable verbose logging.
 #   8. -s -- Whether to shard the weights or not. Defaults to tp1 if not set, else tp8.
-#   9. -x -- If set, compilation is skipped. Useful when only sharding or export is required.
+#   9. -c -- If set, compilation is enabled as a part of the pipeline.
 #  10. -e -- If set, export is called. This allows sharding to be run separately from export.
+#  11. -o -- The value for EXPORT_DIR, where the exported model will be placed
 
 
 print_help_string() {
-  printf "%s\n" "USAGE: run_shark_ai [OPTIONS] ... [EXPORT_DIR]
+  printf "%s\n" "USAGE: run_shark_ai [OPTIONS] ... -o [EXPORT_DIR]
 Exports the model using shark-ai and compiles the IR to produce the VMFB module.
 
-Some options must be madatorily specified.
-    -h            prints the help string. The same output is emitted
-                    when no flags are specified.
+Some options are required to be specified.
+    -h            prints the help string. The same output is emitted when no arguments are given.
     -v            enables verbose logging. When not specified, only errors are logged.
     -w            the location of the GGUF/IRPA file(s) that contain the parameters. (required)
     -a            the location of the artifacts (sharded weights). (defaults to EXPORT_DIR/artifacts)
     -b            batch sizes for the export. Multiple batch sizes should be separated by ','.
                     defaults to (1,4).
-    -i            location of the IREE build directory, to use local build for compilation.
+    -i            the directory containing iree-compile, if from local build, to use for compilation.
                     defaults to the PATH value if not set.
     -s            use to enable sharding. shards to tp8 if set, otherwise tp1.
-    -x            whether to skip compilation and only export the model.
+    -c            whether to also compile the exported model and produce a VMFB file.
     -e            calls export first if set. This allows the export and sharding commands to be
                     run independently.
-
-EXPORT_DIR is a positional argument, to be placed at the end. Specifies where the exported model will be placed."
+    -o            the location for the exported model. (required)"
 }
 
+print_help_string_and_exit() {
+  print_help_string && exit $1
+}
+
+unset -v VERBOSE_LOGGING
+unset -v WEIGHT_FILE_LOC
+unset -v ARTIFACT_LOC
+unset -v EXPORT_BATCH_SIZES
+unset -v EXPORT
+unset -v ROCR_VISIBLE_DEVICES
+unset -v MODEL_CONFIG_LOC
+unset -v IREE_BUILD_DIR
+unset -v SHARD_WEIGHTS
+unset -v COMPILE
+unset -v EXPORT_DIR
+
 VERBOSE_LOGGING=false
-WEIGHT_FILE_LOC=0
-ARTIFACT_LOC=0
 EXPORT_BATCH_SIZES=1,4
 EXPORT=0
 ROCR_VISIBLE_DEVICES=1
 MODEL_CONFIG_LOC=0
-IREE_BUILD_DIR=0
 SHARD_WEIGHTS=0
-SKIP_COMPILE=0
-EXPORT_DIR=0
+COMPILE=0
 
-err() { printf "%s\n" "$*" >&2 && exit 1; }
-log() { $VERBOSE_LOGGING && printf "%s\n" "$*" >&2; }
-
-unset h
-unset i
-unset s
-unset v
-unset x
-unset w
-unset t
-unset a
-unset b
-unset e
+err() { printf "[ERROR] %s\n" "$*" >&2 && exit 1; }
+log() { "$VERBOSE_LOGGING" && printf "%s\n" "$*" >&2; }
+info() { "$VERBOSE_LOGGING" && log "[INFO] $*"; }
+warn() { "$VERBOSE_LOGGING" && log "[WARNING] $*"; }
 
 parse_and_handle_args() {
-  while getopts ":ehvsxw:t:a:b:i:" flag; do
-    case "$flag" in
+  while getopts ":ehvscw:t:a:b:i:o:" flag; do
+    case ""$flag"" in
       h)
-        print_help_string && exit 0
+        print_help_string_and_exit 0
         ;;
       v)
         VERBOSE_LOGGING=true
         ;;
       w)
-        WEIGHT_FILE_LOC=$OPTARG
+        WEIGHT_FILE_LOC="$OPTARG"
         ;;
       a)
-        ARTIFACT_LOC=$OPTARG
-        if [[ ! -d $ARTIFACT_LOC ]]; then
-            log "[WARNING] $ARTIFACT_LOC does not exist, creating..."
-            mkdir -p $ARTIFACT_LOC
+        ARTIFACT_LOC=$(realpath "$OPTARG")
+        if [[ ! -d "$ARTIFACT_LOC" ]]; then
+            warn ""$ARTIFACT_LOC" does not exist, creating..."
+            mkdir -p "$ARTIFACT_LOC"
         fi
         ;;
       b)
-        EXPORT_BATCH_SIZES=$OPTARG
+        EXPORT_BATCH_SIZES="$OPTARG"
         ;;
       i)
-        IREE_BUILD_DIR=$OPTARG
+        IREE_BUILD_DIR=$(realpath "$OPTARG")
         ;;
       s)
         SHARD_WEIGHTS=1
-        log "[INFO] Weights will be sharded with tp 8"
+        info "Weights will be sharded with tp 8"
         ;;
-      x)
-        SKIP_COMPILE=1
-        log "[INFO] Compilation will be skipped"
+      c)
+        COMPILE=1
+        info "Compilation is enabled"
         ;;
       e)
         EXPORT=1
-        log "[INFO] Export is enabled"
-	;;
-      :)
-        err "[ERROR] option -$OPTARG expected an argument"
+        info "Export is enabled"
         ;;
-      \?)   print_help_string && err "[ERROR] INVALID OPTION -$OPTARG"
+      o)
+        EXPORT_DIR=$(realpath "$OPTARG")
+        info "Exported model will be saved at $EXPORT_DIR"
+        ;;
+      :)
+        err "option -$OPTARG expected an argument"
+        ;;
+      \?)   print_help_string && err "INVALID OPTION -$OPTARG"
         ;;
     esac
   done
-  shift $(( $OPTIND-1 ))
-
-  [[ $# -eq 1 ]] || err "[ERROR] expected EXPORT_DIR as exactly one positional argument"
-  EXPORT_DIR=$1
+  shift $(( "$OPTIND"-1 ))
 }
 
 check_valid() {
-  if [[ ! -d $EXPORT_DIR ]]; then
-    log "[WARNING] $EXPORT_DIR does not exist, creating..."
-    mkdir -p $EXPORT_DIR
+
+  for arg in EXPORT_DIR WEIGHT_FILE_LOC; do
+    if [[ -z "${!arg}" ]]; then
+      print_help_string
+      err "Missing required argument for ${arg@Q}"
+    fi
+  done
+
+  if [[ ! -d "$EXPORT_DIR" ]]; then
+    warn "$EXPORT_DIR does not exist, creating..."
+    mkdir -p "$EXPORT_DIR"
   fi
 
-  MODEL_CONFIG_LOC=$EXPORT_DIR/config.json
+  MODEL_CONFIG_LOC="$EXPORT_DIR"/config.json
 
-  [[ -f $WEIGHT_FILE_LOC ]] || (err "[ERROR] Missing argument -w or invalid path to weight file" && exit 1)
-
-  # When we do not want to skip compile.
-  if [[ $IREE_BUILD_DIR -eq 0 ]] && [[ $SKIP_COMPILE -eq 0 ]]; then
-    compiler_loc=$(which iree-compile)
-    [[ ! -z $compiler_loc ]] || err "[ERROR] iree-compile not in PATH, and IREE_BUILD_DIR was not provided through '-i'"
-    IREE_BUILD_DIR=$(dirname $compiler_loc)
-    log "[WARNING] Invalid path to IREE build dir. Defaulting to installed python package"
-    log "[INFO] Using iree-compile from $IREE_BUILD_DIR/tools/"
-  fi
-
-  log "[INFO] Using weight file from $WEIGHT_FILE_LOC"
+  [[ -f "$WEIGHT_FILE_LOC" ]] || (err "Missing argument to -w or invalid path to weight file" && exit 1)
+  info "Using weight file from $WEIGHT_FILE_LOC"
 }
 
 sharktank_export() {
-  log "[INFO] Starting Export..."
-  MLIR_PATH=$EXPORT_DIR/model.mlir
-  log "[INFO] Model MLIR will be saved at $MLIR_PATH"
-  log "[INFO] Generated config.json will be placed in $MODEL_CONFIG_LOC"
+  info "Starting Export..."
+  MLIR_PATH="$EXPORT_DIR"/model.mlir
+  info "Model MLIR will be saved at $MLIR_PATH"
+  info "Generated config.json will be placed in $MODEL_CONFIG_LOC"
 
-  python3 -m sharktank.examples.export_paged_llm_v1 --gguf-file=$WEIGHT_FILE_LOC --output-mlir=$MLIR_PATH --output-config=$MODEL_CONFIG_LOC --bs=$EXPORT_BATCH_SIZES 2>&1
+  python3 -m sharktank.examples.export_paged_llm_v1 \
+    --gguf-file="$WEIGHT_FILE_LOC" \
+    --output-mlir="$MLIR_PATH" \
+    --output-config="$MODEL_CONFIG_LOC" \
+    --bs="$EXPORT_BATCH_SIZES" 2>&1
   if [[ $? -ne 0 ]]; then
-    err "[ERROR] Failed to run export"
+    err "Failed to run export"
   fi
 
-  log "[INFO] Successfully exported model to $EXPORT_DIR"
+  info "Successfully exported model to $EXPORT_DIR"
 
 }
 
 shard() {
-    log "[INFO] Sharding enabled. Sharding weights and saving to $ARTIFACT_LOC"
+    info "Sharding enabled. Sharding weights and saving to $ARTIFACT_LOC"
 
-    if [[ $ARTIFACT_LOC -eq 0 ]]; then
-      ARTIFACT_LOC=$EXPORT_DIR/artifacts
-      mkdir -p $ARTIFACT_LOC
-      log "[WARNING] Invalid or missing path to artifacts dir; Defaulting to $EXPORT_DIR/artifacts"
+    if [[ -z "$ARTIFACT_LOC" ]] || [[ ! -d "$ARTIFACT_LOC" ]]; then
+      ARTIFACT_LOC="$EXPORT_DIR"/artifacts
+      mkdir -p "$ARTIFACT_LOC"
+      warn "Invalid or missing path to artifacts dir; Defaulting to $EXPORT_DIR/artifacts"
     fi
 
-    log "[INFO] Setting artifacts dir to $ARTIFACT_LOC"
+    info "Setting artifacts dir to $ARTIFACT_LOC"
 
-    weight_file_stem=$(echo $WEIGHT_FILE_LOC | rev | cut -d / -f 1 | cut --complement -d . -f 1 | rev)
+    weight_file_stem=$(echo "$WEIGHT_FILE_LOC" | rev | cut -d / -f 1 | cut --complement -d . -f 1 | rev)
     shard_file_suffix="_tp8.irpa"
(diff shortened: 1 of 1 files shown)


vinayakdsci:
The function for compilation still needs an implementation.

vinayakdsci:
Compile is now implemented.

rsuderman (approved):

ScottTodd (changes requested):
Nice start! It definitely helps already seeing each step in sequence as part of a script, and this would immediately help with the number of environment variables and steps in the user guide at https://github.com/nod-ai/shark-ai/blob/main/docs/shortfin/llm/user/llama_serving.md.

My largest pieces of feedback is around what language the script is written in - prefer Python given how many options this has, what sorts of other libraries and tools it interoperates with, and how we will want to distribute this to users in our packages.

Beyond that, my other suggestions are about finer user experience points. We could get by without some of those addressed, if we scope and name the tool appropriately, like with `mlc_llm convert_weight` and `trtllm-build`. Something like "shark ai importllm" (with some punctuation) maybe?

run_shark_ai.sh
```diff
@@ -0,0 +1,297 @@
...
+  info "Successfully compiled the model and saved VMFB at $EXPORT_DIR/model.vmfb"
+}
+
+main() {
+  if [[ $# -eq 0 ]]; then
+    print_help_string_and_exit 1
+  fi
+
+  parse_and_handle_args "$@"
+  check_valid
```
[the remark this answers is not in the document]
vinayakdsci: @ScottTodd thanks for the review! The philosophy behind choosing bash for the script was that since we are combining so many different tools, a bash script would  essentially do not much but just act like a glue between the steps.

In fact, the first option was indeed Python, and I do agree it would have had been a more convenient choice. I can port the script to Python too, no worries.

run_shark_ai.sh
```diff
@@ -0,0 +1,297 @@
...
+  fi
+
+  if [[ "$COMPILE" -eq 1 ]]; then
+    if [[ -z "$IREE_BUILD_DIR" ]]; then
+      compiler_loc=$(which iree-compile)
+      [[ ! -f "$compiler_loc" ]] && err "iree-compile not in PATH, and IREE_BUILD_DIR was not provided through '-i'"
+      IREE_BUILD_DIR=$(dirname "$compiler_loc")
+      warn "Missing argument to IREE build dir. Defaulting to installation found in PATH"
+      info "Using iree-compile from $IREE_BUILD_DIR"
+    fi
```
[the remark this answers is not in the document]
vinayakdsci: Okay. This switch was simply for added convenience to the user, but what you are saying does make sense. I will remove the case (and the flag).

run_shark_ai.sh
```diff
@@ -0,0 +1,297 @@
...
+
+  info "Successfully compiled the model and saved VMFB at $EXPORT_DIR/model_tp8.vmfb"
+}
+
+compile_tp1() {
+  info "Starting compilation"
+  "$IREE_BUILD_DIR"/iree-compile "$EXPORT_DIR"/model.mlir \
+               --iree-hal-target-backends=rocm \
+               --iree-hip-target=gfx942 \
+               -o "$EXPORT_DIR"/model.vmfb 2>&1
```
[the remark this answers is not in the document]
vinayakdsci: This I totally agree with. I will put up a commit that allows switching targets.

run_shark_ai.sh
```diff
@@ -0,0 +1,297 @@
...
+        --iree-dispatch-creation-enable-aggressive-fusion=true     \
+        --iree-global-opt-propagate-transposes=true  \
+        --iree-opt-aggressively-propagate-transposes=true     \
+        --iree-opt-data-tiling=false  \
+        --iree-preprocessing-pass-pipeline='builtin.module(util.func(iree-preprocessing-generalize-linalg-matmul-experimental))'     \
+        --iree-hal-indirect-command-buffers=true  \
+        --iree-stream-resource-memory-model=discrete  \
+        --iree-hip-legacy-sync=false    \
+        --iree-hal-memoization=true  \
+        --iree-opt-strip-assertions \
```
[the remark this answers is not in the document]
vinayakdsci: These flags are very much open to discussion. These are what we currently use for compiling for benchmarking tp8 mode. I am sure some of these can be removed.

run_shark_ai.sh
```diff
@@ -0,0 +1,297 @@
...
+
+
+print_help_string() {
+  printf "%s\n" "USAGE: run_shark_ai [OPTIONS] ... -o [EXPORT_DIR]
+Exports the model using shark-ai and compiles the IR to produce the VMFB module.
+
+Some options are required to be specified.
+    -h            prints the help string. The same output is emitted when no arguments are given.
+    -v            enables verbose logging. When not specified, only errors are logged.
+    -w            the location of the GGUF/IRPA file(s) that contain the parameters. (required)
```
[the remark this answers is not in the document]
vinayakdsci: This also I thought of, but @kumardeepakamd's POV was that the weight file should be handled by the user however they would like, and not be a part of the script in any way, so the idea was dropped.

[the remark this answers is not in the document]
vinayakdsci: Yes. The script does not handle running shortfin (yet). Again the idea was that the user might have a different workflow than what the script enforces, do that was not included.

vinayakdsci:
@ScottTodd thanks a lot for the review. I completely agree with your views on moving to Python, so I mam closing this PR in favor of a new one that implements the script in Python.

ScottTodd:
> @ScottTodd thanks a lot for the review. I completely agree with your views on moving to Python, so I mam closing this PR in favor of a new one that implements the script in Python.

Why close this PR and create a new one? The review history and GitHub crossreferences (linking to https://github.com/nod-ai/shark-ai/issues/402 and https://github.com/nod-ai/shark-ai/issues/691) are all useful to keep here. Splitting to a new PR fragments the discussion. You can rename the PR to better reflect what it is about (e.g. "Introduce new tool for LLM import and compilation"), without being overly specific about one detail of the implementation (a single file written in Bash).

vinayakdsci:
@ScottTodd re-opened the PR.

shark-ai/export_and_serve.py
```diff
@@ -0,0 +1,412 @@
...
+        help="batch sizes for export. Multiple batch sizes should be separated by a ','.",
+    )
+    parser.add_argument(
+        "-i",
+        "--ir",
+        type=Path,
+        help="location for the MLIR to be compiled, if compilation is done independently.",
+    )
+    parser.add_argument(
+        "-p",
```
archana-ramalingam, line 400: Nit: Is there a reason we call this -p instead of -tp, which is more common?

shark-ai/export_and_serve.py
```diff
@@ -0,0 +1,412 @@
...
+
+    base_pipeline.exec_pipeline()
+
+
+def _get_argparser() -> CliParser:
+    msg = "Utility script to combine shark-ai tools"
+    # The argparser should not print the usage when an error occurs.
+    # We handle that ourselves.
+    parser = CliParser(
+        prog="export_and_serve.py",
```
ScottTodd, line 338: A note on serving: plan on there being checkpoint steps after

1. Downloading weights
2. Compiling the model (I'd group the export from PyTorch to .mlir _and_ the .mlir to .vmfb compilation in this, from a user's perspective)
3. Serving the model

A user may want to serve the model on multiple machines, not just the machine that was used to compile. If we can package the artifacts for serving (`program.vmfb`, `weights_shard1.irpa`, `tokenizer_config.json`) in some organized way then we can help users push those artifacts and start the server.

See for example how https://llm.mlc.ai/docs/compilation/compile_models.html is structured:

* There is a single `mlc_llm` tool with mutliple modes: `compile`, `convert_weight`, `gen_config`
* The running/serving step points at the directory of outputs

We could do something similar here using sub-commands: https://docs.python.org/3/library/argparse.html#sub-commands

shark-ai/export_and_serve.py
```diff
@@ -0,0 +1,412 @@
...
+        "--artifact-dir",
+        type=Path,
+        help="the location where the artifacts (sharded weights) should be saved. Defaults to EXPORT_DIR/artifacts/",
+    )
+    parser.add_argument(
+        "-s",
+        "--shard",
+        action="store_true",
+        help="shard the weight file in tp8 mode and export to MLIR.",
+    )
```
ScottTodd, line 386: Seems like `--export` and `--shard` could be replaced with a single `--shard_count` option? Would values other than 1 and 8 make sense? What about 2, 4, or 16?

If the options are kept, should they be mutually exclusive? The branches in the code seem to suggest that both could be set.

shark-ai/export_and_serve.py
```diff
@@ -0,0 +1,412 @@
...
+        help="location for the MLIR to be compiled, if compilation is done independently.",
+    )
+    parser.add_argument(
+        "-p",
+        "--tensor-parallel",
+        type=int,
+        default=1,
+        choices=[1, 8],
+        help="tensor parallel size (required for independent compilation).",
+    )
```
ScottTodd, line 406: What do you mean by "independent compilation"? You could also use argument groups for options like these to keep the help output more organized.

https://docs.python.org/3/library/argparse.html#argument-groups

* Compilation and export options (compilation target, extra flags, sharding, batch sizes)
* Development/debugging options (extra intermediate output)

shark-ai/export_and_serve.py
```diff
@@ -0,0 +1,412 @@
...
+    return parser
+
+
+if __name__ == "__main__":
+    parser = _get_argparser()
+    main(parser.parse_args(args=None if sys.argv[1:] else ["--help"]))
```
ScottTodd, line 412: Oh, this is interesting. I haven't seen this pattern before of default to `--help` if the program is called with no arguments. I think I like that, but you could also just let argparse do its thing and complain about missing required arguments.

shark-ai/export_and_serve.py
```diff
@@ -0,0 +1,412 @@
...
+            )
+            exit(1)
+
+        if not os.path.isfile(self.weight_loc):
+            logger.error(f"File {str(self.weight_loc)} does not exist")
+            raise ValueError(
+                f"Invalid path to weight file {str(self.weight_loc)}. File does not exist."
+            )
+
+        if not os.path.isdir(self.export_dir):
```
ScottTodd, line 97: use `pathlib` not `os`
```suggestion
        if not self.export_dir.is_dir():
```

shark-ai/export_and_serve.py
```diff
@@ -0,0 +1,412 @@
...
+            logger.warning(
+                f"Export directory {str(self.export_dir)} does not exist, creating..."
+            )
+            try:
+                Path.mkdir(self.export_dir, parents=True)
```
ScottTodd, line 102: This is a member function that you can call on the object, rather than pass `self` like this

```suggestion
                self.export_dir.mkdir(parents=True)
```

shark-ai/export_and_serve.py
```diff
@@ -0,0 +1,412 @@
...
+            except:
+                logger.error("Failed to create export directory")
+                raise
+
+        if self.shard and not os.path.isdir(self.artifacts_dir):
+            logger.warning(
+                f"Artifacts directory {str(self.artifacts_dir)} does not exist, creating..."
+            )
+            try:
+                Path.mkdir(self.artifacts_dir, parents=True)
```
ScottTodd, line 112: same comments about `pathlib` usage
```suggestion
        if self.shard and not self.artifacts_dir.is_dir():
            logger.warning(
                f"Artifacts directory {str(self.artifacts_dir)} does not exist, creating..."
            )
            try:
                self.artifacts_dir.mkdir(parents=True)
```

shark-ai/export_and_serve.py
```diff
@@ -0,0 +1,412 @@
...
+            except:
+                logger.critical("Failed to create artifacts directory")
+                raise
+
+        if self.export:
+            self.mlir_path = Path(os.path.join(self.export_dir, "model.mlir"))
+            self.model_config_path = Path(os.path.join(self.export_dir, "config.json"))
```
ScottTodd, line 119: These are already `Path` objects, so you can join them with `/`
```suggestion
            self.mlir_path = self.export_dir / "model.mlir"
            self.model_config_path = self.export_dir / "config.json"
```

See https://docs.python.org/3/library/pathlib.html#operators

shark-ai/export_and_serve.py
```diff
@@ -0,0 +1,412 @@
...
+        logger.info(
+            f"Successfully exported and saved model MLIR and config to {self.export_dir}"
+        )
+
+    def compile_model(self) -> None:
+        if not shutil.which("iree-compile"):
+            logger.error("iree-compile not found in PATH")
+            raise FileNotFoundError(
+                "iree-compile binary is required to be in PATH for compilation."
+            )
```
ScottTodd, line 227: May want to check this during startup, rather than wait until after export

shark-ai/export_and_serve.py
```diff
@@ -0,0 +1,412 @@
...
+            logger.error("Failed to shard model to TP8")
+            print(shard_subp.stderr.decode(), file=sys.stderr)
+            exit(shard_subp.returncode)
+
+        logger.info(
+            f"Successfully sharded weight file; Unranked weight file located at {shard_file}"
+        )
+
+        # TODO(vinayakdsci): Refactor this into a separate function.
+        logger.info("Starting TP8 model export")
```
ScottTodd, line 199: Could log information about how long this is expected to take (ideally there would be a progress meter too)

shark-ai/export_and_serve.py
```diff
@@ -0,0 +1,412 @@
...
+                "python3",
+                "-m",
+                "sharktank.examples.export_paged_llm_v1",
+                f"--irpa-file={shard_file}",
+                f"--output-mlir={str(self.mlir_path)}",
+                f"--output-config={self.model_config_path}",
+                "--bs=4",
+            ],
+            capture_output=True,
+        )
```
ScottTodd, line 211: Prefer to call the Python functions directly in-process instead of using `subprocess`. That way you get integrated error handling and ensure that the same python interpreter is used (this could also use `sys.executable` instead of `python3`

shark-ai/export_and_serve.py
```diff
@@ -0,0 +1,412 @@
...
+
+            logger.info("TP1 export is enabled")
+            logger.info(f"TP1 MLIR will be saved to {self.mlir_path}")
+            logger.info(f"TP1 model config will be saved to {self.model_config_path}")
+
+        if self.shard:
+            self.mlir_path = Path(os.path.join(self.export_dir, "model.mlir"))
+            self.model_config_path = Path(
+                os.path.join(self.export_dir, "config_tp8.json")
+            )
```
ScottTodd, line 129: pathlib
```suggestion
            self.mlir_path = self.export_dir / "model.mlir"
            self.model_config_path = self.export_dir / "config_tp8.json"
            )
```

But to some of my other comments about `export` and `shard` being sorta conflicting and overlapping, I think this could be collapsed down to a single code path. You could encode some information about tp ("tensor parallel") / sharding in either the file name and/or the .json file itself as needed.

shark-ai/export_and_serve.py
```diff
@@ -0,0 +1,412 @@
...
+
+        if self.shard or self.tp == 8:
+            self._compile_tp8()
+        else:
+            self._compile_tp1()
+
+    def _compile_tp1(self):
+        logger.info("Compiling unsharded IR")
+        tp1_vmfb_path = os.path.join(self.export_dir, "model.vmfb")
```
ScottTodd, line 236: pathlib

```suggestion
        tp1_vmfb_path = self.export_dir / "model.vmfb"
```

shark-ai/export_and_serve.py
```diff
@@ -0,0 +1,412 @@
...
+            print(compile_subp.stderr.decode(), file=sys.stderr)
+            exit(compile_subp.returncode)
+
+        logger.info(
+            f"Successfully compiled and saved TP1 VMFB module to {tp1_vmfb_path}"
+        )
+
+    def _compile_tp8(self):
+        logger.info("Compiling sharded IR")
+        tp8_vmfb_path = os.path.join(self.export_dir, "model_tp8.vmfb")
```
ScottTodd, line 265: pathlib

```suggestion
        tp8_vmfb_path = self.export_dir / "model_tp8.vmfb"
```

shark-ai/export_and_serve.py
```diff
@@ -0,0 +1,412 @@
...
+        # TODO(vinayakdsci): Add a flag to support backends other than rocm.
+        cmd += [f"--iree-hal-target-backends={self.hal_target_backend}"]
+
+        # TODO(vinayakdsci): Add a flag to support targets other than gfx942.
+        cmd += [f"--iree-hip-target={self.hip_target}"]
+
+        compile_subp = subprocess.run(
+            cmd,
+            capture_output=True,
+        )
```
ScottTodd, line 252: We may want this to use the IREE compiler API _or_ the `iree.build` API (https://iree-python-api.readthedocs.io/en/latest/compiler/build.html) instead of using a subprocess to call `iree-compile`.

Then:

* we wouldn't need to check for `iree-compile` on the PATH
* developers wanting to bring their own `iree-compile` build would need to build the python bindings
* we'd get the benefits of iree.build for compiling multiple programs in a build graph (as needed... LLMs don't need that as much as SDXL though)
* we'd get integrated error handling instead of checking return codes and sys.stderr

shark-ai/export_and_serve.py
```diff
@@ -0,0 +1,412 @@
...
+            logger.error("Failed to compile unsharded IR")
+            print(compile_subp.stderr.decode(), file=sys.stderr)
+            exit(compile_subp.returncode)
+
+        logger.info(
+            f"Successfully compiled and saved TP1 VMFB module to {tp1_vmfb_path}"
+        )
+
+    def _compile_tp8(self):
+        logger.info("Compiling sharded IR")
```
ScottTodd, line 264: May also want to include a time estimate here, or ideally a progress bar (https://github.com/iree-org/iree/issues/14369)

shark-ai/export_and_serve.py
```diff
@@ -0,0 +1,412 @@
...
+            f"Successfully compiled and saved TP8 VMFB module to {tp8_vmfb_path}"
+        )
+
+
+def main(args: Namespace) -> None:
+    compile = args.compile
+    artifacts_dir = (
+        args.artifact_dir
+        if args.artifact_dir is not None
+        else Path(os.path.join(args.export_dir, "artifacts/"))
```
ScottTodd, line 297: pathlib
```suggestion
        else args.export_dir / "artifacts"
```

shark-ai/export_and_serve.py
```diff
@@ -0,0 +1,412 @@
...
+
+        logger.info(
+            f"Successfully exported and saved model MLIR and config to {self.export_dir}"
+        )
+
+    def export_tp8(self) -> None:
+        shard_file = os.path.join(
+            str(self.artifacts_dir),
+            str(self.weight_loc).split("/")[-1][:-5] + "_tp8.irpa",
+        )
```
ScottTodd, line 176: use pathlib here. Join with `/` and replace the `[-1][:-5]` with some of the functions from https://docs.python.org/3/library/pathlib.html#corresponding-tools (`.stem()`, `.parent()`, etc.)

run_shark_ai.sh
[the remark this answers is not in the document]
ScottTodd: Brainstorming more name choices...

Use the project name as the tool name then have a subcommand for each task:

* `shark-ai compile_llm`
* `shark-ai serve`

Put `-build` in the name, like trtllm:

* `shark-ai-build compile_llm`
* `shark-ai-build compile_sdxl`

I'm not sure if we should try to fit everything that we support into a neat box like those, or if we should embrace having more bespoke pipelines and servers, at least for our current split between LLMs (llama) and image generators (SDXL, Flux).

If we pick a sufficiently specific tool name like then we can wiggle out of it later. If we pick a generic name like `shark-ai` then we should continue to use that across future releases. I definitely want to aim for something that we include as a console script that gets installed as part of the Python packages.

ScottTodd (changes requested):
Thanks for rewriting in Python! This looks much easier to maintain and I see no portable issues. Comments are split between Python style and overall user experience details. Some of this we can iterate on after landing an initial version, but I want to get a solid foundation before we advertise this to users.

shark-ai/export_and_serve.py
```diff
@@ -0,0 +1,412 @@
...
+  -c, --compile                     compile the exported model as part of the pipeline. Default is FALSE.
+  -w, --weight-file WEIGHT_FILE     the location of the GGUF/IRPA file(s) that contain the parameters.
+  -e, --export                      export the model in tp1 mode.
+  -a, --artifact-dir ARTIFACT_DIR   the location where the artifacts (sharded weights) should be saved. Defaults to EXPORT_DIR/artifacts/
+  -s, --shard                       shard the weight file in tp8 mode and export to MLIR.
+  -b, --batch-sizes BATCH_SIZES     batch sizes for export. Multiple batch sizes should be separated by a ','.
+  -i, --ir IR                       location for the MLIR to be compiled, if compilation is done independently.
+  -p, --tensor-parallel {1,8}       tensor parallel size. Used for independent compilation. Defaults to 1.
+        """
+        _ = file.write(help_text + "\n")
```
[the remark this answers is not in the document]
vinayakdsci: @ScottTodd the reason to have a separate help string is that the default format is actually quite unclean and difficult to read. It repeats options against each flag (the abbreviation and the long-form) and then wraps the help string for that flag onto the next line. Since this is supposed to be user-facing, IMO this would be an import part of the UX.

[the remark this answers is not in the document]
ScottTodd: Explicit help text:
```
λ python .\shark-ai\export_and_serve.py --help
usage: export_and_serve.py [-h] [-v] [-c] -w WEIGHT_FILE [-e] [-a ARTIFACT_DIR] [-s] [-b BATCH_SIZES] [-i IR] [-p {1,8}] export_dir

Utility script to combine shark-ai tools

positional arguments:
  export_dir                        the directory where the exported artifacts will be saved.

options:
  -h, --help                        show this help message and exit
  -v, --verbose                     set logging level to INFO. The default logging level is WARNING.

  -c, --compile                     compile the exported model as part of the pipeline. Default is FALSE.
  -w, --weight-file WEIGHT_FILE     the location of the GGUF/IRPA file(s) that contain the parameters.
  -e, --export                      export the model in tp1 mode.
  -a, --artifact-dir ARTIFACT_DIR   the location where the artifacts (sharded weights) should be saved. Defaults to EXPORT_DIR/artifacts/
  -s, --shard                       shard the weight file in tp8 mode and export to MLIR.
  -b, --batch-sizes BATCH_SIZES     batch sizes for export. Multiple batch sizes should be separated by a ','.
  -i, --ir IR                       location for the MLIR to be compiled, if compilation is done independently.
  -p, --tensor-parallel {1,8}       tensor parallel size. Used for independent compilation. Defaults to 1.
```

Default help text (line width 100):
```
λ python .\shark-ai\export_and_serve.py --help
usage: export_and_serve.py [-h] [-v] [-c] -w WEIGHT_FILE [-e] [-a ARTIFACT_DIR] [-s]
                           [-b BATCH_SIZES] [-i IR] [-p {1,8}]
                           export_dir

Utility script to combine shark-ai tools

positional arguments:
  export_dir            the directory where the exported artifacts will be saved.

options:
  -h, --help            show this help message and exit
  -v, --verbose         set logging level to INFO. The default logging level is WARNING.
  -c, --compile         compile the exported model as part of the pipeline. Default is FALSE.
  -w WEIGHT_FILE, --weight-file WEIGHT_FILE
                        the location of the GGUF/IRPA file(s) that contain the parameters.
  -e, --export          export the model in tp1 mode.
  -a ARTIFACT_DIR, --artifact-dir ARTIFACT_DIR
                        the location where the artifacts (sharded weights) should be saved.
                        Defaults to EXPORT_DIR/artifacts/
  -s, --shard           shard the weight file in tp8 mode and export to MLIR.
  -b BATCH_SIZES, --batch-sizes BATCH_SIZES
                        batch sizes for export. Multiple batch sizes should be separated by a ','.
  -i IR, --ir IR        location for the MLIR to be compiled, if compilation is done
                        independently.
  -p {1,8}, --tensor-parallel {1,8}
                        tensor parallel size (required for independent compilation).
```

I prefer the default line wrapping behavior, especially at line width 80, since it left pads the wrapped lines. I see your point about the repetition, but sticking to standard argparse behavior will make the tool output more predictable and simplify maintenance.

PR #738, opened 2025-01-02
[shortfin] Zero out kv cache pages during allocation

rsuderman:
First decode sometimes resulted in bad decode values. This is likely related to bad values in the KV cache. Zeroing should avoid nan / inf corruption for uninitialized memory.


renxida (approved):
lgtm. we should probably get this in regardless of CI status just to eliminate one potential cause of problems.

PR #754, opened 2025-01-04
Update nightly llama benchmarking tests

aviator19941:
- Updates nightly llama benchmarking tests to benchmark input token lengths of 128 and 2048 for llama 8b, 70b, and 405b.
- Switch IREE compile flag from `--iree-hal-target-backends` to `--iree-hal-target-device`

TODO: Add 405b decode benchmark calls to 405b fp16 tests when decode is fixed


.github/workflows/ci-llama-large-tests.yaml
```diff
@@ -7,6 +7,7 @@
 name: Llama Benchmarking Tests
 
 on:
+  pull_request:
```
ScottTodd, line 10: This is still running after almost 4 hours: https://github.com/nod-ai/shark-ai/actions/runs/12639047256/job/35216569084 , resulting in a bit of a queue for the `llama-mi300x-1` runner: https://github.com/nod-ai/shark-ai/actions?query=is%3Aqueued. Is that expected?

archana-ramalingam: This patch is expected to resolve the failing Llama benchmark nightly in sharktank. Since it's running various tests for 70B and 405B it's extremely slow. We are only testing here if the patch works as expected. Looks like it failed.

PR #756, opened 2025-01-06
[tuner] Remove legacy tuner path

Max191:
This PR removes the old code paths for the tuner following the large refactoring done in https://github.com/nod-ai/shark-ai/pull/606 and https://github.com/nod-ai/shark-ai/pull/704.
- The old compilation and benchmarking logic for models and dispatches are now gone, and the unified `benchmark` and `compile` functions should be used instead.
- Much of the dispatch parsing logic is removed, and dispatch parsing is now done by 2 DispatchParser implementations for contraction ops and convolution ops.
- The example tuning clients in `tuner/examples` for dispatches and punet are removed, since they use the old path. The new example to follow is in `tuner/examples/test`.
- The candidate generation `tune` function is removed, and the `generate_configs_and_td_specs` should be used to generate candidates instead.
- Many utility functions, structs, and struct fields are now removed, since they are no longer used.
- All tests testing the old path are removed.


tuner/tuner/libtuner.py
```diff
@@ -66,78 +60,32 @@
...
-    compiled_dispatch_path: Optional[Path] = None
-    compiled_dispatch_hash: Optional[str] = None
-    first_benchmark_time: Optional[float] = None
-    first_benchmark_device_id: Optional[str] = None
     spec_path: Optional[Path] = None
-    compiled_model_path: Optional[Path] = None
-    compiled_model_hash: Optional[str] = None
-    model_benchmark_time: Optional[float] = None
-    model_benchmark_device_id: Optional[str] = None
-    baseline_benchmark_time: Optional[float] = None
```
kuhar, line 89: Why did you remove benchmark times from here?

kuhar (commented):
So much red, fantastic

tuner/tuner/libtuner.py
Max191: This was all used to carry state for benchmark parsing, but since benchmark parsing happens in the iree bindings, it's not necessary anymore. The times are logged in the benchmark function.

tuner/tuner/candidate_gen.py
```diff
@@ -36,65 +31,11 @@
 from .dispatch_constraints import *
 from .dispatch_parser import *
 from .spec_builder import *
+from . import libtuner
```
kuhar, line 34: Isn't this a circular dependency?

tuner/tuner/candidate_gen.py
```diff
@@ -727,20 +268,22 @@ def main():
...
-        args.output,
-        args.limit,
-        args.num_subgroups,
-        args.lhs_dims,
-        args.rhs_dims,
-        args.tile_dims,
-    )
+    with ir.Context() as ctx:
+        tuner_ctx = TunerContext(ctx, tune_logger)
+        mlir_text = libtuner.strip_compilation_info(args.input)
```
kuhar, line 273: Should we move this function to some common header to avoid this dependency on libtuner?

kuhar (commented):
Just one suggestion, looks good otherwise

tuner/tuner/candidate_gen.py
Max191: I chose not to because I would have to move a bunch of functions over. Using absolute imports like I did above does not actually create a circular dependency, though, so it should be fine like this (doing `from .libtuner import strip_compilation_info` would cause the circular dependency). If you'd prefer, I can move everything over and remove the import, but it works fine like this.

PR #758, opened 2025-01-06
Merge fixtures: available_port into llm_server

renxida:
The available_ports fixture seems to rely on the servers booting very quickly to work. Merging it into the llm_server & increasing the server startup timeout fixes the remaining CI failures caused by updating IREE to 1220.


stbaione (approved):
LGTM, assuming CI passes

PR #759, opened 2025-01-06
Drop path filters from shortfin and SDXL CI workflows.

ScottTodd:
We aren't particularly resource constrained and I'd rather have continuous signal for these workflows.

The only workflow that still uses path filters is [`.github/workflows/ci-tuner.yml`](https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/ci-tuner.yml).

### DIFF
--- a/.github/workflows/ci-sdxl.yaml
+++ b/.github/workflows/ci-sdxl.yaml
@@ -9,15 +9,9 @@ name: CI - shortfin - SDXL
 on:
   workflow_dispatch:
   pull_request:
-    paths:
-      - '.github/workflows/ci-sdxl.yaml'
-      - 'shortfin/**'
   push:
     branches:
       - main
-    paths:
-      - '.github/workflows/ci-sdxl.yaml'
-      - 'shortfin/**'
 
 permissions:
   contents: read
--- a/.github/workflows/ci_linux_x64_asan-libshortfin.yml
+++ b/.github/workflows/ci_linux_x64_asan-libshortfin.yml
@@ -9,15 +9,9 @@ name: CI - shortfin - ASan
 on:
   workflow_dispatch:
   pull_request:
-    paths:
-      - '.github/workflows/ci_linux_x64_asan-libshortfin.yml'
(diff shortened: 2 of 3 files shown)


.github/workflows/ci-sdxl.yaml
ScottTodd, line 1: > We aren't particularly resource constrained and I'd rather have continuous signal for these workflows.

Hmm... well, since this workflow has `runs-on: mi300x-4`, it's been queued for 20 minutes, currently stuck behind https://github.com/nod-ai/shark-ai/actions/runs/12637108441/job/35210632193.

I wonder if this can just on any mi300, or if it needs that specific runner. I see we have a bunch of runners with different labels in this repository. Can we consolidate some labels to be able to request just "any MI300"?

@saienduri @amd-chrissosa @yamiyysu

amd-chrissosa: I don't have a lot of context on this specific workflow but theoretically any workload without heavy storage/networking requirements that needs a MI300x should be able to be migrated to what @yamiyysu is porting.

saienduri: Yes, we should be able to port this one with no problem. Also, we can consolidate all 3 of our lab machines onto the same mi300 label. Are all the paths hardcoded in workflow files consistent across lab machines? @archana-ramalingam @aviator19941

archana-ramalingam: On MI300x-1, weights are at `/data/llama3.1/weights/8b`, MI300x-3, they are at `/data/llama3.1/8b`. We need to make this consistent across all systems.

ScottTodd: There should be no hardcoded paths. Everything should use environment variables and cache directories, with scripts being used to populate the cache on any machine as needed.

ScottTodd:
Abandoning this PR for now. Pinned versions together with automated dependency updates should help keep the SDXL workflow passing and using recent versions of the tools/packages.

Issue #760, opened 2025-01-06
Test both pinned and unpinned versions of IREE dependencies


ScottTodd:
We have many workflows testing only unpinned versions of the IREE packages:, such as: https://github.com/nod-ai/shark-ai/blob/5f08cb2354c928476ace3cfdcda3b835dd13cd61/.github/workflows/ci-sharktank.yml#L79-L84

Some workflows explicitly test pinned versions like https://github.com/nod-ai/shark-ai/blob/5f08cb2354c928476ace3cfdcda3b835dd13cd61/.github/workflows/ci-sglang-benchmark.yml#L71-L75

Having inconsistent version pinning results in fragmented PRs updating pins like https://github.com/nod-ai/shark-ai/pull/757, https://github.com/nod-ai/shark-ai/pull/746, and https://github.com/nod-ai/shark-ai/pull/721. 

We should further consolidate where version pins are defined and also refactor some workflows to test both pinned and unpinned versions. Workflows that pin the versions can be marked as "required checks" and give us confidence that workflow failures are a result of the changes in a PR/commit and _not_ due to changes in dependencies. We do still want early and regular signal for upcoming API breaks, regressions, and other issues coming from dependencies though, so these versions of workflows could still run on pull requests or at least on schedules (e.g. nightly).

ScottTodd:
Brainstorming a few strategies for this...

## Possible strategies

### A) Test with only pinned, use dependabot to send PRs that try new versions

This would involve switching all workflows that install packages to use requirements files with pinned versions in them. Then we would have some automation (likely dependabot, but there are other options too) send pull requests at some regular frequency attempting to bump to the latest versions.

References:

* https://docs.github.com/en/code-security/dependabot
* https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates
* https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates
* https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/optimizing-pr-creation-version-updates
* https://docs.github.com/en/code-security/getting-started/dependabot-quickstart-guide
* https://github.blog/news-insights/product-news/keep-all-your-packages-up-to-date-with-dependabot/
* https://github.com/dependabot

### B) Add a `matrix` to each job to run with multiple different versions

[...]

PR #762, opened 2025-01-06
Delegate IREE source fetching to shortfin/CMakeLists.txt.

ScottTodd:
This will make it easier to update the ref. This will slow workflows down a bit since FetchContent always fetches all submodules.

This simplification will help with https://github.com/nod-ai/shark-ai/issues/760.


shortfin/requirements-iree-compiler.txt
```diff
@@ -1,4 +1,4 @@
-# Keep in sync with "ref: iree-" in .github/workflows/* and GIT_TAG in CMakeLists.txt
+# Keep in sync with SHORTFIN_IREE_GIT_TAG in CMakeLists.txt
 -f https://iree.dev/pip-release-links.html
 iree-base-compiler==3.1.0rc20241204
 iree-base-runtime==3.1.0rc20241204
```
ScottTodd, line 4: @marbre did you experiment with having the CMake file pull from this requirements file?

I guess we can also just add a comment in this other location pointing back here too:
https://github.com/nod-ai/shark-ai/blob/5f08cb2354c928476ace3cfdcda3b835dd13cd61/shortfin/CMakeLists.txt#L46-L47

marbre: Yeah, some time ago I sent #502. Happy to provide more context (and do the actual review) later.

ScottTodd: We discussed this a bit offline. I think having two locations that point to each other is pretty manageable without any extra tooling support. Having five locations with only some pointing to the others is not. So for now I think keeping the requirements and CMakeLists only loosely linked via comments is sufficient.

shortfin/CMakeLists.txt
```diff
@@ -44,6 +44,7 @@ add_compile_options("$<$<C_COMPILER_ID:MSVC>:/utf-8>")
 add_compile_options("$<$<CXX_COMPILER_ID:MSVC>:/utf-8>")
 
 # Pins
+# Keep in sync with requirements-iree-compiler.txt.
```
ScottTodd, line 47: Update this if moving the requirements file in https://github.com/nod-ai/shark-ai/pull/765.

marbre (approved):
> * Workflows will be several seconds slower since FetchContent always fetches all submodules

It actually not as bad as we have set
https://github.com/nod-ai/shark-ai/blob/922b1c2f6fae6467c0222cacef46ef3297f58830/shortfin/CMakeLists.txt#L234
as well as
https://github.com/nod-ai/shark-ai/blob/922b1c2f6fae6467c0222cacef46ef3297f58830/shortfin/CMakeLists.txt#L243


> * The `SHORTFIN_IREE_SOURCE_DIR` option is no longer tested

That's the only downside I see but we can read a workflow later if we really need it.

PR #765, opened 2025-01-06
Move requirements-iree-*.txt to top level.

ScottTodd:
This is prep work for https://github.com/nod-ai/shark-ai/issues/760.

I also considered putting the files under `build_tools/` or `shark-ai/`, but we already have a few requirements files in the repository root. Still not as many as https://github.com/vllm-project/vllm though 😛.

### DIFF
--- a/.github/workflows/ci-libshortfin.yml
+++ b/.github/workflows/ci-libshortfin.yml
@@ -119,12 +119,12 @@ jobs:
         cache: "pip"
         cache-dependency-path: |
           'shortfin/requirements-tests.txt'
-          'shortfin/requirements-iree-compiler.txt'
+          'requirements-iree-pinned.txt'
     - name: Install Python packages
       working-directory: ${{ env.LIBSHORTFIN_DIR }}
       run: |
         pip install -r requirements-tests.txt
-        pip install -r requirements-iree-compiler.txt
+        pip install -r ../requirements-iree-pinned.txt
         pip freeze
 
     - name: Build shortfin
--- a/.github/workflows/ci-sdxl.yaml
+++ b/.github/workflows/ci-sdxl.yaml
@@ -65,7 +65,7 @@ jobs:
       working-directory: ${{ env.LIBSHORTFIN_DIR }}
       run: |
         pip install -r requirements-tests.txt
-        pip install -r requirements-iree-compiler.txt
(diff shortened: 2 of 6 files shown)


marbre (approved):
The failing workflow either pull in unpinned versions or if using the pinned version at least install what is expected, thus the patch does what it should.

I am still not 100% convinced as the `requirements-iree-compiler.txt` was supposed to keep optional (!) dependencies for shortfin only and if a user wants to install those, at least I would expect or look in the shortfin subdir. However, for what we aim for right now this is probably a huge improvement even though we might iterate later.

shortfin/CMakeLists.txt
```diff
@@ -43,8 +43,10 @@ endif()
 add_compile_options("$<$<C_COMPILER_ID:MSVC>:/utf-8>")
 add_compile_options("$<$<CXX_COMPILER_ID:MSVC>:/utf-8>")
 
-# Pins
-# Keep in sync with requirements-iree-compiler.txt.
+# Version pins for dependencies.
+# Prefer to keep the IREE git tag synced with the Python package version in the
+# requirements-iree-pinned.txt file. At a minimum, the compiler from those
+# packages must be compatible with the runtime at this source ref.
```
ScottTodd, line 49: FYI @marbre I added on to the comment here

ScottTodd:
> I am still not 100% convinced as the `requirements-iree-compiler.txt` was supposed to keep optional (!) dependencies for shortfin only and if a user wants to install those, at least I would expect or look in the shortfin subdir. However, for what we aim for right now this is probably a huge improvement even though we might iterate later.

How about having a `requirements-iree-unpinned.txt` and `requirements-iree-pinned.txt` in both `shortin/` and `sharktank/`? I'm expecting that we'll keep iterating a bit. Might just start with this PR as-is.

marbre:
> How about having a `requirements-iree-unpinned.txt` and `requirements-iree-pinned.txt` in both `shortin/` and `sharktank/`? I'm expecting that we'll keep iterating a bit. Might just start with this PR as-is.

Seems my comment got lost (?). My concern might be not worth it as we don't pull in requirements via requirements files in our wheel (we rather have it in the `pyproject.toml` files). It what more like to give background on the initial motivation. I think we should keep it easy and lightweight and go with what you have. If we really want to split it up, we can do that in a follow up.

PR #775, opened 2025-01-07
Mark sharktank vae iree test as xfail

IanNod:
Xfails vae until iree runtime issue on the CI machine is fixed. Resolves sharktank data-dependent tests


sharktank/tests/models/vae/vae_test.py
```diff
@@ -100,6 +100,9 @@ def testCompareF16EagerVsHuggingface(self):
 
         torch.testing.assert_close(ref_results, results)
 
+    @pytest.mark.xfail(
+        reason="Waiting on fix for https://github.com/iree-org/iree/issues/19623"
+    )
```
ScottTodd, line 105: It sounds like this test succeeds on some machines? May want to skip instead of xfail in that case, so developers can run the tests locally and get expected results. Or, are we running pytest without strict xfail checking? That may be fine then.

ScottTodd (approved):
nit: Can you type a bit more of a sentence in the PR title?

How about "Mark sharktank VAE test as xfail." ?

sharktank/tests/models/vae/vae_test.py
IanNod: Yeah, only machine I've been able to recreate the failure is on the CI system. We are not running with strict xfail checking. I figured it would be better for developers to run and see it xpass/xfail on different systems to get better test feedback.

archana-ramalingam (approved):

PR #780, opened 2025-01-08
Convert `ci-shark-ai.yml` to use `pkgci_shark_ai.yml` so that we only build packages once

renxida:
This builds on #625, #589 to make progress on issue #584.

This adds a pkgci.yml to run multiple package-based CI tasks after building package using Scott's changes in #667. This gives us the following benefits:

* Integration test workflows are faster because they now use dev packages, without needing to build them from source or use editable installs. Also, if more integration tests are added, they can reuse the built packages.
* Users and developers can access the same dev packages to reproduce CI results
* Only one runner needs the build requirements (potentially including clang, ninja, CMake, Rust, etc.), other runners only need Python.

This also switches to using uv to create venvs, which is faster.

This PR brings shortfin CPU LLM CI time to roughly half an hour on the mi250 runner to a few seconds of package build (fast due to caching) and around 5 minutes of testing.


.github/workflows/pkgci.yml
```diff
@@ -0,0 +1,37 @@
+# Copyright 2024 Advanced Micro Devices, Inc.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+name: PkgCI
+
+on:
+  workflow_dispatch:
```
ScottTodd, line 10: FYI you can test workflows using this `workflow_dispatch` trigger without needing to send a pull request. You can also run on any triggers you want from your fork, so long as you don't try to use self-hosted runners that only exist in this repository.

I usually do that whenever I'm iterating on workflows to avoid generating too much noise in the main repository and triggering unrelated workflows when they aren't configured with something like the "skip ci" that you tried ;). Workflows in IREE are configurable via git trailers: https://iree.dev/developers/general/contributing/#ci-behavior-manipulation, but we haven't built anything like that in this repository yet.

renxida: skipci did work to skip all the workflows, but my problem was that the PkgCI workflow didn't show up for manual triggering : /

ScottTodd: Ah, I think https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/skipping-workflow-runs is all-or-nothing.

renxida:
TODO: add tracing related script changes from #625

renxida:
TODO: add venv creation script improvements scott made in iree

.github/workflows/pkgci_shark_ai.yml
```diff
@@ -0,0 +1,83 @@
...
+        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
+        with:
+          name: snapshot-shortfin-linux-x86_64-cp${{ env.PY_VERSION_NO_DOT }}-cp${{ env.PY_VERSION_NO_DOT }}
+          path: ${{ env.PACKAGE_DOWNLOAD_DIR }}
+
+      - name: Download shark-ai artifacts
+        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
+        with:
+          name: snapshot-shark-ai-linux-x86_64-cp${{ env.PY_VERSION_NO_DOT }}-cp${{ env.PY_VERSION_NO_DOT }}
+          path: ${{ env.PACKAGE_DOWNLOAD_DIR }}
```
ScottTodd, line 67: I'd like to simplify these steps as we add more `pkgci_*.yml` workflows, but this is fine for now.

For the downloads, actions/download-artifact supports downloading multiple artifacts using some pattern: https://github.com/actions/download-artifact?tab=readme-ov-file#download-multiple-filtered-artifacts-to-the-same-directory . Could try with

```yaml
pattern: snapshot-*-cp${{ env.PY_VERSION_NO_DOT }}-*
```

Note that sharktank and shark-ai aren't dependent on Python versions, and audit_wheel actually removes the `cp311` from the names: 
image

So if we keep the versions in the names, we can have a single download step that filters. If we remove the versions, we could either download all artifacts (no filtering at all) or two download steps, one that filters for `cp311` and one that  filters for `py3-none`.

For the `PY_VERSION_NO_DOT`, I don't see a way to get actions/setup-python to use the `cp311` syntax: https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md.

The downloads are fast (less than 10MB total, taking seconds or less), so I'd probably just download all artifacts then make sure `setup_venv.py` respects the selected Python version and only tries to install compatible packages from the download.

.github/workflows/pkgci_shark_ai.yml
```diff
@@ -0,0 +1,83 @@
...
+      - name: Setup venv
+        run: |
+          ./build_tools/pkgci/setup_venv.py ${VENV_DIR} \
+            --artifact-path=${PACKAGE_DOWNLOAD_DIR} \
+            --fetch-gh-workflow=${{ inputs.artifact_run_id }}
+
+      - name: Install nightly IREE packages
+        run: |
+          source ${VENV_DIR}/bin/activate
+          uv pip install -r requirements-iree-pinned.txt
```
ScottTodd, line 78: nit: "pinned" instead of "nightly" just in case the pinned versions aren't nightly releases at some point in the future

```suggestion
      - name: Install pinned IREE packages
        run: |
          source ${VENV_DIR}/bin/activate
          uv pip install -r requirements-iree-pinned.txt
```

build_tools/pkgci/setup_venv.py
```diff
@@ -0,0 +1,378 @@
...
+    # Set up venv using 'uv' (https://docs.astral.sh/uv/).
+    # We could use 'pip', but 'uv' is much faster at installing packages.
+    venv_path = args.venv_dir
+    python_exe = find_venv_python(venv_path)
+
+    if not python_exe:
+        print(f"Creating venv at {str(venv_path)}")
+
+        subprocess.check_call([sys.executable, "-m", "pip", "install", "uv"])
+        subprocess.check_call(["uv", "venv", str(venv_path), "--python", "3.11"])
```
ScottTodd, line 335: I didn't finish this while working on this originally. The version here should be configurable, maybe coming from `sys.version`. When running locally, either only compatible wheels should be downloaded or only compatible wheels should be installed. I suspect if you run this script locally right now it will fail at the install step.

pytorch-cpu-requirements.txt
ScottTodd, line 1: Let's try to land these requirements file changes via the smaller https://github.com/nod-ai/shark-ai/pull/792

ScottTodd (commented):
There are a few rough edges here that we can probably ignore for a bit, since the workflow latency savings are well worth it.

ScottTodd:
I made a few small edits to the PR description. You could also pull some of the text (with updated metrics) from https://github.com/nod-ai/shark-ai/pull/589 and https://github.com/nod-ai/shark-ai/pull/625.

.github/workflows/pkgci_shark_ai.yml
renxida: Made an issue to track this #804 & on my todo list

build_tools/pkgci/setup_venv.py
renxida: fixd!

.github/workflows/pkgci_shark_ai.yml
renxida: fixed!

pytorch-cpu-requirements.txt
renxida: kk, just pinged @marbre

.github/workflows/pkgci_shark_ai.yml
```diff
@@ -0,0 +1,83 @@
...
+jobs:
+  test_shortfin_llm_server:
+    name: "Integration Tests - Shortfin LLM Server"
+    strategy:
+      matrix:
+        version: [3.11]
+      fail-fast: false
+    runs-on: nodai-amdgpu-mi250-x86-64
+    # runs-on: ubuntu-latest # everything else works but this errs during model loading
+    # TODO: make this / a copy of this using tiny llamas run on the standard runners
```
ScottTodd, line 31: Logs on https://github.com/nod-ai/shark-ai/actions/runs/12697418376/job/35393755864?pr=780 are still showing some suspicious timings using

```
Thu, 09 Jan 2025 20:01:40 GMT Current runner version: '2.321.0'
Thu, 09 Jan 2025 20:01:40 GMT Runner name: 'nod-mi250-gpu7'
Thu, 09 Jan 2025 20:01:40 GMT Runner group name: 'presubmit'
Thu, 09 Jan 2025 20:01:40 GMT Machine name: 'gpud4a8'
```

* almost 10 minutes on actions/checkout:

    ```
    Thu, 09 Jan 2025 20:02:34 GMT Cleaning the repository
    Thu, 09 Jan 2025 20:02:34 GMT /usr/bin/git clean -ffdx
    Thu, 09 Jan 2025 20:11:59 GMT Removing .pytest_cache/
    ```

    May still want to add a cleanup step to the runner (https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/running-scripts-before-or-after-a-job)

* 19 minutes in "Setup venv", with slow segments like:

    ```
    Thu, 09 Jan 2025 20:12:46 GMT Resolved 9 packages in 2.12s
    Thu, 09 Jan 2025 20:16:55 GMT Prepared 1 package in 4m 08s
    Thu, 09 Jan 2025 20:16:55 GMT warning: Failed to hardlink files; falling back to full copy. This may lead to degraded performance.
    Thu, 09 Jan 2025 20:16:55 GMT          If the cache and target directories are on different filesystems, hardlinking may not be supported.
    Thu, 09 Jan 2025 20:16:55 GMT          If this is intentional, set `export UV_LINK_MODE=copy` or use `--link-mode=copy` to suppress this warning.
    Thu, 09 Jan 2025 20:24:46 GMT Installed 9 packages in 7m 50s
    Thu, 09 Jan 2025 20:24:46 GMT  + filelock==3.13.1
    Thu, 09 Jan 2025 20:24:46 GMT  + fsspec==2024.2.0
    Thu, 09 Jan 2025 20:24:46 GMT  + jinja2==3.1.3
    Thu, 09 Jan 2025 20:24:46 GMT  + markupsafe==2.1.5
    Thu, 09 Jan 2025 20:24:46 GMT  + mpmath==1.3.0
    Thu, 09 Jan 2025 20:24:46 GMT  + networkx==3.2.1
    Thu, 09 Jan 2025 20:24:46 GMT  + sympy==1.13.1
    Thu, 09 Jan 2025 20:24:46 GMT  + torch==2.3.0+cpu
    Thu, 09 Jan 2025 20:24:46 GMT  + typing-extensions==4.9.0
    ```

    May want to cache the cache location to be on the same filesystem if it isn't already

build_tools/pkgci/setup_venv.py
ScottTodd: Did you test locally? It still looks like all wheels are downloaded and installed, when only wheels compatible with the selected Python version should be downloaded or installed.

renxida:
UV issue solved! Weirdly, even after adding a cleaning step, checkout still takes a long time at `git clean -ffdx`

There's an option [here](https://github.com/actions/checkout#:~:text=%23%20Whether%20to%20execute%20%60git%20clean%20%2Dffdx%20%26%26%20git%20reset%20%2D%2Dhard%20HEAD%60%20before%20fetching%0A%20%20%20%20%23%20Default%3A%20true%0A%20%20%20%20clean%3A%20%27%27) to turn it off but i'm not sure it's a good idea to do that.

What do you think @ScottTodd ? Merge as is or keep trying to make checkout faster?

Incidentally, I tried running it on the standard Ubuntu runners and checkout is blazing fast : /

ScottTodd:
We'd need to debug those runner machines themselves to see why the cleanup step part of the checkout action is taking so long. We can keep investigation separate from this PR. Could also use mi300 instead of mi250 machines...

.github/workflows/pkgci_shark_ai.yml
```diff
@@ -81,3 +93,7 @@ jobs:
         run: |
           source ${VENV_DIR}/bin/activate
           pytest -v -s app_tests/integration_tests/llm/shortfin --log-cli-level=INFO
+
+      - name: Clean up repo to make next checkout faster
+        run: |
+          git clean -ffdx
```
ScottTodd, line 99: This should be part of the runner configuration, not the job.

https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/running-scripts-before-or-after-a-job

.github/workflows/pkgci_shark_ai.yml
```diff
@@ -48,6 +48,18 @@ jobs:
...
           echo "PY_VERSION_NO_DOT=$(echo ${{ matrix.version }} | tr -d '.')" >> $GITHUB_ENV
 
+      - name: Setup UV caching
+        run: |
+          CACHE_DIR="${GITHUB_WORKSPACE}/.uv-cache"
+          echo "UV_CACHE_DIR=${CACHE_DIR}" >> $GITHUB_ENV
+          mkdir -p "${CACHE_DIR}"
+
+      - name: Cache UV packages
+        uses: actions/cache@v3
```
ScottTodd, line 58: match the version we use elsewhere in the project

```suggestion
        uses: actions/cache@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
```

(and pin to a hash, https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies)

.github/workflows/pkgci_shark_ai.yml
```diff
@@ -0,0 +1,99 @@
...
+        with:
+          path: .uv-cache
+          key: ${{ runner.os }}-uv-py${{ matrix.version }}-${{ hashFiles('requirements-iree-pinned.txt') }}
```
ScottTodd, line 61: Should include all the requirements files in the hash key

Here are the files installed by the script
```python
        str(REPO_ROOT / "pytorch-cpu-requirements.txt"),
```

```python
    if "sharktank" in packages:
        requirements_files.append("sharktank/requirements.txt")
        requirements_files.append("sharktank/requirements-tests.txt")
    if "shortfin" in packages:
        requirements_files.append("shortfin/requirements-tests.txt")
```

No harm in overinclusion, but underinclusion could result in restoring an older cache entry and never updating to new versions. I was bit by that the other day: https://github.com/nod-ai/shark-ai/pull/778

renxida: OOF. Fixed this one. TY

.github/workflows/pkgci_shark_ai.yml
renxida: Fixed!

.github/workflows/pkgci_shark_ai.yml
renxida: Gotcha. Going to remove this and collect this into a new issue and ask Sai for help.

ScottTodd:
> I made a few small edits to the PR description. You could also pull some of the text (with updated metrics) from #589 and #625.

Bump ^ can you add more information to the PR description?

.github/workflows/pkgci_shark_ai.yml
```diff
@@ -0,0 +1,95 @@
...
+        default: ""
+
+jobs:
+  test_shortfin_llm_server:
+    name: "Integration Tests - Shortfin LLM Server"
+    strategy:
+      matrix:
+        version: [3.11]
+      fail-fast: false
+    runs-on: mi300x-4
```
ScottTodd, line 29: So much faster on an mi300 machine.

https://github.com/nod-ai/shark-ai/actions/runs/12701070231/job/35405010380?pr=780

Probably fine to stay there given the benefits, but I don't want _all_ of our tests running on the same machines. That makes it too easy to depend on hardcoded paths and persistent files.

ScottTodd (approved):

.github/workflows/pkgci_shark_ai.yml
renxida: yeah; would be nice if we can eventually move as much as possible onto standard runners.

Issue #783, opened 2025-01-08
[Tuner] Run baseline benchmarks first


kuhar:
The tuner typically runs with more candidates than available devices, i.e., we schedule multiple benchmarks onto the same device.

Currently, the baseline runs last, and we should make sure that the baseline benchmarks run at the very beginning. This will allow us to watch logs in real time and see if the tuner is making progress or not. We can still run it at the very end to confirm whether the performance regresses over time or not (e.g., the device is overheating.).

PR #789, opened 2025-01-08
[tuner] Run baseline benchmarks first and check regression

bangtianliu:
This PR is relevant to address the task in: https://github.com/nod-ai/shark-ai/issues/783.


tuner/tuner/libtuner.py
```diff
@@ -848,4 +848,45 @@ def get_speedup(result: BenchmarkResult) -> float:
...
+
+    regression_detected = False
+    for device_id in baseline_times_by_device:
+        baseline_time = baseline_times_by_device[device_id]
+        post_time = post_baseline_times_by_device[device_id]
+
+        if post_time > baseline_time * 1.03:
+            regression_detected = True
+            percentage_slower = ((post_time - baseline_time) / baseline_time) * 100
+            logging.info(
```
Max191, line 884: nit: Use `logging.warning` here instead of `logging.info`.

tuner/tuner/libtuner.py
```diff
@@ -831,6 +813,24 @@ def benchmark(
...
+        if i != 0
+    ]
+    worker_context_queue = create_worker_context_queue(args.devices)
+    candidate_results = multiprocess_progress_wrapper(
+        num_worker=len(args.devices),
+        task_list=task_list,
+        function=run_iree_benchmark_module_command,
+        initializer=init_worker_context,
+        initializer_inputs=(worker_context_queue,),
+    )
```
Max191, line 832: Since we are running essentially the same code with just different candidates, could you outline this code into a separate function that takes the list of candidate indices, and returns the results of the benchmarking? Then you can call the same function for all 3 benchmark sets here.

i.e., for the baseline, the indices would be `[0] * len(args.devices)`, and for the main benchmarks the indices would be `[i for i in compiled_candidates if i != 0]`

tuner/tuner/libtuner.py
```diff
@@ -848,4 +848,45 @@ def get_speedup(result: BenchmarkResult) -> float:
...
+            regression_detected = True
+            percentage_slower = ((post_time - baseline_time) / baseline_time) * 100
+            logging.info(
+                f"Performance regression detected on device {device_id}: "
+                f"Baseline time = {baseline_time}, Post-baseline time = {post_time}, "
+                f"Slower by {percentage_slower:.3f}%"
+            )
+
+    if not regression_detected:
+        logging.info("No performance regressions detected.")
```
Max191, line 891: nit: This can probably be `logging.debug`.

Max191 (commented):
Overall looks good. Just a few nits about the logging and code reuse. Could you also add a more informative PR description (essentially explain the reasoning for benchmarking baselines twice like in the linked issue).

tuner/tuner/libtuner.py
bangtianliu: > Since we are running essentially the same code with just different candidates, could you outline this code into a separate function that takes the list of candidate indices, and returns the results of the benchmarking? Then you can call the same function for all 3 benchmark sets here.
> 
> i.e., for the baseline, the indices would be `[0] * len(args.devices)`, and for the main benchmarks the indices would be `[i for i in compiled_candidates if i != 0]`

Good point! Will do.

bangtianliu:
The tuner crashed when it tried to do model-level baseline benchmarking inside this PR, I am working on fixing the issue.

tuner/tuner/libtuner.py
```diff
@@ -727,6 +727,35 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
     return collision_detected, unique_indexes
 
 
+def benchmark_candidates(candidate_indices, devices, tuning_client, candidate_trackers):
+    """
+    Runs the benchmarking for a given list of candidate indices.
+    """
+    # Create worker context queue
```
kuhar, line 734: When writing comments, focus on **why** we are doing certain things, not **what** the code is doing. The **what** is usually redundant because it carries the same information content as the code.
```suggestion
```

tuner/tuner/libtuner.py
```diff
@@ -727,6 +727,35 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+    worker_context_queue = create_worker_context_queue(devices)
+
+    # Prepare task list
+    task_list = [
```
kuhar, line 738:
```suggestion
    task_list = [
```

tuner/tuner/libtuner.py
```diff
@@ -727,6 +727,35 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+    # Perform benchmarking
+    benchmark_results = multiprocess_progress_wrapper(
+        num_worker=len(devices),
+        task_list=task_list,
+        function=run_iree_benchmark_module_command,
+        initializer=init_worker_context,
+        initializer_inputs=(worker_context_queue,),
+    )
+
+    return benchmark_results
```
kuhar, line 756:
```suggestion
    # Perform benchmarking.
    return multiprocess_progress_wrapper(
        num_worker=len(devices),
        task_list=task_list,
        function=run_iree_benchmark_module_command,
        initializer=init_worker_context,
        initializer_inputs=(worker_context_queue,),
    )
```

tuner/tuner/libtuner.py
```diff
@@ -873,4 +885,38 @@ def get_speedup(result: BenchmarkResult) -> float:
         )
 
     top_candidates = [result.candidate_id for result in best_results]
+
+    # Benchmarking baselines on each involved device again to check performance regression on devices.
```
kuhar, line 889:
```suggestion
    # Benchmarking baselines again to check for performance regressions. These may indicate machine instability, overheating, etc.
```

tuner/tuner/libtuner.py
```diff
@@ -873,4 +885,38 @@ def get_speedup(result: BenchmarkResult) -> float:
...
+        candidate_trackers=candidate_trackers,
+    )
+
+    post_baseline_times_by_device = {}
+    for r in post_baseline_results:
+        post_baseline_times_by_device[r.device_id] = r.time
+
+    assert (
+        baseline_times_by_device.keys() == post_baseline_times_by_device.keys()
+    ), "Error: The device IDs in baseline and post-baseline results do not match."
```
kuhar, line 904:
```suggestion
    ), "Device ID mismatch between baseline runs."
```

tuner/tuner/libtuner.py
```diff
@@ -873,4 +885,38 @@ def get_speedup(result: BenchmarkResult) -> float:
...
+
+    regression_detected = False
+    for device_id in baseline_times_by_device:
+        baseline_time = baseline_times_by_device[device_id]
+        post_time = post_baseline_times_by_device[device_id]
```
kuhar, line 909: What happens when benchmarking fails altogether and there's no time?

tuner/tuner/libtuner.py
```diff
@@ -873,4 +885,38 @@ def get_speedup(result: BenchmarkResult) -> float:
...
+            regression_detected = True
+            percentage_slower = ((post_time - baseline_time) / baseline_time) * 100
+            logging.warning(
+                f"Performance regression detected on device {device_id}: "
+                f"Baseline time = {baseline_time}, Post-baseline time = {post_time}, "
+                f"Slower by {percentage_slower:.3f}%"
+            )
+
+    if not regression_detected:
+        logging.debug("No performance regressions detected.")
```
kuhar, line 921:
```suggestion
```

kuhar (changes requested):

tuner/tuner/libtuner.py
```diff
@@ -873,4 +881,38 @@ def get_speedup(result: BenchmarkResult) -> float:
...
+        post_baseline_times_by_device[r.device_id] = r.time
+
+    assert (
+        baseline_times_by_device.keys() == post_baseline_times_by_device.keys()
+    ), "Device ID mismatch between baseline runs."
+
+    for device_id in baseline_times_by_device:
+        if device_id not in baseline_times_by_device:
+            logging.warning(f"Baseline time missing for device {device_id}")
+            continue
```
Max191, line 905: Is this warning possible to run? I think it can be removed.

Max191 (approved):
Looks good, thanks!

tuner/tuner/libtuner.py
bangtianliu: This was added to address Jakub's comment 
> What happens when benchmarking fails altogether and there's no time?

It addresses the corner case where benchmarking completely fails for a specific device ID (e.g., when one device is broken). While I haven't encountered this scenario, I think it's better to keep this safeguard in place.

tuner/tuner/libtuner.py
```diff
@@ -727,6 +727,31 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+    task_list = [
+        BenchmarkPack(
+            iree_benchmark_module_flags=tuning_client.get_iree_benchmark_module_flags(),
+            benchmark_timeout=tuning_client.get_benchmark_timeout_s(),
+            candidate_tracker=candidate_trackers[idx],
+        )
+        for idx in candidate_indices
+    ]
+
+    # Perform benchmarking
```
kuhar, line 745: Use proper punctuation.

```suggestion
    # Perform benchmarking.
```

tuner/tuner/libtuner.py
```diff
@@ -873,4 +881,38 @@ def get_speedup(result: BenchmarkResult) -> float:
         )
 
     top_candidates = [result.candidate_id for result in best_results]
+
+    # Benchmarking baselines again to check for performance regressions. These may indicate machine instability, overheating, etc.
```
kuhar, line 885: This seems to exceed the column limit, doesn't the formatter complain about it?

tuner/tuner/libtuner.py
```diff
@@ -873,4 +881,38 @@ def get_speedup(result: BenchmarkResult) -> float:
...
+    post_baseline_times_by_device = {}
+    for r in post_baseline_results:
+        post_baseline_times_by_device[r.device_id] = r.time
+
+    assert (
+        baseline_times_by_device.keys() == post_baseline_times_by_device.keys()
+    ), "Device ID mismatch between baseline runs."
+
+    for device_id in baseline_times_by_device:
+        if device_id not in baseline_times_by_device:
```
kuhar, line 903: Have you tested this code? Make sure you exercise the failure conditions

tuner/tuner/libtuner.py
kuhar: I think the correct way to handle this to issue a warning when any of the baseline benchmarks fail, but more importantly try to recover from that: try to use the time from the other baseline run, if that succeeded.

kuhar:
Ideally, these corner cases should be covered by unit tests.

bangtianliu:
WIP:  need to rebase on top of https://github.com/nod-ai/shark-ai/pull/799, and continue to work.

tuner/tuner/candidate_gen.py
```diff
@@ -355,7 +354,7 @@ def main():
             prefetch_shared_memory=args.prefetch_shared_memory_options,
             no_reduce_shared_memory_bank_conflicts=args.no_reduce_shared_memory_bank_conflicts_options,
         )
-        specs: list[ir.Module] = generate_configs_and_td_specs(
+        specs = generate_configs_and_td_specs(
```
kuhar, line 357: Why do we want to remove the type hints?

tuner/tuner/candidate_gen.py
```diff
@@ -369,7 +368,7 @@ def main():
             spec_path = spec_dir / f"{candidate_num}_spec.mlir"
             spec_dir.mkdir(parents=True, exist_ok=True)
             with open(spec_path, "w") as f:
-                local_scope_spec_str: str = spec.operation.get_asm(use_local_scope=True)
+                local_scope_spec_str = spec.operation.get_asm(use_local_scope=True)
```
kuhar, line 371: also here

tuner/tuner/libtuner.py
```diff
@@ -221,6 +221,53 @@ def validate_devices(user_devices: list[str]) -> None:
         )
 
 
+def validate_benchmark_results(
+    benchmark_results: list[BenchmarkResult],
+) -> list[BenchmarkResult]:
```
kuhar, line 226: Based on the function name alone, it's impossible to tell what this code does... Could we call it something like: `get_valid_benchmark_results` and add a one-line docstring explaining what is considered valid?

tuner/tuner/libtuner.py
```diff
@@ -221,6 +221,53 @@ def validate_devices(user_devices: list[str]) -> None:
...
+    filtered_benchmark_results = [r for r in benchmark_results if math.isfinite(r.time)]
+    if len(filtered_benchmark_results) == 0:
+        logging.error("No successful candidate benchmarks.")
+
+    return filtered_benchmark_results
+
+
+def map_baseline_by_device(baseline_results: list[BenchmarkResult]) -> dict[str, float]:
+    return {r.device_id: r.time for r in baseline_results}
```
kuhar, line 235: Should we check that the device IDs are unique? Otherwise this will result in data loss.

tuner/tuner/libtuner.py
```diff
@@ -221,6 +221,53 @@ def validate_devices(user_devices: list[str]) -> None:
...
+
+
+def validate_baselines_device_ids_match(
+    first_baseline_by_device: dict[str, float],
+    second_baseline_by_device: dict[str, float],
+) -> bool:
```
kuhar, line 241: I think it's better to expand this check inline, the helper function doesn't seem to help that much -- it's just a comparison

tuner/tuner/libtuner.py
```diff
@@ -221,6 +221,53 @@ def validate_devices(user_devices: list[str]) -> None:
...
+    return first_baseline_by_device.keys() == second_baseline_by_device.keys()
+
+
+def validate_baseline_regression(
+    first_baseline_by_device: dict[str, float],
+    second_baseline_by_device: dict[str, float],
+) -> list[str]:
```
kuhar, line 248: Similar here, it's impossible to tell what this function does based on the name alone. We need a better name and some comment with an explanation.

tuner/tuner/libtuner.py
```diff
@@ -892,41 +964,49 @@ def benchmark(
...
-        initializer=init_worker_context,
-        initializer_inputs=(worker_context_queue,),
+    baseline_times_by_device = map_baseline_by_device(baseline_results)
+
+    candidate_indices = [i for i in compiled_candidates if i != 0]
+    candidate_results = benchmark_candidates(
+        candidate_indices=candidate_indices,
+        devices=args.devices,
+        tuning_client=tuning_client,
+        candidate_trackers=candidate_trackers,
```
kuhar, line 983: Should we put baseline benchmarking into it's own function that return a map from devices to results?

tuner/tuner/libtuner_test.py
```diff
@@ -233,3 +231,64 @@ def test_select_best_benchmark_results() -> None:
...
+
+
+def test_validate_baselines_device_id_match():
+    first_baseline = {"hip://0": 1000.0, "hip://1": 2000.0}
+    second_baseline = {"hip://1": 1500.0, "hip://2": 2500.0}
+
+    result = libtuner.validate_baselines_device_ids_match(
+        first_baseline, second_baseline
+    )
+    assert result is False
```
kuhar, line 262:
```suggestion
    assert not result
```

tuner/tuner/libtuner_test.py
```diff
@@ -233,3 +231,64 @@ def test_select_best_benchmark_results() -> None:
...
+
+    first_baseline = {"hip://0": 1000.0, "hip://1": 2000.0}
+    second_baseline = {"hip://0": 1500.0, "hip://1": 2500.0}
+
+    result = libtuner.validate_baselines_device_ids_match(
+        first_baseline, second_baseline
+    )
+    assert result is True
```
kuhar, line 270:
```suggestion
    assert result
```

kuhar (commented):
Can you edit the PR description and explain what error conditions you considered and how they are handled?

tuner/tuner/candidate_gen.py
bangtianliu: same

tuner/tuner/candidate_gen.py
kuhar: make this function typed then?

tuner/tuner/libtuner.py
bangtianliu: Yeah, good point!

tuner/tuner/libtuner.py
bangtianliu: No, since the baseline results are used later on: https://github.com/nod-ai/shark-ai/blob/main/tuner/tuner/libtuner.py#L930.

tuner/tuner/libtuner.py
```diff
@@ -221,6 +221,62 @@ def validate_devices(user_devices: list[str]) -> None:
...
+    Filter the benchmark_results list to return list with finite `time` values.
+    """
+    filtered_benchmark_results = [r for r in benchmark_results if math.isfinite(r.time)]
+    if len(filtered_benchmark_results) == 0:
+        logging.error("No successful candidate benchmarks.")
+
+    return filtered_benchmark_results
+
+
+def check_baseline_devices_uniqueness(baseline_results: list[BenchmarkResult]) -> bool:
```
kuhar, line 237: Based on the function name, we can't tell if `True` indicates a success or a failure. (That's why we have the `LogicalResult` type in LLVM, BTW!).

I think we could use this name instead:
```suggestion
def are_benchmark_devices_unique(baseline_results: list[BenchmarkResult]) -> bool:
```

tuner/tuner/libtuner.py
```diff
@@ -221,6 +221,62 @@ def validate_devices(user_devices: list[str]) -> None:
...
+    seen = set()
+    for result in baseline_results:
+        if result.device_id in seen:
+            return False
+        seen.add(result.device_id)
+    return True
```
kuhar, line 243:
```suggestion
    return len(baseline_results) == len(set(map(lambda r: r.device_id, baseline_results))
```
?

tuner/tuner/libtuner.py
```diff
@@ -221,6 +221,62 @@ def validate_devices(user_devices: list[str]) -> None:
...
+def map_baseline_by_device(baseline_results: list[BenchmarkResult]) -> dict[str, float]:
+    return {r.device_id: r.time for r in baseline_results}
+
+
+def detect_baseline_regression(
+    first_baseline_results: list[BenchmarkResult],
+    second_baseline_results: list[BenchmarkResult],
+) -> list[str]:
+    """
+    Detects performance regressions between two sets of baseline results.
```
kuhar, line 255: What is the result of this function?

tuner/tuner/libtuner.py
```diff
@@ -221,6 +221,62 @@ def validate_devices(user_devices: list[str]) -> None:
...
+        first_baseline_time = first_baseline_by_device[device_id]
+        second_baseline_time = second_baseline_by_device[device_id]
+
+        if second_baseline_time > first_baseline_time * 1.03:
+            percentage_slower = (
+                (second_baseline_time - first_baseline_time) / first_baseline_time
+            ) * 100
+            logging.warning(
+                f"Performance regression detected on device {device_id}: "
+                f"Baseline time = {first_baseline_time}, Post-baseline time = {second_baseline_time}, "
```
kuhar, line 272:
```suggestion
                f"First baseline time = {first_baseline_time}, Second baseline time = {second_baseline_time}, "
```

Also, what's the time unit?

tuner/tuner/libtuner.py
```diff
@@ -221,6 +221,62 @@ def validate_devices(user_devices: list[str]) -> None:
...
+    Detects performance regressions between two sets of baseline results.
+    """
+    regression_device_ids = []
+    first_baseline_by_device = map_baseline_by_device(first_baseline_results)
+    second_baseline_by_device = map_baseline_by_device(second_baseline_results)
+    for device_id in first_baseline_by_device:
+        if device_id not in second_baseline_by_device:
+            continue
+        first_baseline_time = first_baseline_by_device[device_id]
+        second_baseline_time = second_baseline_by_device[device_id]
```
kuhar, line 264: What's the time unit? We should make it a suffix of the variable name

tuner/tuner/libtuner.py
```diff
@@ -892,41 +973,51 @@ def benchmark(
...
+        candidate_indices=post_baseline_indices,
+        devices=args.devices,
+        tuning_client=tuning_client,
+        candidate_trackers=candidate_trackers,
     )
 
+    if not check_baseline_devices_uniqueness(post_baseline_results):
+        logging.warning("Duplicate device IDs detected in the second baseline results.")
+
+    first_baseline_by_device = map_baseline_by_device(baseline_results)
```
kuhar, line 1008: Maybe move this uniqueness check to `map_baseline_by_device`? This function is already quite long.

tuner/tuner/libtuner.py
```diff
@@ -749,6 +822,113 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+        function=run_iree_benchmark_module_command,
+        initializer=init_worker_context,
+        initializer_inputs=(worker_context_queue,),
+    )
+    return baseline_results
+
+
+class BaselineResultHandler:
+    def __init__(self) -> None:
+        self.runs: dict[str, list[float]] = defaultdict(list)
```
kuhar, line 876: It's not obvious what are the keys and what are the keys and the value. Can you either rename it to something like `self.device_to_run_results` or add a comment?

tuner/tuner/libtuner.py
```diff
@@ -749,6 +822,113 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+
+    def add_run(self, results: list[BenchmarkResult]) -> None:
+        for result in results:
+            if math.isfinite(result.time):
+                self.runs[result.device_id].append(result.time)
```
kuhar, line 881: Why not store the whole results instead of just the times? Even if it didn't succeed, I think it might be useful to keep this information around instead of silently discarding it. The issue with discarding it here is that the results no longer align, i.e., you can't tell which of the runs succeeded.

tuner/tuner/libtuner.py
```diff
@@ -749,6 +822,113 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+
+    def num_successful_runs(self, device_id: str) -> int:
+        return len(self.runs[device_id])
+
+    def get_average_result(self, device_id: str) -> Optional[float]:
```
kuhar, line 886: What is the time unit? We should make it be a part of the function name, IMO. 
https://users.ece.cmu.edu/~eno/coding/CppCodingStandard.html#units
https://ruudvanasseldonk.com/2022/03/20/please-put-units-in-names

tuner/tuner/libtuner.py
```diff
@@ -749,6 +822,113 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+        times = self.runs.get(device_id, [])
+        if times:
+            return sum(times) / len(times)
+        return None
+
+    def is_valid(self) -> bool:
```
kuhar, line 892: What does it mean for this to be valid? Can you add a docstring?

tuner/tuner/libtuner.py
```diff
@@ -749,6 +822,113 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+
+    def is_valid_for_device(self, device_id: str) -> bool:
+        return bool(self.runs.get(device_id))
+
+    def calculate_speedup(
+        self, candidate_results: list[BenchmarkResult]
+    ) -> dict[int, float]:
+        """
+        Calculate the speedup for a list of candidate results compared to the baseline.
+        Return A dictionary where the key is the candidate_id and the value is the speedup ratio.
```
kuhar, line 903:
```suggestion
        Returns a map from candidate_id to its speedup ratio.
```

tuner/tuner/libtuner.py
```diff
@@ -749,6 +822,113 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+        """
+        baseline_times = [
+            time
+            for times in self.runs.values()
+            for time in times
+            if math.isfinite(time)
```
kuhar, line 909: Why not use `get_average_result` function that you defined above?

tuner/tuner/libtuner.py
```diff
@@ -749,6 +822,113 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+        ]
+        fallback_baseline = (
+            sum(baseline_times) / len(baseline_times) if baseline_times else None
+        )
```
kuhar, line 913: Can we use an early return instead?

tuner/tuner/libtuner.py
```diff
@@ -903,46 +1046,54 @@ def benchmark(
...
+
+    if not are_baseline_devices_unique(second_baseline_result):
+        logging.warning("Duplicate device IDs detected in the second baseline results.")
 
-    best_results: list[BenchmarkResult] = select_best_benchmark_results(
-        candidate_results=candidate_results,
-        baseline_results=baseline_results,
-        num_candidates=num_candidates,
+    regression_devices = detect_baseline_regression(
+        first_baseline_result, second_baseline_result
```
kuhar, line 1086: Why not move it to the new class? It already knows how to deal with baseline results.

tuner/tuner/libtuner.py
```diff
@@ -903,46 +1046,54 @@ def benchmark(
...
+            f"Performance regressions detected for the following devices: {', '.join(regression_devices)}"
+        )
 
-    top_candidates = [result.candidate_id for result in best_results]
+    speedup_result = baseline_handler.calculate_speedup(candidate_results)
+    # If the baseline is valid (`baseline_handler.is_valid()`), `speedup_result` represents the speedup values.
+    # Otherwise, `speedup_result` contains the raw time values.
+    top_candidates = get_top_candidates(
+        speedup_result, num_candidates, baseline_handler.is_valid()
+    )
```
kuhar, line 1098: This can also be moved to the new class IMO.

tuner/tuner/libtuner_test.py
```diff
@@ -188,60 +188,220 @@ def test_get_compilation_success_rate():
...
+    additional_baseline = [
+        libtuner.BenchmarkResult(5, 1.2, "hip://2"),
+        libtuner.BenchmarkResult(6, 0.9, "hip://2"),
+    ]
+    handler.add_run(additional_baseline)
+    assert handler.num_successful_runs("hip://2") == 2
+    assert handler.num_successful_runs("hip://3") == 0
+
+
+def test_get_average_result():
```
kuhar, line 321: We can test multiple methods in one function to avoid repeating the same setup code. Not every function has to have its own test function.

tuner/tuner/libtuner.py
```diff
@@ -749,6 +750,168 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+        self.device_baseline_results: dict[str, list[BenchmarkResult]] = defaultdict(
+            list
+        )
+
+    def add_run(self, results: list[BenchmarkResult]) -> None:
+        for result in results:
+            self.device_baseline_results[result.device_id].append(result)
+
+    def are_baseline_devices_unique(self, results: list[BenchmarkResult]) -> bool:
+        return len(results) == len(set(result.device_id for result in results))
```
kuhar, line 814: `self` is not used, this can be a staticmethod

tuner/tuner/libtuner.py
```diff
@@ -749,6 +750,168 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+
+    def get_valid_time_ms(self, device_id: str) -> list[float]:
+        return [
+            result.time
+            for result in self.device_baseline_results.get(device_id, [])
+            if math.isfinite(result.time)
```
kuhar, line 820: Can we add a method to `BenchmarkResult` that checks if that result is valid? Invalid benchmarks being represented with infinite values is an implementation detail that shouldn't leak here.

tuner/tuner/libtuner.py
```diff
@@ -749,6 +750,168 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+        ]
+
+    def num_successful_runs(self, device_id: str) -> int:
+        return len(self.get_valid_time_ms(device_id))
```
kuhar, line 824: The function name would suggest to me that this tell whether the whole run was valid, not how many times device finished. 

Maybe we don't need this function anymore and can use `get_valid_time_ms` directly?

tuner/tuner/libtuner.py
```diff
@@ -749,6 +750,168 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+
+        return regressions
+
+    def is_valid(self) -> bool:
+        """
+        Check if there are any valid finite baseline time recorded.
+        Return True if at least a valid (finite) baseline time recorded,
+        otherwise False.
+        This method determines whether the baseline data is available for computations
+        such as calculating speedup.
```
kuhar, line 860:
```suggestion
```

tuner/tuner/libtuner.py
```diff
@@ -749,6 +750,168 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+                    f"Slower by {((result.time - baseline_avg) / baseline_avg) * 100:.2f}%"
+                )
+
+        return regressions
+
+    def is_valid(self) -> bool:
+        """
+        Check if there are any valid finite baseline time recorded.
+        Return True if at least a valid (finite) baseline time recorded,
+        otherwise False.
```
kuhar, line 858:
```suggestion
        Return True iff at least a valid (finite) baseline time recorded.
```

tuner/tuner/libtuner.py
```diff
@@ -749,6 +750,168 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+        This method determines whether the baseline data is available for computations
+        such as calculating speedup.
+        """
+        return any(
+            self.get_valid_time_ms(device_id)
+            for device_id in self.device_baseline_results
+        )
+
+    def is_valid_for_device(self, device_id: str) -> bool:
+        return bool(self.get_valid_time_ms(device_id))
```
kuhar, line 868: Maybe we can remove this function and do `num_successful_runs(device_id) != 0` instead?

tuner/tuner/libtuner.py
```diff
@@ -749,6 +750,168 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+    def calculate_speedup(
+        self, candidate_results: list[BenchmarkResult]
+    ) -> dict[int, float]:
+        """
+        Calculate the speedup for a list of candidate results compared to the baseline.
+        Returns a map from candidate_id to its speedup ratio.
+        """
+        if not self.is_valid():
+            logging.warning("No valid baseline times available.")
+            # Use the candidate time directly when no baselines are available
```
kuhar, line 879: nit: Use proper punctuation
```suggestion
            # Use the candidate time directly when no baselines are available.
```

tuner/tuner/libtuner.py
```diff
@@ -749,6 +750,168 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+            return {
+                candidate.candidate_id: candidate.time
+                for candidate in candidate_results
+            }
+
+        # Calculate the fallback baseline as the average of all valid times across devices
```
kuhar, line 885: nit: Use proper punctuation
```suggestion
        # Calculate the fallback baseline as the average of all valid times across devices.
```

tuner/tuner/libtuner.py
```diff
@@ -749,6 +750,168 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+    ) -> dict[int, float]:
+        """
+        Calculate the speedup for a list of candidate results compared to the baseline.
+        Returns a map from candidate_id to its speedup ratio.
+        """
+        if not self.is_valid():
+            logging.warning("No valid baseline times available.")
+            # Use the candidate time directly when no baselines are available
+            return {
+                candidate.candidate_id: candidate.time
```
kuhar, line 881: I think in the absence of any baselines, the speedup should be 1.0, no?

tuner/tuner/libtuner.py
```diff
@@ -749,6 +750,168 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+
+            baseline_avg = self.get_average_result_ms(result.device_id)
+            if baseline_avg is not None and result.time > baseline_avg * threshold:
+                regressions.append(result.device_id)
+                logging.warning(
+                    f"Performance regression detected on device {result.device_id}: "
+                    f"Stored average baseline time = {baseline_avg:.2f} ms, "
+                    f"New baseline time = {result.time:.2f} ms, "
+                    f"Slower by {((result.time - baseline_avg) / baseline_avg) * 100:.2f}%"
+                )
```
kuhar, line 850: The code that uses this function already prints a warning. Maybe we can drop this message?

tuner/tuner/libtuner.py
```diff
@@ -903,46 +1007,62 @@ def benchmark(
...
+    baseline_tracker = candidate_trackers[0]
+    first_baseline_result = benchmark_baseline(
+        devices=args.devices,
+        tuning_client=tuning_client,
+        candidate_tracker=baseline_tracker,
+    )
+    baseline_handler = BaselineResultHandler()
+    baseline_handler.add_run(first_baseline_result)
+    if not baseline_handler.is_valid():
+        logging.warning("Baseline result is not valid after first run")
```
kuhar, line 1020: I don't think it's obvious what it means for baseline results to be not valid at this point. How about this?
```suggestion
        logging.warning("Baseline run failed")
```

tuner/tuner/libtuner.py
```diff
@@ -903,46 +1007,62 @@ def benchmark(
...
+
+    if not baseline_handler.are_baseline_devices_unique(first_baseline_result):
+        logging.warning("Duplicate device IDs detected in the first baseline results.")
```
kuhar, line 1023: I think we should move this check to the code that relies on baseline devices being unique. This code is no longer in this function.

tuner/tuner/libtuner.py
```diff
@@ -903,46 +1007,62 @@ def benchmark(
...
+        logging.warning(
+            f"Performance regressions detected for the following devices: {', '.join(regression_devices)}"
+        )
+    baseline_handler.add_run(second_baseline_result)
+
+    if not baseline_handler.is_valid():
+        logging.warning("Baseline result is not valid after second run")
+
+    if not baseline_handler.are_baseline_devices_unique(second_baseline_result):
+        logging.warning("Duplicate device IDs detected in the second baseline results.")
```
kuhar, line 1050: Same here

tuner/tuner/libtuner.py
```diff
@@ -749,6 +750,165 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+            for time in self.device_baseline_times.get(device_id, [])
+            if math.isfinite(time)
+        ]
+
+    def num_successful_runs(self, device_id: str) -> int:
+        return len(self.get_valid_time_ms(device_id))
+
+    def get_average_result_ms(self, device_id: str) -> Optional[float]:
+        valid_times = self.get_valid_time_ms(device_id)
+        if valid_times:
```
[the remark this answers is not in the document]
bangtianliu:
```python
    def get_average_result_ms(self, device_id: str) -> Optional[float]:
        valid_times = self.get_valid_time_ms(device_id)
        if valid_times:
            return sum(valid_times) / len(valid_times)
        return None
```
The complete code is shown above, `valid_times `is reused multiple times and thus maybe it's better left as is.

tuner/tuner/libtuner.py
kuhar: Can we use some other metric instead like the average across all the other devices?

bangtianliu: Under this condition, the context is all the baseline data is not available.

tuner/tuner/libtuner.py
```diff
@@ -749,6 +753,166 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+            return sum(valid_times) / len(valid_times)
+        return None
+
+    def detect_regressions(
+        self,
+        baseline_results: list[BenchmarkResult],
+        threshold: float = 1.03,
+    ) -> list[str]:
+        """
+        Return a list of device IDs where regressions were detected.
```
kuhar, line 845: This doesn't really explain what we consider a regression
```suggestion
        Returns a list of device IDs where regressions were detected.
```

tuner/tuner/libtuner.py
```diff
@@ -749,6 +753,166 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+            baseline_avg = self.get_average_result_ms(result.device_id)
+            if baseline_avg is not None and result.time > baseline_avg * threshold:
+                regressions.append(result.device_id)
+
+        return regressions
+
+    def is_valid(self) -> bool:
+        """
+        Check if there are any valid finite baseline time recorded.
+        Return True iff at least a valid (finite) baseline time recorded.
```
kuhar, line 861:
```suggestion
        Returns True iff at least one valid (finite) baseline time was recorded.
```

tuner/tuner/libtuner.py
```diff
@@ -749,6 +753,166 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+        )
+
+    def is_valid_for_device(self, device_id: str) -> bool:
+        return len(self.get_valid_time_ms(device_id)) != 0
+
+    def calculate_speedup(
+        self, candidate_results: list[BenchmarkResult]
+    ) -> dict[int, float]:
+        """
+        Calculate the speedup for a list of candidate results compared to the baseline.
```
kuhar, line 875: Can you explain what you understand by speedup? Is it `run_time / avg_baseline_time` ?

tuner/tuner/libtuner.py
```diff
@@ -902,46 +1008,60 @@ def benchmark(
...
+        )
+    baseline_handler.add_run(second_baseline_result)
+
+    if not baseline_handler.is_valid():
+        logging.warning("Baseline run failed.")
+
+    speedup_result = baseline_handler.calculate_speedup(candidate_results)
+    # If the baseline is valid (`baseline_handler.is_valid()`), `speedup_result` represents the speedup values.
+    # Otherwise, `speedup_result` contains the raw time values.
+    top_candidates = baseline_handler.get_top_candidates(speedup_result, num_candidates)
```
kuhar, line 1050: I'd change this API such that `get_top_candidates` returns you **all** the candidates ordered by speedup, as well as the speedup values themselves. This way we don't leak implementation details around what is a speedup. In the code here, we can decide how many of the op candidates to print out to the log.

tuner/tuner/libtuner.py
```diff
@@ -874,6 +877,17 @@ def calculate_speedup(
         """
         Calculate the speedup for a list of candidate results compared to the baseline.
         Returns a map from candidate_id to its speedup ratio.
+
+        Speedup is defined as the ratio of the candidate's runtime to the average baseline time
+        for the corresponding device as:
+
+            speedup = candidate_runtime / avg_baseline_time
+
+        If no valid baseline times are available, the candidate'sruntime is used directly as:
```
kuhar, line 886: typo
```suggestion
        If no valid baseline times are available, the candidate's runtime is used directly as:
```

tuner/tuner/libtuner.py
```diff
@@ -874,6 +877,17 @@ def calculate_speedup(
...
+
+            speedup = candidate_runtime
```
kuhar, line 888: This looks wrong because for very short runtimes this can end up being less then other `candidate_runtime / avg_baseline_time` and dominate the list of top candidates. Please add a test for this.

tuner/tuner/libtuner.py
```diff
@@ -1045,23 +1054,21 @@ def benchmark(
         logging.warning("Baseline run failed.")
 
     speedup_result = baseline_handler.calculate_speedup(candidate_results)
-    # If the baseline is valid (`baseline_handler.is_valid()`), `speedup_result` represents the speedup values.
-    # Otherwise, `speedup_result` contains the raw time values.
-    top_candidates = baseline_handler.get_top_candidates(speedup_result, num_candidates)
+    all_candidates_with_speedup = baseline_handler.get_top_candidates(speedup_result)
```
kuhar, line 1057: Speedup result should be calculated inside get_top_candidates. We should also rename this to something else now that it doesn't produce top N values, and instead, returns the full list. See the full suggestion here: https://github.com/nod-ai/shark-ai/pull/789#discussion_r1920717256

tuner/tuner/libtuner_test.py
```diff
@@ -284,9 +290,15 @@ def test_baseline_result_handler_speedup():
         7: 0.8 / 1.2,
     }
 
-    assert handler.get_top_candidates(speedup) == [5, 7, 6]
-    assert handler.get_top_candidates(speedup, 2) == [5, 7]
-    assert handler.get_top_candidates(speedup, 5) == [5, 7, 6]
+    all_candidates_with_speedup = handler.get_top_candidates(speedup)
+    print(all_candidates_with_speedup)
```
kuhar, line 294: Please remove debug prints

tuner/tuner/libtuner.py
bangtianliu: There is no mixing of results due to the presence of a fallback baseline. If no valid baseline times exist on all devices, we sort based on raw runtime instead fully. Relevant test cases for this scenario have already been added.

tuner/tuner/libtuner.py
```diff
@@ -749,6 +753,175 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+            for result in self.device_baseline_results[device_id]
+            if result.is_valid()
+        ]
+
+        fallback_baseline = sum(valid_baseline_times) / len(valid_baseline_times)
+
+        speedup_by_candidate = {}
+        for candidate in candidate_results:
+            baseline_avg = self.get_average_result_ms(candidate.device_id)
+            if baseline_avg is None or not math.isfinite(baseline_avg):
```
kuhar, line 913: Add a time unit to `baseline_avg`. Also AFAICT, `get_average_result_ms` never returns a non-finite time, does it?

bangtianliu: Yeah, good catch.

tuner/tuner/libtuner.py
```diff
@@ -902,46 +1023,60 @@ def benchmark(
...
+            f"Performance regressions detected for the following devices: {', '.join(regression_devices)}."
+        )
+    baseline_handler.add_run(second_baseline_result)
+
+    if not baseline_handler.is_valid():
+        logging.warning("Baseline run failed.")
+
+    speedup_result = baseline_handler.calculate_speedup(candidate_results)
+    all_candidates_with_speedup = baseline_handler.sort_candidates_with_speedup(
+        speedup_result
```
kuhar, line 1064: We should only have to call `baseline_handler.sort_candidates_with_speedup`, it can calculate the speedup internally. This `speedup_result` variable is only ever used in this function. See the two prior comments that made the same suggestion: https://github.com/nod-ai/shark-ai/pull/789#discussion_r1920717256 and https://github.com/nod-ai/shark-ai/pull/789#discussion_r1920819793 .

tuner/tuner/libtuner.py
```diff
@@ -902,46 +1023,60 @@ def benchmark(
...
+    )
+    top_candidates_with_speedup = all_candidates_with_speedup[:num_candidates]
+
+    if baseline_handler.is_valid():
+        candidate_time_map = {
```
kuhar, line 1069: Can you add the time unit to the variable name? I'd call it `candidate_id_to_time_ms`.

tuner/tuner/libtuner.py
```diff
@@ -902,46 +1023,60 @@ def benchmark(
...
+            result.candidate_id: result.time for result in candidate_results
+        }
+        for candidate_id, speedup in top_candidates_with_speedup:
+            actual_time = candidate_time_map[candidate_id]
```
kuhar, line 1073: What's the time unit?

tuner/tuner/libtuner.py
```diff
@@ -749,6 +753,181 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+        for candidate in candidate_results:
+            baseline_avg_ms = self.get_average_result_ms(candidate.device_id)
+            if baseline_avg_ms is None:
+                baseline_avg_ms = fallback_baseline
+            speedup_by_candidate[candidate.candidate_id] = (
+                candidate.time / baseline_avg_ms
+            )
+        return speedup_by_candidate
+
+    def sort_candidates_with_speedup(
```
kuhar, line 924:
```suggestion
    def get_candidates_ordered_by_speedup(
```
?

tuner/tuner/libtuner.py
```diff
@@ -1059,24 +1052,23 @@ def benchmark(
...
     top_candidates_with_speedup = all_candidates_with_speedup[:num_candidates]
 
     if baseline_handler.is_valid():
-        candidate_time_map = {
+        candidate_id_to_time_ms = {
             result.candidate_id: result.time for result in candidate_results
         }
         for candidate_id, speedup in top_candidates_with_speedup:
-            actual_time = candidate_time_map[candidate_id]
+            time_ms = candidate_id_to_time_ms[candidate_id]
```
kuhar, line 1065: If `get_candidates_ordered_by_speedup` returned `list[tuple[BenchmarkResult, float]]:`, we could get rid of this logic.

kuhar (approved):
LGTM % one suggestion.

Thanks for all the changes, the code looks much more robust now.

tuner/tuner/libtuner.py
```diff
@@ -749,6 +756,168 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+        )
+
+    def is_valid_for_device(self, device_id: str) -> bool:
+        return len(self.get_valid_time_ms(device_id)) != 0
+
+    def get_candidates_ordered_by_speedup(
+        self, candidate_results: list[BenchmarkResult]
+    ) -> Union[list[BenchmarkResult], list[tuple[BenchmarkResult, float]]]:
+        """
+        returns:
```
kuhar, line 881:
```suggestion
        Returns:
```

tuner/tuner/libtuner.py
```diff
@@ -749,6 +756,168 @@ def collision_handler(index_hash_list: list[tuple[int, str]]) -> tuple[bool, lis
...
+        - `list[BenchmarkResult]` sorted by runtime if no valid baselines exist.
+        - `list[tuple[BenchmarkResult, float]]` sorted by speedup when baselines are available.
```
kuhar, line 883: We shouldn't change the return value type based on the input -- can we always return `list[tuple[BenchmarkResult, float]]`. When no baselines are available, you can make the speedup be 1.0, for example, as a placeholder value.

kuhar (changes requested):

kuhar (approved):

PR #792, opened 2025-01-08
Drop unneeded pip flags and switch torch channel

marbre:
Drops `--pre` and `--upgrade` where dependencies are pinned to specific versions. Further switches to unstall torch cpu form the stable release channel.

### DIFF
--- a/pytorch-cpu-requirements.txt
+++ b/pytorch-cpu-requirements.txt
@@ -1,3 +1,2 @@
---pre
---index-url https://download.pytorch.org/whl/test/cpu
+--index-url https://download.pytorch.org/whl/cpu/
 torch==2.3.0
--- a/.github/workflows/ci-sglang-benchmark.yml
+++ b/.github/workflows/ci-sglang-benchmark.yml
@@ -69,7 +69,7 @@ jobs:
           pip install --no-compile -r requirements.txt -e sharktank/ shortfin/
 
           # Pin to known-working versions.
-          pip install -f https://iree.dev/pip-release-links.html --pre --upgrade \
+          pip install -f https://iree.dev/pip-release-links.html
             iree-base-compiler==3.1.0rc20241220 \
             iree-base-runtime==3.1.0rc20241220 \
             "numpy<2.0"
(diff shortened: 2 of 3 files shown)


pytorch-cpu-requirements.txt
```diff
@@ -1,3 +1,2 @@
---pre
---index-url https://download.pytorch.org/whl/test/cpu
+--index-url https://download.pytorch.org/whl/cpu/
 torch==2.3.0
```
ScottTodd, line 2: I bet we can similarly update this file https://github.com/nod-ai/shark-ai/blob/0c53fb0c011b3f80f9177c5d1107598c31062965/pytorch-rocm-requirements.txt#L1-L3

Official docs (https://pytorch.org/get-started/locally/#start-locally) say
```
pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/rocm6.2
```

ScottTodd (approved):

pytorch-cpu-requirements.txt
marbre: I skipped this as I am not sure if we want stable versions or nightly releases here. If we want the latter I assume we still need `--pre` as we do not set an explicit version but I couldn't test as pip was somehow stuck on my VM.

ScottTodd: We have no workflows in this repository using the `pytorch-rocm-requirements.txt` file and only one mention of it in a developer guide. I vote to just have it match the upstream instructions unless we have a compelling reason to diverge.

marbre: Fair point. Let me add this to this PR.

marbre: Well neither
```bash
pip install --index-url https://download.pytorch.org/whl/rocm6.2 torch>=2.3.0
```
nor
```bash
pip install --pre --index-url https://download.pytorch.org/whl/rocm6.2 torch>=2.3.0
```
install anything for me locally right now. However, I would like to test before committing.

marbre: I got to the point that this needs to be handled in a follow up to unblock a PR by @renxida.

PR #794, opened 2025-01-08
Point to iree-turbine nightly releases

marbre:
Advise to install the latest nightly release instead of cloning and installing and editable version of iree-turbine.

### DIFF
--- a/docs/developer_guide.md
+++ b/docs/developer_guide.md
@@ -99,6 +99,14 @@ pip install -f https://iree.dev/pip-release-links.html --upgrade --pre \
   iree-base-compiler iree-base-runtime iree-turbine
 ```
 
+You can also install an editable iree-turbine dep:
+```bash
+# Optionally clone and install the latest editable iree-turbine dep in deps/.
+pip install -f https://iree.dev/pip-release-links.html --upgrade --pre \
+  iree-base-compiler iree-base-runtime --src deps \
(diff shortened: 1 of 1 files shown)


docs/developer_guide.md
```diff
@@ -96,8 +96,7 @@ pip install -r requirements.txt -e sharktank/ shortfin/
 # Optionally clone and install the latest editable iree-turbine dep in deps/,
 # along with nightly versions of iree-base-compiler and iree-base-runtime.
 pip install -f https://iree.dev/pip-release-links.html --upgrade --pre \
-  iree-base-compiler iree-base-runtime --src deps \
-  -e "git+https://github.com/iree-org/iree-turbine.git#egg=iree-turbine"
+  iree-base-compiler iree-base-runtime iree-turbine
```
ScottTodd, line 99: Update the comment to match the code. I think for a developer guide we should mention the editable source install too, but still recommend most people to start with nightly packages.

ScottTodd (changes requested):

docs/developer_guide.md
marbre: I did but missed to save and commit -.-

marbre: I restored / added a comment pointing to the editable iree-turbine.

ScottTodd (approved):

PR #795, opened 2025-01-08
Patch version computations in pypi_deploy.sh.

ScottTodd:
Adding `-stable` fixes
```
Computing local versions for sharktank and shortfin...
usage: compute_local_version.py [-h] [--write-json] (-stable | -rc | -dev | --version-suffix VERSION_SUFFIX) path
compute_local_version.py: error: one of the arguments -stable/--stable-release -rc/--nightly-release -dev/--development-release --version-suffix is required
```

Adding `--write-json` fixes the `write_requirements.py` step from either
* using previously written JSON files if present (resulting in old version pins in the meta package, oops!)
* failing to open files that don't exist (failing to release, but at least not using the wrong versions)

Future changes should put more logging in place and add a confirmation step. I only caught this because I ran _part_ of the script and paused to verify manually.


marbre (approved):
Thanks.

PR #798, opened 2025-01-09
[tuner] Add timeout for compilation in TuningClient

Max191:
This PR adds a timeout for compilation to prevent the tuner from hanging when iree-compile hangs. The `iree.compiler.compile_file` binding is no longer used, since it does not support a timeout. Instead, the command is run as a regular subprocess.

### DIFF
--- a/tuner/examples/simple/simple_tuner.py
+++ b/tuner/examples/simple/simple_tuner.py
@@ -19,6 +19,9 @@ class SimpleTuner(libtuner.TuningClient):
     def get_iree_compile_flags(self) -> list[str]:
         return self.compile_flags
 
+    def get_iree_compile_timeout_s(self) -> int:
+        return 10
+
     def get_iree_benchmark_module_flags(self) -> list[str]:
         return self.benchmark_flags
 
--- a/tuner/examples/simple/simple_tuner.py
+++ b/tuner/examples/simple/simple_tuner.py
@@ -15,12 +15,13 @@ class SimpleTuner(libtuner.TuningClient):
         super().__init__(tuner_context)
         self.compile_flags: list[str] = []
         self.benchmark_flags: list[str] = []
+        self.compile_timeout: int = 10
 
     def get_iree_compile_flags(self) -> list[str]:
         return self.compile_flags
 
     def get_iree_compile_timeout_s(self) -> int:
-        return 10
+        return self.compile_timeout
 
     def get_iree_benchmark_module_flags(self) -> list[str]:
         return self.benchmark_flags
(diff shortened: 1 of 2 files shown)


tuner/tuner/libtuner.py
```diff
@@ -440,30 +445,35 @@ def run_iree_compile_command(compile_pack: CompilePack) -> Optional[int]:
...
+    # TODO(Max191): Make the device in `traget_backend` a command line option
+    # instead of hardcoding rocm.
+    target_backend = "rocm"
+    iree_compile = ireec.binaries.find_tool("iree-compile")
+    compile_command = [
+        iree_compile,
+        input_file,
+        f"--iree-input-type={str(ireec.core.InputType.AUTO)}",
+        f"--iree-vm-bytecode-module-output-format={str(ireec.core.OutputFormat.FLATBUFFER_BINARY)}",
+        f"--iree-hal-target-backends={target_backend}",
```
kuhar, line 462: This should be specified by each client

tuner/tuner/libtuner.py
```diff
@@ -440,30 +445,35 @@ def run_iree_compile_command(compile_pack: CompilePack) -> Optional[int]:
...
-            extra_args=extra_flags,
-            crash_reproducer_path=crash_dump_path,
+    # TODO(Max191): Make the device in `traget_backend` a command line option
+    # instead of hardcoding rocm.
+    target_backend = "rocm"
+    iree_compile = ireec.binaries.find_tool("iree-compile")
+    compile_command = [
+        iree_compile,
+        input_file,
+        f"--iree-input-type={str(ireec.core.InputType.AUTO)}",
```
kuhar, line 460: This is already the default, no?

tuner/tuner/libtuner.py
```diff
@@ -440,30 +445,35 @@ def run_iree_compile_command(compile_pack: CompilePack) -> Optional[int]:
...
+        f"--iree-vm-bytecode-module-output-format={str(ireec.core.OutputFormat.FLATBUFFER_BINARY)}",
```
kuhar, line 461: Also here, do we need to override the defaults?

tuner/examples/simple/simple_tuner.py
```diff
@@ -19,6 +19,9 @@ def __init__(self, tuner_context: libtuner.TunerContext):
     def get_iree_compile_flags(self) -> list[str]:
         return self.compile_flags
 
+    def get_iree_compile_timeout_s(self) -> int:
```
kuhar, line 22: We should make this a field. That we update in the depending on whether we compile dispatches or models. For dispatches, 10s is reasonable, but for models a good starting point may be something like 60s.

kuhar (changes requested):

kuhar (approved):

PR #799, opened 2025-01-09
[tuner] Filter out non finite benchmark times

Max191:
This PR fixes a bug where math.inf benchmark times can be selected as the top candidates. Any non finite times are now filtered out before selecting top candidates.


tuner/tuner/libtuner.py
```diff
@@ -17,6 +17,7 @@
 
 
 import math
+import random
```
kuhar, line 20: This seems unused?

kuhar (approved):

tuner/tuner/libtuner.py
Max191: Yes, I was using it for debugging. Thanks for the catch.

PR #800, opened 2025-01-09
[tuner] Write candidate specs with local scope

Max191:
This PR changes the format of generated TD specs to inline the compilation info into the matcher function. This makes splitting the TD spec and combining with others after tuning much easier.


kuhar (approved):
Great, thanks for fixing this

PR #802, opened 2025-01-09
Bump IREE version pins to 3.2.0rc20250109.

ScottTodd:
See https://github.com/nod-ai/shark-ai/issues/760 for context. We want to stay close to the latest versions while still pinning versions for predictability. Updating version pins is currently a manual process but we plan on automating it in the future.

We can decide how noisy we want these dependency updates to be:

* new PRs daily or less frequently
* do or don't reuse existing PRs
* merge ASAP or let them sit for multiple days

### DIFF
--- a/requirements-iree-pinned.txt
+++ b/requirements-iree-pinned.txt
@@ -3,6 +3,6 @@
 # Keep these versions synced with SHORTFIN_IREE_GIT_TAG in shortfin/CMakeLists.txt
 --pre
 --find-links https://iree.dev/pip-release-links.html
-iree-base-compiler==3.1.0rc20250107
-iree-base-runtime==3.1.0rc20250107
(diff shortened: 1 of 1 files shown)


monorimet (approved):
LGTM

ScottTodd:
Note that this does not yet update the IREE runtime version pin used for shortfin: https://github.com/nod-ai/shark-ai/blob/1e23a74fd1bcad87e0af3c4e2b32a8323c16c7ff/shortfin/CMakeLists.txt#L46-L50

With https://github.com/iree-org/iree/pull/19645 merged, we could try bumping that to https://github.com/iree-org/iree/commit/9055c9d1f6342a061a6747ef9b385816b96a0a8f, then future version updates can try to update both pinned versions.

PR #812, opened 2025-01-10
[shortfin] Add a tracing enabled CI job.

ScottTodd:
Also update the pinned IREE commit used for building shortfin now that a fix for Tracy on gcc has landed.

### DIFF
--- a/.github/workflows/ci-libshortfin.yml
+++ b/.github/workflows/ci-libshortfin.yml
@@ -127,13 +127,13 @@ jobs:
         cmake --build build --target all
 
     - name: pip install shortfin
-      if: ${{ matrix.name != 'Ubuntu (Clang)(host-only)'}}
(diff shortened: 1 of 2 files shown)


marbre (approved):

ScottTodd:
Made a few changes. PTAL @marbre

ScottTodd:
post-merge review should be fine too

.github/workflows/ci-libshortfin.yml
```diff
@@ -64,6 +64,12 @@ jobs:
             python-version: "3.12"
             cmake-options:
               -DCMAKE_C_COMPILER=gcc-14 -DCMAKE_CXX_COMPILER=g++-14
+          - name: Ubuntu (GCC 14, tracing)
+            runs-on: ubuntu-24.04
+            # Only test with GCC 14 and Python 3.12
```
marbre, line 69: As you disabled testing (which is fine) we could rephrase to
```suggestion
            # Only build with GCC 14 and Python 3.12
```
But no strong need.

PR #813, opened 2025-01-10
Switch workflows to use new requirements-iree-*.txt files.

ScottTodd:
Progress on https://github.com/nod-ai/shark-ai/issues/760. We could make the scheduled jobs test both pinned and unpinned versions like on https://github.com/nod-ai/shark-ai/pull/767.

Cleanup included here:

* Dropped the "Installing the PyTorch CPU wheels saves multiple minutes and a lot of bandwidth on runner setup." comments since they are repetitive. Could add them back if people find them useful.
* Stopped installing from the root `requirements.txt` in some workflows, instead opting to just install from the more specific `sharktank/requirements-tests.txt`

I did not test the changes to scheduled workflows. Could do that on request, or just revert if we see issues.


.github/workflows/ci-sglang-benchmark.yml
```diff
@@ -61,18 +61,13 @@ jobs:
...
           pip install --no-compile -r pytorch-cpu-requirements.txt
-          pip install -f https://iree.dev/pip-release-links.html --pre iree-turbine
-          pip install --no-compile -r requirements.txt -e sharktank/ shortfin/
-
           # Pin to known-working versions.
-          pip install -f https://iree.dev/pip-release-links.html \
-            iree-base-compiler==3.1.0rc20241220 \
-            iree-base-runtime==3.1.0rc20241220 \
-            "numpy<2.0"
+          pip install -r requirements-iree-pinned.txt
```
archana-ramalingam, line 69: Benchmark tests are usually `unpinned` to track performance on latest releases. @stbaione can confirm.

archana-ramalingam (approved):

.github/workflows/ci-llama-large-tests.yaml
```diff
@@ -55,18 +55,12 @@ jobs:
...
           pip install --no-compile -r pytorch-cpu-requirements.txt
-
-          # Install nightly IREE packages.
-          # We could also pin to a known working or stable version.
-          pip install -f https://iree.dev/pip-release-links.html --pre --upgrade \
-            iree-base-compiler \
-            iree-base-runtime \
-            iree-turbine
-
-          pip install --no-compile -r requirements.txt -r sharktank/requirements-tests.txt -e sharktank/
```
marbre, line 69: With dropping `requirements.txt` (which I am highly supportive for), this also drops installing [requirements-dev.txt](https://github.com/nod-ai/shark-ai/blob/main/requirements-dev.txt) and (probably more important) [sharktank/requirements.txt](https://github.com/nod-ai/shark-ai/blob/main/sharktank/requirements.txt). The latter might be pulled in via `-e sharktank/` as the requirements file is specified in sharktank's `pyproject.toml`.

.github/workflows/ci_eval.yaml
```diff
@@ -53,18 +53,13 @@ jobs:
           python -m pip install --no-compile --upgrade pip
 
           # Note: We install in three steps in order to satisfy requirements
-          # from non default locations first. Installing the PyTorch CPU
-          # wheels saves multiple minutes and a lot of bandwidth on runner setup.
+          # from non default locations first.
           pip install --no-compile -r pytorch-cpu-requirements.txt
+          pip install -r requirements-iree-unpinned.txt
+          pip install --no-compile \
+            -r sharktank/requirements-tests.txt \
```
marbre, line 60: As above.

.github/workflows/ci-llama-large-tests.yaml
ScottTodd: Yeah I think `sharktank/requirements-tests.txt` and `-e sharktank/` should be sufficient, given that sharktank includes its deps/requirements.

.github/workflows/ci-sglang-benchmark.yml
```diff
@@ -61,18 +61,13 @@ jobs:
...
+          pip install --no-compile -r requirements.txt -e sharktank/ shortfin/
```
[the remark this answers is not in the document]
ScottTodd: Thanks, I'll accept those suggestions. They look correct.

.github/workflows/ci-sglang-benchmark.yml
stbaione: I've been leaving this unpinned, so that we could still get performance numbers, even if a regression is introduced in a nightly IREE version. However, the more I hear the suggestion to leave unpinned and test against nightly for always the latest performance, the more it sounds like a better idea. I think this can be switched to using unpinned

ScottTodd: Switched to unpinned here.

.github/workflows/ci-sglang-benchmark.yml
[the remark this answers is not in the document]
ScottTodd: This doesn't seem to work:
```
-r sharktank/requirements-tests.txt shortfin/requirements-tests.txt
```
This does work:
```
-r sharktank/requirements-tests.txt \
-r shortfin/requirements-tests.txt
```

.github/workflows/ci_eval.yaml
ScottTodd: I'm not following here. What change do you want made?

I believe
```
          pip install --no-compile \
            -r sharktank/requirements-tests.txt \
            -e sharktank/
```

is sufficient, no need to also install `sharktank/requirements.txt` or `sharktank/requirements-dev.txt`

marbre: As commented above with this, this drops installing the dependencies specified in `sharktank/requirements-dev.txt`. Just wanted to point out that this is the same here but that might have been a little to short. No change needed.

.github/workflows/ci-sglang-benchmark.yml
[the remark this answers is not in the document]
marbre: Ah yes, makes sense.

marbre (approved):
Nice, thanks!

Issue #806, opened 2025-01-09
[Tuner] Reduce log file size


kuhar:
The tuner outputs all the logs to a single file `autotune_n.log`, which can easily grow to hundreds of MB in size.

We should reduce the amount of information printed and make it easier to read. Specifically:
- We should not print compilation errors
- We should reduce the precision of fp constants printed (e.g., benchmark times)
- We should not print candidate trackers at the end

Further, we should summarize the most important information, like top candidates (dispatch and model) and paths to their specs in a separate file.

bangtianliu:
Small todo: fix wrong print info in the log like 
> Successfully compiled [6] candidates. Success rate: 1.20

PR #815, opened 2025-01-11
[tuner] add the output file

bangtianliu:
This PR addresses the task described in https://github.com/nod-ai/shark-ai/issues/806.

It adds a separate output file to the tuner to summarize the most important information, such as the top candidates (dispatch and model) and paths to their specifications.

Also, it fixes the issue with incorrect reporting of the compilation success rate in the log.

### DIFF
--- a/tuner/examples/simple/simple_tuner.py
+++ b/tuner/examples/simple/simple_tuner.py
@@ -80,7 +80,7 @@ def main():
     stop_after_phase: str = args.stop_after
 
     print("Setup logging")
-    tune_logger = libtuner.setup_logging(args, path_config)
+    root_logger = libtuner.setup_logging(args, path_config)
     print(path_config.run_log, end="\n\n")
 
     if not args.dry_run:
@@ -93,12 +93,15 @@ def main():
         args.simple_model_benchmark_flags_file
     )
 
+    summary_log_file = path_config.base_dir / "summary.log"
+    path_config.set_summary_log(summary_log_file)
+    summary_handler = logging.FileHandler(summary_log_file)
+    summary_handler.setLevel(logging.INFO)
+    summary_handler.setFormatter(
+        logging.Formatter("%(asctime)s - %(levelname)s - %(message)s")
+    )
     print("Generating candidate tuning specs...")
-    with TunerContext(logger=tune_logger) as tuner_context:
-        summary_log_file = path_config.base_dir / "summary.log"
-        path_config.set_summary_log(summary_log_file)
-        summary_handler = logging.FileHandler(summary_log_file)
-        summary_handler.setLevel(logging.INFO)
+    with TunerContext(logger=root_logger) as tuner_context:
         tuner_context.add_logging_handler(summary_handler)
         simple_tuner = SimpleTuner(tuner_context)
         candidates = libtuner.generate_candidate_specs(
@@ -119,7 +122,7 @@ def main():
             return
 
         print("Benchmarking compiled dispatch candidates...")
-        tune_logger.info(f"Summarization about top dispatch candidates:")
+        logging.info(f"Summarization about top dispatch candidates:")
         simple_tuner.benchmark_flags = ["--input=1", "--benchmark_repetitions=3"]
         top_candidates = libtuner.benchmark(
             tuner_context,
@@ -130,9 +133,9 @@ def main():
             simple_tuner,
             args.simple_num_dispatch_candidates,
         )
-        tune_logger.info(f"Top dispatch candidates: {top_candidates}")
+        logging.info(f"Top dispatch candidates: {top_candidates}")
         for id in top_candidates:
-            tune_logger.info(f"{candidate_trackers[id].spec_path.resolve()}")
+            logging.info(f"{candidate_trackers[id].spec_path.resolve()}")
         if stop_after_phase == libtuner.ExecutionPhases.benchmark_dispatches:
             return
 
@@ -151,7 +154,7 @@ def main():
             return
 
         print("Benchmarking compiled model candidates...")
-        tune_logger.info(f"Summarization about top model candidates:")
+        logging.info(f"Summarization about top model candidates:")
         simple_tuner.benchmark_flags = model_benchmark_flags
         simple_tuner.benchmark_timeout = 60
         top_model_candidates = libtuner.benchmark(
@@ -163,9 +166,9 @@ def main():
             simple_tuner,
             args.simple_num_model_candidates,
         )
-        tune_logger.info(f"Top model candidates: {top_model_candidates}")
+        logging.info(f"Top model candidates: {top_model_candidates}")
         for id in top_model_candidates:
-            tune_logger.info(f"{candidate_trackers[id].spec_path.resolve()}")
+            logging.info(f"{candidate_trackers[id].spec_path.resolve()}")
         print(f"Top model candidates: {top_model_candidates}")
 
         print("Check the detailed execution logs in:")
--- a/tuner/examples/simple/simple_tuner.py
+++ b/tuner/examples/simple/simple_tuner.py
@@ -75,6 +75,7 @@ def main():
 
     path_config = libtuner.PathConfig()
     path_config.base_dir.mkdir(parents=True, exist_ok=True)
+    path_config._set_tune_output(Path("autotune_output.txt"))
     # TODO(Max191): Make candidate_trackers internal to TuningClient.
     candidate_trackers: list[libtuner.CandidateTracker] = []
     stop_after_phase: str = args.stop_after
@@ -156,3 +157,13 @@ def main():
 
         print("Check the detailed execution logs in:")
         print(path_config.run_log.resolve())
+
+        print("Check the tuning results in:")
+        print(path_config.tune_output.resolve())
+        with open(path_config.tune_output, "w") as file:
+            file.write(f"Top dispatch candidates: {top_candidates}\n")
+            for id in top_candidates:
+                file.write(f"{candidate_trackers[id].spec_path.resolve()}\n")
+            file.write(f"Top model candidates: {top_model_candidates}\n")
+            for id in top_model_candidates:
+                file.write(f"{candidate_trackers[id].spec_path.resolve()}\n")
--- a/tuner/examples/simple/simple_tuner.py
+++ b/tuner/examples/simple/simple_tuner.py
@@ -81,7 +81,8 @@ def main():
 
     print("Setup logging")
     libtuner.setup_logging(args, path_config)
-    print(path_config.run_log, end="\n\n")
+    print(path_config.run_log, end="\n")
+    print(path_config.summary_log, end="\n\n")
 
     if not args.dry_run:
         print("Validating devices")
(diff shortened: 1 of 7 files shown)


tuner/examples/simple/simple_tuner.py
```diff
@@ -75,6 +75,7 @@ def main():
 
     path_config = libtuner.PathConfig()
     path_config.base_dir.mkdir(parents=True, exist_ok=True)
+    path_config._set_tune_output(Path("autotune_output.txt"))
```
kuhar, line 78: This appears to be a private function (it starts with an underscore), so we shouldn't call it directly in the current form.

tuner/examples/simple/simple_tuner.py
```diff
@@ -156,3 +157,13 @@ def main():
...
         print(path_config.run_log.resolve())
+
+        print("Check the tuning results in:")
+        print(path_config.tune_output.resolve())
+        with open(path_config.tune_output, "w") as file:
+            file.write(f"Top dispatch candidates: {top_candidates}\n")
+            for id in top_candidates:
+                file.write(f"{candidate_trackers[id].spec_path.resolve()}\n")
+            file.write(f"Top model candidates: {top_model_candidates}\n")
+            for id in top_model_candidates:
```
kuhar, line 168: We should write these as they become available, not at the very end. This is in case something fails / we interrupt the tuner -- in these situations partial results are still useful.

tuner/examples/simple/simple_tuner.py
kuhar, line 78: Also, can we give it a more descriptive name than `output`? It would not be clear to what to expect based on this name. Maybe `summary.log`?

tuner/tuner/libtuner.py
```diff
@@ -801,7 +805,7 @@ def compile(
         num_worker=num_worker, task_list=task_list, function=run_iree_compile_command
     )
     compiled_candidates = [c for c in compiled_candidates if c is not None]
-    success_rate = float(len(compiled_candidates)) / float(len(candidates))
+    success_rate = float(len(compiled_candidates)) / float(len(task_list))
```
kuhar, line 808: Nice catch! I wish we had a test for this too...

kuhar (changes requested):
Could you also add the time information (absolute time and % of the baseline)?

tuner/tuner/libtuner.py
```diff
@@ -86,6 +87,7 @@ def __post_init__(self):
         object.__setattr__(self, "candidates_dir", self.base_dir / "candidates")
         object.__setattr__(self, "compiled_dir", self.candidates_dir / "compiled")
         object.__setattr__(self, "specs_dir", self.candidates_dir / "specs")
+        object.__setattr__(self, "output_dir", self.base_dir / "summary.log")
```
kuhar, line 90: This is not a directory -- it's a path to a file

kuhar (commented):
We should open the file once and keep appending to it

kuhar:
Instead of opening a plaintext file, we can have multiple loggers configured with different logging level, such that the summary log contains only the most important messages: https://docs.python.org/3/howto/logging-cookbook.html#logging-to-multiple-destinations

tuner/tuner/libtuner.py
```diff
@@ -717,10 +723,16 @@ def generate_candidate_specs(
...
         raise
 
-    logging.info(f"Generated [{len(candidates) - 1}] candidates")
+    logging.debug(f"Generated [{len(candidates) - 1}] candidates")
     return candidates
 
 
+def get_compilation_success_rate(compiled_candiates: list[Any]) -> float:
+    successful_candidates = [c for c in compiled_candiates if c is not None]
+    success_rate = float(len(successful_candidates)) / float(len(compiled_candiates))
```
kuhar, line 732: This can cause division by zero, no? Can we add a unit test for this?

tuner/tuner/libtuner.py
```diff
@@ -717,10 +723,16 @@ def generate_candidate_specs(
         tune_logger.exception("Error in candidate_gen.py:")
         raise
 
-    logging.info(f"Generated [{len(candidates) - 1}] candidates")
+    logging.debug(f"Generated [{len(candidates) - 1}] candidates")
     return candidates
 
 
+def get_compilation_success_rate(compiled_candiates: list[Any]) -> float:
```
kuhar, line 730: Why are we using `Any` for the element type?

tuner/tuner/libtuner.py
```diff
@@ -868,18 +881,20 @@ def get_speedup(result: BenchmarkResult) -> float:
     best_results = sorted(filtered_candidate_results, key=sorting_key)[
         :num_top_candidates
     ]
-    logging.info(f"Selected top[{len(best_results)}]:")
+    tuner_context.logger.info(f"Selected top[{len(best_results)}]:")
```
kuhar, line 884: What happens when we use the root logger here? Why do we have to explicitly log via the logger in the context?

tuner/tuner/libtuner_test.py
```diff
@@ -176,7 +188,18 @@ def test_validate_devices_with_invalid_device() -> None:
...
-def test_select_best_benchmark_results() -> None:
+def test_get_compilation_success_rate():
+    compiled_candidates = [0, None, 2, None, 4]
+    assert libtuner.get_compilation_success_rate(compiled_candidates) == 3.0 / 5.0
+
+    compiled_candidates = [0, 1, 2, 3, 4]
+    assert libtuner.get_compilation_success_rate(compiled_candidates) == 1.0
+
+    compiled_candidates = [None, None, None]
+    assert libtuner.get_compilation_success_rate(compiled_candidates) == 0.0
```
kuhar, line 199: Missing testcase: empty input list

tuner/tuner/libtuner_test.py
```diff
@@ -10,13 +10,25 @@
...
 
 
+@pytest.fixture
+def tuner_ctx() -> Generator[common.TunerContext, None, None]:
+    from logging import Logger
+    from unittest.mock import MagicMock
+
+    mock_logger = MagicMock(spec=Logger)
+    with common.TunerContext(logger=mock_logger) as ctx:
+        yield ctx
```
kuhar, line 29: This appears unused

tuner/tuner/libtuner.py
bangtianliu: No, compiled_candidates will never be empty. Even if all attempts fail, it will still be a list containing only None values.

tuner/tuner/libtuner.py
bangtianliu: if we use root, it will not output to the summary log.

tuner/tuner/libtuner_test.py
bangtianliu: It does use for test_select_best_benchmark_results.

tuner/tuner/libtuner.py
kuhar: Why is that? I thought that if we register the logging handler for summary, it will be attached to the root logger.

tuner/tuner/libtuner.py
kuhar: I can see this happening if we for whatever reason select the max number of candidates to generate to 0

tuner/tuner/libtuner.py
bangtianliu: Yes, if we use root, it will go to the root logger but not to the summary log. Using tuner_context.logger can ensure that it goes to both the root and summary loggers. Not sure, I guess it is relevant to context management.

tuner/tuner/libtuner.py
bangtianliu: Sure, I will see how to support this case.

kuhar (commented):
Can you upload a sample summary log file so that we know what to expect after merging this?

bangtianliu:
> Can you upload a sample summary log file so that we know what to expect after merging this?

[summary.log](https://github.com/user-attachments/files/18412138/summary.log)

tuner/examples/simple/simple_tuner.py
```diff
@@ -114,15 +122,20 @@ def main():
             return
 
         print("Benchmarking compiled dispatch candidates...")
+        logging.info(f"Summarization about top dispatch candidates:")
```
kuhar, line 125: No need for fstring, we don't print any variables here. Also, the grammar is weird. The summary is not ready at this point, so how about we just print the same thing as the print above?

tuner/examples/simple/simple_tuner.py
```diff
@@ -141,18 +154,25 @@ def main():
             return
 
         print("Benchmarking compiled model candidates...")
+        logging.info(f"Summarization about top model candidates:")
```
kuhar, line 157: Same here

tuner/examples/simple/simple_tuner.py
```diff
@@ -141,18 +154,25 @@ def main():
...
-
+        logging.info(f"Top model candidates: {top_model_candidates}")
+        for id in top_model_candidates:
+            logging.info(f"{candidate_trackers[id].spec_path.resolve()}")
         print(f"Top model candidates: {top_model_candidates}")
 
         print("Check the detailed execution logs in:")
         print(path_config.run_log.resolve())
+
+        print("Check the tuning results in:")
```
kuhar, line 177:
```suggestion
        print("Check the summary in:")
```

tuner/tuner/libtuner.py
```diff
@@ -875,11 +889,13 @@ def get_speedup(result: BenchmarkResult) -> float:
...
         else:
             speedup = "baseline unavailable"
-        logging.info(f"Candidate {r.candidate_id} time: {r.time:.2f} ({speedup})")
+        result = f"Candidate {r.candidate_id} time: {r.time:.2f} ms ({speedup})"
+        logging.info(result)
     return best_results
 
 
 def benchmark(
+    tuner_context: TunerContext,
```
kuhar, line 898: Do we still need this?

tuner/tuner/common.py
```diff
@@ -46,6 +46,9 @@ def __enter__(self) -> "TunerContext":
         self.mlir_ctx.__enter__()
         return self
 
+    def add_logging_handler(self, handler: logging.Handler) -> None:
+        self.logger.addHandler(handler)
```
kuhar, line 50: I don't think we need this helper, users can access the logger directly.

kuhar (commented):
Overall looks good, but it seems like there's some stuff to clean up still

tuner/tuner/libtuner.py
```diff
@@ -92,9 +93,12 @@ def _name_base_dir(self) -> Path:
         base_dir = Path(f"./tuning_{timestamp}")
         return base_dir
 
-    def _set_run_log(self, run_log: Path):
+    def set_run_log(self, run_log: Path):
         object.__setattr__(self, "run_log", run_log)
 
+    def set_summary_log(self, summary_log: Path):
+        object.__setattr__(self, "summary_log", summary_log)
```
kuhar, line 100: Do we need to keep track of the summary log here? I'd think that the simple tuner has the path and that `libtuner` doesn't have to know about it

bangtianliu: Yeah, good point!

kuhar (approved):
LGTM

PR #825, opened 2025-01-14
Added experimental mm padding for cache behavior

rsuderman:
Padding to cache lines can significantly improve performance by separating out cache lines between iterations.


rsuderman:
Include sample invocation:
```
python3 -m sharktank.examples.export_paged_llm_v1 --block-seq-stride=32 --attention-kernel=torch --bs=4 --irpa-file irpa/llama.8b.fp16.irpa --output-mlir /tmp/llama.mlir --experimental-mm-cache-size=128 --^Cperimental-mm-cache-s
ets=4
```
=

PR #826, opened 2025-01-14
Fix shortfin cpu llm concurrency test problems by zero-fill start positions instead of one-filling

renxida:
PRs in the history of this problem: #665, #723

#665 is supposed to fix a NaN cache corruption issue by 1-filling seq_len instead of 0-filling. Its supposed to 1-fill seq_len for decode and prefill, but I mistakenly 1-filled seq_len for decode only, and also 1-filled the start_position for decode instead of prefill seq_len.

#723 adds 1-filling for prefill, and this PR removes the mistaken start_positions 1-filling for decode.

After this PR shortfin concurrent tests should be working properly.

Up next: a failing trie kv sharing test case.

### DIFF
--- a/shortfin/python/shortfin_apps/llm/components/service.py
+++ b/shortfin/python/shortfin_apps/llm/components/service.py
@@ -435,11 +435,21 @@ class InferenceExecutorProcess(sf.Process):
                 args = [tokens, seq_lens, start_positions, seq_block_ids]
             else:
                 args = [tokens, seq_lens, seq_block_ids]
-            args.extend(self.page_tables)
+            for page_table in self.page_tables:
+                args.append(sfnp.disable_barrier(page_table))
             logger.info(
                 "INVOKE %r: %s",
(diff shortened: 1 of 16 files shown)


shortfin/python/shortfin_apps/llm/components/service.py
```diff
@@ -394,7 +394,7 @@ async def run(self):
                 start_positions_host = start_positions.for_transfer()
                 with start_positions_host.map(discard=True) as m:
                     m.fill(
-                        1
+                        0
                     )  # Pad unused requests. Must pad with nonzero value because division by 0 floods clobber page (page 0) in cache with NaN values.
```
stbaione, line 398: Should remove/change comment if it no longer applies

renxida: missed that! ty

stbaione (approved):

PR #827, opened 2025-01-14
Add script+workflow to update IREE requirement pins.

ScottTodd:
Progress on https://github.com/nod-ai/shark-ai/issues/760, built off of the work in https://github.com/iree-org/iree-turbine/pull/388.

This adds a new workflow that runs once a day to update all pinned IREE versions. I also looked into using Dependabot but found that it struggles with `--find-links`, `--index-url`, and with there being multiple `requirements.txt` files in a repository. While I would love to not need to reinvent this wheel, I do like keeping full control over the process.

This PR includes:

* A new `build_tools/update_iree_requirement_pins.py` script handles updating the pins in `requirements-iree-pinned.txt` and `shortfin/CMakeLists.txt`. The script also sets some variables in `GITHUB_ENV`.
* A new `.github/workflows/update_iree_requirement_pins.yml` workflow runs that script then calls https://github.com/peter-evans/create-pull-request to create or update a pull request if there are local changes after running that script. The commit message and pull request body are constructed using the variables set by the script.

Test action run: https://github.com/ScottTodd/shark-ai/actions/runs/12777789320
Test pull request: https://github.com/ScottTodd/shark-ai/pull/1


ScottTodd:
@renxida I figured you might want to review this (and/or https://github.com/iree-org/iree-turbine/pull/388), given your prior work on https://github.com/iree-org/iree/blob/main/.github/workflows/bump_torch_mlir.yml

.github/workflows/update_iree_requirement_pins.yml
```diff
@@ -0,0 +1,50 @@
...
+        uses: peter-evans/create-pull-request@67ccf781d68cd99b580ae25a5c18a1cc84ffff1f # v7.0.6
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          base: main
+          branch: integrates/iree
+          delete-branch: true
+          signoff: true
+          title: "Bump IREE requirement pins to their latest versions."
+          body: |
+            Diff: https://github.com/iree-org/iree/compare/iree-${{ env.CURRENT_IREE_BASE_COMPILER_VERSION }}...iree-${{ env.LATEST_IREE_BASE_COMPILER_VERSION }}
```
renxida, line 47: Nifty!

renxida (approved):

.github/workflows/update_iree_requirement_pins.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: "Setting up Python"
+        uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.0
+        with:
+          python-version: 3.11
+
+      # This sets a few environment variables used below.
+      - name: Update IREE requirement pins
+        run: build_tools/update_iree_requirement_pins.py
+
+      - name: Create or update pull request
```
renxida, line 32: did you have to set up keys / perms or anything like that to auto-create PRs in your own personal repo? or is it as simple as just setting the default branch to this?

ScottTodd: I have these settings in my forks:
image

I see that we don't have the same settings in this repository though. May need to change that for this to work, or create a token with broader permissions and pass that around via secrets. I can run some tests on my fork with more limited settings to see what happens.

ScottTodd: Good catch. Needed to add
```
permissions:
  contents: write
  pull-requests: write
```
to work with the current default permissions. We do also need that "Allow GitHub Actions to create and approve pull requests" setting checked. I'll check if we can flip that here.

build_tools/update_iree_requirement_pins.py
```diff
@@ -0,0 +1,195 @@
...
+        "--find-links",
+        "https://iree.dev/pip-release-links.html",
+    ]
+    latest_compiler_version = get_latest_version("iree-base-compiler", nightly_pip_args)
+    latest_runtime_version = get_latest_version("iree-base-runtime", nightly_pip_args)
+    latest_turbine_version = get_latest_version("iree-turbine", nightly_pip_args)
+    # TODO(scotttodd): Get this from git? It should generally be in sync with
+    #     the python packages and follow a naming convention. If that isn't
+    #     true, such as right after a stable release, then this may break.
+    latest_git_tag = f"iree-{latest_runtime_version}"
```
marbre, line 111: I thought through this a little longer and I think there is no reason this should not be in sync, even if patch releases for iree-base-compiler get published. The tag + the latest wheel version should always be in sync.

marbre (approved):

build_tools/update_iree_requirement_pins.py
ScottTodd: Consider the 3.2.0 release. There will be a window between when we publish 3.2.0 to PyPI and when we publish the first 3.3.0rcYYYYMMDD release during which the latest version fetched by this script will be 3.2.0. This will try to use the git tag `iree-3.2.0` which will not exist if we stick to the new `vX.Y.Z` tag naming (e.g. https://github.com/iree-org/iree/releases/tag/v3.1.0). That would not have been an issue back when we used `iree-X.Y.Z` (e.g. https://github.com/iree-org/iree/releases/tag/iree-3.0.0)

ScottTodd: One workaround would be to change the `get_latest_version()` function to scrape https://iree.dev/pip-release-links.html instead of calling `pip index`. The release links page _only_ contains prereleases, while `pip index` searches that page _and_ PyPI.

I'm not too concerned about this edge case though, since we'll be paying pretty close attention to the repositories and package versions around when we push a stable release. Where it will matter is if we decide to change the tag naming convention in the future, since the script expects the `iree-PACKAGE_VERSION` naming scheme.

marbre: We can also push `iree-X.Y.Z` + `vX.Y.Z` for stable releases ;)

ScottTodd:
Holding off on merging until a change like https://github.com/iree-org/iree-turbine/pull/391 is ready for the nod-ai organization.

Without that, or another of the workarounds from https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs, pull requests created by this action will not run other actions.

ScottTodd:
Ready with the GITHUB_TOKEN changes but need some version of this action on the `main` branch to test in prod. Merging then will send a follow-up PR that fixes the token permissions.

PR #831, opened 2025-01-15
Fix Flux transformer numerics and add test comparing eager against Hugging Face

sogartar:
* Fix layer norm to use the correct normalized_shape.
* Fix attention to not be causal.
* Make GELU use Tanh approximation.

Add tests that compare eager bf16 and f32 against Huggging Face's transformer.

### DIFF
--- a/sharktank/tests/models/flux/flux_test.py
+++ b/sharktank/tests/models/flux/flux_test.py
@@ -62,7 +62,7 @@ iree_compile_flags = [
 ]
 
 
-def covert_dtype_if_dtype(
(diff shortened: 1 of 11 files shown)


sharktank/tests/models/flux/flux_test.py
```diff
@@ -58,6 +62,14 @@
 ]
 
 
+def covert_dtype_if_dtype(
```
KyleHerndon, line 65: nit: convert

KyleHerndon (approved):
Looks great, nice finds!

PR #832, opened 2025-01-16
[llama] Generate causal mask better

Groverkss:


Groverkss:
Turning into draft since I want to see if we can make the compiler propagate the slice.

PR #838, opened 2025-01-17
[shortfin_apps/llm] Add locks to protect trie nodes, node reference counts, and the trie structure itself from race conditions

renxida:


renxida:
The problem I set out to fix in this was fixed by #874. Still going to merge something like this later but lower priority now.

renxida:
Closing for now - need to focus on testing changes.

PR #840, opened 2025-01-17
Fast server startup, using file_handle to load model param files

stbaione:
# Description

Currently, when we load model parameter files, we load the entire contents of the files into memory, then mmap that data to the devices. This would cause very long server startup times. For example, 70b (~130 GB) took me 10 minutes to start the server, and 405b (~750 GB) took me over `5 hours` to start the server.

As an alternative, this PR uses `iree_io_file_handle_open` to obtain a handle to the parameter files, then streams that data to the devices, insteading of mmaping it. After this change, we are able to start the server for `70b` and `405b` within seconds.

We default to the new method and add a private function `LoadMmap` for cases where `mmap == true`. This should improve the startup time for both `LLM` and `SDXL`, especially when loading large files.


monorimet (approved):
LGTM.

PR #842, opened 2025-01-17
[shortfin][python] Skip tokenization in sd.server if request holds `input_ids`

vinayakdsci:
`sd.server` currently performs tokenization irrespective of whether the request has `input_ids` attached to it.

This patch adds a check to skip tokenization if the request has tokenized inputs attached to it, and adds `input_ids` to `gen_inputs` so that such requests can be created from `InferenceExecRequests.from_batch()`.


monorimet (approved):
LGTM. Thanks!

vinayakdsci:
Thanks for the approval! Will merge when the CI finishes.

PR #844, opened 2025-01-18
Add trace_tensor op and refactor tensor tracing

sogartar:
We don't have an op that dispatches to the underlying iree.turbine.ops.iree.trace_tensor.

With this change we can trace into safetensors files when executing both in eager and with IREE.
Routes all tracing through the new tracing op. The user can set a desired sink.

Added some functionality to prefix trace keys based on the module structure.

Something that I could not preserve is the tensor trace counter. It is a global variable and when exporting we pass multiple times through the traced function, which makes the counter inconsistent. The only way to distinguish traced tensors is through the trace keys.

The distinction between tensor tracing and goldens tracing has been remove at the point of issuing a trace. If the user needs different treatment they can customize the trace sink.

### DIFF
--- a/sharktank/sharktank/utils/testing.py
+++ b/sharktank/sharktank/utils/testing.py
@@ -7,6 +7,7 @@
 from typing import Optional
 import contextlib
 from pathlib import Path
+from os import PathLike
 import os
 import shutil
 import tempfile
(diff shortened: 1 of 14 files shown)


sharktank/sharktank/utils/testing.py
```diff
@@ -169,7 +215,15 @@ def print_stats(label, t):
         for name in ref_f.keys():
             actual = actual_f.get_tensor(name)
             ref = ref_f.get_tensor(name)
-            torch.testing.assert_close(actual, ref, msg=name)
+            try:
```
dan-garvey, line 218: isn't this redundant to what torch.testing.assert_close already does?

dan-garvey (approved):
Small nit, otherwise looks great. Thanks for adding good tests

sharktank/sharktank/utils/testing.py
sogartar: If `fail_fast` is `False` then we will keep going and print the failures. I did this to allow for a full report that shows all discrepancies.

PR #849, opened 2025-01-20
[Llama] Remove inplace read for KVCache

Groverkss:
The in-place read for KVCache doesn't seem to be used and is just used to tell the KVCache number of partitions to read from. The number of partitions for KVCache is in reality fixed (as from the name, K/V cache) and we are adding overhead by passing this parameter.


dan-garvey (approved):
+15, -263 is what I like to see. If we don't need all this thats great.

Groverkss:
The flux test failure seems unrelated, merging

PR #854, opened 2025-01-21
[Llama] Remove DirectKVCache

Groverkss:
From my reading of it, it looks like DirectKVCache was used when PagedKVCache wasn't implemented properly, but now we have different llm layers for Paged variants and Direct variants. Paged variants explicitly use PagedKVCache, while Direct variants don't really need a cache, just need a cache k/v tensor passed to them through the direct layer.

My main problem with DirectKVCache was it's being used inconsistently, with paged variants trying to if case if it's a DirectKVCache.


sharktank/sharktank/layers/llama_attention_block.py
```diff
@@ -5,6 +5,7 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 from typing import Optional
+import math
```
dan-garvey, line 8: any reason to change torch.sqrt to math? historically there are problems tracing this

dan-garvey (changes requested):
Other than the one comment it looks good.

sharktank/sharktank/layers/llama_attention_block.py
Groverkss: Its to take sqrt of a constant (head dim value). This will always be constant at tracing time.

dan-garvey (approved):

PR #856, opened 2025-01-21
[Llama] Do not allow configurable partitions for KVCache

Groverkss:
This has been inconsistent in multiple places. KVCache generally only needs 2 partitions, one for Key and another for Value. Some places assume there are two partitions, other allow multiple partitions. This PR makes it that the cache is always partitioned into 2.

### DIFF
--- a/sharktank/sharktank/layers/kv_cache.py
+++ b/sharktank/sharktank/layers/kv_cache.py
@@ -55,7 +55,6 @@ class PagedKVCache:
         transformer_block_count: int,
         attn_head_count: int,
         attn_head_dim: int,
-        cache_partition_count: int = 2,
         block_seq_stride: int = 16,
         dtype: torch.dtype = torch.float32,
         device: Optional[torch.device] = None,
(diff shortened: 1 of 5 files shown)


sharktank/sharktank/layers/kv_cache.py
```diff
@@ -64,7 +63,6 @@ def __init__(
         self.transformer_block_count = transformer_block_count
         self.attn_head_count = attn_head_count
         self.attn_head_dim = attn_head_dim
-        self.cache_partition_count = cache_partition_count
```
dan-garvey, line 67: Lets go ahead and keep the member variable but remove the argument instead setting this as 2 to avoid using unnamed `2`'s below. You can add a comment that it should remain fixed if you feel that its ambiguous.

dan-garvey (changes requested):
small nit, otherwise great simplification

sharktank/sharktank/layers/kv_cache.py
Groverkss: done

renxida:
Would this affect deepseek's MLA where instead of caching K and V, we cache only one thing (the latent low rank compression that it uses to generate both K and V)?

PR #725, opened 2024-12-20
Switch eval and llama tests to use the default hip device.

ScottTodd:
Some of these workflows have been failing with
```
>           hal_device_id = haldriver.query_available_devices()[device_idx]["device_id"]
E           IndexError: list index out of range
sharktank/sharktank/utils/vmfb_runner.py:38: IndexError
```

Example logs: https://github.com/nod-ai/shark-ai/actions/workflows/ci_eval_short.yaml?query=branch%3Amain

Rather than assume that self-hosted runners will have multiple GPUs available and having each workflow use a specific device index, we can use the default device and have the runners themselves choose which devices to make visible.

PR #870, opened 2025-01-27
Add Flux transformer benchmarking

sogartar:
Add also some more general functionality to check benchmark results against baseline results. This requires the Google Benchmark compare.py tool that is not a part of the pip package. That is why I added the repo as a git submodule. This tool does a statistical comparison between benchmarks with proper p-value calculation. I don't think we should roll out our own.

Adds a new nightly CI job that should contain nightly tests and benchmarks that are not in their own category like Llama is now.


sogartar:
This PR depends on https://github.com/iree-org/iree-turbine/pull/418.

sogartar:
I am not sure if we should put the benchmarks in this repo or at https://github.com/nod-ai/SHARK-TestSuite.

.github/workflows/ci-sharktank.yml
```diff
@@ -85,7 +85,11 @@ jobs:
       - name: Run sharktank tests
         if: ${{ !cancelled() }}
         run: |
-          pytest -n 4 sharktank/ --durations=10
+          pytest \
+            -n 4 \
+            --durations=10 \
+            -m "not expensive" \
+            sharktank/
```
ScottTodd, line 92: Be careful when adding new tests, particularly if they are expensive. Developers should be able to run `pytest` and expect a reasonable default set of tests to run. Developers will not remember opt-out marks like this.

The filtering also didn't work here? https://github.com/nod-ai/shark-ai/actions/runs/13079961026/job/36501056330?pr=870 is still running on standard github-hosted runners after 2h+.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "third_party/benchmark"]
+	path = third_party/benchmark
+	url = https://github.com/google/benchmark
```
ScottTodd, line 3: Is this actually used? You have `google-benchmark` in Python test requirements already.

A source dependency on the C++ library could be added to shortfin as needed, probably via FetchContent here: https://github.com/nod-ai/shark-ai/blob/4eac34e1455f6518d5813299f712137af6a96573/shortfin/CMakeLists.txt#L326-L340

.github/workflows/ci-sharktank-nightly.yaml
ScottTodd, line 1: I'd much prefer we consolidate existing nightly workflows (eval, sglang benchmark, llama large, etc.) before adding a new one. That being said, I like the "sharktank-nightly" name better than model-specific names...

.github/workflows/ci-sharktank-nightly.yaml
```diff
@@ -0,0 +1,87 @@
...
+          pip freeze
+
+      - name: Run benchmarks
+        run: |
+          source ${VENV_DIR}/bin/activate
+          pytest \
+            --verbose \
+            --capture=no \
+            --iree-hip-target=gfx942 \
+            --iree-device=hip://6 \
```
ScottTodd, line 75: Match what other workflows do with hip devices: https://github.com/nod-ai/shark-ai/pull/725

.github/workflows/ci-sharktank-nightly.yaml
```diff
@@ -0,0 +1,87 @@
...
+            -m="benchmark and expensive" \
+            --html=out/benchmark/index.html \
+            sharktank/tests
+
+      - name: Deploy to GitHub Pages
+        uses: peaceiris/actions-gh-pages@4f9cc6602d3f66b9c108549d475ec49e8ef4d45e # v4.0.0
+        with:
+          github_token: ${{ secrets.SHARK_PLATFORM_GH_TOKEN }}
+          publish_dir: ./out/benchmark
+          destination_dir: ./benchmark
```
ScottTodd, line 86: "benchmark" isn't a very specific name, and it also doesn't match "sharktank nightly". Consider how you want this to slot in to

* https://github.com/nod-ai/shark-ai/tree/gh-pages
* https://nod-ai.github.io/shark-ai/

ScottTodd (changes requested):

.github/workflows/ci-sharktank-nightly.yaml
sogartar: Do we actually allocate GPUs to CI jobs?

.github/workflows/ci-sharktank.yml
sogartar: The test failure is due to this required PR https://github.com/iree-org/iree-turbine/pull/418.

sogartar: I changed the default marker selection in `pyproject.toml` to exclude `expensive`.

.gitmodules
sogartar: I wrongly left the pip package dependency. It is removed now. Unfortunately, the script to compare benchmark results is not a part of the pip package. This script is used in the python test.
See https://github.com/google/benchmark/blob/main/docs/tools.md#comparepy
That is why the submodule is used.

.github/workflows/ci-sharktank-nightly.yaml
sogartar: I think the CI jobs are due for a refactoring. To reorganize them and to reduce code duplication. I want to do that next.

.github/workflows/ci-sharktank-nightly.yaml
sogartar: I change it to used `hip://0`.

ScottTodd: Thanks. Generally, CI jobs should be written to run on any compatible runner machine and be easy to reference and run locally. The runners themselves should be responsible for assigning devices.

.github/workflows/ci-sharktank-nightly.yaml
archana-ramalingam: The advantage of keeping them separately is the ease of tracking various nightly across llm, flux models and sharktank/ shortfin regressions. If there are tools to do this, or a CI summary in github-pages, that would be great.

.github/workflows/ci-sharktank-nightly.yaml
```diff
@@ -0,0 +1,87 @@
...
+        run: |
+          source ${VENV_DIR}/bin/activate
+          pytest \
+            --verbose \
+            --capture=no \
+            --iree-hip-target=gfx942 \
+            --iree-device=hip://0 \
+            --with-flux-data \
+            -m="benchmark" \
+            --html=out/sharktank-nightly/benchmark/index.html \
```
archana-ramalingam, line 78: Updated this to `sharktank_nightly` [here](https://github.com/nod-ai/shark-ai/tree/gh-pages/sharktank_nightly/benchmark). We can change this here as well.

.github/workflows/ci-sharktank-nightly.yaml
```diff
@@ -0,0 +1,87 @@
...
+            sharktank/tests
+
+      - name: Deploy to GitHub Pages
+        uses: peaceiris/actions-gh-pages@4f9cc6602d3f66b9c108549d475ec49e8ef4d45e # v4.0.0
+        with:
+          github_token: ${{ secrets.SHARK_PLATFORM_GH_TOKEN }}
+          publish_dir: ./out/sharktank-nightly
+          destination_dir: ./sharktank-nightly
```
archana-ramalingam, line 86: I believe `publish_dir` should be `./out/sharktank_nightly/benchmark`. Same for `destination_dir` 
Also, we need to update `sharktank-nightly` to `sharktank_nightly`

.github/workflows/ci-sharktank-nightly.yaml
sogartar: I changed it.

.github/workflows/ci-sharktank-nightly.yaml
sogartar: I renamed it.
I wanted to have one step to push all sharktank_nightly artifacts. This workflow in the future may have more jobs that produce artifacts. I would be surprised if the action does not support whole directory trees.

archana-ramalingam: Sure, were you able to test this nightly on this PR?

sogartar: The workflow is not yet available and I can't trigger it manually. I will trigger it once it is on main.

sharktank/sharktank/utils/benchmark.py
ScottTodd, line 1: > I am not sure if we should put the benchmarks in this repo or at https://github.com/nod-ai/SHARK-TestSuite.

@geomin12 has also started refactoring the benchmarks from https://github.com/iree-org/iree/tree/main/experimental/regression_suite/shark-test-suite-models by moving them to https://github.com/iree-org/iree-test-suites (possibly in this subfolder: https://github.com/iree-org/iree-test-suites/tree/main/sharktank_models).

Given the similarity between FLUX and SDXL, we may want some of this infrastructure to move there too. Starting directly next to the code in this nod-ai/shark-ai repository probably makes sense in the near term, then as model support matures we can move some code around.

I'm mainly concerned about there being a proliferation of different test and benchmark implementations.
* There should be one well documented way to export, compile, run, and benchmark each model.
* There should be one central location to view the latest test and benchmark results across all projects.
* Any developer or user should be able to reproduce test or benchmark results on their own hardware.

I think the way we get there is to streamline our export/compile/run/etc. flows and keep our release process running smoothly.

---

Here are a few parts of vLLM's benchmark suite to reference:
* https://github.com/vllm-project/vllm/tree/main/.buildkite/nightly-benchmarks
* https://github.com/vllm-project/vllm/blob/main/.buildkite/nightly-benchmarks/tests/latency-tests.json
* https://github.com/vllm-project/vllm/blob/main/.buildkite/run-benchmarks.sh

Note that test cases only specify a model name in the standard huggingface repository naming style then the project knows how to serve that model. We should be able to do the same for SDXL, Flux, Llama, etc.

If we build a serving-focused benchmark suite like vLLM's focused on datacenter GPUs, I could see it living in nod-ai/shark-ai or nod-ai/SHARK-TestSuite. If we want to support other backends or benchmark finer grained details in the IREE compiler and runtime then maybe iree-org/iree-test-suites would be better 🤔.

sharktank/sharktank/utils/benchmark.py
```diff
@@ -0,0 +1,118 @@
...
+    )
+
+    out = process_result.stdout.decode()
+    err = process_result.stderr.decode()
+
+    if process_result.returncode != 0:
+        raise RuntimeError(f"stderr:\n{err}\nstdout:\n{out}")
+
+    if err != "":
+        print(err, file=sys.stderr)
```
ScottTodd, line 34: We have other test suites that log the compile/run/benchmark commands before issuing them. I don't see that done here, but maybe it should be. Otherwise, these error messages would be missing some context.

You could also use `logging` instead of `print()` and use debug/info/warning/error levels if you expect that logging to be noisy enough that a developer might want to silence some parts of it.

sharktank/tests/models/flux/flux_test.py
```diff
@@ -17,12 +17,14 @@
     export_flux_transformer_from_hugging_face,
     export_flux_transformer,
     import_flux_transformer_dataset_from_hugging_face,
+    iree_compile_flags,
```
ScottTodd, line 20: Keep the compile flags in this test file, or maybe add a `compile_model.py` alongside `export.py` so there is a standard way of compiling the model that the test can call into.

See also my other comment about the tools requiring less effort to use, along with issues like https://github.com/nod-ai/shark-ai/issues/691. Generally speaking, tests should be simple: https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html. If the test isn't simple, the system being tested should be changed first. The software we ship (sharktank / shortfin / shark-ai) should know how to compile Flux. These flags in a test show that the software still needs help with that.

sharktank/sharktank/utils/testing.py
```diff
@@ -21,6 +22,8 @@
 from ..types import *
 from .math import cosine_similarity
 
+is_mi300x = pytest.mark.skipif("config.getoption('iree_hip_target') != 'gfx942'")
```
ScottTodd, line 25: Moving this into the utils file is nice. Could pull that out into its own PR if you want it merged sooner.

Could add similar marks for other cases too.

ScottTodd (changes requested):

sharktank/sharktank/utils/benchmark.py
ScottTodd: To be clear, I am quite happy to see more benchmarks being worked on and added. I'm just worried about there being 7 different test suites, benchmark suites, etc. all working in slightly different ways. Each one has its own advantages and disadvantages so I think we should learn what we can from each and then work to consolidate somewhere. Just need to make those decisions and stand by them.

geomin12: let me take a look at this code and see if there are any ways to combine (from a quick skim, seems almost identical where it's a compile -> benchmark -> compare results)! I'll start writing up my benchmark generic tests then find a simple way to get flux on there (this can be a test to make sure the benchmark code is generic enough to onboard other models!)

sogartar: @geomin12, do you suggest we migrate the generic parts of this PR to [iree-test-suites](https://github.com/iree-org/iree-test-suites)?
The important element here is the use of Google benchmark [compare.py](https://github.com/google/benchmark/blob/main/docs/tools.md#comparepy) to do a statistically sound comparison of benchmark results. I think just comparing mean and median is not good enough. Doing this sort of comparison requires keeping the complete baseline benchmark with all repetition results, which may be a lot as you need 20-30 runs to reduce variance.

sharktank/sharktank/models/flux/export.py
```diff
@@ -17,6 +17,26 @@
 
 flux_transformer_default_batch_sizes = [1]
 
+iree_compile_flags = [
+    "--iree-hal-target-device=hip",
+    "--iree-hip-target=gfx942",
```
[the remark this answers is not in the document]
sogartar: I moved the args to a new `compile.py` file. I agree that refactoring the compilation process is needed, but I think it is outside the scope of this PR. The model is currently being tuned and these flags are a first best guess as to what would be good.

sharktank/tests/models/flux/flux_benchmark_test.py
```diff
@@ -0,0 +1,66 @@
...
+
+        benchmark_compare_result = iree_benchmark_compare(
+            [
+                f"--dump_to_json={benchmark_compare_result_file_path}",
+                "benchmarks",
+                str(baseline_benchmark_result_file_path),
+                benchmark_result_file_path,
+            ]
+        )
+        print(benchmark_compare_result)
```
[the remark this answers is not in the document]
sogartar: I added an example output to the PR description.
I think `compare.py` is essential to do a proper comparison to determine if 2 benchmarks come from the same distribution.

sharktank/sharktank/utils/benchmark.py
sogartar: I improved logging and the exception message.

sharktank/tests/models/flux/flux_test.py
sogartar: I agree that refactoring is required, but I would like to not include it in this test.

sharktank/sharktank/utils/benchmark.py
geomin12: yes! I've been working on [this in the iree-test-suites](https://github.com/geomin12/iree-test-suites/tree/main/sharktank_models/test_suite/benchmarks), so it will be a lot simplier to onboard a model and start doing threshold tests. 

I've got some work to do to include hugging face and pulling models from there, and I can also incorporate the more sophisticated `compare.py` tool. Right now, in that repo, we just use golden time/dispatch/size

I think for now, this can stay here (so it doesn't block this PR), then I can work on migrating the flux benchmarks over to the iree-test-suite

sogartar: @geomin12, roughly when do you expect to be able to port this benchmarking?
@ScottTodd, should we merge this in the mean time?

geomin12: will be done next week!

sogartar: @geomin12, is there any progress on migrating this benchmark?

geomin12: plan on doing the official iree migration today!

sogartar: One thing that is important for the benchmark repo is that it regularly and automatically bumps up the sharktank version. Probably a daily bump. We want changes in the model to quickly make their way to the benchmark results.
This means that [iree-test-suites](https://github.com/iree-org/iree-test-suites) would have to take sharktank as a dependency.

PR #871, opened 2025-01-27
Make Flux Transformer RoPE use a custom IREE kernel

sogartar:
We assume that the custom kernel would yield better performance instead of using PyTorch ops.

### DIFF
--- a/sharktank/sharktank/layers/rotary_embedding.py
+++ b/sharktank/sharktank/layers/rotary_embedding.py
@@ -220,12 +220,9 @@ class RotaryEmbeddingLayer(BaseLayer):
         return xt_out.type_as(xt)
 
     def _compute_rotary_embed_table(self, t):
-        dim = self.rope_dimension_count
-        freqs = 1.0 / (
-            self.rope_freq_base ** ((torch.arange(0, dim) // 2).float() / dim * 2.0)
+        return compute_rotary_embedding_table(
+            t, self.rope_dimension_count, self.rope_freq_base, torch.float32
         )
-        freqs = torch.outer(t, freqs).float()
-        return freqs
 
     def _create_rotary_embed_table(self):
         t = torch.arange(self.max_seqlen, device=self.device)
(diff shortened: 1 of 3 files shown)


sogartar:
To match the required dimension order, axis permutation is performed.
I am not sure if this can be optimized out. Needs some investigation of
the compilation passes.

This change introduces a significant (10X) deterioration in performance.
Baseline performance is 552 ms.
With this change it is 6131 ms.

sogartar:
The large performance problem has been addressed by https://github.com/iree-org/iree/issues/19822.

renxida (approved):

sharktank/sharktank/layers/rotary_embedding.py
```diff
@@ -238,3 +235,17 @@ def _replicate(self, t):
             t = ops.replicate(t, self.tensor_parallelism_size)
 
         return t
+
+
+def compute_rotary_embedding_table(
+    positions: torch.Tensor,
```
renxida, line 241: Nit: would it make more sense to just rename _compute_rotary_embedding_table?

sogartar: I want to use this function outside of  the class.

renxida: Yes, i mean, instead of:

- copying compute_rotary_embedding_table so we can use it outside the class
- make old _compute_rotary_embedding_table redirect to compute_rotary_embedding_table

Just

- rename _compute_rotary_embedding_table to compute_rotary_embedding_table and use it outside the class
- change all referenes to _compute_rotary_embedding_table to use compute_rotary_embedding_table instead

The latter requires an IDE and is slightly more work, but does not leave a stub/redirect function behind.

sogartar:
This PR is waiting on https://github.com/iree-org/iree/pull/19829.

sharktank/sharktank/layers/rotary_embedding.py
sogartar: The class variant got updated to handle the hugging face case so it has more hair now.

PR #873, opened 2025-01-28
Fix Masking-related token corruption issue that oddly only manifests when trie is enabled

renxida:
Earlier, I changed the generation requests to store the whole token history instead of just the most recent token.

Missed a spot, which caused seq_len to be a lot longer than it should be, which caused the kvcache read to be wrong.


renxida:
(the problem is fixed and the CI is only failing because of a different issue fixed in #874 )

stbaione (approved):
Let's goooooooo! Failure output makes sense now, if the start_positions offsets were getting screwy.

Remember to add back the xfail before merging

renxida:
> Let's goooooooo! Failure output makes sense now, if the start_positions offsets were getting screwy.

Yup! Except! It's actually the seq_lens that were the problem - - it's used to mask out un-used kv cache slots in the current page!

PR #875, opened 2025-01-28
Simplify pkgci_shark_ai workflow using wildcard downloads

### LINKED ISSUE #804: Simplify Package Download in `pkgci_shark_ai.yml`
Comments from Scott on #780 

I'd like to simplify these steps as we add more `pkgci_*.yml` workflows, but this is fine for now.

For the downloads, actions/download-artifact supports downloading multiple artifacts using some pattern: https://github.com/actions/download-artifact?tab=readme-ov-file#download-multiple-filtered-artifacts-to-the-same-directory . Could try with

```yaml
pattern: snapshot-*-cp${{ env.PY_VERSION_NO_DOT }}-*
```

Note that sharktank and shark-ai aren't dependent on Python versions, and audit_wheel actually removes the `cp311` from the names: 
image

So if we keep the versions in the names, we can have a single download step that filters. If we remove the versions, we could either download all artifacts (no filtering at all) or two download steps, one that filters for `cp311` and one that  filters for `py3-none`.

For the `PY_VERSION_NO_DOT`, I don't see a way to get actions/setup-python to use the `cp311` syntax: https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md.

The downloads are fast (less than 10MB total, taking seconds or less), so I'd probably just download all artifacts then make sure `setup_venv.py` respects the selected Python version and only tries to install compatible packages from the download.

renxida:
closes #804

### DIFF
--- a/.github/workflows/pkgci_shark_ai.yml
+++ b/.github/workflows/pkgci_shark_ai.yml
@@ -44,10 +44,6 @@ jobs:
         with:
           python-version: ${{matrix.version}}
 
-      - name: Set Python version without dot
-        run: |
-          echo "PY_VERSION_NO_DOT=$(echo ${{ matrix.version }} | tr -d '.')" >> $GITHUB_ENV
-
       - name: Setup UV caching
         run: |
           CACHE_DIR="${GITHUB_WORKSPACE}/.uv-cache"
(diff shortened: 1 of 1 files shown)


ScottTodd:
Looks like the workflow code did the right thing? Can you check if the test failures are unrelated?
https://github.com/nod-ai/shark-ai/actions/runs/13021274920/job/36322354352?pr=875

renxida:
Ack it's cause I forgot to click merge on #874. Now it should be all good.

ScottTodd (approved):
LGTM assuming the workflow passes. Thanks!

renxida:
pkgci passed!

PR #882, opened 2025-01-29
run cpu llm tests on cpu server

renxida:


renxida:
This is part of #883

renxida:
(the pending / cancelled sharktank tests aren't relevant for this PR)

PR #889, opened 2025-01-30
Make IREE Bump PRs Tag Rotation Engineers

renxida:
Add assignees to `update_iree_requirements_pins.yml`.

Right now it's just me on the rotation schedule but it should keep working when we put more people in.


renxida:
Tested on fork; it works!

https://github.com/renxida/shark-ai/pull/2

.github/workflows/update_iree_requirement_pins.yml
```diff
@@ -25,6 +25,25 @@ env:
   GIT_BRANCH_NAME: ${{ inputs.branch-name || 'integrates/iree' }}
 
 jobs:
+  determine-duty-engineer:
+    runs-on: ubuntu-latest
```
ScottTodd, line 29: Use the same pinned runner as in the other workflows. Then we can more easily do a find-replace across the entire repository to update to newer versions and aren't at the whim of whatever GitHub changes "latest" to mean.
```suggestion
    runs-on: ubuntu-24.04
```

Actually now I'm noticing that I set both 22.04 and 24.04 in this file. Should be able to switch all to 24.04.

ScottTodd (approved):
Cool! There are other ways to build a rotation schedule, like having a separate text file or a spreadsheet with lines for each date range, but this looks fine. Have to start somewhere, and there is only one member right now :D

We can also create a github team if we want an entire group of people to get notified.

.github/workflows/update_iree_requirement_pins.yml
renxida: Done!

renxida:
Reading from a spreadsheet would be super nice!

Let's do that when more people sign on!

PR #890, opened 2025-01-30
move list of models to model_management.py

renxida:


app_tests/integration_tests/llm/device_settings.py
```diff
@@ -15,3 +15,33 @@ class DeviceSettings:
...
+        "--iree-hip-target=gfx942",
+    ),
+    server_flags=("--device=hip",),
+)
+
+
+def get_device_based_on_env_variable():
+    import os
+
+    device_name = os.environ.get("SHORTFIN_INTEGRATION_TEST_DEVICE", "cpu").lower()
```
renxida, line 31: I wonder if there's a better name for this env var / a better way to pass the device selection through.

app_tests/integration_tests/llm/model_management.py
```diff
@@ -275,4 +290,37 @@ def process_model(self, config: ModelConfig) -> ModelArtifacts:
...
+
+
+TEST_MODELS = {
+    "open_llama_3b": ModelConfig(
+        source=ModelSource.HUGGINGFACE,
+        repo_id="SlyEcho/open_llama_3b_v2_gguf",
+        model_file="open-llama-3b-v2-f16.gguf",
+        tokenizer_id="openlm-research/open_llama_3b_v2",
+        batch_sizes=(1, 4),
+        device_settings=device_settings.get_device_based_on_env_variable(),
```
renxida, line 304: Not sure if this is the best way to set up the device, but doing this because ModelConfig can be used as a cache key for the stored model artifacts, and the device settings matter for caching compiled artifacts.

renxida:
Currently the new tests show as

[Test shark-ai / Integration Tests - Shortfin LLM Server (3.11, gfx942)](https://github.com/nod-ai/shark-ai/actions/runs/13063895371/job/36452867772?pr=890#logs)

in the CI tasks.

Might be a good idea to name them shorter so that the cpu/gfx942 matrix becomes easier to read.

renxida:
The CPU version takes ~30min to run while the GPU ver only takes ~10 min.

While CPU is slower, I'm pretty sure this also has to do with caching.

I'm currently sharing sharktank's `hf_datasets.py` to download hf models. So if we fix that, it would enable caching almost everywhere.

app_tests/integration_tests/llm/device_settings.py
renxida: Also, maybe don't default to CPU and throw an error instead, to force explicit device specification.

.github/workflows/pkgci_shark_ai.yml
```diff
@@ -25,10 +25,9 @@ jobs:
     strategy:
       matrix:
         version: [3.11]
+        device: ["cpu", "gfx942"]
       fail-fast: false
-    runs-on: azure-cpubuilder-linux-scale
-    # runs-on: ubuntu-latest # everything else works but this throws an "out of resources" during model loading
-    # TODO: make a copy of this that runs on standard runners with tiny llama instead of a 8b model
+    runs-on: ${{ matrix.device == 'cpu' && 'azure-cpubuilder-linux-scale' || 'mi300x-3' }}
```
stbaione, line 30: Is there another system we can use instead of `mi300x-3`? That system gets so much activity. Might end up causing some runs to fail from other usage/tests

app_tests/integration_tests/llm/device_settings.py
stbaione: I think a cli arg `--test_device` would server better than an env var here. Also makes sense to not default and force explicit specification

stbaione:
> The CPU version takes ~30min to run while the GPU ver only takes ~10 min.
> 
> While CPU is slower, I'm pretty sure this also has to do with caching.
> 
> I'm currently sharing sharktank's `hf_datasets.py` to download hf models. So if we fix that, it would enable caching almost everywhere.

That's pretty significant. The last run on main took 6 mins: https://github.com/nod-ai/shark-ai/actions/runs/13077565296/job/36493491268

renxida:
> > The CPU version takes ~30min to run while the GPU ver only takes ~10 min.
> > While CPU is slower, I'm pretty sure this also has to do with caching.
> > I'm currently sharing sharktank's `hf_datasets.py` to download hf models. So if we fix that, it would enable caching almost everywhere.
> 
> That's pretty significant. The last run on main took 6 mins: https://github.com/nod-ai/shark-ai/actions/runs/13077565296/job/36493491268

The speed difference has two main causes:
1. Model downloads (~6 min) aren't being cached, artificially inflating CPU runtime
2. Switched from Ollama (3B) to Meta Llama (8B) which mildly increases runtime due to larger size

The previous 30 min CPU runtime was likely due to cluster load, since recent CPU runs take ~12 min before caching.

I think the best way to do this is probably:

- keep CPU cluster run on CPU cluster
- move GPU run ~maybe to MI250
- add a much smaller much faster model for quick ci verification (e.g. [this very small version of llama3](https://huggingface.co/Amu/supertiny-llama3-0.25B-v0.1/discussions))

.github/workflows/pkgci_shark_ai.yml
renxida: Thinking about our mi250x machines. I don't remember anybody mentioning them for a while now, but I'm not sure if they're just free, or being actively used in the background in one of the other repos.

app_tests/integration_tests/llm/device_settings.py
renxida: Awesome. Implemented both suggestions!

app_tests/integration_tests/llm/model_management.py
renxida: Using a get_device_settings_by_name function instead.

.github/workflows/pkgci_shark_ai.yml
renxida: Ty for suggesting smc300x-ccs-aus-GPUA30F.

Now using tag mi300 for gfx942.

renxida:
Not caching. Working on that.

.github/workflows/pkgci_shark_ai.yml
```diff
@@ -25,17 +25,25 @@ jobs:
...
       fail-fast: false
-    runs-on: azure-cpubuilder-linux-scale
-    # runs-on: ubuntu-latest # everything else works but this throws an "out of resources" during model loading
-    # TODO: make a copy of this that runs on standard runners with tiny llama instead of a 8b model
+    runs-on: >-
+      ${{
+        matrix.test_device == 'cpu' && 'azure-cpubuilder-linux-scale' ||
+        matrix.test_device == 'gfx942' && 'linux-mi300-gpu-1' ||
+        matrix.test_device == 'gfx90a' && 'nodai-amdgpu-mi250-x86-64'
+      }}
```
ScottTodd, line 35: How about an explicit matrix instead of putting this logic in?

For example: https://github.com/iree-org/iree/blob/5f7b471f88e9c8e145637bc1d2f7a4c1d8b4462e/.github/workflows/pkgci_regression_test.yml#L29-L56

```yml
strategy:
  matrix:
    - test_device: cpu
      runs-on: azure-cpubuilder-linux-scale
    - test_device: gfx90a
      runs-on: nodai-amdgpu-mi250-x86-64
```

Docs: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow

.github/workflows/pkgci_shark_ai.yml
```diff
@@ -25,17 +25,25 @@ jobs:
...
       run:
         shell: bash
     env:
       PACKAGE_DOWNLOAD_DIR: ${{ github.workspace }}/.packages
       VENV_DIR: ${{ github.workspace }}/.venv
     steps:
+      - name: Run rocminfo
+        continue-on-error: true
+        run: |
+          rocminfo || echo "rocminfo not available"
```
ScottTodd, line 46: Only run on GPU runners instead of using `continue-on-error`?

```suggestion
      - name: Run rocminfo
        if: contains(matrix.test_device, 'gfx')
        run: rocminfo
```

.github/workflows/pkgci_shark_ai.yml
renxida: thanks! that looks MUCH cleaner!

stbaione (approved):
LGTM once CI passes. Seems like the timeout from the last run was a fluke

stbaione:
Another good future improvement we could make on this: Build all model artifacts at startup in parallel

renxida:
>  [Test shark-ai / Integration Tests - Shortfin LLM Server (cpu, azure-cpubuilder-linux-scale, cpu, 3.11)](https://github.com/nod-ai/shark-ai/actions/runs/13188367368/job/36815908091?pr=890#logs)
succeeded 1 hour ago in 5m 46s

Issue #793, opened 2025-01-08
Migrate workflows to OSSCI


yamiyysu:
Currently we have workflows that are running on our lab and we want to migrate them to our OSSCI clusters.

Two types of workflows

1. w/o storage affinity: a [PR](https://github.com/iree-org/iree/pull/19517) is in progress to migrate to Conductor
2. w/ storage affinity: this needs a solution to move over the storage before we can migrate the workflows and we need to decide where to migrate to.

ScottTodd:
Repository | Workflow | Migrated? | Special considerations
----------- | ----------- | ----------- | -------------------------
iree-org/iree | [`pkgci_test_amd_mi300.yml`](https://github.com/iree-org/iree/blob/main/.github/workflows/pkgci_test_amd_mi300.yml) | ✔️ Yes, https://github.com/iree-org/iree/pull/19517 | <ul><li>Builds the IREE runtime from source, needs clang+cmake+ninja</li></ul>
iree-org/iree | [`pkgci_test_amd_mi250.yml`](https://github.com/iree-org/iree/blob/main/.github/workflows/pkgci_test_amd_mi250.yml) | ❌ No | <ul><li>Needs an MI250 accelerator</li><li>Builds the IREE runtime from source, needs clang+cmake+ninja</li></ul>
iree-org/iree | [`pkgci_regression_test.yml`](https://github.com/iree-org/iree/blob/main/.github/workflows/pkgci_regression_test.yml) | ❌ No | <ul><li>Jobs for CPU, MI250, and MI300</li><li>Local cache of SDXL and SD3 model weights ([sample sources here](https://github.com/iree-org/iree/blob/main/experimental/regression_suite/shark-test-suite-models/sdxl/test_clip.py)), around 20GB</li></ul>

yamiyysu:
[ci-tk.yaml](https://github.com/iree-org/iree-turbine/blob/main/.github/workflows/ci-tk.yaml) has local cache reference as seen from the [PR build](https://github.com/iree-org/iree/pull/19738)

renxida:
I'd like [pkgci_shark_ai.yml](https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/pkgci_shark_ai.yml) to be on the new cluster soon, if possible.

This CI file performs llama integration tests, and has been failing / hanging for a while.

Here's why it's important:
[...]

PR #891, opened 2025-01-31
Make get_iree_devices read IREE_DEVICE env var if provided

sogartar:
This allows to inject what exact IREE device(s) are to be used without propagating all the way to program arguments.
Example:
```
IREE_DEVICE=hip://5,hip://3 pytest sharktank/tests
```

### DIFF
--- a/sharktank/sharktank/utils/iree.py
+++ b/sharktank/sharktank/utils/iree.py
@@ -8,6 +8,7 @@ import iree.runtime
 from typing import List, Tuple, Optional, Union
 from pathlib import Path
 import torch
+import os
 import numpy as np
 import collections.abc
 from collections import OrderedDict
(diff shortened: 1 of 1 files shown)


archana-ramalingam (commented):
If this change is specifically targeting CIs, our runners pick whichever gpu is available atm. We had a [PR ](https://github.com/nod-ai/shark-ai/pull/725) making this change for llama benchmark/ppl CIs. @ScottTodd  can confirm.

ScottTodd:
> It is not just for CIs, but when running locally as well. There should be some way of overriding the device list. Are CI machines actually allocating devices to different jobs? Where is this code hosted?

I believe on servers with multiple GPUs, we have one runner per GPU, using `ROCR_VISIBLE_DEVICES=1` or equivalent on runner 1, `ROCR_VISIBLE_DEVICES=2` on runner 2, etc. We also have a cluster of runners with labels like `linux-mi300-gpu-1` getting set up as part of https://github.com/nod-ai/shark-ai/issues/793. I think we'd create a `linux-mi300-gpu-2` label for workflows that request two GPUs?

That's my understanding anyways. @yamiyysu and @saienduri would know more, and we should write that all down somewhere and get the code used to configure the runners checked in too.

sharktank/sharktank/utils/iree.py
```diff
@@ -22,7 +23,28 @@
 from .tree import Tree
 
 
-def get_iree_devices(driver: str, device_count: int) -> List[iree.runtime.HalDevice]:
+def get_iree_devices(
+    *, driver: str | None = None, device_count: int = 1
+) -> List[iree.runtime.HalDevice]:
```
ScottTodd, line 28: Can you add a docstring to this function and mention the `IREE_DEVICE` environment variable with a few examples? You can lift some text from the PR description

> This allows to inject what exact IREE device(s) are to be used without propagating all the way to program arguments. Example:
> 
> ```
> IREE_DEVICE=hip://5,hip://3 pytest sharktank/tests
> ```

````python
def get_iree_devices(
    *, driver: str | None = None, device_count: int = 1
) -> List[iree.runtime.HalDevice]:
    """Gets a list of IREE HAL devices for the given driver.
    
    The first available device_count devices will be created,
    unless the IREE_DEVICE environment variable is set to an
    explicit list of device URIs.

    For example, to select HIP devices 5 and 3:
    ```
    export IREE_DEVICE=hip://5,hip://3
    python ... device=hip # (something here)
    ```
    """
````

(could wordsmith that a bit, just getting the general idea across)

File: sharktank/sharktank/utils/iree.py [hunk 2]
```diff
@@ -22,7 +23,28 @@
 from .tree import Tree
 
 
-def get_iree_devices(driver: str, device_count: int) -> List[iree.runtime.HalDevice]:
+def get_iree_devices(
```
ScottTodd, line 26: May also want to put pytest details in https://github.com/nod-ai/shark-ai/blob/main/docs/developer_guide.md#running-tests

sogartar: I would, but unfortunately our code is fragmented and not all tests use this function to create the devices.
I would need to refactor them first to use this function.

sharktank/sharktank/utils/iree.py
sogartar: Done.

ScottTodd (approved):

PR #892, opened 2025-01-31
[Shortfin] SD server indexing fix

IanNod:
Fixes a bug in the sd server that was not connecting the appropriate fibers to workers resulting in unbalanced workers threads


jinchen62 (approved):
LGTM, thanks!

PR #894, opened 2025-01-31
Fix llama 8b fp8 benchmarking test

aviator19941:
Benchmarking test for fp8 was failing the Export phase, but was expected to fail for the Benchmarking phase and was getting killed.

### DIFF
--- a/sharktank/tests/models/llama/benchmark_amdgpu_test.py
+++ b/sharktank/tests/models/llama/benchmark_amdgpu_test.py
@@ -66,10 +66,10 @@ class BenchmarkLlama3_1_8B(BaseBenchmarkTest):
     def setUp(self):
         super().setUp()
         # TODO: add numpy files to Azure and download from it
-        self.artifacts_dir = Path("/shark-dev/data/llama3.1/weights/8b")
-        self.artifacts_dir_2048 = Path("/shark-dev/8b")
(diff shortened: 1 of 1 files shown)


sharktank/tests/models/llama/benchmark_amdgpu_test.py
```diff
@@ -66,10 +66,12 @@ class BenchmarkLlama3_1_8B(BaseBenchmarkTest):
     def setUp(self):
         super().setUp()
         # TODO: add numpy files to Azure and download from it
-        self.artifacts_dir = Path("/shark-dev/data/llama3.1/weights/8b")
-        self.artifacts_dir_2048 = Path("/shark-dev/8b")
-        self.irpa_path = self.artifacts_dir / "fp16/llama3.1_8b_fp16.irpa"
-        self.irpa_path_fp8 = self.artifacts_dir / "f8/llama3.1_8b_fp8.irpa"
+        self.artifacts_dir = Path("/shark-dev/8b")
```
archana-ramalingam, line 69: Do we have a separate PR to update the 70B/405B instruct paths as well?

archana-ramalingam (approved):

sharktank/tests/models/llama/benchmark_amdgpu_test.py
aviator19941: yeah will be a separate PR

PR #902, opened 2025-02-03
Add real page pool tests for trie_attention_cache

renxida:
Previously, we were testing with mocked page pools so the tests run faster. In this PR, I split trie_attention_cache_tests.py into 2 files:

trie_attention_cache/mock_pool_tests.py contains the old tests, and we continue to test with a mocked-up page pool to verify that the trie correctly does accounting for the pages and the evictions.

trie_attention_cache/real_pool_tests.py will contain new tests for page-copying prefix sharing, so that we won't have to recompute the entire last page's worth of KV if branching on a token. Since we're copying the page, the tests will need to not mock the page pool and actually allocate the buffer, which will make them slower. I opted to do this separately from the old tests so that we won't have to take 5-ish seconds to set up the buffer for each of the 30 ish tests.

This PR also replaces some of the nuisance print statements with logging.debug.

~~This is a step on the way to implement beam search (required by MLPerf).~~

Edit: [MLPerf only requires beam search for GPT-J](https://github.com/mlcommons/inference_policies/blob/master/inference_rules.adoc#:~:text=Q%3A%20What%20algorithm,uses%20greedy%20search.). Thanks @stbaione


shortfin/tests/apps/llm/components/kvcache/trie_attention_cache/real_pool_tests.py
```diff
@@ -0,0 +1,101 @@
...
+    def _publish_sequence(tokens: List[int]) -> None:
+        alloc = trie_cache.acquire_pages_for_tokens(tokens, extra_token_slots=0)
+        alloc.publish_pages_for_tokens(alloc.tokens)
+        alloc.release_pages()
+
+    return _publish_sequence
+
+
+@pytest.mark.xfail(reason="Partial page reuse is not yet implemented.", strict=True)
+def test_partial_page_publishing(trie_cache):
```
stbaione, line 81: Should we add a simple non-xfailed case, like `test_single_page_publishing` just to fill the tests out a little more and have something that could catch a regression?

renxida: My reasoning was that partial page publishing would be the only reason why the underlying buffer (page table & page pool) might matter for testing, hence this being the first test case.

But now that you mention it I think I should probably put in a test case for writing some data to a single page and reading back to make sure that it didn't change.

I think it'd be best if I put it in separately in a different smaller pr though.

stbaione (approved):

PR #905, opened 2025-02-03
Shard input_mask for Llama

stbaione:
Currently, 405b OOMs when using long input prompts (issue [here](https://github.com/iree-org/iree/issues/19832)).

This PR implements a suggested fix for this, which is to shard the `input_mask`. This makes sense, since the OOM error is dependent on the length of the input.

I'm new to sharktank and still seeing the issue with current implementation, so wanted to have it double-checked to make sure it's properly implemented.

MLIR [here](https://sharkpublic.blob.core.windows.net/sharkpublic/stephen/llama3.1_405b_tp8/mlir/iree_19832/llama_405b.mlir)

### DIFF
--- a/sharktank/sharktank/models/llama/llama.py
+++ b/sharktank/sharktank/models/llama/llama.py
@@ -129,6 +129,8 @@ class PagedLlamaModelV1(BaseCausalLMModel):
         # [bs, batch_seq_len]
         tokens: Union[torch.Tensor, ReplicatedTensor],
         *,
+        # [bs, batch_seq_len]
+        input_mask: Union[torch.Tensor, ReplicatedTensor],
         # [1, 1, batch_seq_len, batch_seq_len]
         attention_mask: Optional[Union[torch.Tensor, ReplicatedTensor]],
         # [bs, batch_seq_len // block_seq_stride]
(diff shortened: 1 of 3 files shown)


sogartar (commented):
One thing that seems unaddressed is the [scaled_dot_product_attention](https://github.com/nod-ai/shark-ai/blob/c10a5de6143e0b7f75536bbe865aee5392c8b50c/sharktank/sharktank/layers/paged_llama_attention_block.py#L205-L206). It also applies the mask. How would that be reconciled? If `is_causal=False` does it not apply any mask?

sharktank/sharktank/models/llama/llama.py
```diff
@@ -129,20 +129,25 @@ def prefill(
...
         self._assert_device(tokens)
+        self._assert_device(input_mask)
         self._assert_device(attention_mask, dtype=self.activation_dtype)
         self._assert_device(seq_block_ids)
         self._assert_device(*cache_state, dtype=self.activation_dtype)
 
         h = self.token_embedding(tokens)
         self.trace_tensor("llama.token_embedding", h)
 
+        h *= input_mask.unsqueeze(-1)
```
sogartar, line 149: I don't think that if we assume this should somehow substitute the attention mask the math checks out.

stbaione:
Stale

PR #914, opened 2025-02-05
Add a bf16 datatype to sfnp

KyleHerndon:
I don't fully understand the surrounding code, but I made this thin wrapper and it seems to mostly work. Let me know if you see any glaring issues, because I'm debugging Flux BF16 on top of this and can't pinpoint the issue.

It probably makes sense to move the type definition to another file, I was just hacking on it. Trying to figure out where things belong too.

(For reference I tried working with the __bf16 clang type, but it didn't have enough support so I decided to make this from scratch/uint16 instead)


KyleHerndon:
We determined the numerical issue is elsewhere. This is ready for review.

monorimet (approved):
LGTM. Can we add the case to the sfnp op/dtype tests where applicable?

PR #915, opened 2025-02-05
Llama support table + sharding docs

stbaione:
This PR adds a support table to our `llama_serving.md` guide, specifying our supported variants of `llama3.1-8b, llama3.1-70b, llama3.1-405b`. 

I used [vLLM Supported Models](https://docs.vllm.ai/en/v0.4.2/models/supported_models.html#supported-models) as a reference. I changed the structure a bit to make more sense for our server, but it's relatively similar to how they set their table up.

For sharding instructions, I added a section at the end of our doc. I debated between creating a separate `md` file for it, but I think it actually makes sense where it is. It gives our doc a flow where we give detailed descriptions of what's going on while the user gets setup with the lowest-barrier model. Following that, you get to the more advanced sharding section. The details aren't as specific and it's not an exact copy + paste flow like above. The assumption is that after reading above section, the user shouldn't have to be hand-held as much in this section.

Currently, I know we can shard 405b with an upper-bound of `tp8`. I need to run some tests to see what the supported lower-bound is.

This also further highlights that we need to streamline our export/compile process. We should allow user to specify just the huggingface repo when starting the server, while we take care of downloading safetensors, exporting, and compiling. We should do this while still allowing for specific local files to be specified: #402, #691

### DIFF
--- a/docs/shortfin/llm/user/llama_serving.md
+++ b/docs/shortfin/llm/user/llama_serving.md
@@ -390,9 +390,6 @@ iree-compile /path/to/output/llama3.1-405b.mlir \
 
 ### Run the server
 
-Now that we have compiled our sharded model, we can run the server with the 
-`--parameters` flag to specify the path to our sharded model files:
-
 > [!NOTE]
 > For running a sharded model, we must specify each irpa file in `--parameters`,
 > and the number of devices in `--device_ids` should be equal to the 
--- a/docs/shortfin/llm/user/llama_serving.md
+++ b/docs/shortfin/llm/user/llama_serving.md
@@ -302,9 +302,21 @@ ps -f | grep shortfin
 
 <!-- TODO(#402): Streamline the way that models are sharded/exported/compiled for server. -->
 
+Sharding, in the context of LLMs, refers to splitting the model’s parameters
+across multiple machines or GPUs so that each device only handles a portion of
+the overall weight matrix. This technique allows large models to fit into
+memory and be trained or inferred upon more efficiently by distributing the
+computational load.
+
+For a more detailed explanation of sharding and different sharding + optimization
+techniques, see [Efficient Training on Multiple GPUs](https://huggingface.co/docs/transformers/v4.48.2/en/perf_train_gpu_many).
+
 For models that require sharding, like [Llama-3.1-405b](#supported-models), we
 will use the [`sharktank.examples.sharding.shard_llm_dataset`](https://github.com/nod-ai/shark-ai/blob/main/sharktank/sharktank/examples/sharding/shard_llm_dataset.py)
 script, which exports our model as sharded `irpa` files.
+
+Specifically, we use the [Tensor Parallelism](https://huggingface.co/docs/transformers/v4.48.2/en/perf_train_gpu_many#tensor-parallelism)
+technique in `sharktank`.
 
 > [!NOTE]
 > The `--tensor-parallelism-size` argument specifies the number of shards to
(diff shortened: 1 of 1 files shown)


docs/shortfin/llm/user/llama_serving.md
```diff
@@ -283,7 +298,124 @@ If you want to find the process again:
 ps -f | grep shortfin
 ```
 
-## Server Options
+## 5. Running with sharded models
+
+<!-- TODO(#402): Streamline the way that models are sharded/exported/compiled for server. -->
+
+For models that require sharding, like [Llama-3.1-405b](#supported-models), we
```
ScottTodd, line 305: Should we explain briefly what "sharding" means in this context, or do you expect enough users will understand without extra help?

ScottTodd (approved):
Thanks! This looks good to merge with maybe a few edits. Agreed about the need for some tools that help put together the command sequences.

docs/shortfin/llm/user/llama_serving.md
```diff
@@ -283,7 +298,124 @@ If you want to find the process again:
...
+  --iree-hal-target-device=hip[4] \
+  --iree-hal-target-device=hip[5] \
+  --iree-hal-target-device=hip[6] \
+  --iree-hal-target-device=hip[7] \
+  --iree-hip-target=gfx942 \
+  --iree-dispatch-creation-enable-aggressive-fusion=true \
+  --iree-global-opt-propagate-transposes=true \
+  --iree-opt-aggressively-propagate-transposes=true \
+  --iree-opt-data-tiling=false \
+  --iree-preprocessing-pass-pipeline='builtin.module(util.func(iree-preprocessing-generalize-linalg-matmul-experimental))' \
```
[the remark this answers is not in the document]
stbaione: Haha, a well founded Eek. Yeah, this was just ctrl+c of my day-to-day command. I can remove the experimental flags

docs/shortfin/llm/user/llama_serving.md
stbaione: What comes to mind is that if they're looking to run a sharded model, they probably know what a sharded model is. Although, someone new to the space might just stumble across this section after reading above section. Always best to encourage that kind of curiosity. 

Could only improve docs if I add a 1-2 sentence description and link to an article that goes more in depth

stbaione: Added 2 sentence description, with a link to a huggingface doc for more details. Also specified which specific technique is used in sharktank

docs/shortfin/llm/user/llama_serving.md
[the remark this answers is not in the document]
stbaione: Removed grossness

docs/shortfin/llm/user/llama_serving.md
ScottTodd: LGTM. Those docs are specific to training and huggingface transformers, but they get the point across.

docs/shortfin/llm/user/llama_serving.md
```diff
@@ -1,5 +1,19 @@
 # Llama end to end serving instructions
```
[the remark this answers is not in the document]
stbaione: Yeah, I really like this `shark-compile` idea. I was thinking of something like this, or moving it to the server command itself and adding a mutually exclusive list of args. Kinda like how [sglang](https://docs.sglang.ai/start/send_request.html#Launch-A-Server) does it.

For example:

```bash
python -m shortfin_apps.llm.server \
   --model-path https://huggingface.co/meta-llama/Llama-3.1-405B-Instruct
   --device hip \
   --tp 8 \
   --compile-target gfx942 \
   --device_ids 0 1 2 3 4 5 6 7
```

We would default model artifacts to be saved to `.cache/shark/Llama-3.1-405b-Instruct`, and add an optional arg `--cache-dir` if they wanna save them somewhere else. Maybe we also include some flags for regenerating cached artifacts, like `--ensure-export`, `--ensure-compile`.

We would still wanna support the case of pre-compiled artifacts, so maybe we keep around our original set of args for that scenario:

```bash
python -m shortfin_apps.llm.server \
   --tokenizer_json /path/to/output/tokenizer.json \
   --model_config /path/to/output/llama3.1-405b.config.json \
   --vmfb /path/to/output/llama3.1-405b.vmfb \
   --parameters \
      /path/to/output/llama3.1-405b.irpa \
      /path/to/output/llama3.1-405b.rank0.irpa \
      /path/to/output/llama3.1-405b.rank1.irpa \
      /path/to/output/llama3.1-405b.rank2.irpa \
      /path/to/output/llama3.1-405b.rank3.irpa \
      /path/to/output/llama3.1-405b.rank4.irpa \
      /path/to/output/llama3.1-405b.rank5.irpa \
      /path/to/output/llama3.1-405b.rank6.irpa \
      /path/to/output/llama3.1-405b.rank7.irpa \
   --device=hip \
   --device_ids 0 1 2 3 4 5 6 7
```

PR #896, opened 2025-02-02
[sharktank] restore custom matmul kernel

dan-garvey:
fix https://github.com/iree-org/iree/issues/19859#issuecomment-2628589074

```
python3 -m sharktank.examples.export_paged_llm_v1 --irpa-file=/sharedfile/llama3_8b_fp8.irpa \
--output-mlir=fp8_dan1.mlir \
--output-config=config1.json \
--bs=1 --attention-kernel torch \
--attention-dtype=float8_e4m3fnuz --activation-dtype=bfloat16
```

Issue #888, opened 2025-01-30
Job timeouts in "CI - sharktank / Data-dependent Tests" after updating IREE versions


ScottTodd:
Recent attempts to update the versions of IREE used in shark-ai have resulted in 6 hour job timeouts in the "CI - sharktank / Data-dependent Tests" job, source in this file: https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/ci-sharktank.yml.

First observed on version `3.2.0rc20250124` with https://github.com/nod-ai/shark-ai/pull/867.

Logs with IREE version `3.2.0rc20250129` can be seen on https://github.com/nod-ai/shark-ai/pull/879, for example https://github.com/nod-ai/shark-ai/actions/runs/13035503409/job/36364799644?pr=879

```
  source ${VENV_DIR}/bin/activate
  pytest \
    --with-clip-data \
    --with-flux-data \
    --with-t5-data \
    --with-vae-data \
    sharktank/tests/models/clip/clip_test.py \
    sharktank/tests/models/t5/t5_test.py \
    sharktank/tests/models/flux/flux_test.py \
    sharktank/tests/models/vae/vae_test.py \
    --durations=0 \
    --timeout=600
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    VENV_DIR: /shark-dev/actions-runner-llama/_work/shark-ai/shark-ai/.venv
    HF_HOME: /data/huggingface
    pythonLocation: /shark-dev/actions-runner-llama/_work/_tool/Python/3.11.11/x64
    PKG_CONFIG_PATH: /shark-dev/actions-runner-llama/_work/_tool/Python/3.11.11/x64/lib/pkgconfig
    Python_ROOT_DIR: /shark-dev/actions-runner-llama/_work/_tool/Python/3.11.11/x64
    Python2_ROOT_DIR: /shark-dev/actions-runner-llama/_work/_tool/Python/3.11.11/x64
    Python3_ROOT_DIR: /shark-dev/actions-runner-llama/_work/_tool/Python/3.11.11/x64
    LD_LIBRARY_PATH: /shark-dev/actions-runner-llama/_work/_tool/Python/3.11.11/x64/lib
============================= test session starts ==============================
platform linux -- Python 3.11.11, pytest-8.0.0, pluggy-1.5.0
rootdir: /shark-dev/actions-runner-llama/_work/shark-ai/shark-ai/sharktank
configfile: pyproject.toml
[...]

PR #916, opened 2025-02-05
Bump pinned iree-turbine version to 3.2.0rc20250205

sogartar:


sogartar:
I want this for PR #896. We generally would like to bump IREE and Turbine together, but there has been some issues preventing the IREE bump #888 that are not resolved yet.

ScottTodd (approved):

rsuderman (approved):

PR #925, opened 2025-02-06
[Tuner] Allow passing extra TD specs during tuning

Max191:
Adds a new flag to libtuner `--extra-spec-file`, which allows specifying a path for an additional tuning spec to use during compilation. This allows the tuner to account for already tuned shapes when compiling models, which can have an affect on what the optimal candidate is.

Another flag is added to the simple tuner (`--simple-best-spec-output-path`), which specifies a path to dump the final best spec after tuning. If stop-after is set to stop after dispatch benchmarking, then it will dump the best dispatch tuning spec, and otherwise the top model tuning spec will be saved.

These 2 flags together now allows tuning loops to be set up where a full tuning spec can be generated automatically/directly from tuning several dispatches in a loop, by dumping the spec from the previous dispatch, and passing it to the tuning for the next dispatch. This has 2 benefits:

1. We can account for other tuned dispatches when building a spec.
2. The concatenated spec will be automatically generated by the tuner, removing the manual concatenation step at the end of the tuning process.

### DIFF
--- a/tuner/examples/simple/simple_tuner.py
+++ b/tuner/examples/simple/simple_tuner.py
@@ -6,6 +6,7 @@
 
 import argparse
 from pathlib import Path
+import shutil
 from tuner import libtuner
 from tuner.common import *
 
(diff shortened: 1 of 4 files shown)


Max191:
Leaving this as a draft for now, since it needs tests.

Issue #926, opened 2025-02-06
Set up HF_HUB_CACHE for OSSCI Machines


renxida:
Our integration tests currently download weights for many huggingface models using

[hf_datasets.py](https://github.com/nod-ai/shark-ai/blob/56f3d217e871389c9cd11481b8a4316a82c90353/sharktank/sharktank/utils/hf_datasets.py)

Currently, caching is not working on:
- the new ossci cluster, labeled as `linux-mi300-gpu-1`
- the `azure-cpubuilder-linux-scale`

Which causes 20+ minute model weight re-downloads on [every CI run for shortfin llm](https://github.com/nod-ai/shark-ai/actions/runs/13185254507/job/36805854084?pr=890)

To address this, we need:
- a writeable space
- with enough space to hold a significant portion of the models in `/home/xidaren2/shark-ai/sharktank/sharktank/utils/hf_datasets.py` (my estimation is on the scale of hundreds of GBs)
- supports symlinking to the CI working dir
- (optionally) periodically cleared or downsized to evict old models

And set this directory as an environment variable
HF_HUB_CACHE
for CI tasks that uses it. 

This would speed up sharktank and shortfin integration tests by a lot. In my previous experiments, caching hf_hub can cause a 40 minute test to complete in 6 minutes.

renxida:
List of ci files that depend on this:
- [ ] sharktank data-dependent tests: https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/ci-sharktank.yml
- [ ] shortfin llm integration tests: https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/pkgci_shark_ai.yml

PR #932, opened 2025-02-07
[CI][sharktank] Move Sharktank Data-Dependent Tests to OSSCI Cluster

renxida:
We've been getting memory-allocation related errors possibly due to oversaturation on shark-mi300x-3.

### DIFF
--- a/.github/workflows/ci-sharktank.yml
+++ b/.github/workflows/ci-sharktank.yml
@@ -93,7 +93,7 @@ jobs:
     strategy:
       matrix:
         python-version: [3.11]
-        runs-on: [llama-mi300x-3]
+        runs-on: [linux-mi300-gpu-1]
       fail-fast: false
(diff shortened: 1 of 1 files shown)


renxida:
Need #926

sogartar:
These tests require some data to be available on the runner.

The premise that we don't have enough resources on the machine seems wrong.
What we need is a proper allocation of GPUs to jobs/runners. Also users to adhere to a mandatory reservation system. Meaning that if you don't request a GPU you don't see any. For example when running `iree-run-module --list_devices`.
I sometimes catch myself not specifying the correct GPU.
Maybe we could make a service that utilizes the `ROCR_VISIBLE_DEVICES` env var. Or to utilize Linux groups to control access.

sogartar:
I would also want to move these test soon to nightly as we should strive to have only fast test on every PR.

PR #936, opened 2025-02-07
[shortfin llm integration tests] Add tinystories llama2 25m model as a smoke test

### LINKED ISSUE #933: Tiny model for shortfin CI
Need a model that runs under a minute in the integration tests so we can get an early waning of failures & use it to test other things.

renxida:
This is a tiny model that would take less than a minute to download / export / compile.

The smoke test is intended to run before we run the full integration CI suite. This would give us early visibility into shortfin failures & save CI resources. 

This PR adds an integration_tests function to:
- download hf repo with `model.safetensors`
- convert to gguf 
- convert to irpa

Closes #933

### DIFF
--- a/app_tests/integration_tests/llm/model_management.py
+++ b/app_tests/integration_tests/llm/model_management.py
@@ -433,6 +433,7 @@ Dataset(
     (
         RemoteFile(
             file_id = "model.safetensors",
+            filename= "model.safetensors",
             repo_id = "Mxode/TinyStories-LLaMA2-25M-256h-4l-GQA",
             extra_filenames =  (
                 "config.json",
--- a/app_tests/integration_tests/llm/model_management.py
+++ b/app_tests/integration_tests/llm/model_management.py
@@ -453,36 +453,3 @@ TEST_MODELS["tinystories_llama2_25m"] = ModelConfig(
     batch_sizes=(1, 4),
     device_settings=None,
 )
-
-
-# Usage example
-# This uses a small model that takes roughly 2 minutes to run
-# from pathlib import Path
-# import sys
-# sys.path.append("/home/xidaren2/shark-ai")
-
-
-# from app_tests.integration_tests.llm.model_management import ModelConfig, ModelSource, ModelProcessor
-# from app_tests.integration_tests.llm.device_settings import GFX942
-
-# # Setup base directory
-# base_dir = Path("./model_artifacts")
-# base_dir.mkdir(exist_ok=True)
-
-# # Configure model
-# config = ModelConfig(
-#     source=ModelSource.HUGGINGFACE_TO_GGUF_TO_IRPA,
-#     repo_id="Mxode/TinyStories-LLaMA2-25M-256h-4l-GQA",
-#     model_file="model.irpa",
-#     tokenizer_id="Mxode/TinyStories-LLaMA2-25M-256h-4l-GQA",
-#     batch_sizes=(1, 4),
-#     device_settings=GFX942
-# )
-
-# # Process model
-# processor = ModelProcessor(base_dir)
-# artifacts = processor.process_model(config)
-
-# # Print results
-# print(f"Artifacts location: {base_dir}")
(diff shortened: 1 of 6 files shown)


app_tests/integration_tests/llm/model_management.py
```diff
@@ -143,12 +186,67 @@ def _download_from_huggingface(self) -> Path:
 
         return model_path
 
+    def _download_and_convert_from_huggingface(self) -> Path:
```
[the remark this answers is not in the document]
renxida: yeah. the huggingface stuff between here and sharktank is a little bit of a mess. I will submit another PR to address this. There is significant code replication here.

app_tests/integration_tests/llm/model_management.py
```diff
@@ -294,33 +394,62 @@ def process_model(self, config: ModelConfig) -> ModelArtifacts:
...
+    tokenizer_id="openlm-research/open_llama_3b_v2",
+    batch_sizes=(1, 4),
+    device_settings=None,
+)
+
+Dataset(
+    "Mxode/TinyStories-LLaMA2-25M-256h-4l-GQA",
+    (
+        RemoteFile(
+            filename,
```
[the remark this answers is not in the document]
renxida: just realized RemoteFile really should be RemoteFiles because it supports downloading multiple files. Fixed that.

renxida:
Comments addressed. Re-review pls.

Coming next:
- dedupe more code between here and sharktank's hf_datasets.py
- maybe move model_management and related files into shark-ai repo. Not sure what's the plan / if it's ok to make it less of  a stub

app_tests/integration_tests/llm/model_management.py
```diff
@@ -13,6 +16,37 @@
 logger = logging.getLogger(__name__)
 
 
+def get_llama_cpp_path() -> Path:
```
stbaione, line 19: Nit:

Should the download for `llama_cpp` be a function here or should it be moved to the github workflow? Not a blocker, just wondering

app_tests/integration_tests/llm/model_management.py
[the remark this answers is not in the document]
stbaione: Follow up PR sounds good

app_tests/integration_tests/llm/model_management.py
renxida: - Performance: doesn't matter. This takes less than 5 seconds, so no reason to cache it / pre-download it.
- Readability: I put it in model_management because it's used nowhere else. I would totally move it out into an appropriate place though if anything starts depending on it.

app_tests/integration_tests/llm/shortfin/tinystories_llama2_25m_test.py
```diff
@@ -0,0 +1,125 @@
...
+    [
+        ["tinystories_llama2_25m", {"prefix_sharing": "none"}],
+    ],
+    indirect=True,
+)
+
+
+# goldens are generated in: https://colab.research.google.com/drive/1pFiyvyIxk1RsHnw5gTk_gu9QiQNy9gfW?usp=sharing
+GOLDEN_PROMPT = "Once upon a time"
+GOLDEN_RESPONSE = ", there was a little girl named Lily."
```
stbaione, line 30: Just double checking, this is actually something I've wondered. I know we use greedy search for our generation. Do we have a purely deterministic implementation? I'm not sure if our temperature setting for requests actually have an impact on the results, so I've assumed this, have been curious.

Just wanna make sure that equality comparison for this prompt is reliable enough moving forward, and we don't flag ", in a galaxy far far away" as a failure... Unless it should be

renxida: Yup, this is purely deterministic. Or it should be.

We're only using greedy search. If the completion changed, it should be flagged as a failure. Let me add a note about this.

stbaione (approved):

PR #937, opened 2025-02-07
Update shortfin user instructions to use up to the supported torch version of 2.5.1+cpu

aviator19941:
Update shortfin user instructions to use up to the supported torch version of 2.5.1+cpu. The exported IR using torch2.6.0+cpu fails to compile and needs further investigating.

### DIFF
--- a/docs/shortfin/llm/user/llama_serving.md
+++ b/docs/shortfin/llm/user/llama_serving.md
@@ -80,7 +80,7 @@ following either https://pytorch.org/get-started/locally/ or our recommendation:
 
 ```bash
 # Fast installation of torch with just CPU support.
-pip install torch --index-url https://download.pytorch.org/whl/cpu
+pip install torch --index-url https://download.pytorch.org/whl/cpu torch==2.3.0
 ```
 
 ### Prepare a working directory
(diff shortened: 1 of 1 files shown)


docs/shortfin/llm/user/llama_serving.md
```diff
@@ -80,7 +80,7 @@ following either https://pytorch.org/get-started/locally/ or our recommendation:
 
 ```bash
 # Fast installation of torch with just CPU support.
-pip install torch --index-url https://download.pytorch.org/whl/cpu
+pip install torch --index-url https://download.pytorch.org/whl/cpu torch==2.3.0
```
ScottTodd, line 83: Does `"torch>=2.3.0,<2.6.0"` work? That's what we pin to in iree-turbine: https://github.com/iree-org/iree-turbine/blob/main/pytorch-cpu-requirements.txt

We also only test 2.3, 2.4, and 2.5 in the sharktank CI: https://github.com/nod-ai/shark-ai/blob/2c61420ff5b54de8990a7f2f8941f987023cd690/.github/workflows/ci-sharktank.yml#L30

aviator19941: Oh sorry didn't see this, let me try

aviator19941: Confirmed that using torch 2.5, compilation of exported IR succeeds for both TP1 and TP8.

ScottTodd (approved):
Could update the PR body and commit message to reflect the latest changes. Otherwise LGTM

PR #943, opened 2025-02-10
Adding conv igemm layout support - step 1: conv op matcher

jerryyin:
The convolution op matcher is originally implemented by @Max191. I've added bare minimal amount of implementation to allow it to be tested via dispatch parser test. This is a subset of https://github.com/nod-ai/shark-ai/pull/898 and as an attempt to reduce amount of changes in a single PR to facilitate code reviews.

Summary of changes:
 - common.py: new data structure `ConvolutionDimensions` to host convolution fields information; `ProblemSize` that will capture lhs/rhs and result affine expressions as well as conv dimension
 - op_matchers.py: Implemented the new `ConvolutionInterfaceMatcher`
 - dispatch_parser.py: hooked up with the new convolution matcher to populate additional fields necessary to populate igemm information

### DIFF
--- a/tuner/tuner/op_matchers.py
+++ b/tuner/tuner/op_matchers.py
@@ -93,7 +93,10 @@ class GenericOpMatcher(NamedOpMatcher):
 
         maps_attr = None
         for attr in op.opview.attributes:
-            if attr.name == "indexing_maps" and isinstance(attr.attr, ir.ArrayAttr):
+            if attr.name in (
+                "indexing_maps",
+                "linalg.memoized_indexing_maps",
+            ) and isinstance(attr.attr, ir.ArrayAttr):
                 maps_attr = attr.attr
         if maps_attr is None:
             return False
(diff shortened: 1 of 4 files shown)


tuner/tuner/op_matchers.py
```diff
@@ -93,7 +93,10 @@ def match(self, op: ir.Operation) -> bool:
 
         maps_attr = None
         for attr in op.opview.attributes:
-            if attr.name == "indexing_maps" and isinstance(attr.attr, ir.ArrayAttr):
+            if attr.name in (
+                "indexing_maps",
+                "linalg.memoized_indexing_maps",
+            ) and isinstance(attr.attr, ir.ArrayAttr):
```
Max191, line 99: I know this came from my change, but it is actually a bit confusing to check for memoized_indexing_maps in the `GenericOpMatcher`. Could you create another class `LinalgOpMatcher`, which extends `NamedOpMatcher`, and has all the implementation from `GenericOpMatcher`? Then you can make `GenericOpMatcher` extend the `LinalgOpMatcher` with its constructor setting the op names to "linalg.generic".

tuner/tuner/op_matchers.py
```diff
@@ -124,6 +127,113 @@ def get_map_result_dim_positions(map: ir.AffineMap):
     return exprs
 
 
+def get_convolution_dims(
```
Max191, line 130: This whole `get_convolution_dims` logic is temporary, with the expectation that we will swap this out with a call to MLIR python bindings for linalg::inferConvolutionDims. Could you add a TODO here that says we want to drop this logic once we expose this binding?

Max191 (commented):
Overall looking good, just a couple new comments

tuner/tuner/op_matchers.py
jerryyin: That's a good idea, thanks for the clear instructions!

tuner/tuner/op_matchers.py
jerryyin: comment added in latest commit

PR #947, opened 2025-02-11
shortfin_apps.llm.components: Architecture documentation + a timelines diagram for service.py and generate.py

renxida:
Experimenting with Mermaid for documenting our systems.

View the diagram at:

https://github.com/nod-ai/shark-ai/blob/b0ef8ba4b89273270b1cb6a62f7c17657306af5a/shortfin/python/shortfin_apps/llm/ARCHITECTURE.md


stbaione:
This is super useful from learning/reference standpoint, but when I see it as a code artifact, the first thing that comes to mind is that it'll either be tedious to keep up to date, or will quickly become stale.

Would this be better in https://github.com/nod-ai/playbook?

renxida:
good point. this is more useful for a few months for just linking & sending people i think.

i'd actually just keep it here in a closed PR; I am already able to link it via https://github.com/nod-ai/shark-ai/blob/b0ef8ba4b89273270b1cb6a62f7c17657306af5a/shortfin/python/shortfin_apps/llm/ARCHITECTURE.md

PR #948, opened 2025-02-11
Add metrics to sdxl phases

daveliddell:
Added logging via `@measure` (and other metrics.py functions) for the individual phases of SDXL to the sdxl shortfin server, including an average for the step portion of unet.


monorimet (approved):
LGTM

Issue #934, opened 2025-02-07
Bad results when serving Llama3.1_8b_f16 with shortfin


Alex-Vasile:
Following the [serving guide](https://github.com/nod-ai/shark-ai/blob/main/docs/shortfin/llm/user/llama_serving.md) and running llama3.1_8b_f16 at tp8 I am getting incoherent results. 

For example:
```
curl http://localhost:8000/generate \
    -H "Content-Type: application/json" \
    -d '{
        "text": "Name the capital of the United States.",
        "sampling_params": {"max_completion_tokens": 50}
    }'
````

Produces:
```
data: !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

**Comment by dan-garvey** [comment]
@stbaione have you seen this before

**Comment by stbaione** [comment]
> [@stbaione](https://github.com/stbaione) have you seen this before

This is new to me, as of this morning when Alex found out. I'm looking into it

**Comment by stbaione** [comment]
At HEAD for shark-ai/IREE. Using gguf from our user docs:

# Unsharded

```bash
curl http://localhost:8003/generate     -H "Content-Type: application/json"     -d '{
        "text": "Name the capital of the United States.",
        "sampling_params": {"max_completion_tokens": 50}
    }'

data:  Washington D.C.
What is the capital of the United States? Washington D.C.
What is the capital of the United States? Washington D.C.
What is the capital of the United States? Washington D.C.
What is the capital of the United States
```

# Sharded

I got a memory access fault on this on decode. This is new to me. Earliest I've heard this reported was this morning from Avi.

He reported that the unsharded 8b inputs looked good in testing yesterday. My suspicion is that this regression is related to us updating our pinned IREE versions. The sharding issue seems like something else. Still looking into it...

ScottTodd:
> My suspicion is that this regression is related to us updating our pinned IREE versions.

If you have an easy repro, you could try a partial revert of https://github.com/nod-ai/shark-ai/commit/2c61420ff5b54de8990a7f2f8941f987023cd690, just in `shortfin/CMakeLists.txt`. So switch shortfin to build against the older IREE runtime code but still use the latest IREE compiler code / .vmfb generation.

stbaione:
Looks like the output Avi got did actually show repetitiveness. So may not specifically be related to that requirements bump.

Alex found NaNs in KVCache post prefill. KVCache has no NaNs prior to prefill. Looking further

[...]

Issue #921, opened 2025-02-05
[WIP] LLM Server Release v3.3.0


stbaione:
Issue used for determining and tracking the issues for the Shortfin LLM Server, related to the `v3.3.0` release. 

For a more granular view into issues and progress for `v3.3.0`, see the [Shortfin LLM Projects](https://github.com/orgs/nod-ai/projects/29/views/7).

# Goals

### Artifacts Builder 
- Currently, our process to download safetensors from hf and generate the artifacts needed for the server is disgusting, see [llama_serving](https://github.com/nod-ai/shark-ai/blob/main/docs/shortfin/llm/user/llama_serving.md). We should write a painless way for users to launch our server with the supported model of their choice.
- My most recent idea for how this should look can be found [here](https://github.com/nod-ai/shark-ai/issues/691#issuecomment-2637650376).
- We should use the `iree.build` API, similar to how SDXL uses it ([reference](https://github.com/nod-ai/shark-ai/blob/main/shortfin/python/shortfin_apps/sd/components/builders.py)).

### OpenAI Compatible API
- This would be a really nice implementation from the user side. If we implement this, we could use things like the [openai sdk](https://github.com/openai/openai-python). I've seen other packages that also are capable of communicating with OpenAI compatible backends out of the box.
- I think similar to the sglang frontend language, this could give us a couple "free features" from the user perspective, because the sdk usually has some fancy stuff in it.
- Would also be a useful guide in what we should implement and how we should implement it. Given that it's the most used LLM sdk (to my knowledge), there's a reason that they have the routes that they do, structure their responses the way that they do, etc. 
- [Reference](https://github.com/openai/openai-openapi)

# Issues
- Artifacts Builder
    - #402 
    - #691 
- OpenAI API
    - #245 
    - #691

stbaione:
> Benchmarking Dashboard

There's already a repo and site set up for this: https://github.com/iamakanshab/SGLang_Automation/tree/main/data_pipeline

I was busy with 405b when it got spun up, but it just needs to be hooked up with the CI action

PR #951, opened 2025-02-11
Use llama chat tags in example requests

stbaione:
Use llama chat tags in example requests.

More details on this can be found [here](https://github.com/nod-ai/shark-ai/issues/934#issuecomment-2649718739)

### DIFF
--- a/docs/shortfin/llm/user/llama_serving.md
+++ b/docs/shortfin/llm/user/llama_serving.md
@@ -262,7 +262,7 @@ Next, let's send a generation request:
 curl http://localhost:8000/generate \
     -H "Content-Type: application/json" \
     -d '{
-        "text": "Name the capital of the United States.",
+        "text": "<|begin_of_text|>Name the capital of the United States.<|eot_id|>",
         "sampling_params": {"max_completion_tokens": 50}
     }'
 ```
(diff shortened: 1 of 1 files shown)


docs/shortfin/llm/user/llama_serving.md
```diff
@@ -262,7 +262,7 @@ Next, let's send a generation request:
 curl http://localhost:8000/generate \
     -H "Content-Type: application/json" \
     -d '{
-        "text": "Name the capital of the United States.",
+        "text": "<|begin_of_text|>Name the capital of the United States.<|eot_id|>",
```
ScottTodd, line 265: Should the server be inserting these itself? Maybe for a specific endpoint? Right now we just have "generate"? 

Other projects have endpoints like `chat/completions` that don't include these tags:
* https://docs.sglang.ai/start/send_request.html
* https://docs.vllm.ai/en/latest/getting_started/quickstart.html#openai-completions-api-with-vllm

https://github.com/sgl-project/sglang/blob/bb418ced802c6dbb6b0ae0d65218327129148769/python/sglang/srt/conversation.py#L49-L279

Are we expecting shortfin to sit under a frontend like sglang that will insert the tags for us? Or should we be putting similar logic in our apps too?

stbaione: Yeah, they're called [`ChatTemplates`](https://docs.sglang.ai/references/custom_chat_template.html). It's a feature that we should add. For now, our SGLang serving docs use the llama tag and apply it through the sglang frontend language. 

By default, if a chat_template is not specified, they apply it based on model name/path: https://github.com/sgl-project/sglang/blob/main/python/sglang/lang/backend/runtime_endpoint.py#L48

Shouldn't be a hard thing to add, just not related to this PR, which is aimed at cleaning up the user doc experience

stbaione: I see that as an issue to add to https://github.com/nod-ai/shark-ai/issues/921

ScottTodd (approved):
Good enough for now (though see the discussion)

PR #957, opened 2025-02-11
Reduce buffer copying by using one device to reduce and distribute

Alex-Vasile:
For tensor parallelization (TP) > 1, this reduces buffer copies from (TP-1)*TP copies down to (TP-1)*2. Setup work for future improvements. On its own, it only results in minor speedups since all devices have to wait for the reduction result.


stbaione:
Should we hold off on merging this until the sharding accuracy is stable again?

rsuderman (approved):

PR #959, opened 2025-02-12
Add integration test harness that bypasses the FastAPI server to directly test the batcher 

renxida:
Previously, to test concurrent generation, we quickly submit multiple requests.

This does not consistently result in the same batching, and timing differences across shortfin can cause e.g. 4 requests to go from being all batches together to boarding a batch of 3 and then a batch of 1. This caused some very difficult-to-replicate errors and accounted for some of the regressions we observed leading up to the v3.2.0 release.

This harness and it's associated tests will eliminate inconsistent batching as an untested point of failure, allowing us to detect and replicate future errors that depend on certain batching conditions to appear.

### DIFF
--- a/app_tests/benchmark_tests/llm/sglang_benchmarks/conftest.py
+++ b/app_tests/benchmark_tests/llm/sglang_benchmarks/conftest.py
@@ -80,7 +80,7 @@ def server(model_artifacts, request):
     )
 
     server_instance = ServerInstance(server_config)
-    server_instance.start()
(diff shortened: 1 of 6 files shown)


renxida:
Note that this introduces some code duplication in server_management.py.

Will put in a follow up PR to dedupe the code there, and between model_management.py and sharktank's hf_datasets.py

renxida:
Temporarily dropping this in favor of a micro-server like SDXL's.

PR #961, opened 2025-02-12
Add metrics to sdxl phases (#948)

daveliddell:
Added logging via `@measure` (and other metrics.py functions) for the individual phases of SDXL to the sdxl shortfin server, including an average for the step portion of unet.

Same changes as on main branch.


daveliddell:
python3.11 is messed up for pre-commit on my machine, getting that in order, so I can run pre-commit before checking in! :-)

PR #963, opened 2025-02-13
[shortfin-sd] Reusable service command buffers

monorimet:
This PR non-trivially restructures the SDXL inference process to interface with reusable command buffers.

At service startup, we attach to each fiber a set of reusable command buffers (a set of empty device arrays to be reused throughout the inference process), and upon assigning each exec request a fiber (now the "meta_fiber" equipped with command buffers), we pop an appropriate command buffer from the meta fiber, mount the command buffer on the request, and return the command buffer to the meta_fiber for reuse.

Each meta_fiber is equipped with an explicit number of command buffers, and if this count is exceeded by the number of InferenceExecProcesses assigned to the fiber, a new command buffer will be initialized ad-hoc. This _shouldn't_ happen if service is managed properly, as we shouldn't oversubscribe fibers too heavily (thundering herd).

These changes also simplify the request structure throughout the inference processor, operating on a _single_ request as opposed to a bundle. This greatly simplifies the work required to enable larger batch sizes, and reduces how much we need to handle program I/O during execution.

We also remove all `await device` usage where it wasn't required. It should only be used after VAE decode, where we are done with the GPU and need to use the result contents of a device->host transfer. fyi @daveliddell

### DIFF
--- a/shortfin/python/shortfin_apps/sd/components/messages.py
+++ b/shortfin/python/shortfin_apps/sd/components/messages.py
@@ -79,12 +79,10 @@ class InferenceExecRequest(sf.Message):
 
         # Denoise phase.
         self.sample = sample
+        self.guidance_scale = guidance_scale
         self.steps = steps
 
         # Decode phase.
-        self.denoised_latents = denoised_latents
-
-        # Postprocess.
         self.image_array = image_array
 
         self.result_image = None
@@ -166,15 +164,11 @@ class InferenceExecRequest(sf.Message):
                 meta = [self.width, self.height]
                 return required, meta
             case InferencePhase.DENOISE:
-                required = not self.denoised_latents
+                required = True
                 meta = [self.width, self.height, self.steps]
                 return required, meta
             case InferencePhase.ENCODE:
-                p_results = [
-                    self.prompt_embeds,
-                    self.text_embeds,
-                ]
-                required = any([inp is None for inp in p_results])
+                required = True
                 return required, None
             case InferencePhase.PREPARE:
                 p_results = [self.sample, self.input_ids]
--- a/shortfin/python/shortfin_apps/sd/components/messages.py
+++ b/shortfin/python/shortfin_apps/sd/components/messages.py
@@ -49,7 +49,10 @@ class InferenceExecRequest(sf.Message):
         steps: int | None = None,
         guidance_scale: float | list[float] | sfnp.device_array | None = None,
         seed: int | list[int] | None = None,
-        input_ids: list[list[int]] | list[list[list[int]]] | list[sfnp.device_array] | None = None,
+        input_ids: list[list[int]]
+        | list[list[list[int]]]
+        | list[sfnp.device_array]
+        | None = None,
         sample: sfnp.device_array | None = None,
         prompt_embeds: sfnp.device_array | None = None,
         text_embeds: sfnp.device_array | None = None,
@@ -150,7 +153,7 @@ class InferenceExecRequest(sf.Message):
 
         self.command_buffer = cb
         return
-        
+
     def post_init(self):
         """Determines necessary inference phases and tags them with static program parameters."""
         for p in reversed(list(InferencePhase)):
--- a/shortfin/python/shortfin_apps/sd/components/messages.py
+++ b/shortfin/python/shortfin_apps/sd/components/messages.py
@@ -103,7 +103,7 @@ class InferenceExecRequest(sf.Message):
         # Input IDs for CLIP if they are used as inputs instead of prompts.
         if self.input_ids is not None:
             # Take a batch of sets of input ids as ndarrays and fill cb.input_ids
-            host_arrs = [None] * 4
+            host_arrs = [None] * len(cb.input_ids)
             for idx, arr in enumerate(cb.input_ids):
                 host_arrs[idx] = arr.for_transfer()
                 for i in range(cb.sample.shape[0]):
--- a/shortfin/python/shortfin_apps/sd/components/messages.py
+++ b/shortfin/python/shortfin_apps/sd/components/messages.py
@@ -54,11 +54,6 @@ class InferenceExecRequest(sf.Message):
         | list[sfnp.device_array]
         | None = None,
         sample: sfnp.device_array | None = None,
-        prompt_embeds: sfnp.device_array | None = None,
-        text_embeds: sfnp.device_array | None = None,
-        timesteps: sfnp.device_array | None = None,
-        time_ids: sfnp.device_array | None = None,
-        denoised_latents: sfnp.device_array | None = None,
         image_array: sfnp.device_array | None = None,
     ):
         super().__init__()
@@ -83,14 +78,8 @@ class InferenceExecRequest(sf.Message):
         self.input_ids = input_ids
 
         # Denoise phase.
-        self.prompt_embeds = prompt_embeds
-        self.text_embeds = text_embeds
         self.sample = sample
         self.steps = steps
-        self.steps_arr = None
-        self.timesteps = timesteps
-        self.time_ids = time_ids
-        self.guidance_scale = guidance_scale
 
         # Decode phase.
         self.denoised_latents = denoised_latents
--- a/shortfin/python/shortfin_apps/sd/components/messages.py
+++ b/shortfin/python/shortfin_apps/sd/components/messages.py
@@ -112,7 +112,7 @@ class InferenceExecRequest(sf.Message):
 
     def set_command_buffer(self, cb):
         # Input IDs for CLIP if they are used as inputs instead of prompts.
-        if self.input_ids:
+        if self.input_ids is not None:
             # Take a batch of sets of input ids as ndarrays and fill cb.input_ids
             host_arrs = [None] * 4
             for idx, arr in enumerate(cb.input_ids):
@@ -127,7 +127,7 @@ class InferenceExecRequest(sf.Message):
                 cb.input_ids[idx].copy_from(host_arrs[idx])
 
         # Same for noisy latents if they are explicitly provided as a numpy array.
-        if self.sample:
+        if self.sample is not None:
             sample_host = cb.sample.for_transfer()
             with sample_host.map(discard=True) as m:
                 m.fill(self.sample.tobytes())
--- a/shortfin/python/shortfin_apps/sd/components/messages.py
+++ b/shortfin/python/shortfin_apps/sd/components/messages.py
@@ -10,6 +10,7 @@ import logging
 
 import shortfin as sf
 import shortfin.array as sfnp
+import numpy as np
 
 from .io_struct import GenerateReqInput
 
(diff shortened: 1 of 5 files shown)


shortfin/python/shortfin_apps/sd/components/messages.py
```diff
@@ -58,8 +59,9 @@ def __init__(
         image_array: sfnp.device_array | None = None,
     ):
         super().__init__()
+        self.command_buffer = None
         self.print_debug = True
-
+        self.batch_size = 1
```
IanNod, line 67: Why are we hardcoding this? Also I don't see it used anywhere

shortfin/python/shortfin_apps/sd/components/messages.py
```diff
@@ -74,14 +76,15 @@ def __init__(
...
-        # This is a list of sequenced positive and negative token ids and pooler token ids.
+        # This is a list of sequenced positive and negative token ids and pooler token ids (tokenizer outputs)
         self.input_ids = input_ids
 
         # Denoise phase.
         self.prompt_embeds = prompt_embeds
         self.text_embeds = text_embeds
         self.sample = sample
         self.steps = steps
+        self.steps_arr = None
```
IanNod, line 90: Doesn't look to be used anywhere

shortfin/python/shortfin_apps/sd/components/messages.py
```diff
@@ -104,36 +107,50 @@ def __init__(
...
+        cb.steps_arr.copy_from(steps_host)
+
+        num_step_host = cb.num_steps.for_transfer()
+        num_step_host.items = [self.steps]
+        cb.num_steps.copy_from(num_step_host)
+
+        guidance_host = cb.guidance_scale.for_transfer()
+        with guidance_host.map(discard=True) as m:
+            # TODO: do this without numpy
+            np_arr = np.asarray(self.guidance_scale, dtype="float16")
```
IanNod, line 149: is guidance_scale always fp16 or only for fp16/punet

IanNod (approved):
few nit comments but otherwise I think it looks good

shortfin/python/shortfin_apps/sd/components/messages.py
monorimet: self.exec_request.batch_size is used in the inference process. I'll fix the hardcode once larger batch sizes are actually possible.

shortfin/python/shortfin_apps/sd/components/messages.py
monorimet: Oops. left hardcoded to avoid parsing the sfnp.DType to np.dtype. Will fix.

shortfin/python/shortfin_apps/sd/components/service.py
```diff
@@ -350,366 +359,293 @@ def __init__(
         service: GenerateService,
         fiber,
```
daveliddell, line 360: A meta fiber is the same thing as an EquippedFiber, right?  If so, can you use one or the other name consistently?  Also, it would be nice if function arguments taking the thing would be called meta_fiber or equipped_fiber to distinguish it from raw sf fibers.

shortfin/python/shortfin_apps/sd/components/service.py
```diff
@@ -350,366 +359,293 @@ def __init__(
...
             if phases[InferencePhase.POSTPROCESS]["required"]:
-                await self._postprocess(device=device0, requests=self.exec_requests)
-            await device0
-            for i in range(req_count):
-                req = self.exec_requests[i]
-                req.done.set_success()
+                await self._postprocess(device=device)
+            await device
+            self.exec_request.done.set_success()
+            self.meta_fiber.command_buffers.append(self.exec_request.command_buffer)
```
daveliddell, line 401: Does returning the command buffer to the meta fiber (equipped fiber) need to be done outside the try block, so that an exception doesn't cause the command buffer to be lost?

shortfin/python/shortfin_apps/sd/components/service.py
```diff
@@ -350,366 +359,293 @@ def __init__(
...
-                        logger.error("Executor process recieved disjoint batch.")
-                phase = req.phase
-            phases = self.exec_requests[0].phases
-            req_count = len(self.exec_requests)
-            device0 = self.fiber.device(0)
+            if not self.exec_request.command_buffer:
+                self.assign_command_buffer(self.exec_request)
+
+            device = self.fiber.device(0)
+            await device
```
daveliddell, line 387: What is this wait for? If it's for metrics, we might need to break the actual "inference process" out of the run function:

```
async def run(self):
    ...
    device = self.fiber.device(0)
    await device
    run_inference(device)

@measure(...)
aync def run_inference(device):
    phases = ...
    _prepare, _encode, etc
```

shortfin/python/shortfin_apps/sd/components/service.py
```diff
@@ -350,366 +359,293 @@ def __init__(
...
-                device, [req_bs, self.service.model_params.max_seq_len], sfnp.sint64
-            ),
-            sfnp.device_array.for_device(
-                device, [req_bs, self.service.model_params.max_seq_len], sfnp.sint64
-            ),
-            sfnp.device_array.for_device(
-                device, [req_bs, self.service.model_params.max_seq_len], sfnp.sint64
-            ),
-        ]
         host_arrs = [None] * 4
```
daveliddell, line 431: another hard-coded 4. Should it be len(cb.input_ids}?

shortfin/python/shortfin_apps/sd/components/service.py
```diff
@@ -350,366 +359,293 @@ def __init__(
...
+        generator = sfnp.RandomGenerator(seed)
+
+        sample_host = cb.sample.for_transfer()
+        with sample_host.map(discard=True) as m:
+            m.fill(bytes(1))
+
+        sfnp.fill_randn(sample_host, generator=generator)
+
+        cb.sample.copy_from(sample_host)
+        await device
```
daveliddell, line 456: This is a new await device that wasn't in the original code.  Is it needed?

shortfin/python/shortfin_apps/sd/components/service.py
```diff
@@ -350,366 +359,293 @@ def __init__(
...
             "INVOKE %r: %s",
             fn,
-            "".join([f"\n  0: {latents.shape}"]),
+            "".join([f"\n  0: {cb.latents.shape}"]),
         )
-        (image,) = await fn(latents, fiber=self.fiber)
-
+        (cb.images,) = await fn(cb.latents, fiber=self.fiber)
+        cb.images_host.copy_from(cb.images)
         await device
```
daveliddell, line 554: I pulled this await device out of my (now obsolete) PR, as we're not accessing the actual bytes of the buffer, only the buffer metadata, which you can do without waiting.

shortfin/python/shortfin_apps/sd/components/service.py
```diff
@@ -350,366 +359,293 @@ def __init__(
...
+                await self._prepare(device=device)
             if phases[InferencePhase.ENCODE]["required"]:
-                await self._encode(device=device0, requests=self.exec_requests)
+                await self._encode(device=device)
             if phases[InferencePhase.DENOISE]["required"]:
-                await self._denoise(device=device0, requests=self.exec_requests)
+                await self._denoise(device=device)
             if phases[InferencePhase.DECODE]["required"]:
-                await self._decode(device=device0, requests=self.exec_requests)
+                await self._decode(device=device)
```
daveliddell, line 396: Right in here is where I had added (in my now obsolete PR):

```
            # Postprocessing needs the output data to be on the host.  Even
            # without postprocessing, we're done with the GPU, so we wait for
            # it to finish here.
            await device
```
And I also removed the await device from below and from _decode.  Basically, none of the phases except _postprocess are using the data from the device buffers, so the whole thing can free run up to this point.

daveliddell (approved):
Very elegant solution to the management of the device arrays!

All of my comments are at your discretion to disregard, as always. :-)

shortfin/python/shortfin_apps/sd/components/service.py
monorimet: When we assign a command buffer to the request in the beginning of the inference process, we copy some host arrays (anything user-submitted like input_ids as numpy arrays) and await here for the copy to device to finish.
The await after the phases is extraneous and will be removed.

shortfin/python/shortfin_apps/sd/components/service.py
monorimet: For SDXL inference this is a very specific shape that isn't tied to a config variable. I don't think it should be driven by the shape of a command buffer element -- it won't ever change for SDXL

shortfin/python/shortfin_apps/sd/components/service.py
monorimet: yes, as it succeeds a host to device copy.

shortfin/python/shortfin_apps/sd/components/service.py
monorimet: The decode step transfers the VAE output to host, then creates a numpy array with the result. Doesn't the numpy array instantiation need the buffer contents and not just its metadata? Hence the await device

shortfin/python/shortfin_apps/sd/components/service.py
monorimet: Meh... since we're doing h2d here it probably makes sense just to align them.

shortfin/python/shortfin_apps/sd/components/service.py
daveliddell: Do need to wait for the copy to device to finish before starting a program?  I thought that as long as they're sequenced correctly on a fiber, shortfin and/or the GPU command buffer guarantees correct order of execution

shortfin/python/shortfin_apps/sd/components/service.py
daveliddell: Looks like np.frombuffer does a shallow copy, which would mean the data is free to come in later.  Anyway, Stella was the one who pointed out that none of the await devices should be necessary until postprocess, where you actually use the data.  I had consistent results doing just one await device before preprocess, not that I was pounding on the server particularly hard. :-)

shortfin/python/shortfin_apps/sd/components/service.py
```diff
@@ -281,15 +297,16 @@ def board_flights(self):
...
-            if len(self.service.idle_fibers) == 0:
+            if len(self.service.idle_meta_fibers) == 0:
+                logger.debug("Waiting for an idle fiber...")
                 return
-            fiber = self.service.idle_fibers.pop()
-            fiber_idx = self.service.fibers.index(fiber)
-            worker_idx = self.service.get_worker_index(fiber)
-            logger.debug(f"Sending batch to fiber {fiber_idx} (worker {worker_idx})")
-            self.board(batch["reqs"], fiber=fiber)
+            fiber = self.service.idle_meta_fibers.pop(0)
```
daveliddell, line 296: Rename this `fiber` variable to `meta_fiber`?  I thought (improbably) that I had caught a type mismatch on line 307 when I saw `meta_fiber=fiber`.  :-)

daveliddell (approved):

shortfin/python/shortfin_apps/sd/components/service.py
stellaraccident: You shouldn't need to wait for it to finish. The only time you should need to await the device is if you copy back to the host and want to inspect the results (on the host).

shortfin/python/shortfin_apps/sd/components/service.py
stellaraccident: Shouldn't need await on h2d so long as you are not mutating the host side buffer while the pipeline is progressing.

shortfin/python/shortfin_apps/sd/components/service.py
stellaraccident: Yeah, you only need to await when observing the bytes on the host. Just juggling views aren't touching the contents. With that said, if you were trying to have a "safety await" and didn't know for sure that the numpy API you are using only takes a view, then it would be a reasonable thing to do.

shortfin/python/shortfin_apps/sd/components/service.py
monorimet: Thanks for the catch!

PR #967, opened 2025-02-14
Don't use GGUF but convert T5 directly from Hugging Face

sogartar:
We don't want the stack to depend on the conversion tool from the lamma.cpp repo. Also the conversion to GGUF would not convert all tensors to bf16, but leave some in f32. We would like to control that ourselves if needed.

This change makes any previously generated IRPA files obsolete.

### DIFF
--- a/sharktank/sharktank/layers/configs/llm_configs.py
+++ b/sharktank/sharktank/layers/configs/llm_configs.py
@@ -14,7 +14,7 @@ When in question, we draw from the vocabulary and normalization they have done
 (and indeed, can bootstrap these off of GGUF files).
 """
 
-from dataclasses import asdict, dataclass, field
+from dataclasses import asdict, dataclass, field, fields
 from typing import Any, Optional
 import torch
 from transformers import T5Config as T5ConfigHf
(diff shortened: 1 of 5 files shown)


sharktank/sharktank/layers/configs/llm_configs.py
```diff
@@ -231,53 +233,57 @@ def __post_init__(self):
...
-            {
-                config_name: properties[gguf_name]
-                for gguf_name, config_names in gguf_to_config_names_map.items()
-                for config_name in config_names
-            }
-        )
-
-        gguf_to_optional_config_names_map = {
-            "t5.decoder_start_token_id": ["decoder_start_token_id"],
+    def from_hugging_face_config(
```
KyleHerndon, line 236: Not blocking on this, but can these keys be discovered automatically?

sharktank/tests/models/t5/t5_test.py
```diff
@@ -498,19 +493,19 @@ def testCompareAgainstTransformers(
 
         theta = Theta(
             {
-                "attn_q.weight": DefaultPrimitiveTensor(
+                "q.weight": DefaultPrimitiveTensor(
```
KyleHerndon, line 496: What is the motivation with all the renames? Is it just the corresponding names from huggingface's layout instead of gguf's?

KyleHerndon (approved):

sharktank/tests/models/t5/t5_test.py
sogartar: Yes.

sharktank/sharktank/layers/configs/llm_configs.py
sogartar: I shortened it.

PR #918, opened 2025-02-05
Update IREE to 3.2.0rc20250129 and disable t5 sharktank-data-dependent tests 

renxida:
Diff: https://github.com/iree-org/iree/compare/iree-3.2.0rc20250120...iree-3.2.0rc20250129

PR #969, opened 2025-02-14
[sharktank][ci] Re-enable t5 data-dependent tests

renxida:
They were timing out & causing other sharktank tests to not run at all, so we disabled them in #918 to uncover other failures.

Pending @AWoloszyn's work in #888  and the fix in https://github.com/iree-org/iree/pull/19999, we might be good to reenable them.


marbre:
We should be able to merge this after we bumped IREE to `3.3.0rc20250219` or beyond.

PR #975, opened 2025-02-17
SDXL app: move await device to fix empty results

daveliddell:
The output array was coming out empty.  We discovered that the `await device` needs to be moved after the transfer to host but before the device array's `items` is read.


monorimet (approved):
LGTM. Thanks

PR #976, opened 2025-02-17
Improve T5 encoder tests with more prompts and static context length

sogartar:
The set of prompts is not big enough for statistically sound testing of
the T5 encoder. This is true for other text encoders.
With the expansion of the prompt set the bf16 numerical difference
between eager and IREE vanished. IREE is even more accurate.

In tests the tokenizer padding has been change to produce always max
length token sequence. This is in line how T5 is used int the Flux
pipeline. The T5 encoder export has been expanded with an option to
export with a static token sequence length.

The tests were refactored to share tolerance values for f32 and bf16.


sogartar:
This PR is on top of #967, which must be merged first.

sharktank/sharktank/utils/testing.py
```diff
@@ -344,4 +344,8 @@ def decorator(test_item: Callable):
     "The horse went into the river",
     "We need at least one sentence long enough so that it spans more than one padding block which by default is of size 16.",
     "Make the batch size 4",
+    "In data analysis, cosine similarity is a measure of similarity between two non-zero vectors defined in an inner product space. Cosine similarity is the cosine of the angle between the vectors; that is, it is the dot product of the vectors divided by the product of their lengths. It follows that the cosine similarity does not depend on the magnitudes of the vectors, but only on their angle.",
```
archana-ramalingam, line 347: Can we move the prompts to a separate `.txt.` file? 

Also, to improve the prompt/dataset variety, it's advisable to use a standard dataset to run evaluations of this sort, like we do for perplexity tests [here](https://github.com/nod-ai/shark-ai/blob/main/sharktank/sharktank/evaluate/perplexity_iree.py#L209).

KyleHerndon: +1. I also don't want to worry about the copyright of some of this text :)

KyleHerndon (approved):

sharktank/sharktank/utils/testing.py
archana-ramalingam: Great, as the prompts can sometimes be non-sensical shorter phrases/titles, it's useful to add preprocessing for the prompts, like [this](https://github.com/nod-ai/shark-ai/blob/main/sharktank/sharktank/evaluate/perplexity_iree.py#L219C9-L223C10)

PR #985, opened 2025-02-19
[shortfin llm] Simplify interface between llm specific code and fastapi webapp

renxida:
Currently it's a little hard to trace through the server initialization code because a lot of global variables are involved. This PR eliminates them and uses a singe location to put a context manager for init-deinit.


stbaione (approved):
Very nice refactor

PR #986, opened 2025-02-19
Move the vae reshaping logic back to its correct location

KyleHerndon:


KyleHerndon:
Subsumed by [989](https://github.com/nod-ai/shark-ai/pull/989)

PR #1000, opened 2025-02-25
Fix llama fp8 benchmark test and remove decomposed benchmarking tests

aviator19941:
Fixes llama fp8 benchmark test, updates the prefill and decode paths, and removes decomposed benchmarking tests.


archana-ramalingam (approved):
For the pre-submit llama benchmark test failure, it uses [pinned](https://github.com/nod-ai/shark-ai/blob/main/requirements-iree-pinned.txt) IREE, which needs to be bumped to `3.3.0rc20250219` version that includes the fix for this error.

sharktank/tests/models/llama/benchmark_amdgpu_test.py
```diff
@@ -807,7 +826,7 @@ def testBenchmark405B_f16_TP8_Non_Decomposed_Input_Len_2048(self):
         )
         output_shard_file_name = (
             self.artifacts_dir
-            / f"fp16/tp8/llama3.1_405b_fp16_tp{self.tensor_parallelism_size}_parameters.irpa"
+            / f"tp8/llama3_405b_instruct_fp16_tp{self.tensor_parallelism_size}.irpa"
```
AmosLewis, line 829: `/ f"tp8/llama3.1_405b_fp16_tp{self.tensor_parallelism_size}_parameters.irpa"`

AmosLewis: @Alex-Vasile I think the path issue start from here

PR #1008, opened 2025-02-25
Pipeline parallelism support

Alex-Vasile:
**Goal:** Introduce pipeline parallelism without requiring a change to the weight irpa files or the forward passes for the different layers (see PPFFN.forward in the example file).

**Changes**
- ShardedTensor now explicitly store what device each of their shards should live on in a `.device` attribute. Previously the implicit convection was that shard `i` lived on device `i`.
- ShardedTensor can also be pinned to specific devices, such as for weights, or left unpinned to signal that it should be moved if needed using a `.pinned` attribute.
- operators call a helper function to see if either tensor needs to be transferred such that all shards are on matching devices. E.g. `ops.foo(t1 on devs [1,2,3], t2 pinned on devs[5,6,7])` would transfer the shards of t1 onto devices [5,6,7] before performing the operation.
- Several helper functions in `ops` can take in a `torch.Tensor` and therefore won't know what devices to place them on, e.g. `def replicate(input: AnyTensor, count: int) -> ShardedTensor:`. I've added `devices` and `pinned` as extra parameters and used defaults to keep the current behaviour unchanged.
- Added a wrapper to all functions important from `ops.signatures` into `ops` to handle transfer and pinning automatically when called with `ShardedTensor` subclasses. Making device parallelism work, mostly, invisibly without needing to modify the functions in `sharded_impls.py`
    - One downside is that IDEs don't appear to be able to do tab completion anymore for `ops.___`

**Discussion points**
- Overall thoughts on approach?
- Both `device` and `pinned` are required parameters. Should either, especially `pinned`, be optional and have defaults?
- Exactly how should the different unary ops like `ops.replicate` handle the extra parameters needs more thought.
- Should `is_deep_equal()` consider `.devices` and `.pinned`? Enabling causes a few tests to fail, such as `testReplicatedLhsShardedParallelDimRhs`


sharktank/sharktank/types/tensors.py
```diff
@@ -752,10 +752,29 @@ class ShardedTensor(InferenceTensor):
     """
 
     def __init__(
-        self, *, shape: list[int], shard_dim: int | None, name: str = UnnamedTensorName
+        self, *, 
+        shape: list[int], 
+        shard_dim: int | None, 
+        name: str = UnnamedTensorName,
+        devices: Tuple[int],
+        devices_pinned: bool,
```
rsuderman, line 760: Make this an `Optional[bool]` set to  `None`. From there we can check some behaviors. We may be able to check if `torch.Tensor` is a `AbstractTensor` or one associated with the value to pin.

Overall we want to think of this default as "if this is real values, has a DeviceTensorTrait, etc lets make it pinned".

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -30,6 +30,88 @@
...
+                    else barrier_on_logical_device(shard, d_pinned[j])
+                )
+                for j, shard in enumerate(tensor.shards)
+            )
+            new_tensors.append(copy_w_new_shards_and_devices(tensor, shards, d_pinned))
+
+    return tuple(new_tensors)
+
+
+def override_w_tranfer(operation, *override_args):
```
rsuderman, line 102: We may want to include `**kwargs` as well. Operators could have some keyword arg we want propagated. Its also likely the new override could have some special return cases. E.g. pinning results.

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -30,6 +30,88 @@
...
+    def decorator(f):
+        @operation.override(*override_args)
+        def wrapper(*args: Tuple, **kwargs: Dict[str, Any]):
+            args = list(args)
+            t_i = [i for i, arg in enumerate(args) if isinstance(arg, ShardedTensor)]
+            if len(t_i) > 0:
+                t_vals = transfer_if_needed(*(args[i] for i in t_i))
+                for i, t in zip(t_i, t_vals):
+                    args[i] = t
+            return f(*args, **kwargs)  # TODO: Should I convert func_args back to tuple after modifying it?
```
rsuderman, line 112: We may also want to add a special case that if the input is pinned and its an unary tensor operator we would also want to pin. Its a bit special but almost guaranteed.

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -68,16 +150,16 @@ def all_reduce_split_or_unreduced(
...
-                    else transfer_to_logical_device(shard, i)
+                    else transfer_to_logical_device(shard, input.devices[i])
                 )
                 for j, shard in enumerate(input.shards)
             ],
         )
         for i in range(input.shard_count)
     ]
-    return ReplicatedTensor(ts=shards)
+    return ReplicatedTensor(ts=shards, devices=input.devices, devices_pinned=input.devices_pinned)
```
rsuderman, line 162: I wouldn't set any result of an all_reduce or all_gather as pinned. We are already transferring around so moving it has no cost.

rsuderman (changes requested):

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -30,6 +30,88 @@
...
+            name=tensor.name,
+            devices=new_devices,
+            devices_pinned=tensor.devices
+        )
+    elif isinstance(tensor, UnreducedTensor):
+        return UnreducedTensor(
+            ts=new_shards,
+            name=tensor.name,
+            shape=tensor.shape,
+            devices=new_devices
```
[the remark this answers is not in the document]
Alex-Vasile: Yup, missed it and I guess that branch never got hit

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -30,6 +30,88 @@
...
+    
+    # Check if all tensors are on the same devices.
+    devices_0 = tensors[0].devices
+    for tensor in tensors[1:]:
+        if not all(devices_0[j] == tensor.devices[j] for j in range(len(devices_0))):
+            break
+    else:
+        return tensors  # All tensors already on same device
+
+    i_pinned = tuple(i for i, t in enumerate(tensors) if t.devices_pinned)
```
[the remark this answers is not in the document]
Alex-Vasile: Agree. I originally re-used them later one, but was easier to read without reusing, forgot to clean up after.

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -30,6 +30,88 @@
...
+
+def transfer_if_needed(*tensors: Tuple[ShardedTensor, ...]) -> Tuple[ShardedTensor, ...]:
+    if len(tensors) <= 1:
+        return
+    assert all(isinstance(tensor, ShardedTensor) for tensor in tensors)
+    
+    # Check if all tensors are on the same devices.
+    devices_0 = tensors[0].devices
+    for tensor in tensors[1:]:
+        if not all(devices_0[j] == tensor.devices[j] for j in range(len(devices_0))):
```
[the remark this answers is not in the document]
Alex-Vasile: I think I understood what you meant, let me know if the change I made is what you had in mind.

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -30,6 +30,88 @@
...
+    return tuple(new_tensors)
+
+
+def override_w_tranfer(operation, *override_args):
+    def decorator(f):
+        @operation.override(*override_args)
+        def wrapper(*args: Tuple, **kwargs: Dict[str, Any]):
+            args = list(args)
+            t_i = [i for i, arg in enumerate(args) if isinstance(arg, ShardedTensor)]
+            if len(t_i) > 0:
```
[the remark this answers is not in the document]
Alex-Vasile: Good catch. Forgot to change that after redoing transfer_if_needed for arbitrary number of tensors.

sharktank/sharktank/examples/pipeline/export_ppffn_net.py
```diff
@@ -0,0 +1,121 @@
+# Copyright 2024 Advanced Micro Devices, Inc.
```
rsuderman, line 1: Just as a minor note we will want the right year here.

sharktank/sharktank/examples/sharding/export_ffn_net.py
```diff
@@ -30,16 +30,16 @@ def create_theta(
     gate_weight = torch.rand(hidden_dim, primary_dim, dtype=torch.float16)
     up_weight = torch.rand(hidden_dim, primary_dim, dtype=torch.float16)
     sharded_gate_weight = SplitPrimitiveTensor(
-        name="ffn_gate.weight", shard_dim=0, ts=gate_weight.split(split_size, dim=0)
+        name="ffn_gate.weight", shard_dim=0, ts=gate_weight.split(split_size, dim=0), devices=tuple(range(shard_count)), devices_pinned=True
```
rsuderman, line 33: Lets avoid setting device placement information on non pipeline parallel programs. Overall we should be able to default safely without changing things.

Something to consider is that authors for a tensor-parallel program should not need to think about the pipeline parallel behavior. Same as unsharded programs should be able to not think about sharding.

sharktank/sharktank/layers/kv_cache.py
```diff
@@ -98,7 +98,7 @@ def unflatten_page_table(
             shards = [
                 shard.unflatten(1, self.sub_page_dims) for shard in page_slab.shards
             ]
-            return SplitPrimitiveTensor(ts=shards, shard_dim=4)
+            return SplitPrimitiveTensor(ts=shards, shard_dim=4, devices=tuple(range(self.shard_count)), devices_pinned=True)  # TODO
```
rsuderman, line 101: Lets avoid changing these sections and use them as essentially tests that the default path works.

One day I could us representing `ReplicatedTensor` differently. Storing the shards in a map instead of a list. Then replicated tensor interactions would lookup the corresponding replicated shard. Not required in the short term.

sharktank/sharktank/ops/sharded_impls.py
Alex-Vasile: Are you thinking?:
```Python
res = f(*args, **kwargs)
if (
    len(args) == 1 
    and len(t_i) == 1
    and isinstance(res, ShardedTensor)
):
    res.pinned = True
return res
```

sharktank/sharktank/types/tensors.py
```diff
@@ -752,10 +752,29 @@ class ShardedTensor(InferenceTensor):
     """
 
     def __init__(
-        self, *, shape: list[int], shard_dim: int | None, name: str = UnnamedTensorName
+        self, *, 
+        shape: list[int], 
+        shard_dim: int | None, 
+        name: str = UnnamedTensorName,
+        devices: Tuple[int],
```
rsuderman, line 759: Devices should be `Optional[Tuple[int]] = None`. Then when not set we can initialize according to the number of shards.

sharktank/sharktank/types/tensors.py
```diff
@@ -752,10 +752,29 @@ class ShardedTensor(InferenceTensor):
...
+        devices_pinned: bool,
     ):
         super().__init__(name=name, shape=shape)
         self.shard_dim = shard_dim
+        self._devices = devices
+        self._devices_pinned = devices_pinned
```
rsuderman, line 765: We can probably just rename this to `pinned`. Its more succinct and conceptually its basically the only definition is memory pinning.

sharktank/sharktank/types/tensors.py
```diff
@@ -1171,6 +1211,8 @@ def add_to_archive(self, builder: ShardedArchiveBuilder) -> InferenceTensorMetad
             {"": self.name},
             extra_properties={
                 "shard_count": len(self._shards),
+                "devices": self.devices,
```
rsuderman, line 1214: Nope, we shouldn't be putting this in the archive. Pipeline parallelism won't have any importance in comparison to sharding needing to know count/orientation.

sharktank/sharktank/types/tensors.py
```diff
@@ -1190,6 +1232,8 @@ def create(
         extra_properties: dict[str, Any],
     ) -> "InferenceTensor":
         shard_count = int(extra_properties["shard_count"])
+        devices = extra_properties["devices"]
```
rsuderman, line 1235: Same here.

sharktank/sharktank/types/tensors.py
```diff
@@ -1198,13 +1242,12 @@ def create(
...
-            # TODO This should be changed to assigned affinities
-            for i in range(shard_count):
-                DeviceTensorTrait(i).set(ts[i])
+            for i, device in enumerate(devices):
+                DeviceTensorTrait(device).set(ts[i])
 
         except KeyError as e:
             raise IOError(f"Missing component tensor '' in {raw_tensors.keys()}") from e
-        return cls(name=name, ts=ts)
+        return cls(name=name, ts=ts, devices=devices, devices_pinned=devices_pinned)
```
rsuderman, line 1250: We should rely on default behavior and override elsewhere.

sharktank/sharktank/types/tensors.py
```diff
@@ -1255,11 +1300,14 @@ def __init__(
         ts: list[torch.Tensor],
         name: str = UnnamedTensorName,
         shape: Optional[list[int]] = None,
+        devices: List[int],
```
rsuderman, line 1303: See comments about default behavior.

sharktank/tests/export_test.py
```diff
@@ -31,7 +31,7 @@
 
 class ExportTest(TestCase):
     def testFlattenSignature(self):
-        expected_a = [SplitPrimitiveTensor(ts=[torch.tensor([1])], shard_dim=0)]
+        expected_a = [SplitPrimitiveTensor(ts=[torch.tensor([1])], shard_dim=0, devices=(0,), devices_pinned=True)]
```
rsuderman, line 34: Lets make these tests allow for our default behavior. New tests should check that device placement can be explicitly specified.

sharktank/tests/layers/sharded_rotary_embedding_test.py
```diff
@@ -39,8 +39,8 @@ def test_sharded_rotary_table():
     ok = default_layer(xt=xk, start_index=0)
 
     # Then we can shard the same inputs and layer
-    xq = SplitPrimitiveTensor(ts=xq, shard_dim=2, shard_count=4)
-    xk = SplitPrimitiveTensor(ts=xk, shard_dim=2, shard_count=4)
+    xq = SplitPrimitiveTensor(ts=xq, shard_dim=2, shard_count=4, devices=tuple(range(4)), devices_pinned=True)
```
rsuderman, line 42: Ditto on default testing.

sharktank/tests/ops/sharded_test.py
```diff
@@ -25,7 +25,7 @@ def testAllGather(self):
         ]
         expected_result = torch.cat(shards, dim=shard_dim)
 
-        sharded = SplitPrimitiveTensor(shard_dim=shard_dim, ts=shards)
+        sharded = SplitPrimitiveTensor(shard_dim=shard_dim, ts=shards, devices=tuple(range(shard_count)), devices_pinned=False)
```
rsuderman, line 28: Here and below. Make all the existing tests work by default rather than changing their behavior. We want to make PP invisible when possible

rsuderman (changes requested):

sharktank/sharktank/ops/sharded_impls.py
rsuderman: Yest mostly. It should also check whether the input tensor was pinned. The idea is that if we do reshape/transpose on a pinned tensor its almost guaranteed we want the result pinned. We won't want it to pin a value that wasn't pinned though.

Alex-Vasile: Whoops messed up the logic there, yes on the same page about it.

sharktank/sharktank/types/tensors.py
Alex-Vasile: I think the `None` cases for `devices` and `devices_pinned` should be dealt with in `ShardedTensorBase.__init__`.

`ShardedTensor.__init__` is called from it before the shards are set, so we won't have access to the underlying `torch.Tensors` to infer anything.

sharktank/sharktank/types/tensors.py
Alex-Vasile: See discussion on your other point: https://github.com/nod-ai/shark-ai/pull/1008#discussion_r1972422563

sharktank/sharktank/types/tensors.py
Alex-Vasile: If we don't save devices, how should we handle the for-loop setting DeviceTensorTrait below?

sharktank/tests/ops/sharded_test.py
```diff
@@ -667,10 +667,10 @@ def testReplicatedLhsShardedParallelDimRhs(self):
         b = torch.rand(3, 6, dtype=torch.float32)
         shard_count = 3
         unsharded_result = torch.matmul(a, b)
-        expected_result = ops.reshard_split(unsharded_result, dim=2, count=shard_count)
+        expected_result = ops.reshard_split(unsharded_result, dim=2, count=shard_count)  # TODO: How to know this should also not be pinned
         b_sharded = ops.reshard_split(b, dim=1, count=shard_count)
         a_sharded = ops.replicate(a, count=shard_count)
-        actual_result = ops.matmul(a_sharded, b_sharded)
+        actual_result = ops.matmul(a_sharded, b_sharded)  # GOOD: Should NOT be pinned
         assert ops.equal(expected_result, actual_result)
```
Alex-Vasile, line 674: Adding `.pinned` and `.devices` to `is_deep_equals`, which `ops.equals` calls, causes this test to fail.

`actual_result.pinned == False` which is correct. But `expected_result.pinned` would end up being `True` if any heuristic is used to convert the default `None` value into a bool: it's a concrete `torch.Tensor` and would be indistinguishable from one loaded from a file, i.e. a weight.

How to handle this? Should `is_deep_equal` be looking at `.devices` and `.pinned`? I feel like it should given its docstring and current behaviour.

sharktank/sharktank/ops/__init__.py
```diff
@@ -33,3 +33,5 @@
 # Comment this out to completely disable optimized quantized implementations.
 from . import qconv_impls
 from . import qlinear_impls
+
+from .sharded_impls import transfer_if_needed  # TODO: Hack just to get tests running, figure out properly later
```
Alex-Vasile, line 37: TODO

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -30,6 +30,100 @@
 from .shape import broadcast_dims, broadcast_dim, unbroadcast_dim
 from ..utils import longest_equal_range
 
+def copy_w_new_shards_and_devices(tensor: ShardedTensor, new_shards: List[torch.Tensor], new_devices: Tuple[int]) -> ShardedTensor:
+    # TODO: What does transfrom_globals need from this function?
```
Alex-Vasile, line 34: TODO

sharktank/sharktank/ops/signatures.py
```diff
@@ -824,11 +824,19 @@ def replicate(input: AnyTensor, count: int) -> ShardedTensor:
...
+        devices = devices if devices is not None else tuple(range(count))
+        pinned = pinned if pinned is not None else False
+    else:
+        # TODO: Is this correct? Will use data on `input`.
+        assert devices is None
+        assert pinned is None
+    
     for override in d.find_overrides(tensors):
-        result = override(input, count=count)
+        result = override(input, count=count, devices=devices, pinned=pinned)
```
Alex-Vasile, line 839: TODO

sharktank/tests/ops/pipeline_parallelized_test.py
```diff
@@ -0,0 +1,192 @@
...
+        assert isinstance(res_sharded, SplitPrimitiveTensor)
+        assert res_sharded.shard_dim == 1
+        assert res_sharded.shard_count == shard_count
+        for i in range(shard_count):
+            assert res_sharded.devices[i] == a_sharded.devices[i]  # A is pinned, result should be on its device
+        actual_result = ops.sharded_cat(res_sharded)
+        torch.testing.assert_close(actual_result, expected_result)
+
+# TODO: Tests needed
+# 1. Pinned input for unary ops should return a pinned result.
```
Alex-Vasile, line 192: TODO

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -30,6 +30,110 @@
...
+        (
+            transfer_to_logical_device(shard, new_devices[j])
+            if new_devices[j] != tensor.devices[j]
+            else barrier_on_logical_device(shard, new_devices[j])
+        )
+        for j, shard in enumerate(tensor.shards)
+    )
+
+    if isinstance(tensor, SplitPrimitiveTensor ):
+        return SplitPrimitiveTensor(
```
rsuderman, line 49: One thing we may want to do is create a `clone` operation that takes all of the normal constructor args then overrides with the changed ones. This way you could do
```
tensor.clone(ts=new_shards, pinned=False, devices=new_devices)
```
and be guaranteed it just makes a new cloned version with the right devices and pinning

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -102,12 +180,12 @@ def cat_split(
...
         ]
     )
-
+    # TODO: Need to transfer tensors to same devices
     shard_dim = tensors[0].shard_dim
     shard_count = tensors[0].shard_count
     if dim != shard_dim:
         shards = [cat(shards, dim) for shards in zip(*[t.shards for t in tensors])]
-        return SplitPrimitiveTensor(ts=shards, shard_dim=shard_dim)
+        return SplitPrimitiveTensor(ts=shards, shard_dim=shard_dim,devices=tensors[0].devices, pinned=False)
```
rsuderman, line 188: We can remove all of these sets now right? We want the wrapper to set the right device information above.

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -31,6 +31,84 @@
...
+            for i, t in zip(t_i, t_vals):
+                args[i] = t
+            args = tuple(args)
+            res = f(*args, **kwargs)
+            if (
+                len(t_i) == 1
+                and t_i[0].pinned
+                and isinstance(res, ShardedTensor)
+            ):
+                res.pinned = True
```
rsuderman, line 107: We should set device information as well. The idea is the override should set all the additional device information based on the invocation.

sharktank/tests/ops/sharded_test.py
rsuderman: Blah, I can see both cases. We may want to update to include an option to include comparisons of pinning. `ops.equals` is such a special case that I can see us wanting to compare both numerics and placement information. For our tests though we mostly just want to compare numerics and pinning comparison is mostly metadata and not value. If this is the only case we can always just manually compare.

rsuderman (changes requested):

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -31,6 +31,106 @@
...
+            if (
+                len(t_vals) == 1
+                and t_vals[0].pinned
+                and isinstance(res, ShardedTensor)
+            ):
+                res = res.clone(pinned=True)
+            return res
+        
+        if hasattr(f, "override"):  # Needed for ops like .gelu_tanh_approximation
+            wrapper.override = f.override
```
Alex-Vasile, line 70: Should this be in here or applied on the output at the call site?

sharktank/sharktank/ops/sharded_impls.py
Alex-Vasile: We can drop the pinned and let the wrapper set it to whatever the input value was. But devices we can't. The wrapper only transfers the input shards and sets devices on the input, it won't set it on the output. If we don't have it in here, then the `SplitPrimitiveTensor` will default to `devices=tuple(range(len(shard_count))` rather than the location of the pinned tensor in the input.

Although now that I look at cat longer, it should always through an error on the input since we're unlike to ever cat 1 pinned tensor with N unpinned tensors. Would it not be used primarily to cat unpinned tensors?

sharktank/sharktank/ops/__init__.py
```diff
@@ -18,7 +18,75 @@
 
 from . import _registry
 from ..types.tensors import unbox_tensor
-from .signatures import *
+
+def import_and_wrap_signatures():
```
Alex-Vasile, line 22: This ended up having to go in the \_\_init\_\_ file so that any import of ops would wrap the functions.

One problem with the approach: a subsequent run of `from .signatures import *` will override the wrapped versions.

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -31,6 +31,57 @@
 from ..utils import longest_equal_range
 
 
+def transfer_if_needed(*tensors: Tuple[ShardedTensor]) -> List[ShardedTensor]:
```
Alex-Vasile, line 34: I would like to put this in \_\_init\_\_ at well since that is not the only place it's being used. But the top level type hints for this functions will make the imports messy.

sharktank/tests/ops/pipeline_parallelized_test.py
```diff
@@ -0,0 +1,350 @@
...
+        shard_count = 4
+        shard_shape = [3, 4]
+        devices = tuple(2+i for i in range(shard_count))
+        shards = [torch.rand(shard_shape, dtype=torch.float32) for _ in range(shard_count)]
+        pre = SplitPrimitiveTensor(shard_dim=1, ts=shards, devices=devices, pinned=False)
+
+        post = pre.T
+        assert all(d_pre == d_post for d_pre, d_post in zip(pre.devices, post.devices))
+        # TODO: post gets pinned since resulting ShardedTensor is made with torch.Tensor shards which are assumed to always be pinned
+        # assert post.pinned == pre.pinned
```
Alex-Vasile, line 339: This may be an issue

sharktank/sharktank/ops/__init__.py
rsuderman: I am pretty certain we don't want to do this. Only the `sharded_impls` should be the one to consider transfer information. This is pretty specific to sharded impls. We only want to wrap the importers for the sharded cases.

sharktank/tests/ops/sharded_test.py
```diff
@@ -667,10 +667,12 @@ def testReplicatedLhsShardedParallelDimRhs(self):
         b = torch.rand(3, 6, dtype=torch.float32)
         shard_count = 3
         unsharded_result = torch.matmul(a, b)
-        expected_result = ops.reshard_split(unsharded_result, dim=2, count=shard_count)
+        expected_result = ops.reshard_split(
+            unsharded_result, dim=2, count=shard_count
+        )  # TODO: How to know this should also not be pinned
```
rsuderman, line 672: Remove the TODOs here and below

sharktank/sharktank/ops/sharded_impls.py
rsuderman: We want to keep in sharding as all of these wrapping behavior should be sharded specific.

sharktank/sharktank/ops/__init__.py
rsuderman: Also, try to name this `import_sharded_signatures` or something to that effect.

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -26,9 +26,188 @@
...
+
+            args = unflatten_args(args, t_i_args, t_vals[: len(t_vals_args)])
+            kwargs = unflatten_args(kwargs, t_i_kwargs, t_vals[len(t_vals_args) :])
+            res = f(*args, **kwargs)
+            if isinstance(res, ShardedTensor) and len(t_vals) > 0:
+                pinned = (
+                    res.pinned
+                    or (len(t_vals) == 1 and t_vals[0].pinned)
+                    or "_like_" in f.__name__
+                )  # TODO: How to handle reshard_like (and future similar ones properly)
```
Alex-Vasile, line 105: @rsuderman

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -26,9 +26,188 @@
...
+        """
+        new_shards = tuple(
+            (
+                transfer_to_logical_device(shard, new_devices[j])
+                if new_devices[j] != tensor.devices[j]
+                else barrier_on_logical_device(shard, new_devices[j])
+            )
+            for j, shard in enumerate(tensor.shards)
+        )
+        return tensor.clone(ts=new_shards, devices=new_devices)
```
Alex-Vasile, line 179: @rsuderman 

What, if anything, do I need to do about `transfrom_globals` and DeviceTensorTraits in this function or in tensor.clone?

sharktank/sharktank/ops/signatures.py
```diff
@@ -824,11 +826,23 @@ def replicate(input: AnyTensor, count: int) -> ShardedTensor:
...
+        devices = devices if devices is not None else tuple(range(count))
+        pinned = pinned if pinned is not None else False
+    else:
+        # TODO: Is this correct? Will use data on `input`.
+        assert devices is None
+        assert pinned is None
+
     for override in d.find_overrides(tensors):
-        result = override(input, count=count)
+        result = override(input, count=count, devices=devices, pinned=pinned)
```
Alex-Vasile, line 845: @rsuderman

sharktank/sharktank/types/tensors.py
```diff
@@ -752,10 +752,31 @@ class ShardedTensor(InferenceTensor):
...
     ):
         super().__init__(name=name, shape=shape)
         self.shard_dim = shard_dim
+        self._devices = devices
+        self._pinned = pinned
+
+    @abstractmethod
+    def clone(self, **kwargs) -> "ShardedTensor":
+        """Create a clone of this tensor with the given properties overridden."""
+        ...  # TODO: transform_globals, DeviceTensorTrait, etc. What else needs to be copied?
```
Alex-Vasile, line 771: @rsuderman 

Would this be better if I instead modified and made use of `._clone_with_globals`?

sharktank/sharktank/ops/sharded_impls.py
rsuderman: Yeah we will probably have to custom handle reshard like. Its a pretty good example of a transfer based operation so not really feasible to rely on the wrapper.

sharktank/sharktank/ops/__init__.py
```diff
@@ -18,6 +18,8 @@
 
 from . import _registry
 from ..types.tensors import unbox_tensor
+
```
rsuderman, line 21: Delete these extra lines as they are a non functional change

sharktank/sharktank/types/tensors.py
rsuderman: This is probably good enough. We can always come back after and clean up if its not an issue.

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -26,9 +26,188 @@
 )
 from ..types.tensors import unbox_tensor
 from ._registry import AllOfType, AllOfExprsVariadic, IsOfType
-from .signatures import *
 from .shape import broadcast_dims, broadcast_dim, unbroadcast_dim
 from ..utils import longest_equal_range
+from .signatures import *
+
+
+def wrap_override_decorator():
```
rsuderman, line 34: Rename to `sharded_wrap_override` as this is specific to sharding behavior.

sharktank/sharktank/ops/sharded_impls.py
rsuderman: We wouldn't want to set the `DeviceTensorTrait` in these cases. Those should only be manipulated during placement examples, so no worries on clone.

rsuderman (changes requested):
Just a couple nits. Clean these up and make it an official draft.

rsuderman (changes requested):
Just a couple nits. Clean these up and make it an official draft.

sharktank/sharktank/ops/signatures.py
Alex-Vasile: Does not work for tracing sharded llama since incoming tensor is a fake tensor. This behavious may be better handled by doing `ops.replicate(torch.Tensor).clone(pinned=False)` if we wish to override.

sharktank/sharktank/types/tensors.py
```diff
@@ -801,10 +822,19 @@ def __init__(
...
-        super().__init__(name=name, shape=shape, shard_dim=shard_dim)
+        if devices is None:
+            devices = tuple(range(len(ts)))
+        assert len(ts) == len(devices)
+        if pinned is None:
+            pinned = not isinstance(ts[0], torch._subclasses.fake_tensor.FakeTensor)
+        super().__init__(
+            name=name, shape=shape, shard_dim=shard_dim, devices=devices, pinned=pinned
+        )
         self._shards: tuple[DefaultPrimitiveTensor] = tuple(
```
Alex-Vasile, line 838: How come self._shards is being set here rather than inside `ShardedTensor.__init__`? Moving looks like it would simplify things both for where this PR needs to set `.pinned` and in terms of the repetition for the other subclasses.

sharktank/tests/ops/pipeline_parallelized_test.py
rsuderman: Meh, pinning is more for noting information. Comparing this is not really necessary as its not the intention of correctness.

sharktank/sharktank/types/tensors.py
rsuderman: Probably should have been and was overseen before review. Lets avoid fixing it in this PR as its already quite large.

rsuderman (approved):

rsuderman:
You will want to rebase and check the presubmit tests pass before squashing and merging.

sharktank/sharktank/types/tensors.py
Alex-Vasile: I'll open up a separate PR to clean that up.

Alex-Vasile:
> You will want to rebase and check the presubmit tests pass before squashing and merging.

One of the failing tests is because of hugging face logins issues unrelated to this PR. The other has tests failing in CI but not on local. Investigating.

PR #1011, opened 2025-02-27
Shortfin llm beam search

stbaione:
Initial implementation of `beam_search` for the LLM server. Putting it up as a draft for now to get some feedback on the code, and because it needs unit/integration tests. Too big/important of a change to merge without it.

The beam_search specific logic is contained with `beam_manager.py`, while `generate.py` just contains logic for managing the `InferenceExecRequest`s and orchestrating the overall flow.

This also keeps all of the logic above the Batcher level, which minimized changes a lot from what I previously tried.

At a high level, the idea is that we:

1. Run prefill normally. Get the kvcache initialized and obtain first token.
2. Create `n_beams` InferenceExecRequests. Here we `replicate` the initial req that we used for prefill, including replicating the KVCaches pages.
3. Group each of these reqs under a `BeamGroup`. This is a helper class that handles actually performing beam_search token selection, and tracking the reqs.
4. Submit all reqs to batch, and wait for them all to finish.
5. For each req select the `top_k` tokens, based on cumulative log probability.
6. Select the overall `top_k` tokens, from all reqs.
7. Update our beams. Do any replication/beam collapses if needed.
8. If a req generates an `eos_token`, add it to the set of `completed_reqs`.
9. Repeat until either, `max_completion_tokens` is reached, or all beams generated an `eos_token`
10. When we return, we either select the highest cumulative_probability beam, or return all beams, depending on the request params.

## Selecting top-k

Extra attention in this section would be appreciated.

For each beam:

<begin_loop>
1. Obtain logits from decode invocation.
2. Apply `log_softmax` to logits. By taking the log of softmax, we can use addition to track the cumulative probabilities, instead of multiplication with the raw probabilities. If we did multiplication with raw probabilities, our cumulative probabilities will become smaller and smaller, until we lose precision. [source](https://medium.com/@jessica_lopez/understanding-greedy-search-and-beam-search-98c1e3cd821d) (search for 'From products to addition`)
3. Select the `top_k` values and tokens from the `-1` axis. Track the cumulative log probs for each possible token.
<end_loop>

We then return the top possible selections, in sorted order, based on which potential tokens would yield the beams with the highest probability.


shortfin/python/shortfin_apps/llm/components/kvcache/page_pool.py
```diff
@@ -115,6 +115,20 @@ def free_pages(self, pages: list[PageInfo]):
...
+        if dst_pages is None:
+            return None
+
+        for src_page, dst_page in zip(src_pages, dst_pages):
+            for page_table in self.page_tables:
+                # View of source and destination pages
+                src_view = page_table.view(src_page.index)
+                dst_view = page_table.view(dst_page.index)
+                # Copy the data
+                dst_view.copy_from(src_view)
```
rsuderman, line 129: We shouldn't be copying every page but only incomplete pages. I.e. the final page if it is incomplete. The rest should be tracked with the retain count system.

@renxida should be able to point to it

shortfin/python/shortfin_apps/llm/components/beam_manager.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def topk(
+        self, logits: np.array, k: int, axis: int
+    ) -> Tuple[List[float], List[int]]:
+        # TODO: Move this to sfnp.array
+        indices = np.argpartition(logits, -k, axis=axis)
+        topk_indices = indices[axis][-k:]
+        topk_values = logits[axis][topk_indices]
+
+        sorted_indices = np.argsort(topk_values)[::-1]
```
rsuderman, line 30: Most importantly when moving we want to use a topk algorithm and not sort and select topk. This is easy 20k+ elements for most models so even simple top-k implementations will outperform sorting the whole array.

sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -118,6 +118,7 @@ def generate_params_json(
                 "block_seq_stride": llama_config.block_seq_stride,
                 "device_block_count": args.device_block_count,  # so that this makes its way into the config file & can be edited.
             },
+            "n_beams": llama_config.n_beams,
```
rsuderman, line 121: This should be just handled by a default loader. The export `json` is just an example IR, our dataclass loaders should be resilient to missing optional values.

sharktank/sharktank/utils/cli.py
```diff
@@ -127,6 +127,12 @@ def add_model_options(parser: argparse.ArgumentParser):
         type=int,
         default=512,
     )
+    parser.add_argument(
```
rsuderman, line 130: Beam search is a service level control so there is no benefit to including it during export.

shortfin/python/shortfin_apps/llm/components/generate.py
```diff
@@ -68,20 +73,70 @@ async def run(self):
...
+            exec.input_token_ids.append(token_int)
+            if self.n_beams > 1:
+                await self.beam_search_decode_loop(exec)
+            else:
+                await self.greedy_decode_loop(exec)
         finally:
+            logger.info(f"Freeing cache pages: {exec.rid}")
             exec.free_cache_pages()
 
+    async def greedy_decode_loop(self, exec_req: InferenceExecRequest):
```
rsuderman, line 85: These functions should be abstracted into a helper componsite class instead of integrated into generate. The idea should be that you create the class for the type of search you do.

E.g.

```
class DecodeManager():
  def __init__():
    pass
   
   def decode():
      ...
 
class GreedyDecoder():
    def decode():
       ....
       
       
class BeamSearchDecoder():
    def decode():
       ....
```

rsuderman: Essentially we want this class to define how someone would write different sampling methods and we can just dependency inject the different behavior into the generator.

shortfin/python/shortfin_apps/llm/components/messages.py
```diff
@@ -54,6 +62,21 @@ def reset(self, phase: InferencePhase):
...
         self.result_logits = None
 
+    def replicate_self(self) -> "InferenceExecRequest":
+        new_exec_req = InferenceExecRequest(
+            self.phase,
+            copy.deepcopy(self.input_token_ids),
+            self.rid,
+        )
+        new_exec_req.start_position = self.start_position
+        result_logits: sfnp.device_array = self.result_logits.for_transfer()
```
rsuderman, line 72: It concerns me that we see result_logits being retained / surviving. If we do want it to survive this buffer should be held via retainment and not copied.

shortfin/python/shortfin_apps/llm/components/beam_manager.py
```diff
@@ -0,0 +1,149 @@
...
+        for key in sorted_keys[: self.n_beams - len(self.completed_reqs)]:
+            exec_req, token = log_prob_map[key]
+            exec_req.cumulative_log_prob = key
+            exec_req_selections.append(
+                (
+                    key,
+                    *log_prob_map[key],
+                )
+            )
+
```
rsuderman, line 78: Post selecting the K beams you want to continue down you should normalize the running decode. Float addition can accumulate error so if you find the minimum value and subtract that from all continued beams we can continue down the hypothesis train. There will need to be some correction for different length hypothesis to accommodate for them but it just means tracking what the accumulated normalization was.

shortfin/python/shortfin_apps/llm/components/beam_manager.py
stbaione: This avoids sorting the 20k+ elements in the array, unless I missed something somewhere.

I use [argpartition](https://numpy.org/doc/stable/reference/generated/numpy.argpartition.html#numpy.argpartition) to ensure that the `k` last elements of the array are the largest. This runs in O(n) time. It doesn't guarantee that those k last elements are sorted though.
[source](https://stackoverflow.com/questions/6910641/how-do-i-get-indices-of-n-maximum-values-in-a-numpy-array)

From there, I slice off those maximum k indices. Then I do an index view into logits to obtain `topk_values`.

So, `len(topk_values) == k`.

This argsort is just sorting the k top values (i.e. 4 top values), not values of the entire array.

Will double-check the shape in debugger

shortfin/python/shortfin_apps/llm/components/kvcache/page_pool.py
stbaione: Messaged @renxida and asked if he can provide context for the page copying. I think I see what you're saying though. If a page is full, we just need to read from it, not write to it. So, we should be able to share the full pages, which is all except the last one. That last non-full one is where we may actually see differences, so needs to be copied?

shortfin/python/shortfin_apps/llm/components/generate.py
stbaione: Great idea, will make this refactor

shortfin/python/shortfin_apps/llm/components/messages.py
stbaione: Could you elaborate on the issue and diff between retainment and coped? I think there's some background/context that I'm missing

sharktank/sharktank/utils/cli.py
stbaione: Will remove all reference in sharktank

shortfin/python/shortfin_apps/llm/components/beam_manager.py
stbaione: Yeah, double-checked, len(topk_values) == k

shortfin/python/shortfin_apps/llm/components/messages.py
rsuderman: Two queries may have some overlapping sections. E.g. when we initially branch there may be 4 blocks of contents but the first 3 are identical. The idea with the page table is that we do not duplicate the first blocks that match, we just use the same pages. The retainment idea is we track how many queries use each page and only release it once no active queries use it.

shortfin/python/shortfin_apps/llm/components/beam_manager.py
rsuderman: Ah I misundestood. In that case we don't need to both including this sort. You don't have to remove but typically we don't actually care that the concurrent hypothesis are sorted just that we iterate on the right K. Then once we have the final K options we sort post all decode steps and return. Basically no point to sort per-step in this case, we just sort at the end.

shortfin/python/shortfin_apps/llm/components/beam_manager.py
stbaione: Gotcha, so IIUC, for each decode iteration, shift all log_probs to the right by subtracting the min(all_log_probs).

```python
cumulative_log_prob += (log_prob - min(all_log_probs))
```

For each beam, track the total accumulated normalization: 

```python
accumulated_normalization += abs(min(all_log_probs))
```

When a beam reaches eos, add the total accumulated normalization to it's cumulative_log_prob.

After our decode iterations are finished, we apply length normalization to our top beams.

So, our final score for each beam is calculated like this:

```python
final_score = (cumulative_log_prob + accumulated_normalization) / len(beam)
```

Does that make sense, or should I be doing the correction and/or length normalization while I'm in the decode loop?

shortfin/python/shortfin_apps/llm/components/messages.py
renxida: i got pulled off the prefix sharing before i could actually make it work - - right now trie cache will give subsequent requests sharing the same tokens the same pages

but because we don't actually do fast-forwarding, we still start prefill from 0 and generate onwards

we really need:
prefill funciton with a start_position
some changes in generate.py to check with the cache and fast-forward / skip some prefill steps
Before we can really make the cache handle the block copying instead

so i think right now it's good if we just merged stephen's maual page copying and later replaced it with the proper cache

shortfin/python/shortfin_apps/llm/components/beam_manager.py
stbaione: Removed the per-step sort, and moved the sort at the end to its own function

sharktank/sharktank/examples/export_paged_llm_v1.py
stbaione: Removed

sharktank/sharktank/utils/cli.py
stbaione: Removed all references in Sharktank. n_beams is now an argument via server_config, server cli, or request `sampling_params`

stbaione:
Requested updates have been made, aside from the last page copying. Having a little bit of difficulty figuring that out. The rest of the code can be reviewed though. Next steps, if code is approved, last page copying and unit tests.

renxida:
> Requested updates have been made, aside from the last page copying. Having a little bit of difficulty figuring that out. The rest of the code can be reviewed though. Next steps, if code is approved, last page copying and unit tests.

Sounds good. Reviewing code now, honestly i'm happy to get this in even if it copies all the pages, just so we can have unittests for beam search running.

shortfin/python/shortfin_apps/llm/server.py
```diff
@@ -136,6 +136,16 @@ def parse_args(argv):
...
+    parser.add_argument(
+        "--n_beams",
+        type=int,
+        default=1,
+        help="""The number of beams to use during the decode sequence.\n\n
+
+        If `n_beams == 1`, `greedy` decode strategy is used.\n
+        If `n_beams > 1`, `beam_search` decode strategy is used.\n
+        """,
+    )
```
renxida, line 148: Might be a good idea to actually make this

None: greedy
int: beam_search

this way we can test n_beams=1 beam search and make sure it's equiv. to greedy.

shortfin/tests/apps/llm/components/messages_test.py
```diff
@@ -21,17 +21,19 @@ def test_inference_exec_request_repr(mock_void_future):
 
     Patches shortfin.VoidFuture with a mock because we're not running this testcase on a worker thread.
     """
-    req = LlmInferenceExecRequest(InferencePhase.PREFILL, [1, 2, 3, 4], rid="test123")
+    req = InferenceExecRequest(InferencePhase.PREFILL, [1, 2, 3, 4], rid="test123")
```
renxida, line 24: I think kyle's refactoring renamed InferenceExecRequest to LlmInferenceExecRequest

shortfin/python/shortfin_apps/llm/components/kvcache/page_pool.py
```diff
@@ -115,6 +115,21 @@ def free_pages(self, pages: list[PageInfo]):
         with self._lock:
             self.available_pages.extend(pages)
 
+    def copy_pages(self, src_pages: List[PageInfo]) -> List[PageInfo]:
```
renxida, line 118: Let's maybe call this temp_copy_pages and add a docstring / comment to mark this as a temporary measure & not performant.

renxida (commented):
lgtm.  minor nits to clean up, otherwise i'm happy if we test & merge this.

stbaione:
Old and dropped in favor of #1161

PR #1012, opened 2025-02-27
Refactor the SystemManager across shortfin apps

KyleHerndon:


shortfin/python/shortfin_apps/utils.py
```diff
@@ -4,11 +4,69 @@
...
+from shortfin.interop.support.device_setup import get_selected_devices
+
+
+class SystemManager:
+    def __init__(
+        self,
+        device="local-task",
+        device_ids=None,
+        async_allocs=True,
+        amdgpu_allocators=None,
```
sogartar, line 21: Could you add a type hint?

sogartar, line 21: Could you add type hints?

sogartar (approved):

PR #1013, opened 2025-02-27
[sharktank] Add common model config, export and compile

sogartar:
We don't have a standard way to configure, export and compile sharktank models.

Here is introduced such mechanism and for demonstration the CLIP text model is refactored to utilize this new approach.

`config.json`:
```
{
    "model_type": "my_package.MyModel",
    "mlir_path": "model.mlir",
    "parameters_path": "model.irpa",
    "iree_module_path": "model.vmfb",
    "compile_args": ["--iree-hal-target-device=local"],
    "export_functions": [
        {
            "function": "forward",
            "batch_sizes": [1, 2, 3]
        }
    ]
    "my_model_arg": 5,
}
```

usage
```
model = create_model("config.json")
model.export()
model.compile()
```

### DIFF
--- a/sharktank/sharktank/layers/base.py
+++ b/sharktank/sharktank/layers/base.py
@@ -4,20 +4,28 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-from typing import Dict, Optional
+from typing import Any, Dict, Optional
 from collections import OrderedDict
 from collections.abc import Mapping
+from abc import ABCMeta
 import torch
 import torch.nn as nn
+from os import PathLike
+import logging
 
-from ..types import InferenceTensor, Theta, AnyTensor
+from ..types import InferenceTensor, Theta, AnyTensor, Dataset
 from ..utils import debugging
-from .. import ops
+from .configs import ModelConfig, ExportFunctionConfig, DynamicBatchSize
 
 __all__ = [
     "BaseLayer",
     "ThetaLayer",
+    "create_model",
+    "get_model_type_id",
+    "model_registry",
 ]
+
+logger = logging.getLogger(__name__)
 
 
 def _set_recursively_submodules_default_trace_tensor_key_prefix(
@@ -33,12 +41,83 @@ def _set_recursively_submodules_default_trace_tensor_key_prefix(
         )
 
 
-class BaseLayer(nn.Module):
+def get_model_type_id(model_type: type["BaseLayer"]) -> str:
+    """Get a string representation of the model type."""
+    return f"{model_type.__module__}.{model_type.__name__}"
+
+
+def create_model(config: PathLike | ModelConfig | Mapping[str, Any], /) -> "BaseLayer":
+    """
+    Create model from a configuration.
+    Example
+
+    config.json:
+    ```
+    {
+        "config_version": "0.1.0",
+        "model_type": "MyModel",
+        "mlir_path": "model.mlir",
+        "export_parameters_path": "model.irpa",
+        "iree_module_path": "model.vmfb",
+        "compile_args": ["--iree-hal-target-device=local"],
+        "export_functions": [
+            {
+                "function": "forward",
+                "batch_sizes": [1, 2, 3]
+            }
+        ]
+    ]
+    }
+    ```
+
+    usage
+    ```
+    model = create_model("config.json")
+    model.export()
+    model.compile()
+    ```
+    """
+    if isinstance(config, Mapping):
+        config = ModelConfig.create(**config)
+    elif not isinstance(config, ModelConfig):
+        config = ModelConfig.load(config)
+
+    return config.model_type.from_config(config)
+
+
+model_registry: dict[str, type["BaseLayer"]] = {}
+"""Registry of all model types.
+This is used to dispatch when construction a model form a config."""
+
+
+class BaseLayerMetaClass(ABCMeta):
+    def __init__(cls, clsname, bases, methods):
+        super().__init__(clsname, bases, methods)
+        model_registry[get_model_type_id(cls)] = cls
+
+
+class BaseLayer(nn.Module, metaclass=BaseLayerMetaClass):
     """Base class of all of our layers."""
 
-    def __init__(self):
+    def __init__(self, config: ModelConfig | None = None):
         super().__init__()
         self._trace_tensor_key_prefix = ""
+        self.config = config
+
+        # Can be overridden is derived classes.
+        self.default_export_function = "forward"
+        self.default_export_batch_sizes = [1]
+
+    @classmethod
+    def from_config(cls, config: ModelConfig, /) -> "BaseLayer":
+        """Create a model from config.
+        Override in derived classes if any special behavior is desired."""
+        return cls(config=config)
+
+    @classmethod
+    def config_type(cls) -> type[ModelConfig]:
+        """Return the type of the config for this model."""
+        raise NotImplementedError()
 
     def set_recursively_submodules_default_trace_tensor_key_prefix(self):
         """All submodules get a trace key prefix that reflects their nesting with
@@ -107,8 +186,8 @@ class BaseLayer(nn.Module):
                     raise AssertionError(f"Tensor contains nans! {t}")
 
     def sample_inputs(
-        self, batch_size: int = 1, function: Optional[str] = None
-    ) -> tuple[tuple[AnyTensor], OrderedDict[str, AnyTensor]]:
+        self, batch_size: int | None = 1, function: Optional[str] = None
+    ) -> tuple[tuple[AnyTensor, ...], OrderedDict[str, AnyTensor]]:
         """Return sample inputs that can be used to run the function from the model.
         If function is None then layer is treated as the callable.
         E.g.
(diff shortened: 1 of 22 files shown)


sharktank/sharktank/layers/configs/config.py
ScottTodd, line 1: I'll need to set up some time aside to review this in detail. Can you tag related issues on this PR and connect this back to the overall goals? Bonus points if you could also comment on the issues and help coordinate their full resolution. Cross-cutting work like this goes more smoothly if the plans are communicated ahead of time, before jumping straight to code.

These issues appear related:

* https://github.com/nod-ai/shark-ai/issues/245
* https://github.com/nod-ai/shark-ai/issues/402
* https://github.com/nod-ai/shark-ai/issues/691

sogartar: Yes the are related issues.

sogartar:
This PR require some modification to make it more interoperable with IREE build pipeline tooling.

renxida:
I have some py functions / classes for managing LLMs in `app_tests/integration_tests/llm/model_management.py`.

I like the class / function structure currently in there, but they use a very hacky bash-script-in-python-script kind of solution to call into sharktank and IREE to shard / export / compile models.

Would be very cool if I could switch to using this instead.

sharktank/sharktank/layers/base.py
```diff
@@ -33,12 +41,82 @@ def _set_recursively_submodules_default_trace_tensor_key_prefix(
         )
 
 
-class BaseLayer(nn.Module):
+def get_model_type_id(model_type: type["BaseLayer"]) -> str:
+    """Get a string representation of the model type."""
+    return f"{model_type.__module__}.{model_type.__name__}"
+
+
+def create_model(config: PathLike | ModelConfig | Mapping[str, Any], /) -> "BaseLayer":
```
ScottTodd, line 49: Is `sharktank/layers/base.py` the right spot for these new `create_model` and `model_registry` symbols? 

Alternatives to consider:

* https://github.com/nod-ai/shark-ai/tree/main/sharktank/sharktank/models
* https://github.com/nod-ai/shark-ai/tree/main/sharktank/sharktank/tools
* https://github.com/nod-ai/shark-ai/tree/main/sharktank/sharktank/utils

I would expect a hierarchy of

1. Ops
2. Kernels
3. Layers
4. Models
5. Pipelines

Where this tooling would fit better under "Models" or "Pipelines".

sharktank/sharktank/layers/base.py
```diff
@@ -121,15 +199,159 @@ def sample_inputs(
...
+        if self.config.export_functions is not None:
+            export_functions = self.config.export_functions
+
+        function_batch_size_pairs = {
+            export_function.function
+            or self.default_export_function: export_function.batch_sizes
+            or self.default_export_batch_sizes
+            for export_function in export_functions
+        }
+        from ..export import export_model_mlir
```
ScottTodd, line 235: Relating to my previous comment, https://github.com/nod-ai/shark-ai/blob/main/sharktank/sharktank/export.py is sort of orphaned outside of a subpackage at the moment. Is that intentional? Can we find it a better home too?

sharktank/sharktank/layers/base.py
```diff
@@ -121,15 +199,159 @@ def sample_inputs(
...
+        if theta is None:
+            theta = self.generate_random_theta()
         self.theta = theta
 
     def theta_tensor(self, name: str) -> InferenceTensor:
         # TODO: We may need to do some bookkeeping here to ensure export
         # tracks all of these.
         return self.theta.tensor(name)
+
+    def shard_theta(self, theta: Theta) -> Theta:
```
ScottTodd, line 281: Are these new `shard_theta` and `load_theta` functions necessary for the "common model config, export and compile"? Could this change be split into a few smaller PRs?

Ah, I guess they use `ModelConfig`. Can we separate that out so `ThetaLayer` doesn't need to know about the levels of the hierarchy above it? Flip the flow of information there: Models _tell_ layers what to do, rather than have layers _ask_ a model config what they should do.

You could, for example, have a "layer config" and then have the models forward their "model config" settings into the layer configs. I might not be seeing all the details at the moment though. is sharding really done per "layer", or is it really per "parameter group"? I guess the "theta layer" concept (https://github.com/nod-ai/shark-ai/blob/main/docs/programming_guide.md#core-concepts) is a bit different from what I'm imagining from other systems.

sharktank/sharktank/layers/base.py
```diff
@@ -33,12 +41,82 @@ def _set_recursively_submodules_default_trace_tensor_key_prefix(
...
+
+
+def create_model(config: PathLike | ModelConfig | Mapping[str, Any], /) -> "BaseLayer":
+    """
+    Create model from a configuration.
+    Example
+
+    config.json:
+    ```
+    {
```
ScottTodd, line 56: We may want to version these configs somehow. We should at least think through forwards/backwards compatibility a bit (could be "only supported within the same version").

sharktank/sharktank/layers/base.py
```diff
@@ -33,12 +41,82 @@ def _set_recursively_submodules_default_trace_tensor_key_prefix(
...
+        "model_type": "MyModel",
+        "mlir_path": "model.mlir",
+        "export_parameters_path": "model.irpa",
+        "iree_module_path": "model.vmfb",
```
ScottTodd, line 60: I presume these paths are relative to the JSON file, or they could be absolute?

edit: I see some notes in another file. If these files reach broader adoption it would be worth centralizing the documentation and linking to that one location.

sharktank/tests/models/clip/clip_test.py
```diff
@@ -165,23 +161,31 @@ def runTestCompareIreeAgainstTorchEagerWithInputTokens(
...
-            target_iree_parameters_output_path=parameters_path,
+            extra_config_kwargs={
+                "config_path": None,
+                "export_functions": export_functions,
+                "mlir_path": mlir_path,
+                "export_parameters_path": parameters_path,
+                "iree_module_path": iree_module_path,
+                "compile_args": [
+                    "--iree-hal-target-device=hip",
+                    "--iree-hip-target=gfx942",
```
ScottTodd, line 179: This is just moved from elsewhere in the file, but it would be good to get this target from the environment somehow instead of hardcoding it. These tests should be runnable on other GPUs (and CPU, even if slow).

ScottTodd (commented):
Found some time to make a review pass. I like the direction this is going. Would be great to get reviews from other developers who are more actively working in this project.

sharktank/sharktank/layers/base.py
KyleHerndon: I agree that this seems like it'd fit better somewhere under the models directory. 

Tangent: ops and kernels are actually adjacent in the hierarchy, and that probably means we should rename that directory/do some reorganizing.

KyleHerndon:
This is great work! I need to take some time to do a deeper dive into the inner workings of this, but I also love the direction of this.

One concern I have is that I did some working with the iree build action library system for shortfin, and I found it very clunky and bad at reporting errors (Normal multiprocess/multithreading issues). Do we know how actively maintained/developed the Build stuff is? If it is not been actively worked on, I think we should at least consider alternatives.

sharktank/sharktank/layers/configs/config.py
```diff
@@ -0,0 +1,253 @@
...
+    """Export MLIR to this path."""
+    iree_module_path: Path | None = None
+
+    parameters_path: Path | None = None
+    """Load parameters from this path. IRPA, GGUF, etc."""
+    export_parameters_path: Path | None = None
+    """Location of parameters when exporting. IRPA, GGUF, etc."""
+
+    hugging_face_repo_id: str | None = None
+    hugging_face_revision: str | None = None
```
IanNod, line 85: May need an hf subfolder for specific models/cases

sharktank/sharktank/layers/configs/config.py
```diff
@@ -0,0 +1,253 @@
...
+            self.export_functions = [
+                export_function
+                if isinstance(export_function, ExportFunctionConfig)
+                else ExportFunctionConfig(**export_function)
+                for export_function in self.export_functions
+            ]
+
+    @classmethod
+    def create(cls, model_type: type["BaseLayer"] | str, **kwargs) -> "ModelConfig":
+        """Crate a config with type associated with the model_type."""
```
IanNod, line 119: create instead of crate

sharktank/sharktank/layers/base.py
```diff
@@ -33,12 +41,82 @@ def _set_recursively_submodules_default_trace_tensor_key_prefix(
...
+
+def create_model(config: PathLike | ModelConfig | Mapping[str, Any], /) -> "BaseLayer":
+    """
+    Create model from a configuration.
+    Example
+
+    config.json:
+    ```
+    {
+        "model_type": "MyModel",
```
IanNod, line 57: Is model_type just a colloquial name for the general model like "llama" or "unet"?

sharktank/sharktank/layers/configs/config.py
```diff
@@ -0,0 +1,253 @@
...
+            )
+
+        if parameters_path is not None:
+            config_dict_from_parameters_file = load_properties(
+                Path(config_path) / config_dict["parameters_path"]
+            )
+            config_dict_from_parameters_file.update(config_dict)
+            config_dict = config_dict_from_parameters_file
+            if "SHARK_DATASET_VERSION" in config_dict:
+                config_dict.pop("SHARK_DATASET_VERSION")
```
IanNod, line 145: Any particular reason we don't want to keep SHARK_DATASET_VERSION?

sharktank/sharktank/models/clip/clip.py
```diff
@@ -13,6 +13,7 @@
 import torch
 from torch import nn
 import transformers
+from os import PathLike
```
IanNod, line 16: I don't see PathLike used here

ScottTodd:
> One concern I have is that I did some working with the iree build action library system for shortfin, and I found it very clunky and bad at reporting errors (Normal multiprocess/multithreading issues). Do we know how actively maintained/developed the Build stuff is? If it is not been actively worked on, I think we should at least consider alternatives.

It was built for projects like shark-ai to use, particularly for pipelines involving multiple submodels like SDXL. See https://github.com/nod-ai/shark-ai/issues/402#issuecomment-2450819988:
> Hang tight for a bit. More tooling is coming that will make this all one command. Building it out for sdxl first.

It's roughly as "actively worked on" as any of the alternatives and we should be using it and investing in it more.

There are some API docs at https://iree-python-api.readthedocs.io/en/latest/compiler/build.html already (tracking issue for that: https://github.com/iree-org/iree/issues/19019), but more examples/docs/etc. would be useful.

sharktank/sharktank/layers/base.py
sogartar: Right now the default behavior is that `model_type` would be the fully qualified python class name. E.g. `sharktank.models.llama.PagedLlamaModelV1`. There is an auto registration mechanism that we can potentially override so that we have other names. The name would have to correspond to a python object/class. There we dispatch the model creation interface calls.

sharktank/sharktank/layers/base.py
sogartar: They can be relative to the config file directory or be absolute. There is doc describing this in the `ModelConfig` class.

sharktank/sharktank/layers/base.py
sogartar: I guess this ties to the confusion between module, layer and model. The point of this method is to have a common entry point for model theta sharding.
Potentially layers could have configs and know how to shard their own theta. Right now there are theta layer sharding specs for [common layers](https://github.com/nod-ai/shark-ai/blob/4de14dfc9f8458f5bcd425fb223c97c5c9849117/sharktank/sharktank/types/sharding.py#L76-L209), [llama](https://github.com/nod-ai/shark-ai/blob/4de14dfc9f8458f5bcd425fb223c97c5c9849117/sharktank/sharktank/models/llama/sharding.py
) and some [punet layers](https://github.com/nod-ai/shark-ai/blob/4de14dfc9f8458f5bcd425fb223c97c5c9849117/sharktank/sharktank/models/punet/sharding.py). They live parallel to layers and layers do not know about them.
Some models may have additional config parameters aside from `tensor_parallelism`. For example if we have competing sharding hypothesis of what would constitute a good sharding for a particular model.
At some point we would have to reconcile tensor parallelism sharding with pipeline parallelism sharding.

sharktank/sharktank/layers/configs/config.py
sogartar: We assume that the dataset metadata contains config parameters, which are loaded with lower precedence than the fields in this config file. Meaning, they will be overwritten with fields from the main config file if such exists.
Here we are not loading the dataset itself just the metadata. `SHARK_DATASET_VERSION` may be important when loading the parameters file. I am actually not sure if the dataset loading mechanism should return a clean properties without being polluted by the `SHARK_DATASET_VERSION` field. It makes sense that the loading mechanism would have handled any backwards compatibility and present a current object. If this is not done there the rest of the code would have to deal with the different dataset formats.
Do you think we should always remove `SHARK_DATASET_VERSION` when loading the dataset, so that
```
d = Dataset(properties=..., ...)
d.save(path)
d2 = Dataset.load(path)
assert d.properties == d2.properties
```
holds?

sharktank/sharktank/layers/base.py
sogartar: `load_theta` is used by the clip model. `shard_theta` is not used as there is only a default trivial implementation. It is there to specify the interface.

sharktank/sharktank/layers/base.py
sogartar: I added versioning.

sharktank/sharktank/layers/configs/config.py
sogartar: Thank you for the suggestion. I added it.

sharktank/sharktank/layers/configs/config.py
sogartar: Done.

sharktank/sharktank/models/clip/clip.py
sogartar: Removed.

sharktank/sharktank/layers/base.py
```diff
@@ -121,15 +200,159 @@ def sample_inputs(
...
+    def export(
+        self,
+        mlir_path: PathLike | None = None,
+        parameters_path: PathLike | None = None,
+        /,
+        *args,
+        **kwargs,
+    ):
+        super().export(mlir_path)
+        self.export_parameters(parameters_path)
```
archana-ramalingam, line 358: Do we have to `self.export_parameters()` by default, or only when `self.config.hugging_face_repo_id is not None`?

sharktank/tests/models/clip/clip_test.py
```diff
@@ -192,7 +199,7 @@ def runTestCompareIreeAgainstTorchEagerWithInputTokens(
         )
         expected_outputs = flatten_for_iree_signature(reference_result_dict)
 
-        iree_devices = get_iree_devices(driver="hip", device_count=1)
+        iree_devices = [iree.runtime.get_device(self.iree_device)]
```
archana-ramalingam, line 202: Can we use `get_iree_devices()` from `sharktank/sharktank/utils/iree.py` here?

sharktank/sharktank/layers/base.py
sogartar: I fixed it such that if no export path is specified either as argument or in config, then no parameters export occurs.

sharktank/tests/models/clip/clip_test.py
sogartar: I expanded `get_iree_devices` to accept also a specific device, not just driver name.

sharktank/sharktank/models/clip/testing.py
```diff
@@ -88,11 +83,8 @@ def export_clip_toy_text_model_iree_test_data(
     expected_last_hidden_state_output_path: PathLike,
 ):
     reference_config = clip_toy_text_model_config(reference_dtype)
-    input_ids = make_random_input_token_sequences(
-        batch_size=batch_size, config=reference_config
-    )
-    reference_theta = make_clip_text_model_random_theta(reference_config)
-    reference_model = ClipTextModel(theta=reference_theta, config=reference_config)
+    reference_model = ClipTextModel(config=reference_config)
+    input_ids = reference_model.sample_inputs(batch_size=batch_size)[1]["input_ids"]
```
monorimet, line 87: Will this be the canonical method of supplying input arguments, by utilizing model class attributes? I imagine it would be more flexible to allow the model config to define input shapes/dtypes, but that could be a follow-up patch if we want such functionality.

monorimet: I see that your example has `"my_model_arg": 5,` which I assume is what I was asking about. Disregard

sogartar: I view the input shape as a notion derived from other higher level config arguments. That is why I added the `sample_inputs` method. With it you can interrogate the model about the shape as well.
I was thinking that it should also provide a plausible sample from the expected input distribution, but this is probably too big of an ask. At least it should provide a random sample.

sogartar:
> Do we know how actively maintained/developed the Build stuff is?

@KyleHerndon, it seems that it `iree.build` needs more auto tests. I am especially concerned about potential concurrency problems as they are hard to debug.

sharktank/sharktank/layers/base.py
monorimet: I think this would make the most sense for a hierarchy:

1. ops
2. kernels
3. modules
    1. layers
    2. models
    3. (hypothetical) compiled pipelines -- in the event we do something like CompiledSDXLPipeline where "models" are assembled by invoking their entrypoints from within a MLIR "pipeline" -- will likely be more relevant with deployment outside of Python
4. pipelines

I think the refactor into "module" base-classing should be a separate (follow-up) PR.

monorimet (approved):
Thanks for this @sogartar.  I think this is a great start for canonicalizing sharktank model exports. Can you file or link to an issue that addresses module/model/layer naming convention concerns? Otherwise, LGTM

PR #1023, opened 2025-03-03
Rewriter `write_timestep` to fuse the indexing writes

rsuderman:
It is possible to fuse the timestep writing with the index computation if we manually compute the offset. This reduces the scatter write to a single dispatch avoiding the elementwise write operations.


sharktank/sharktank/layers/kv_cache.py
```diff
@@ -241,6 +241,7 @@ def write_timestep(
         """
         device = self.device
         page_table = self.unflatten_page_table(state)  # 6D
+        page_table = page_table.flatten(0, 3)
```
dan-garvey, line 244: kinda nitty but if we are unflattening and then reflattening would it be better to just change unflatten page table to give us the desired container?

dan-garvey (approved):
Is this a pattern that the compiler could recognize? I feel like there is some kind of exploitable pattern here

sharktank/sharktank/layers/kv_cache.py
rsuderman: Maybe? I thought about that as well but I view it as expanding the total page behavior and removing the ones we don't want is sufficient. Its more a helper so no one needs to know the layout directly. Overall I am biasing towards simpler interfaces with each writer/reader handling the weirdness.

PR #1024, opened 2025-03-03
[types][shortfin_apps] add missing annotations

bjacobgordon:

PR #1025, opened 2025-03-03
feature(shortfin_apps): move png construction to server side

bjacobgordon:


bjacobgordon:
Stacked on #1024

monorimet (changes requested):
Generally looks good to me except for one thing:
If we're not sending string-encoded image data anymore, we don't need to use base64 at all.
Let's fix the postprocessing here and get rid of base64 since we aren't string-encoding images to get it through fastapi anymore: https://github.com/nod-ai/shark-ai/blob/main/shortfin/python/shortfin_apps/sd/components/service.py#L483-L486

bjacobgordon:
@monorimet Gotcha, I've removed the need for the base64 encoding in the `_postprocess` hook! How's it look now?

monorimet (approved):
LGTM. So I can verify I understand this correctly -- we're still sending data as a string-encoded base64 image, but with this we have a class `Base64CharacterEncodedByteSequence` that handles this response data conversion on serverside, and optionally on clientside?

Overall this looks better than before, so I'm happy to merge as-is. Thanks

bjacobgordon:
> So[, ]we're still sending data as a string-encoded base64 image, but with this we have a class `Base64CharacterEncodedByteSequence` that handles this response data conversion on serverside, and optionally on clientside?

Yup, exactly! It matches StabilityAI's API v1 for SDXL, namely responding with a png image in this exact encoding.

That's one of the three major ideas applied in this PR:
1. formalize the idea of "a byte sequence encoded as base64 characters" by defining it as a `str` subclass
    - helps capture and validate existing usage
2. move the string encoding concerns out of the implementation for inference and into that of the server implementation
      - lets us make all the encoding decisions in one place
3. move the png decoding concerns out of the implementation of the client and into that of the server
      - lets us see the encoding decisions that get reverted
      - avoids the overhead of sending an uncompressed image over the web

### Before:
1. image is constructed
2. image is encoded in rgb

(inference/server boundary)

3. encoding loaded into response

(server/web boundary)

4. encoding crosses the web (rgb is really heavy)

(web/client boundary)

5. rgb encoding converted to png encoding with custom implementation (not common in JS)
6. png encoding consumed directly to display image in UI

### After:
1. image is constructed

(inference/server boundary)

2. image is encoded in png
3. encoding loaded into response

(server/web boundary)

4. encoding crosses the web (png is much lighter)

(web/client boundary)

5. png encoding consumed directly to display image in UI

PR #1028, opened 2025-03-04
Resolve CPU llm smoke / integration test hangs by temporarily removing 8-request testcases

renxida:
Had to do this because the cpu integration tests were flaking out.

Also moves the cpu smoke test to standard github runner `azure-cpubuilder-linux-scale` because it's small enough mem-wise.

Issue created to add these back after we fix the problem in #1030

### DIFF
--- a/.github/workflows/pkgci_shark_ai.yml
+++ b/.github/workflows/pkgci_shark_ai.yml
@@ -28,7 +28,7 @@ jobs:
       matrix:
         include:
           - name: cpu
-            runs-on: ubuntu-latest
+            runs-on: ubuntu-24.04
             test_device: cpu
             python-version: 3.11
           - name: amdgpu_rocm_mi300_gfx942
--- a/.github/workflows/pkgci_shark_ai.yml
+++ b/.github/workflows/pkgci_shark_ai.yml
@@ -28,7 +28,7 @@ jobs:
       matrix:
         include:
           - name: cpu
-            runs-on: azure-cpubuilder-linux-scale
+            runs-on: ubuntu-latest
             test_device: cpu
             python-version: 3.11
           - name: amdgpu_rocm_mi300_gfx942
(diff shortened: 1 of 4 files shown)


renxida:
@ScottTodd would like your eye on this for caution's sake since this removes test coverage. Issue already created to track & add it back in.

.github/workflows/pkgci_shark_ai.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       matrix:
         include:
           - name: cpu
-            runs-on: azure-cpubuilder-linux-scale
+            runs-on: ubuntu-latest
```
ScottTodd, line 31: Pin dependencies, including runner types. Use `ubuntu-24.04` here instead of `ubuntu-latest`

app_tests/integration_tests/llm/shortfin/open_llama_3b_llm_server_test.py
```diff
@@ -64,7 +64,7 @@ def test_basic_generation_input_ids(
                 message=f"Generation did not match expected pattern.\nExpected to start with: {expected_prefix}\nActual response: {response}",
             )
 
-    @pytest.mark.parametrize("concurrent_requests", [2, 4, 8])
+    @pytest.mark.parametrize("concurrent_requests", [2, 4])
```
ScottTodd, line 67: We probably also want some soak tests (https://en.wikipedia.org/wiki/Soak_testing), e.g. run for 30 minutes with a variable number of requests coming in throughout that period. These tests are unlikely to catch memory leaks or concurrency issues.

I'd check what vLLM, SGLang, and llama.cpp do too.

ScottTodd (approved):

app_tests/integration_tests/llm/shortfin/open_llama_3b_llm_server_test.py
renxida: #1044

PR #1029, opened 2025-03-04
[Shortfin][SD] Error out on invalid ports

kuhar:
Check that specified ports are valid. Otherwise the server and the client will hang forever without any errors or warnings.

### DIFF
--- a/shortfin/python/shortfin_apps/sd/server.py
+++ b/shortfin/python/shortfin_apps/sd/server.py
@@ -275,6 +275,16 @@ def get_modules(args, model_config, flagfile, td_spec):
     return vmfbs, params
 
 
+def is_port_valid(port):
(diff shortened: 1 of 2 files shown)


shortfin/python/shortfin_apps/sd/server.py
```diff
@@ -403,6 +413,9 @@ def main(argv, log_config=UVICORN_LOG_CONFIG):
         help="Use tunings for attention and matmul ops. 0 to disable.",
     )
     args = parser.parse_args(argv)
+    if not is_port_valid(args.port):
+        exit(3)
```
rsuderman, line 417: Getting flashbacks of "error 37" from a couple years ago but this is good enough for now.

rsuderman (approved):

PR #1031, opened 2025-03-04
Reduce code duplication for ci-llama tests

erman-gurses:
Currently it is a trial using `action.yml`


.github/actions/test-setup/action.yml
```diff
@@ -0,0 +1,38 @@
...
+runs:
+  using: "composite"
+  steps:
+    - name: Get Current Date
+      id: date
+      shell: bash
+      run: echo "date=$(date +'%Y-%m-%d')" >> $GITHUB_ENV
+
+    - name: Set up Python
+      uses: actions/setup-python@v5.4.0
```
marbre, line 22:
```suggestion
      uses: actions/setup-python@42375524e23c412d93fb67b49958b491fce71c38 # v5.4.0
```
We're pinned with hashes as suggested by OpenSSF Scorecard, see https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies.

.github/workflows/ci-llama-tests.yaml
```diff
@@ -0,0 +1,79 @@
...
+        run: |
+          source ${{ github.workspace }}/.venv/bin/activate
+          pytest sharktank/tests/models/llama/benchmark_amdgpu_test.py -v -s \
+            --run-nightly-llama-tests --iree-hip-target=gfx942 --iree-device=hip://0 \
+            --html=out/llm/llama/benchmark/index.html
+
+      - name: Deploy to GitHub Pages
+        uses: peaceiris/actions-gh-pages@v4.0.0
+        with:
+          github_token: ${{ secrets.SHARK_PLATFORM_GH_TOKEN }}
```
marbre, line 70: As you don't have access to the secret from a fork your workflow will fail. Maybe comment this out and enable if / after landing.

.github/actions/test-setup/action.yml
```diff
@@ -0,0 +1,38 @@
...
+      with:
+        python-version: ${{ inputs.python-version }}
+
+    - name: Create Python Virtual Environment
+      shell: bash
+      run: python -m venv ${{ inputs.venv-dir }}
+
+    - name: Install Dependencies
```
marbre, line 30: You might want to cache those.

marbre (commented):
Some first comments.

.github/actions/test-setup/action.yml
erman-gurses: Could you please elaborate little more with examples?

renxida:
The tests are currently failing but this should be good to merge if it's the same failure as on main. LGTM but I think you should get @ScottTodd 's review as well.

This might come in as a separate PR, but might be a good idea to eventually merge this action with 

https://github.com/nod-ai/shark-ai/blob/main/.github/actions/pkgci-setup/action.yml

and make the different behavior between the two depend on an option
such that if we e.g. add caching / migrate to UV, changing one place suffices.

ScottTodd (changes requested):

.github/workflows/ci-llama-tests.yaml
```diff
@@ -0,0 +1,79 @@
+name: "Llama Benchmarking Tests"
```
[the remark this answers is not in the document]
erman-gurses: That is my exact plan.

renxida:
Got a question. Can this be pkgci rather than CI?

So that we only build the packages once per PR?

ScottTodd:
> Got a question. Can this be pkgci rather than CI?
> 
> So that we only build the packages once per PR?

Yes. All of my suggestions in https://github.com/nod-ai/shark-ai/pull/1031#discussion_r1981798392 involve moving into pkgci.

PR #1033, opened 2025-03-04
[sharktank] Add exporting of sharded toy sized Resnet block IREE test data

sogartar:
Export test data to add a test in the IREE test suite.

Refactor the test in sharktank to utilize the shared export function. Make the test not xfail due to numerical precision as the error is around 1e-5 for f32 which should be good enough.

2 other tests required adjusting the numerical tolerance due to a change in random theta generation such that it does not generate in range [0, 1], but in range [-1, 1].

### DIFF
--- a/sharktank/tests/models/punet/sharded_resnet_block_with_iree_test.py
+++ b/sharktank/tests/models/punet/sharded_resnet_block_with_iree_test.py
@@ -6,25 +6,28 @@
 
 import pytest
 import sys
-
+import iree.runtime
+from typing import List
 from pathlib import Path
 import tempfile
-
 import torch
+import iree.compiler
+import gc
 
-
-from iree.turbine import aot
-from sharktank.models.punet.testing import make_resnet_block_2d_theta
-from sharktank.models.punet.layers import ResnetBlock2D
-from sharktank.models.punet.sharding import ResnetBlock2DSplitOutputChannelsSharding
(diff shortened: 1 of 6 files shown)


sogartar:
The failing test `test_toy_size_sharded_resnet_block_with_iree` passes with IREE version [6ecafc10149a0ee5fa7bcba171dd3774211b7577](https://github.com/iree-org/iree/commit/6ecafc10149a0ee5fa7bcba171dd3774211b7577).
This change would need to wait for an IREE bump.

sharktank/tests/models/punet/sharded_resnet_block_with_iree_test.py
```diff
@@ -6,273 +6,127 @@
 
 import pytest
 import sys
-
+import iree.runtime
```
rsuderman, line 9: Could we cleanup the import order a bit? I usually try to do alphabetical order with `from` and `import` separated out, e.g.

```
from pathlib import Path
from typing import List

import iree.compiler
import iree.runtime
import pytest
import sys
import torch
```

etc

rsuderman (approved):

sharktank/tests/models/punet/sharded_resnet_block_with_iree_test.py
sogartar: Done.

PR #1046, opened 2025-03-07
Make sharktank export prefill/decode batch size seperately

rsuderman:
Prefill and decode have different preferable batch sizes. We should export them separately.


renxida (approved):
LGTM

PR #1048, opened 2025-03-07
[sd] Clean up server flags. NFC.

kuhar:
Clarify the format of device IDs.

### DIFF
--- a/shortfin/python/shortfin_apps/sd/server.py
+++ b/shortfin/python/shortfin_apps/sd/server.py
@@ -297,7 +297,7 @@ def main(argv, log_config=UVICORN_LOG_CONFIG):
         type=str,
         required=True,
         choices=["local-task", "hip", "amdgpu"],
-        help="Primary inferencing device",
+        help="Primary inferencing device.",
     )
     parser.add_argument(
         "--target",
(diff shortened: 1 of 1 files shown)


shortfin/python/shortfin_apps/sd/server.py
```diff
@@ -312,7 +312,7 @@ def main(argv, log_config=UVICORN_LOG_CONFIG):
         type=str,
         nargs="*",
         default=None,
-        help="Device IDs visible to the system builder. Defaults to None (full visibility). Can be an index or a sf device id like amdgpu:0:0@0",
+        help="Device IDs visible to the system builder. Defaults to None (full visibility). Can be an IREE index or a device id like GPU-66613339-3934-3261-3131-396338323735.",
```
ScottTodd, line 315: I'm not sure "IREE index" carries any more meaning than just "index". The device ID example looks good to me though.

We could also document these better. Sorta covered by https://iree.dev/guides/deployment-configurations/gpu-rocm/#check-for-hip-devices and https://iree.dev/guides/deployment-configurations/gpu-rocm/#run-a-compiled-program but those are "guide" style sections and what we really want here is a "reference" (to use terminology from https://diataxis.fr/)

ScottTodd (approved):

shortfin/python/shortfin_apps/sd/server.py
kuhar: I implies you can get it from iree runtime, e.g., with `--dump_devices`. This is just a brief help in a flag, we probably need more verbose documentation in markdown down the line.

PR #1049, opened 2025-03-07
[sharktank] Add a more convenient way to export Flux transformer MLIR

sogartar:
E.g.

```
python -m sharktank.models.flux.export_flux_transformer_mlir \
  --output-path=model.mlir \
  --parameters-path=model.irpa
```


PhaneeshB (approved):
LGTM

PR #1056, opened 2025-03-07
[refactor][shortfin_apps]: remove unused `img_metadata` property

bjacobgordon:


monorimet (approved):
Eventually we want to embed image generation metadata into PNG output like other software does. Ok to remove for now

PR #1058, opened 2025-03-07
[types][shortfin_apps.sd] narrow `InferenceExecRequest` to `SDXLInferenceExecRequest`

bjacobgordon:


renxida:
Not code owner so not going to give explicit approval but:

Having separate names for similar classes that do different things across apps would make my life easier when i do fuzzy finds

IanNod (approved):

PR #1060, opened 2025-03-07
Disable flag that is causing the flux compilation test to fail

KyleHerndon:


sharktank/tests/models/flux/flux_test.py
```diff
@@ -49,7 +49,9 @@
     "--iree-opt-const-eval=false",
     "--iree-opt-strip-assertions=true",
     "--iree-global-opt-propagate-transposes=true",
-    "--iree-dispatch-creation-enable-fuse-horizontal-contractions=true",
+    # TODO: We want the following flag to be enable eventually, but there's
```
IanNod, line 52: Is there an IREE issue filed that we can link against in this comment?

IanNod (approved):
Minor nit comment but otherwise looks good

PR #1064, opened 2025-03-11
Fixes harness crash issue when runs with model_weights for the first time

amd-vivekag:
Fixes harness crash issue when runs with model_weights for the first time

Fixes harness crash issue when runs with model_weights for the first time.  This fixes the following issue: https://github.com/nod-ai/2024q2-sdxl-mlperf-sprint/issues/190


amd-vivekag:
@pdhirajkumarprasad , @vinayakdsci  is it the right way to fix the issue in harness.py? 

The issue is that because of new line missing at the end of the json dump, file path is not correct from the std output.

monorimet (approved):
LGTM

PR #1070, opened 2025-03-11
Fix llm model tests to set dtype

rsuderman:
Other tests are changing the default dtype causing flakes on RNG and KVache tests. Setting to `torch.float32` before initializing data guarantees correct type / value generation.

### DIFF
--- a/sharktank/tests/models/deepseek/test_deepseek.py
+++ b/sharktank/tests/models/deepseek/test_deepseek.py
@@ -12,12 +12,8 @@ import pytest
 import torch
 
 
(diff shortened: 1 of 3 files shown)


Alex-Vasile (commented):
No fails even after 25 runs.

I think it would be worth apply the same changes to the other mode tests. E.g. `test_grok.py` has an almost identical structure in terms of setup. I don't know why it isn't as flaky, but it may only be a matter of time before it acts up.

Alex-Vasile (approved):
(Approved)

PR #1071, opened 2025-03-11
[Do not review] To understand logical_devices_per_physical_device

dezhiAmd:
This PR is used to discuss the meaning of logical_devices_per_physical_device with Agrawal, Vivek

### DIFF
--- a/shortfin/src/shortfin/local/systems/amdgpu.cc
+++ b/shortfin/src/shortfin/local/systems/amdgpu.cc
@@ -65,13 +65,6 @@ void AMDGPUSystemBuilder::InitializeDefaultSettings() {
       config_options().GetInt("amdgpu_tracing_level", /*non_negative=*/true);
   default_device_params_.stream_tracing = tracing_level ? *tracing_level : 2;
 
-  // Override logical_devices_per_physical_device if present.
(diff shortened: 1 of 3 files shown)


shortfin/src/shortfin/local/systems/amdgpu.cc
```diff
@@ -195,47 +188,27 @@ SystemPtr AMDGPUSystemBuilder::CreateSystem() {
...
-          device.for_output()));
-      DeviceAddress address(
-          /*system_device_class=*/SYSTEM_DEVICE_CLASS,
-          /*logical_device_class=*/LOGICAL_DEVICE_CLASS,
-          /*hal_driver_prefix=*/HAL_DRIVER_PREFIX,
-          /*instance_ordinal=*/instance_ordinal,
-          /*queue_ordinal=*/0,
-          /*instance_topology_address=*/{logical_index});
-      ConfigureAllocators(amdgpu_allocator_specs_, device, address.device_name);
-      lsys->InitializeHalDevice(std::make_unique<AMDGPUDevice>(
```
dezhiAmd, line 233: This is the real action. When there are multiple logical_device_per_physical_device,

PR #1072, opened 2025-03-12
Split prefill and decode batchers to separate implementations

rsuderman:
Prefill and decode should be batched separately as their preferred batch sizes for each piece of hardware.


renxida (approved):
lgtm. Should we do anything about the base BatcherProcess class?

I remember earlier stella had a note that we should re-do the strobing, but I don't know what a better approach might be.

stbaione (approved):
LGTM

Issue #464, opened 2024-11-09
Improve App Integration Test Ergonomics


stbaione:
# Description

The `SD` and `LLM` app integration tests require deps that are outside of the necessary dependencies for `shortfin`. If you attempt to run `shortfin/tests` with only necessary dependencies, you receive a `Shortfin is missing a dependency` error.

We should move all integration tests with non-default dependencies to their own tree, similar to the location of the `cpu_llm_server_test.py` integration test: https://github.com/nod-ai/SHARK-Platform/blob/main/build_tools/integration_tests/llm/cpu_llm_server_test.py

We should also ensure that tests like this get skipped if the non-required dependency isn't present.

## Repro Steps

Create a fresh venv:
```bash
python -m venv .venv
```

Install only necessary requirements:
```bash
cd shortfin
pip install .
```

Attempt to run pytest:
```bash
pytest -v tests
```

You will receive an error output for packages required for testing apps, but not required for building shortfin:

```bash
E   shortfin.support.deps.ShortfinDepNotFoundError: Shortfin is missing a dependency to use shortfin_apps.llm. This is typically available via `pip install tokenizers`
```

PR #1076, opened 2025-03-12
Add `log_softmax` op to `shortfin.array`

stbaione:
# Desciption

Opening as a draft while I write the tests.

This adds `log_softmax` to the `shortfin.array` API, which is a required function for `beam_search`, `sampling`, and any other selection method that relies on cumulative probability.

`numpy` doesn't have a `log_softmax` function out of the box, which is what `xtensor` typically [maps to](https://xtensor.readthedocs.io/en/latest/numpy.html).

So, I wrote it and did some local manual tests to ensure that it matches the same behavior as `torch.log_softmax`


renxida (approved):
Looks good.

Couple of questions, more about my curiosity than the engineering quality:
- would this run faster than if we used separate sfnp ops sum, log, and exp to implement this?
- how often do we get non float 16 or 32 logits / how hard would it be to add support? I don't have information and I'm not pushing for support - asking out of curiosity.

stbaione:
> Looks good.
> 
> Couple of questions, more about my curiosity than the engineering quality:
> 
> * would this run faster than if we used separate sfnp ops sum, log, and exp to implement this?
> * how often do we get non float 16 or 32 logits / how hard would it be to add support? I don't have information and I'm not pushing for support - asking out of curiosity.

Both good question.

1. I can't say for certainty unless I ran a test, but I think it would be faster to do it this way, than separate sfnp ops. The reason is, when we call an op from a `device_array`, it uses `xtensor` to actually perform the operation. This means that we need to convert to an `xtensor` in the function, then store our results on a `device_array`, either a new one, or one that we pass in explicitly. So, it seems like it'd be less overhead to convert it to an `xtensor` once, do all operations on that `xtensor`, then create or fill one single `device_array` at the end, instead of having to do that for each function call.

2. Definitely conceivable, but don't have enough context of the work streams for other dtypes to give a good answer. Basically, I just matched the supported dtypes of the ops that were already there. I don't think it'd be too difficult to extend the ops if we hit that issue

rsuderman (changes requested):
Can we add a numerical test? We should run  some basic comparisons via `numpy` as a unittest.

stbaione:
> Can we add a numerical test? We should run some basic comparisons via `numpy` as a unittest.

Yeah, working on those now. Will be done shortly

stbaione:
> Can we add a numerical test? We should run some basic comparisons via `numpy` as a unittest.

Couple smaller tests to write to cover all bases, but just pushed the numerical tests.

IIUC, we can't use `numpy` in the shortfin unit tests. From [this issue](https://github.com/nod-ai/shark-ai/issues/464) that I forgot about until be reminded of it by that comment. Basically, there was an annoyance with `app tests`, because there were packages required by `shortfin_apps` that are not required by core `shortfin`.

`numpy` is one of those packages that [aren't required](https://github.com/nod-ai/shark-ai/blob/main/shortfin/python/shortfin/array/__init__.py#L118) for `shortfin.array`, but can be installed.

Therefore, I wanted to have some numerical tests that run independently of whether or not we have numpy. To accomplish this, I ran numerical checks against common identities of the `log_softmax` function, which I think should be sufficient to say the accuracy is good.

I could also add an optional test that only runs when numpy is installed though.

shortfin/tests/api/array_ops_test.py
```diff
@@ -283,6 +283,128 @@ def test_argpartition_error_cases(device):
...
+
+    output_array = sfnp.device_array(device, src.shape, dtype=sfnp.float32)
+    result_out = sfnp.log_softmax(src, axis, out=output_array)
+    result_no_out = sfnp.log_softmax(src, axis)
+
+    assert result_out.shape == src.shape
+    assert result_out.dtype.name == src.dtype.name
+    out_items = result_out.items.tolist()
+    no_out_items = result_no_out.items.tolist()
+    assert out_items == no_out_items
```
rsuderman, line 366: This should be done with approximate equivalency - we shouldn't assume exact numerical match.

shortfin/tests/api/array_ops_test.py
```diff
@@ -283,6 +283,128 @@ def test_argpartition_error_cases(device):
...
+        src.view(i).items = data[i]
+    if axis is not None:
+        result = sfnp.log_softmax(src, axis)
+    else:
+        result = sfnp.log_softmax(src)
+    results = []
+    for i in range(len(data)):
+        vals = [round(val, 3) for val in result.view(i).items.tolist()]
+        results.append(vals)
+    assert results == expected
```
rsuderman, line 349: This should be done with approximate equivalency - we shouldn't assume exact numerical match.

stbaione: Gotcha, yeah I was thinking the `round` would take care of small diffs, but better to just throw an epsilon in there

shortfin/tests/api/array_ops_test.py
stbaione: Should this not test exact numerical match? My thinking was that if they're calling the same function, with the same src array, then the array that is created within the function (not providing `out` argument), and the array that is passed to the function (provided `out` argument), should have the same results.

Is there some interesting stuff under the hood that could cause drift?

shortfin/tests/api/array_ops_test.py
rsuderman: You shouldn't have the round, it is better to use approximately equal. It effectively has the same effect but avoids the fuzzy check from bleeding across multiple lines.

shortfin/tests/api/array_ops_test.py
stbaione: Added and used `approximately_equal` function for comparison

rsuderman (approved):

PR #1078, opened 2025-03-12
Update sharded export command

stbaione:
We missed a spot when we updated the `llama_serving` guide after enabling `prefill` and `decode` to export with different batch sizes


renxida (approved):
Nice catch!

PR #1079, opened 2025-03-12
Add CLI client for shortfin LLM server

rsuderman:
For testing purposes its easier to invoke a single CLI that can batch run values. This can be reworked to include benchmarking the throughput and batch processing local files.


stbaione (approved):
LGTM

PR #1086, opened 2025-03-14
Cpu sharded integration tests

renxida:
A version of #995 but we test on CPU instead


renxida:
We now have a CI-passing sharded integration test, but it's a little bit of a mess and I want to chop it up, clean it up, and merge in bite-sized chunks. See #1154

PR #1089, opened 2025-03-14
BasePagedAttentionCache ref counting

stbaione:
## Description

We need to add `ref_counting` to the pages in `BasePagedAttentionCache` to support beam_search, sampling, etc.

When we `replicate` an `LlmInferenceExecRequest`, we will want to `fork` the kvcache pages. The way this works is if `req_1` has n pages, and we are forking the cache pages to `req_2`, then we will `share` cache pages `1, ..., n-1`, and `copy` the nth page to a new page. This is because pages `1, ..., n-1` will only be read from throughout the rest of the generation process, while the last `incomplete` page may be written too.

When we `free` pages from an `LlmInferenceExecRequest`, we will want to ensure that no other reqs are sharing any pages. If any pages are shared, they should not be freed until their reference count hits `0`. The pages that are not being shared (ref_count is already 0), can be freed immediately.

Special attention to the use of the `threading.Lock` is appreciated.


shortfin/python/shortfin_apps/llm/components/kvcache/page_pool.py
```diff
@@ -126,17 +126,19 @@ def copy_page(self, src_page: PageInfo) -> PageInfo:
...
         """
         # Allocate new page
-        (dst_page,) = self.acquire_free_pages(1)
+        dst_page = self.acquire_free_pages(1)
 
-        # fill src page with data
+        if dst_page is None:
+            return None
+
+        (dst_page,) = dst_page
```
rsuderman, line 134: You can just do `dst_page = dst_page[0]`. Pretty sure thats cleaner.

rsuderman (approved):

PR #1091, opened 2025-03-14
(shortfin-sd) Fix model config behavior in SDXL server.

monorimet:
Somehow this slipped by when merging RDNA4 changes -- we were defaulting to sdxl_config_i8.json if the path provided via CLI was not found, which was unintended. Now, if the path doesn't exist, we try to match against downloaded configs and error out if unsuccessful.


kuhar (approved):
LGTM but we should probably start adding unit tests around this code

PR #1094, opened 2025-03-14
[sharktank] Reenable T5 tests

sogartar:
Since https://github.com/nod-ai/shark-ai/issues/888 is fixed we can enable again the T5 tests.

Remove the testV1_1XxlBf16AgainstFluxGolden test. Other tests cover in a
better way.

### DIFF
--- a/.github/workflows/ci-sharktank.yml
+++ b/.github/workflows/ci-sharktank.yml
@@ -144,6 +144,7 @@ jobs:
             --with-flux-data \
             --with-vae-data \
             --with-quark-data \
+            --with-t5-data \
             --iree-hal-target-device=hip \
             --iree-hip-target=gfx942 \
             --iree-device=hip://0 \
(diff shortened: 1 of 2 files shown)


sogartar:
I did not notice there is a PR for that already https://github.com/nod-ai/shark-ai/pull/969.
The difference is that this PR removes 1 test.

renxida:
all good; let's merge this one instead.

@saienduri Could you help @sogartar figure out how to get around the "secrets unavailable" issue for this PR?

(from [this slack thread](https://xilinx.slack.com/archives/C08BB8QTV7Z/p1741924525689919?thread_ts=1741921133.192279&cid=C08BB8QTV7Z) it looks like all we have to do is to source the model weights from `/shark-cache/data/huggingface`)

renxida (approved):

saienduri:
Please see the full discussion. There is a bit more work needed due to the online checks that happen before pulling a model. But, what we can do is have a conditional logic setup so that if it is a PR from a fork, we pass a `local_files` variable set to `True` to the tests and then we use the `local_files_only` so there is no online check required for pulling hf model. Otherwise, we pass `local_files` as `False` and pull with online checks as usual, so our cache stays updated. If someone can get to implementing this, that would be great :)

sogartar:
This particular tests that are reenabled do not require secrets.

PR #1095, opened 2025-03-14
Group system args and add device reuse

rsuderman:
For testing sharding it can be useful to test reusing the same device multiple times. By default this should be set to false but for testing it can be enabled so sharding can be run without multiple devices.


stbaione (approved):
LGTM

PR #907, opened 2025-02-04
[Sharktank][LLama][fp8] Add support for fp8 attn in Llama

dan-garvey:
```
 python3 -m sharktank.examples.export_paged_llm_v1 --irpa-file=/home/chi/src/test/llama/dan/fp8_attn.irpa \
--output-mlir=/home/chi/src/test/llama/dan/f8_attn_chi_castf32_roctorch.mlir \
--output-config=/home/chi/src/test/llama/dan/config_attn_chi.json \
--bs=1 --attention-kernel sharktank \
--attention-dtype=float8_e4m3fnuz --activation-dtype=bfloat16 --use-attention-mask --use-hf --kv-cache-dtype=float8_e4m3fnuz
```
**IRPA weight file**:
`sudo cp /home/dan/SHARK-Platform/fp8_attn.irpa /sharedfile/attn/fp8_attn.irpa`

**Export 0319 update**:
```
python3 -m sharktank.examples.export_paged_llm_v1 --irpa-file=/sharedfile/attn/fp8_attn.irpa \
--output-mlir=/sharedfile/attn/128/fp8_attn.mlir \
--output-config=/sharedfile/attn/128/config_attn.json \
--bs-prefill=4 --bs-decode=4 --attention-kernel sharktank \
--attention-dtype=float8_e4m3fnuz --activation-dtype=bfloat16 --use-attention-mask --use-hf --kv-cache-dtype=float8_e4m3fnuz
```

PR #1098, opened 2025-03-15
[sharktank] Refactor PagedKVCache to PagedAttention

IanWood1:
This change moves reading/writing of the KV cache + calls to `attention` into a single `PagedAttention` class. This helps decouple the implementation of `PagedAttention` (caching mechanism & attention) from its users. Instead, the page size (`get_page_size()`) is used to allocate the correct amount and `forward_*` methods are provided. This will enable further work that will change the layout of the cache + implement custom kernels for attention.


Groverkss (commented):
Nice, I like the structure of it. This is exactly what I was looking for.

rsuderman (approved):

rsuderman:
@archana-ramalingam just an FYI. This should help cleanup the attention specific work between layers. Eventually we will want something similar for latent attention.

AmosLewis:
https://github.com/nod-ai/shark-ai/pull/907

PR #1104, opened 2025-03-17
Add Integration Test that uses iree-tracy-capture to dump a trace of a simple shortfin llm run

renxida:


app_tests/integration_tests/llm/shortfin/tracing_test.py
```diff
@@ -0,0 +1,153 @@
...
+    "model_artifacts,server",
+    [
+        ["tinystories_llama2_25m", {"prefix_sharing": "none"}],
+    ],
+    indirect=True,
+)
+
+# Same golden values as in tinystories_llama2_25m_test.py
+GOLDEN_PROMPT = "Once upon a time"
+GOLDEN_RESPONSE = ", there was a little girl named Lily."  # this assumes purely deterministic greedy search
```
ScottTodd, line 31: I don't think we should need to run a full LLM (even a tiny version) to test tracing. Can we use something like https://github.com/nod-ai/shark-ai/tree/main/shortfin/examples/python/mobilenet_server ?

Prefer to keep tests focused on a single behavior.

app_tests/integration_tests/llm/shortfin/tracing_test.py
```diff
@@ -0,0 +1,153 @@
...
+
+        tracy_process = subprocess.Popen(
+            ["iree-tracy-capture", "-o", str(capture_file)],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+
+        try:
+            # Wait a moment for tracy capture to initialize
+            time.sleep(3)
```
ScottTodd, line 67: This sleep shouldn't be needed. What issues did you see without it?

app_tests/integration_tests/llm/shortfin/tracing_test.py
```diff
@@ -0,0 +1,153 @@
...
+                    actual=response,
+                    message=f"Generation did not match expected pattern.\nExpected to start with: {expected_prefix}\nActual response: {response}",
+                )
+
+            # Clean up environment variable if we set it
+            if not tracy_env_was_set:
+                os.environ.pop("SHORTFIN_PY_RUNTIME", None)
+
+            # Wait a moment for trace data to be collected
+            time.sleep(5)
```
ScottTodd, line 94: Same here. You might want to set `TRACY_NO_EXIT`. See also https://iree.dev/developers/performance/profiling-with-tracy/#quickstart. I would hope you could remove all the timeouts and waits from this test and just rely on pytest-timeout.

Another thing to consider will be having multiple test processes or user processes using the same ports. May want to be explicit with each test and the capture tool so ensure no overlap.

app_tests/integration_tests/llm/shortfin/tracing_test.py
```diff
@@ -0,0 +1,153 @@
...
+        process, port = server
+        assert process.poll() is None, "Server process terminated unexpectedly"
+
+        # Determine output directory for artifacts (use environment variable if set, otherwise tmp_path)
+        output_dir = os.environ.get("TEST_OUTPUT_DIR")
+        if output_dir:
+            output_dir = Path(output_dir)
+            output_dir.mkdir(parents=True, exist_ok=True)
+        else:
+            output_dir = tmp_path
```
ScottTodd, line 53: Is this a common pattern in shark-ai? Python and pytest both provide ways to get temp paths that could be used instead

app_tests/integration_tests/llm/shortfin/tracing_test.py
```diff
@@ -0,0 +1,153 @@
...
+
+        # Start tracy capture
+        capture_file = output_dir / "capture.tracy"
+        logger.info(f"Tracy capture will be saved to: {capture_file}")
+
+        tracy_process = subprocess.Popen(
+            ["iree-tracy-capture", "-o", str(capture_file)],
```
ScottTodd, line 60: This assumes that `iree-tracy-capture` exists, which is not true on Windows and macOS at the moment. Could skip the test on those platforms or if the tool does not appear on `$PATH`. This issue tracks building on all platforms and bundling for consistency: https://github.com/iree-org/iree/issues/19841

.github/workflows/ci-shortfin-developer-tool-tests.yml
```diff
@@ -0,0 +1,87 @@
...
+on:
+  workflow_dispatch:
+  schedule:
+    # Run weekly to ensure our tools remain functional
+    - cron: '0 0 * * 0'
+  pull_request:
+    paths:
+      - '.github/workflows/ci-shortfin-developer-tool-tests.yml'
+      - 'app_tests/integration_tests/llm/shortfin/tracing_test.py'
+      - 'shortfin/**'
```
ScottTodd, line 18: Do we need a new workflow for this? Can this go in an existing unit test or integration test workflow instead?

.github/workflows/ci-shortfin-developer-tool-tests.yml
```diff
@@ -0,0 +1,87 @@
...
+  cancel-in-progress: true
+
+jobs:
+  shortfin_tracing_test:
+    name: "Shortfin Tracing Test"
+    strategy:
+      matrix:
+        version: [3.11]
+      fail-fast: false
+    runs-on: linux-mi300-1gpu-ossci
```
ScottTodd, line 35: Sanity checking a developer tool should be possible on a github-hosted CPU runner, no GPU needed.

.github/workflows/ci-shortfin-developer-tool-tests.yml
```diff
@@ -0,0 +1,87 @@
...
+          export SHORTFIN_PY_RUNTIME=tracy
+          # Run only the tracing test
+          TEST_OUTPUT_DIR=${ARTIFACTS_DIR} pytest -v app_tests/integration_tests/llm/shortfin/tracing_test.py --test_device=cpu --log-cli-level=INFO
+
+      - name: Upload Trace Artifact
+        uses: actions/upload-artifact@5d5d22a31266ced268874388b861e4b58bb5c2f3 # v4.3.1
+        with:
+          name: tracy-capture
+          path: ${{ env.ARTIFACTS_DIR }}/capture.tracy
+          if-no-files-found: error
```
ScottTodd, line 87: Separate unit tests for functionality from benchmarking/performance profiling. This test and workflow are performing double duty right now. Either tracing_test should be a unit test that exercises the profiling support or the existing benchmark tests should produce traces (run once with tracing and once without tracing, get metrics from tracing disabled but then upload the traces too)

ScottTodd (changes requested):
Drive-by comments

app_tests/integration_tests/llm/shortfin/tracing_test.py
renxida: i do eventually want to regularly trace the LLM server just so i can have a history of traces

But yeah that should be a completely separate PR.

renxida:
@ScottTodd comments addressed in #1106.

Closing this one

PR #1105, opened 2025-03-17
Gate debug messages in dump_gguf.py behind `-v` flag

renxida:
I need this because I've been using `dump_gguf.py` for some integration tests (see https://github.com/nod-ai/shark-ai/blob/41a09a836c4558745ec1ca01e80a819cd761e608/app_tests/integration_tests/llm/model_management.py#L223-L239)


stbaione (approved):
Really nice ergonomic improvement. Terminal gets spammed from the prints without this change

PR #1106, opened 2025-03-17
Test tracy tracing with our existing mobilenet vmfb in shortfin unit tests

renxida:
Supercedes #1104


renxida:
Superceded by 1108

PR #1107, opened 2025-03-17
Support multirequest benchmarking in `cli`

rsuderman:
To support taking rough benchmarks of the service throughput we can use the CLI client to generate a set of sample queries to run through the service. This gives both an average measured latency and average number of requests per second.


shortfin/python/shortfin_apps/llm/cli.py
```diff
@@ -26,9 +28,9 @@
...
     group = parser.add_argument_group("Input Source", "Inputs to select from")
-    group = group.add_mutually_exclusive_group()
-    parser.add_argument("--prompt")
-    parser.add_argument("--prompt-file")
+    group = group.add_mutually_exclusive_group(required=True)
+    group.add_argument("--prompt")
+    group.add_argument("--prompt-file")
 
 
 def add_service_args(parser):
```
renxida, line 36: neat. love how this is organized. new trick acquired (for organizing argparse args into groups)

renxida (approved):

PR #1108, opened 2025-03-17
Test tracy tracing with our existing mobilenet vmfb in shortfin unit tests

renxida:
Recreating 1106 on a branch owned by nod-ai instead of renxida because github secrets aren't available to prs from branches.


.github/workflows/ci-libshortfin.yml
```diff
@@ -127,18 +128,24 @@ jobs:
         cmake --build build --target all
 
     - name: pip install shortfin
-      if: ${{ matrix.name != 'Ubuntu (Clang)(host-only)' && !contains(matrix.name, 'tracing') }}
+      if: ${{ matrix.name != 'Ubuntu (Clang)(host-only)' }}
       working-directory: ${{ env.LIBSHORTFIN_DIR }}
       run: |
-        pip install -v -e build/
+        SHORTFIN_ENABLE_TRACING=${{ matrix.tracing && 'ON' || 'OFF' }} pip install -v -e build/
```
rsuderman, line 134: I believe we can set environment variables via the matrix as a yamal setting. @ScottTodd has chided me on this a couple times.

renxida: on it. i think i know what you're talking about

ScottTodd: I just gave @Eliasj42 similar feedback at https://github.com/iree-org/iree/pull/20260#discussion_r2002062392. Let us know what you find :D

renxida: looked through some github docs. couldn't quite figure it out. Do i just need to set something like

```yaml
      matrix:
        name: ["Ubuntu (Clang)(full)", "Ubuntu (Clang)(host-only)", "Windows (MSVC)"]
        python-version: ["3.10", "3.11", "3.12"]
        include:
          - name: Ubuntu (Clang)(full)
            runs-on: ubuntu-24.04
            cmake-options:
              -DCMAKE_C_COMPILER=clang-18 -DCMAKE_CXX_COMPILER=clang++-18 -DCMAKE_LINKER_TYPE=LLD -DSHORTFIN_ENABLE_TOKENIZERS=ON
            additional-packages: clang lld
            env:
              SHORTFIN_ENABLE_TRACING: OFF
              SHORTFIN_PY_RUNTIME: default
          - name: Ubuntu (Clang)(host-only)
            runs-on: ubuntu-24.04
            # In this configuration, also build static+dynamic in order to verify
            # that path structurally works.
            cmake-options:
              -DCMAKE_C_COMPILER=clang-18 -DCMAKE_CXX_COMPILER=clang++-18 -DCMAKE_LINKER_TYPE=LLD -DSHORTFIN_HAVE_AMDGPU=OFF -DSHORTFIN_BUILD_STATIC=ON -DSHORTFIN_BUILD_DYNAMIC=ON
            additional-packages: clang lld
            env:
              SHORTFIN_ENABLE_TRACING: OFF
              SHORTFIN_PY_RUNTIME: default
```

Or do i need to do more to make it actually put matrix.env out in the environment variables? @ScottTodd

ScottTodd: Might need to forward from `matrix.env` to `GITHUB_ENV`. Something like this?

```yml
name: Forward matrix environment
run:
  echo "SHORTFIN_ENABLE_TRACING=${{ matrix.env.SHORTFIN_ENABLE_TRACING }}" >> $GITHUB_ENV`
  echo "SHORTFIN_PY_RUNTIME=${{ matrix.env.SHORTFIN_PY_RUNTIME}}" >> $GITHUB_ENV`
```

renxida:
iree.dev blocked. If it doesn't get resolved by tomorrow i'm going in and changing all the --find-links

.github/workflows/ci-libshortfin.yml
```diff
@@ -127,18 +151,24 @@ jobs:
...
       working-directory: ${{ env.LIBSHORTFIN_DIR }}
       run: |
         ctest --timeout 30 --output-on-failure --test-dir build
         pytest -s --durations=10 --timeout=30
 
+    - name: Test tracing shortfin
+      if: matrix.tracing
+      working-directory: ${{ env.LIBSHORTFIN_DIR }}
+      run: |
+        pytest -s --durations=10 --timeout=30 tests/invocation/mobilenet_tracing_test.py
```
ScottTodd, line 170: How about these flags instead? There is only one test running, so we might as well get as much output from it as we can.

```suggestion
      run: |
        pytest \
          --durations=0 \
          --timeout=30 \
          --log-cli-level=info \
          tests/invocation/mobilenet_tracing_test.py
```

shortfin/tests/invocation/mobilenet_tracing_test.py
```diff
@@ -0,0 +1,251 @@
...
+        # Clean up
+        del device_output
+        del host_output
+        logger.info("MobileNet inference completed")
+
+    system.run(run_model())
+
+    # Clean up
+    system.shutdown()
+    logger.info("MobileNet model execution finished")
```
ScottTodd, line 112: I wonder if we can hide some of this system setup code behind a nicer API, especially if we add other integration tests like this one that just want "some server running a model for one request".

ScottTodd (approved):
Very nice! I like this setup, especially how the enable/skip settings are plumbed through, the test logs what it is doing (let's get those logs in the CI output), and many different failure cases are accounted for. Nice job!

rsuderman:
One last thought - could we make a super simple sample model instead of running `mobilenet`. For something this simple we should just export, compile, and run a really simple module (e.g. one matmul). This avoids all of the overhead of using a real model and should drastically decrease the runtime (especially on CPU).

ScottTodd:
> This avoids all of the overhead of using a real model and should drastically decrease the runtime (especially on CPU).

cpu on standard github runner: https://github.com/nod-ai/shark-ai/actions/runs/13984524003/job/39155824528#step:12:36
```
4.38s setup    tests/invocation/mobilenet_tracing_test.py::test_tracing_mobilenet
0.75s call     tests/invocation/mobilenet_tracing_test.py::test_tracing_mobilenet
```

already fast enough for me, but agreed about the general principle

renxida:
> One last thought - could we make a super simple sample model instead of running `mobilenet`. For something this simple we should just export, compile, and run a really simple module (e.g. one matmul). This avoids all of the overhead of using a real model and should drastically decrease the runtime (especially on CPU).

I would agree if this took a significant amount of time to run, but incl. pytest's setup + teardown, the actual test step takes less than 5s to run.

I'd like to keep it simple and use the already-existing mobilenet artifacts that is used by other testcases too.

shortfin/tests/invocation/mobilenet_tracing_test.py
renxida: One thing that keeps coming up with the LLAMA models is ...

"Oh no! we have a problem!"
"Does it replicate on iree-run-module?"
"Ug it's a lot of work to get the mlir and the inputs and the flags"

Now if we had an interface like that, that also, in an optional debug mode, dumps the inputs and the flags for iree-run-module....

renxida: Specifically I'm thinking about something where

Let's say we have a program function fn in a shortfin_app

and we have an `await fn(args)`

By setting an enviornmental variable

We can make it dump all the inputs & the flags / options used to set up the system and the fibers and devices.

----

Once we have the dump, we can use an interface `shortfin-run-module` to run it, and that'd be the `some server running a model for 1 request`

Or just test with `iree-run-module` to make sure that the problem isn't just a shortfin problem.

renxida: A `shortfin-run-module` interface would be pretty neat.

Would also make it easier to test discrepancies between sf and iree run module

.github/workflows/ci-libshortfin.yml
ScottTodd: New logs look good:
https://github.com/nod-ai/shark-ai/actions/runs/13998942911/job/39200752810#step:12:34
```
tests/invocation/mobilenet_tracing_test.py::test_tracing_mobilenet 
-------------------------------- live log setup --------------------------------
INFO     root:binaries.py:185 Invoke IREE Tool: /opt/hostedtoolcache/Python/3.12.9/x64/lib/python3.12/site-packages/iree/compiler/tools/../_mlir_libs/iree-compile /tmp/pytest-of-runner/pytest-0/mobilenet_onnx0/mobilenet.mlir --iree-input-type=onnx --iree-vm-bytecode-module-output-format=flatbuffer-binary -o=/tmp/pytest-of-runner/pytest-0/mobilenet_onnx0/mobilenet_cpu.vmfb --mlir-print-debuginfo --mlir-print-op-on-diagnostic=false --iree-hal-target-device=llvm-cpu --iree-llvmcpu-target-cpu=host
-------------------------------- live log call ---------------------------------
INFO     tests.invocation.mobilenet_tracing_test:mobilenet_tracing_test.py:196 Tracy capture will be saved to: /tmp/pytest-of-runner/pytest-0/test_tracing_mobilenet0/capture.tracy
INFO     tests.invocation.mobilenet_tracing_test:mobilenet_tracing_test.py:198 Tracy capture will be using port 55139
INFO     tests.invocation.mobilenet_tracing_test:mobilenet_tracing_test.py:203 Tracy capture will be saved to: /tmp/pytest-of-runner/pytest-0/test_tracing_mobilenet0/capture.tracy
INFO     tests.invocation.mobilenet_tracing_test:mobilenet_tracing_test.py:219 Running MobileNet model with tracing enabled via subprocess
INFO     tests.invocation.mobilenet_tracing_test:mobilenet_tracing_test.py:135 Launching subprocess to run MobileNet model
INFO     tests.invocation.mobilenet_tracing_test:mobilenet_tracing_test.py:225 Waiting for tracy to finish collecting data...
INFO     tests.invocation.mobilenet_tracing_test:mobilenet_tracing_test.py:243 Tracy capture file size: 88402 bytes
PASSED                                                                   [100%]

============================== slowest durations ===============================
3.77s setup    tests/invocation/mobilenet_tracing_test.py::test_tracing_mobilenet
0.77s call     tests/invocation/mobilenet_tracing_test.py::test_tracing_mobilenet

(1 durations < 0.005s hidden.  Use -vv to show these durations.)
============================== 1 passed in 4.56s ===============================
```

PR #1121, opened 2025-03-19
[sharktank] Add toy VAE tests

sogartar:
Add eager and IREE tests for f32 and bf16.

The IREE test fails compilation for the particular model shape. See https://github.com/iree-org/iree/issues/20307

### DIFF
--- a/sharktank/sharktank/models/vae/config.py
+++ b/sharktank/sharktank/models/vae/config.py
@@ -9,9 +9,9 @@
 # While much was a simple reverse engineering of the config.json and parameters,
 # code was taken where appropriate.
 
-from typing import List, Optional, Sequence, Tuple, Union
+from typing import Any, Sequence
 
-from dataclasses import dataclass
+from dataclasses import dataclass, asdict
 import inspect
 import warnings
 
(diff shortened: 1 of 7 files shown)


KyleHerndon (approved):
Looks good!

sharktank/sharktank/models/vae/config.py
```diff
@@ -22,24 +22,23 @@
...
         "UpDecoderBlock2D",
         "UpDecoderBlock2D",
     )
     layers_per_block: int = 2
+    latent_channels: int = 4
     norm_num_groups: int = 32
     scaling_factor: float = 0.13025
     use_post_quant_conv: bool = True
     shift_factor: float = 0.0
+    sample_size: int | tuple[int, int] | list[int, int] = 1024
```
IanNod, line 41: previously the output shape was essentially implicitly determined from the input shape coming from unet or the transformer model. Would this still work given potentially conflicting input shapes and sample size?

sogartar: I am not sure. It will probably cry about a dimension mismatch in the rearrange as `h*w` would have an unexpected size. We really can't get away without it as the dimensions come packed. I see why you put this in the model as it is a small unpack operation that sits between the 2 models in the Flux pipeline and should be attached to one of them for performance.

Generally, if we want introspective models that can describe their own export, we need such config parameters.

PR #1122, opened 2025-03-19
Correcting usage of `--iree-hal-target-device=llvm-cpu`.

### LINKED ISSUE #1119: Replace usage of `--iree-hal-target-device=llvm-cpu` with `local`.
There was a hack in IREE to support the deprecated `--iree-hal-target-backends=` flag where `--iree-hal-target-device=` would accept the same values. Sharktank looks like it's been relying on that by specifying `--iree-hal-target-device=llvm-cpu` explicitly and that is going to stop working in #20295 (the tech debt is finally being cleaned up, yay).

Sharktank needs to replace its usage of `--iree-hal-target-device=llvm-cpu` with `--iree-hal-target-device=local --iree-hal-local-target-device-backends=llvm-cpu`.

benvanik:
https://github.com/iree-org/iree/pull/20295 will remove the legacy compatibility hack that made this flag work. Due to a compiler API issue it must land first and this PR should land once it has been integrated.

Fixes #1119.


benvanik:
https://github.com/iree-org/iree/pull/20295 has landed in IREE. I'm leaving this here for someone on the project to land whenever it makes sense.

ScottTodd (approved):
The changes LGTM. Looks like we may need to merge together with a version update like https://github.com/nod-ai/shark-ai/pull/1126

PR #1123, opened 2025-03-19
Temporary mitigation for iree.dev down

renxida:
See:
https://github.com/iree-org/iree/issues/20309#issuecomment-2737163087


renxida:
Don't think we need this anymore now that iree.dev is back up

PR #1128, opened 2025-03-20
(shortfin-apps)(flux) Align flux response handling with shark-ui compatibility requirements

monorimet:
Mirrors changes from the following sd-focused PRs:

https://github.com/nod-ai/shark-ai/pull/1025
https://github.com/nod-ai/shark-ai/pull/1068
https://github.com/nod-ai/shark-ai/pull/1069


shortfin/python/shortfin_apps/flux/components/generate.py
```diff
@@ -41,13 +52,20 @@ def __init__(
...
         await exec.done
-        self.result_image = exec.result_image
+        self.output = (
+            TextToImageInferenceOutput(exec.response_image)
+            if exec.response_image
+            else None
+        )
+
+
+Item = TypeVar("Item")
```
KyleHerndon, line 68: This looks unused.

shortfin/python/shortfin_apps/flux/server.py
```diff
@@ -104,6 +105,14 @@ async def generate_request(gen_req: GenerateReqInput, request: Request):
 app.post("/generate")(generate_request)
 app.put("/generate")(generate_request)
 
+# -------- MIDDLEWARE --------
+app.add_middleware(
+    CORSMiddleware,
+    allow_origins=["*"],
```
KyleHerndon, line 111: Is it possible to not use wildcards? I don't understand why this is necessary, given we should be the only one serving content...
It would be nice to keep the security implications smaller when possible.

shortfin/python/shortfin_apps/flux/simple_client.py
```diff
@@ -24,39 +29,15 @@
...
-            indexed = request[arg][0]
-        else:
-            indexed = request[arg][idx]
-    else:
-        indexed = request[arg]
-    return indexed
-
-
-async def send_request(session, rep, args, data):
+async def send_request(session: aiohttp.ClientSession, rep, args, data):
```
KyleHerndon, line 40: Would you mind adding type hints for the other arguments?

shortfin/python/shortfin_apps/flux/server.py
monorimet: There was significant discussion about this in the original PR for adding CORSMiddleware: https://github.com/nod-ai/shark-ai/pull/965

> It would be nice to keep the security implications smaller when possible.
I generally agree with this.

shortfin/python/shortfin_apps/flux/simple_client.py
```diff
@@ -37,8 +37,11 @@
 }
 
 
-async def send_request(session: aiohttp.ClientSession, rep, args, data):
+async def send_request(
+    session: aiohttp.ClientSession, rep: int, args: argparse.Namespace, data: dict
+):
     print("Sending request batch #", rep)
+    breakpoint()
```
bjacobgordon, line 44: Looks like a `breakpoint` snuck into the commit!

shortfin/python/shortfin_apps/flux/components/TextToImageInferenceOutput.py
bjacobgordon, line 1: l'm glad that this was easy enough to find, @monorimet! I believe there's an opportunity to keep flux and sd in step.

What if we moved this to a new "shortfin_apps/text_to_image" directory to keep the common, model-agnostic components separate? In the future, if needed, we can just subclass this for aspect specific to SDXL or Flux.

What do you think?

shortfin/python/shortfin_apps/flux/server.py
bjacobgordon: It _is_ possible, but would require some adjustments, and it would probably be a no brainer if/when we unify the text to image server implementations! At the moment, this is just a reflection of the current sd server.

shortfin/python/shortfin_apps/flux/components/service.py
```diff
@@ -730,8 +729,9 @@ async def _postprocess(self, device, requests):
             sfnp.transpose(images_planar, (1, 2, 0), out=permuted)
             permuted = sfnp.multiply(127.5, (sfnp.add(permuted, 1.0)))
             out = sfnp.round(permuted, dtype=sfnp.uint8)
-            image_bytes = bytes(out.map(read=True))
-
-            image = base64.b64encode(image_bytes).decode("utf-8")
-            req.result_image = image
+            out_bytes = bytes(out.map(read=True))
+            processed_image = Image.frombytes(
```
bjacobgordon, line 733: Would using `Image.fromarray` allow us to bypass casting to `bytes` and avoid having to read `out.shape[ :2]`?

shortfin/python/shortfin_apps/flux/simple_client.py
```diff
@@ -24,39 +29,17 @@
...
-        else:
-            indexed = request[arg][idx]
-    else:
-        indexed = request[arg]
-    return indexed
-
-
-async def send_request(session, rep, args, data):
+async def send_request(
+    session: aiohttp.ClientSession, rep: int, args: argparse.Namespace, data: dict
```
bjacobgordon, line 41: I think there's an opportunity for a separate 30sec PR right here! I also have some more thoughts on this, but don't wanna clog up this review haha.

Could you bubble this signature change to its own commit at the beginning of this branch?

bjacobgordon (changes requested):
This is looking great so far!

shortfin/python/shortfin_apps/flux/components/TextToImageInferenceOutput.py
monorimet: I'd like to defer this to the coming refactor patches.

shortfin/python/shortfin_apps/flux/components/service.py
monorimet: `out` here is a shortfin device array so we just take the bytes data -- using the bytes with the size data saves us a numpy dependency here.. The alternative would be to create a new numpy array for compatibility with Image.fromarray.

shortfin/python/shortfin_apps/flux/simple_client.py
monorimet: This seems unnecessary given the commits will be squashed before merge.

shortfin/python/shortfin_apps/flux/components/TextToImageInferenceOutput.py
bjacobgordon: We can defer if you'd like, totally!

But I recommend doing this now mainly because:
1. It shrinks the diff for this particular PR (renaming an existing file with no line changes rather than adding a few dozen lines in the form of a new file)
2. It's a simple, easy thing we can do right now that will shorten the list of things to refactor.
3. Which as a result can simplify or completely avoid the merge conflicts we might encounter along the way.

Would you still like to defer?

shortfin/python/shortfin_apps/flux/components/service.py
bjacobgordon: Ah gotcha, thanks for the context!

shortfin/python/shortfin_apps/flux/simple_client.py
bjacobgordon: On that basis, totally unnecessary, for sure! I'd say it's more about PR granularity, diff size, merge conflict avoidance, etc. We can skip it for now!

The thing I was gonna mention is making the arguments multiline for easier diffing:
```suggestion
    session: aiohttp.ClientSession,
    rep: int,
    args: argparse.Namespace,
    data: dict,
```

monorimet: Yeah that seems reasonable.

monorimet: The pre-commit actually undoes this change... let's keep aligned with the automatic linter in this case.

shortfin/python/shortfin_apps/flux/components/TextToImageInferenceOutput.py
bjacobgordon: Great question!

The intent for that directory was more like "here's the stuff that shortfin uses, but it's actually shortfin-agnostic", so probably not `shortfin_apps/types`. Now that you mention it, we'll probably wanna rename it to `shortfin_apps/library` the next time we mess with that directory. Not in this PR of course.

`TextToImageInferenceOutput` is definitely "aware" of shortfin's primary objective, and the name suggests that we'll eventually define `TextToImageInferenceInput`, so reserving `shortfin_apps/text_to_image` would help us begin with the end in mind!

And then when `TextToTextInferenceOutput`/`TextToTextInferenceInput` enter the chat, we can update our directories accordingly haha.

shortfin/python/shortfin_apps/flux/simple_client.py
bjacobgordon: Just tried it, and it seems to only undo the change if you leave off the dangling comma! Does it pass pre-commit for you if you put the comma after `dict`?

sogartar (approved):
I think in a subsequent PR `GenerateImageProcess` and `ClientGenerateBatchProcess` should be refactored and probably completely shared between Flux and SD.
If the `service` and `client` init args have (and they should) a sufficient base interface then it should be easy to merge the implementations.

bjacobgordon (commented):
The code's trending in the right direction, @monorimet! Gonna try it with the UI, will then let you know whether it's good to merge.

bjacobgordon (approved):

AmosLewis:
@monorimet  @bjacobgordon  The ppl test and smoke test fail since this patch on TOM. Could you guys take a look at and fix?

Issue #810, opened 2025-01-10
[Tuner] Automatic merging of tuner-generated td specs


kuhar:
We should be able to automatically append the best spec selected by the tuner to an existing tuning spec file.

This can be done either through textual manipulation of a spec file with some markers that indicate where new tuning specs should go, or as an iree pass similar to the one that links tuning specs. The main difference compared to the latter one is that we would like to keep adding new specs to an existing `transform.foreach_match in %arg0 ` op in the main named sequence called `__kernel_config`. For the maintainability reasons, this should be done as a compiler pass.

bangtianliu:
I think this issue is done.

PR #1136, opened 2025-03-23
[tuner] merge default td specs

bangtianliu:
This PR adds the utility functions on the tuner side, which call the LinkTuningSpecsPass (`--iree-codegen-link-tuning-specs`) to merge the default td specs. 

Two helper functions are added:
-  `combine_tuning_specs`:  put two modules in one outer module using Python bindings.
- `link_tuning_specs`: invokes linking/merging pass (`--iree-codegen-link-tuning-specs`) to merge td_specs inside the module.

Issue: #810

### DIFF
--- a/tuner/tuner/common.py
+++ b/tuner/tuner/common.py
@@ -10,10 +10,12 @@ from enum import Enum
 from types import TracebackType
 from typing import Optional
 from typing import Any
+import subprocess
 
 from iree.compiler import ir  # type: ignore
 
 from iree.compiler.dialects import iree_gpu  # type: ignore
+import iree.compiler as ireec  # type: ignore
 
 
 class CommonTypes:
@@ -266,3 +268,41 @@ class MLIRTransformation:
     template: list[str]
     modified: str
     embeddable: str
+
+
+def combine_tuning_specs(
+    tuner_ctx: TunerContext, td_specs: list[ir.Module]
+) -> ir.Module:
+    with tuner_ctx.mlir_ctx as ctx, ir.Location.unknown():
+        top_module = ir.Module.create()
+        top_module.operation.attributes[
+            "transform.with_named_sequence"
+        ] = ir.UnitAttr.get()
+
+        for td_spec in td_specs:
+            top_module.body.append(td_spec.operation.clone())
+        return top_module
+
+
+def link_tuning_specs(tuner_ctx: TunerContext, td_specs: list[ir.Module]) -> ir.Module:
+    module = combine_tuning_specs(tuner_ctx, td_specs)
+    iree_opt = ireec.binaries.find_tool("iree-opt")
+
+    result = subprocess.run(
+        [
+            iree_opt,
+            "--split-input-file",
+            "--verify-diagnostics",
+            "--iree-codegen-link-tuning-specs",
+            "-",
+        ],
+        input=str(module),  # Provide MLIR text directly via stdin.
+        capture_output=True,
+        text=True,
+    )
+
+    if result.returncode != 0:
+        raise RuntimeError(f"iree-opt failed: {result.stderr}")
+
+    linked_module = ir.Module.parse(result.stdout, tuner_ctx.mlir_ctx)
+    return linked_module
--- a/tuner/tuner/common.py
+++ b/tuner/tuner/common.py
@@ -270,6 +270,8 @@ class MLIRTransformation:
     embeddable: str
 
 
+# Puts multiple input modules `td_specs` into a single top-level container module.
+# This function does *not* attempt to merge or link `td_specs` across modules.
 def combine_tuning_specs(
     tuner_ctx: TunerContext, td_specs: list[ir.Module]
 ) -> ir.Module:
@@ -284,6 +286,12 @@ def combine_tuning_specs(
         return top_module
 
 
+# Links multiple input modules (`td_specs`) into a single tuning specification module.
+# First, the input modules are combined into a container module. Then, the external
+# `iree-opt` tool is invoked with the `--iree-codegen-link-tuning-specs` pass to
+# link or merge the individual tuning specs. When all input specs are marked with the
+# default attribute `iree_codegen.tuning_spec_with_default_entrypoint`, they are merged
+# into one tuning spec.
 def link_tuning_specs(tuner_ctx: TunerContext, td_specs: list[ir.Module]) -> ir.Module:
     module = combine_tuning_specs(tuner_ctx, td_specs)
     iree_opt = ireec.binaries.find_tool("iree-opt")
@@ -304,5 +312,4 @@ def link_tuning_specs(tuner_ctx: TunerContext, td_specs: list[ir.Module]) -> ir.
     if result.returncode != 0:
         raise RuntimeError(f"iree-opt failed: {result.stderr}")
 
-    linked_module = ir.Module.parse(result.stdout, tuner_ctx.mlir_ctx)
-    return linked_module
+    return ir.Module.parse(result.stdout, tuner_ctx.mlir_ctx)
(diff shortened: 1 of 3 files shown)


tuner/tuner/common.py
```diff
@@ -266,3 +268,41 @@ class MLIRTransformation:
     template: list[str]
     modified: str
     embeddable: str
+
+
+def combine_tuning_specs(
```
kuhar, line 273: Why `combine` instead of `merge`? I think we should stick with consistent naming if we can

tuner/tuner/common.py
```diff
@@ -266,3 +268,41 @@ class MLIRTransformation:
...
+            "--iree-codegen-link-tuning-specs",
+            "-",
+        ],
+        input=str(module),  # Provide MLIR text directly via stdin.
+        capture_output=True,
+        text=True,
+    )
+
+    if result.returncode != 0:
+        raise RuntimeError(f"iree-opt failed: {result.stderr}")
```
kuhar, line 305: I don't think this is covered by any tests?

tuner/tuner/common.py
```diff
@@ -266,3 +268,41 @@ class MLIRTransformation:
...
+
+    result = subprocess.run(
+        [
+            iree_opt,
+            "--split-input-file",
+            "--verify-diagnostics",
+            "--iree-codegen-link-tuning-specs",
+            "-",
+        ],
+        input=str(module),  # Provide MLIR text directly via stdin.
```
kuhar, line 299: I think this may be very annoying to reproduce if it ever fails. Why not serialize it to disk and put under a temp directory?

tuner/tuner/common.py
```diff
@@ -266,3 +268,41 @@ class MLIRTransformation:
...
+        capture_output=True,
+        text=True,
+    )
+
+    if result.returncode != 0:
+        raise RuntimeError(f"iree-opt failed: {result.stderr}")
+
+    linked_module = ir.Module.parse(result.stdout, tuner_ctx.mlir_ctx)
+    return linked_module
```
kuhar, line 308:
```suggestion
    return ir.Module.parse(result.stdout, tuner_ctx.mlir_ctx)
```

tuner/tuner/common.py
kuhar: Could you explain what this function does as a comment so that it's easy to understand the difference between combine/link/merge?

tuner/tuner/common_test.py
```diff
@@ -206,3 +206,142 @@ def test_get_lowering_config(tuner_ctx: common.TunerContext) -> None:
...
+    ), f"Expected first matcher to be 'match', got '{foreach_match_op.matchers[0].value}'"
+    assert (
+        foreach_match_op.matchers[1].value == "inner_module_b_match"
+    ), f"Expected second matcher to be 'inner_module_b_match', got '{foreach_match_op.matchers[1].value}'"
+    assert (
+        foreach_match_op.actions[0].value == "apply_op_config"
+    ), f"Expected first action to be 'apply_op_config', got '{foreach_match_op.actions[0].value}'"
+    assert (
+        foreach_match_op.actions[1].value == "inner_module_b_apply_op_config"
+    ), f"Expected second action to be 'inner_module_b_apply_op_config', got '{foreach_match_op.actions[1].value}'"
```
kuhar, line 347: We shouldn't need to repeat the same tests for the logic already tested in IREE. Here, I'd only check that that there are no nested modules and that there is a `__kernel_config` op.

kuhar (commented):
I think it would be useful to expose this as a small utility executable in the tuner that can be invoked on demand to merge tuning specs. But this can be done in a separate PR

tuner/tuner/common.py
```diff
@@ -266,3 +268,48 @@ class MLIRTransformation:
     template: list[str]
     modified: str
     embeddable: str
+
+
+# Puts multiple input modules `td_specs` into a single top-level container module.
+# This function does *not* attempt to merge or link `td_specs` across modules.
```
kuhar, line 274: This should be a docstring

tuner/tuner/common.py
```diff
@@ -266,3 +268,48 @@ class MLIRTransformation:
...
+            top_module.body.append(td_spec.operation.clone())
+        return top_module
+
+
+# Links multiple input modules (`td_specs`) into a single tuning specification module.
+# First, the input modules are combined into a container module. Then, the external
+# `iree-opt` tool is invoked with the `--iree-codegen-link-tuning-specs` pass to
+# link or merge the individual tuning specs. When all input specs are marked with the
+# default attribute `iree_codegen.tuning_spec_with_default_entrypoint`, they are merged
+# into one tuning spec.
```
kuhar, line 294: This should be a docstring

tuner/tuner/common_test.py
```diff
@@ -206,3 +206,136 @@ def test_get_lowering_config(tuner_ctx: common.TunerContext) -> None:
...
+            transform.named_sequence @__kernel_config(%arg0: !transform.any_op {transform.consumed})
+            -> (!transform.any_op) attributes { iree_codegen.tuning_spec_entrypoint } {
+                %res = transform.foreach_match in %arg0 @match -> @apply_op_config
+                    : (!transform.any_op) -> (!transform.any_op)
+                transform.yield %res : !transform.any_op
+            }
+        }
+    """
+
+    second_module_str = """
```
kuhar, line 267: Could we reduce duplication by providing the same module twice?

tuner/tuner/common_test.py
```diff
@@ -206,3 +206,136 @@ def test_get_lowering_config(tuner_ctx: common.TunerContext) -> None:
...
+    """
+
+    module = ir.Module.parse(module_str, context)
+    module.operation.attributes[
+        "iree_codegen.tuning_spec_with_default_entrypoint"
+    ] = ir.UnitAttr.get()
+    with pytest.raises(RuntimeError) as exc_info:
+        common.link_tuning_specs(tuner_ctx, [module])
+
+    assert "iree-opt failed" in str(exc_info.value)
```
kuhar, line 341: This assert should be inside the `with` statement

tuner/tuner/common_test.py
```diff
@@ -206,3 +206,136 @@ def test_get_lowering_config(tuner_ctx: common.TunerContext) -> None:
...
+            }
+        }
+    """
+
+    module = ir.Module.parse(module_str, context)
+    module.operation.attributes[
+        "iree_codegen.tuning_spec_with_default_entrypoint"
+    ] = ir.UnitAttr.get()
+    with pytest.raises(RuntimeError) as exc_info:
+        common.link_tuning_specs(tuner_ctx, [module])
```
kuhar, line 339: Can you explain why you expect this to fail?

tuner/tuner/common_test.py
```diff
@@ -206,3 +206,120 @@ def test_get_lowering_config(tuner_ctx: common.TunerContext) -> None:
...
+        # iree-opt should fail due to missing named sequence @__kernel_config entrypoint required
+        # by `iree_codegen.tuning_spec_with_default_entrypoint` attribute.
```
kuhar, line 324:
```suggestion
        # by the `iree_codegen.tuning_spec_with_default_entrypoint` attribute.
```

tuner/tuner/common_test.py
```diff
@@ -206,3 +206,120 @@ def test_get_lowering_config(tuner_ctx: common.TunerContext) -> None:
...
+    module_str = """
+        module @inner_module_a
+            attributes { transform.with_named_sequence } {
+            transform.named_sequence @match(%arg: !transform.any_op {transform.readonly}) -> (!transform.any_op) {
+                transform.yield %arg : !transform.any_op
+            }
+
+            transform.named_sequence @apply_op_config(%op: !transform.any_op {transform.readonly}) {
+                transform.yield
+            }
```
kuhar, line 313: Do we need these two named sequences to make it fail because of missing __kernel_config?

kuhar (approved):

PR #1142, opened 2025-03-24
Improving batching behavior for LLM inference

rsuderman:
Batching for LLM inference performed immediate dispatches on requests
before attempting to group. This meant that almost immediately things
were split into individual requests (instead of groups of requests).

### DIFF
--- a/shortfin/python/shortfin_apps/llm/cli.py
+++ b/shortfin/python/shortfin_apps/llm/cli.py
@@ -213,14 +213,12 @@ async def main(argv):
             return self.responder.timer.elapsed()
 
     logger.log(msg=f"Setting up a tasklist of {len(prompts)} items", level=logging.INFO)
-    queue = asyncio.Queue()
     tasks = []
     for p in prompts:
         task = Task(p)
         tasks.append(task)
-        queue.put_nowait(task)
 
-    async def worker(name, queue):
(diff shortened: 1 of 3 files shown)


shortfin/python/shortfin_apps/llm/cli.py
```diff
@@ -233,16 +231,23 @@ async def worker(name, queue):
...
-    global_timer = Timer()
-    global_timer.start()
-
     logger.log(msg=f"Setting up {args.workers} workers", level=logging.INFO)
     workers = []
+    queue = asyncio.Queue()
     for i in range(args.workers):
-        w = asyncio.create_task(worker(f"worker-{i}", queue))
+        name = f"worker-{i}"
+        workerr = service.sysman.ls.create_worker(name)
```
stbaione, line 239: Typo

rsuderman: Nope, not a typo. Its so we don't stop the `async def worker` declared above

stbaione: Gotcha. I think I've typically seen the `_` convention used to prevent shadowing variables (`_worker = ...`), but this is fine

stbaione (approved):
LGTM

PR #1144, opened 2025-03-25
[sharktank] Enable sample token generation for LLM

archana-ramalingam:
Enable sample token generation for LLM inference in IREE


sharktank/sharktank/examples/paged_llm_v1.py
```diff
@@ -153,23 +203,49 @@ def prefill(self):
             attention_mask = replicate(attention_mask, tp)
             seq_block_ids_tensor = replicate(seq_block_ids_tensor, tp)
 
-        if self.dump_bins:
-            write_ndarray_to_bin(
-                token_ids.numpy(),
-                f"prefill_token_ids_{'x'.join([str(x) for x in token_ids.shape])}xi64.bin",
-            )
+        if self.dump_path is not None:
```
rsuderman, line 206: Could we move the dumper work to a separate function. It would be helpful to have a simplified branch more like:

```
if self.dump_path is not None:
	self.dump("token_ids", token_ids)
	self.dump("seq_len", self.seq_len)
	...
```

We can do something similar for decode step where we embed the step number into the string. This reduces down all the replicated dump code

sharktank/sharktank/examples/paged_llm_v1.py
```diff
@@ -32,34 +34,73 @@ def __init__(
...
-    def begin_batch(self, prompts: list[str]):
-        token_ids, seq_lens = self.tokenizer.encode(
-            prompts, pad_to_multiple_of=self.model.cache.pad_sequence_stride
-        )
-
-        token_ids = torch.tensor(token_ids, device=self.model.device)
-        seq_lens = torch.tensor(seq_lens, device=self.model.device)
-        if self.shared_cache_state is not None:
-            cache_state = self.shared_cache_state
+    def process_prompts(
```
rsuderman, line 48: Could we change this to `preprocess_prompts`. Its closer what this step actually does.

rsuderman (changes requested):

rsuderman (approved):

PR #1147, opened 2025-03-25
Limit SplitTensors to >1 shards

Alex-Vasile:
SplitTensors should only be used for sharding across multiple devices.

This PR adds asserts to their `__init__` to ensure they are constructed with at least 2 shards.

### DIFF
--- a/sharktank/sharktank/types/tensors.py
+++ b/sharktank/sharktank/types/tensors.py
@@ -1027,13 +1027,18 @@ class SplitPrimitiveTensor(ShardedTensorBase):
             from ..ops import transfer_to_logical_device
 
             assert shard_count is not None
+            assert (
+                shard_count > 1
+            ), "SplitTensor must have at least 2 shards. Use ReplicatedTensor for 1 shard."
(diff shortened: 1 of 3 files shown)


sogartar:
I think that it is important to support the degenerate case from a mathematical standpoint. I can imagine some tests may want exactly that.
In reality in this scenario the replicated and split tensors are the same, but our type system does not know that. Maybe `isinstance(split_tensor_with_1_shard, ReplicatedTensor)` should be `True`.

I assume your goal is to make sure the user does not do something stupid and catch this usually unwanted effect.

Alex-Vasile:
> I think that it is important to support the degenerate case from a mathematical standpoint. I can imagine some tests may want exactly that.

I'm still getting up to speed, so why some tests want split with 1 shard versus 2, or a different tensor type?

> I assume your goal is to make sure the user does not do something stupid and catch this usually unwanted effect.

Exactly. And pushing all 1 shard usage onto ReplicatedTensor also makes the internal logic easier.

sogartar:
You could argue similarly that doing a `torch.split` should also throw if splitting into 1 piece. I guess if we ever want this special case we will remove the restriction.

sogartar (approved):

PR #1149, opened 2025-03-25
Pipeline parallelism refactor

Alex-Vasile:
- Remove .pinned parameter, requiring manually specifying devices and transferring between them with an inter-block callback function.
- Simplify .clone to stop it from transferring shards between devices.

### DIFF
--- a/sharktank/tests/ops/pipeline_parallelized_test.py
+++ b/sharktank/tests/ops/pipeline_parallelized_test.py
@@ -99,6 +99,9 @@ class AllReduceTest(unittest.TestCase):
 
 
 class CatTest(unittest.TestCase):
+    @pytest.mark.skip(
+        reason="reshard_split implementation needs to be changed to support pipeline parallelism",
+    )
     def testCatSplitDimPinned(self):
         """Concatenation along the sharded split dimension."""
         shard_dim = 1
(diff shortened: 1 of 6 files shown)


sharktank/tests/ops/pipeline_parallelized_test.py
```diff
@@ -294,39 +141,34 @@ def testCatNonSplitDimPinned(self):
             unsharded_result, count=shard_count, dim=shard_dim
         )
         expected_result = expected_result.clone(
-            devices=tuple(4 + i for i in range(shard_count)), pinned=True
+            devices=tuple(4 + i for i in range(shard_count))
```
rsuderman, line 144: Usually I recommend putting a `variable = 4` above then using the variable so the value has a defined explanation.

rsuderman (approved):

sharktank/tests/ops/pipeline_parallelized_test.py
Alex-Vasile: Will fix when I re-enable. What I'm trying to do is ensure that there isn't overlap between the specified devices and the default devices. I am changing it to `tuple(range(shard_count, 2 * shard_count))` to guarantee the lack of overlap and I'll add a comment.

PR #1150, opened 2025-03-25
Implement variations of sharded ops needed for pipeline parallelism

Alex-Vasile:


sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -595,6 +595,14 @@ def equal_split(a: SplitPrimitiveTensor, b: AnyTensor) -> bool:
     return a.is_deep_equal(b)
 
 
+@expand.override(ReplicatedTensor)
+def expand_replicated(tensor: ReplicatedTensor, shape: List[int]) -> ReplicatedTensor:
+    assert len(shape) == len(tensor.shape)
```
sogartar, line 600: Technically we don't need this assert. If something is wrong the underlying `torch.Tensor.expand` op will fail.
Also this assert restricts some valid expands where new leading dimensions are inserted. Without the assert we would support this for free.

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -852,6 +879,22 @@ def linear_sharded(
 # Sharded matmuls.
 
 
+@matmul.override(ReplicatedTensor, ReplicatedTensor)
+def matmul_replicated(
+    lhs: ReplicatedTensor, rhs: SplitPrimitiveTensor, *, transpose_rhs: bool
```
sogartar, line 884: The `rhs` type hit is wrong.

sogartar (approved):
For a while I wanted to make a generic replicated implementation. Most ops should just call the underlying op on the shards. We would then be able to get rid of all these implementations.

Alex-Vasile:
> For a while I wanted to make a generic replicated implementation. Most ops should just call the underlying op on the shards. We would then be able to get rid of all these implementations.

That sounds like a good idea. I guess we could hook that into `SignatureDispatcher`, pattern match if a torch op with the same name and signature exists, and if so apply it to the shards.

PR #1157, opened 2025-03-25
Fix the missing of --use-hf in llama benchmark test

AmosLewis:


aviator19941 (commented):
It seems like we need to update the compilation flags too to here:
```
../iree-build-no-trace/tools/iree-compile f8_attn.mlir \
  --iree-hip-target=gfx942 \
  -o=f8_attn.vmfb \
  --iree-hal-target-device=hip \
  --iree-opt-level=O3 \
  --iree-hal-indirect-command-buffers=true \
  --iree-stream-resource-memory-model=discrete \
  --iree-hal-memoization=true
  ```

AmosLewis:
Use this instead https://github.com/nod-ai/shark-ai/pull/1191 to get rid of CI error from  HF token as a github secret to be able to download.

PR #1109, opened 2025-03-18
[sharktank] Enable Perplexity test to use GPU in eager mode

archana-ramalingam:
Enable Perplexity test to use GPU in eager mode to speed up nightly CIs taking it from 1 hour to 2 mins for bs=100.

PR #1158, opened 2025-03-26
Relayout KV cache for internal reduction dimensions

rsuderman:
Reduction dimensions should be on the inner most dimensions. Currently implementation wrote heads post timesteps which places a parallel dimension post the reduction dimension.


sharktank/sharktank/layers/paged_attention.py
```diff
@@ -201,12 +203,13 @@ def read(
         page_stride = self.transformer_block_count
 
         transformer_block_index = torch.full(
-            (bs, block_seq_len), transformer_block_index, device=self.device
+            (bs, block_seq_len), transformer_block_index
```
archana-ramalingam, line 206: Added `device` in this [PR](https://github.com/nod-ai/shark-ai/pull/1109/files#diff-03b0443ffcc482ea8232db76055286ac72397091ec7a134da31f457b2aa9409cR204) to enable GPU for eager mode. Operation in the next line fails without it.

PR #1161, opened 2025-03-26
Shortfin llm beam search decode

stbaione:
- Contains implementation + unit testing for `beam_search` token selection strategy.
- Small refactor in `BeamGroup` to support the `beam_search` flow + fix existing bug
- Introduce the `Beam` class as a layer of abstraction between the `LlmInferenceExecRequest` and the `BeamGroup` operations
- Fix to `log_softmax`, so that it doesn't hang for large `device_arrays` w/ corresponding test

## Potential Improvements:
- Optimize `log_softmax`. For a device_array representative to the logits of a decode invocation, torch was taking ~.3ms where sf was taking ~2.7ms.
    - `torch` seems to use an `aten` kernel for `log_softmax`, which has some further optimizations, like parsing and performing operations on slices of the tensor in parallel ([implementation](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/SoftMax.cpp#L153))


shortfin/python/array_host_ops.cc
```diff
@@ -906,29 +907,29 @@ void BindArrayHostOps(py::module_ &m) {
...
-          // axis, and c (maxExpanded) is the maximum value along that axis.
-          auto sum_expression = xt::sum(xt::exp(*input_t - maxExpanded), axis);
-          auto sum_axis_expanded = xt::expand_dims(sum_expression, axis);
-          auto log_sum_expression = xt::log(sum_axis_expanded);
-
-          auto result = *input_t - maxExpanded - log_sum_expression;
+
+          auto max_vals =
+              xt::amax(*input_t, axis, xt::evaluation_strategy::immediate);
+          auto max_shape = max_vals.shape();
```
rsuderman, line 913: You should be making an `std::vector` of these and then performing insertions. There is no guarantee that `max_vals.shape()` would return a mutable type (and potentially dangerous if it does).

rsuderman (changes requested):

shortfin/python/array_host_ops.cc
stbaione: Thanks! Updated

shortfin/python/shortfin_apps/utils.py
```diff
@@ -105,6 +105,60 @@ async def run(self):
...
+    frac = value & frac_mask
+
+    # Bias calculation
+    bias = (1 << (exp_bits - 1)) - 1
+
+    # Compute the float value
+    if exponent == 0:
+        if frac == 0:
+            return -0.0 if sign else 0.0  # Zero
+        else:
```
rsuderman, line 149: You shouldn't have elses when the if contains a return just do

```
if ...
   return ...
return ...
```

shortfin/python/shortfin_apps/llm/components/messages.py
```diff
@@ -51,6 +51,11 @@ def __init__(self, phase: InferencePhase, input_token_ids: list[int], rid=None):
         self._cache: BasePagedAttentionCache | None = None
         self.allocation: PageAllocation | None = None
 
+        # Fields that must be tracked for `multi_beam` related selection
+        # strategies.
+        self.score: float = 0.0
```
rsuderman, line 56: Why is this being tracked in a request? This seems like state tracking during decoding and should be store in that, not a ExecuRequest.

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/beam_group.py
```diff
@@ -47,41 +51,52 @@ async def wait(self):
         done_signals = [req.done for req in self.active_exec_reqs]
         return await gather(*done_signals)
 
+    def _update_exec_req(self, selection: ExecRequestSelection):
```
rsuderman, line 54: We should not be updating the execRequest, we should have a stateful beam / hypothesis instead. Mutating the input request feels extremely questionable.

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/beam_search_token_selection_strategy.py
```diff
@@ -0,0 +1,220 @@
...
+                cumulative_log_prob = exec_req.score + value
+
+                # Insert into sorted array
+                selection = ExecRequestSelection(
+                    exec_req,
+                    token,
+                    score=cumulative_log_prob,
+                    normalization_function=self._normalize_exec_req,
+                )
+                index = bisect.bisect_left(
```
rsuderman, line 114: Just place all hypothesis in a map, sort the keys, then return the top K. The number of responses are relatively small given we already did topk.

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/beam_search_token_selection_strategy.py
```diff
@@ -0,0 +1,220 @@
...
+        """
+        config = self.token_selection_strategy_config
+        k = config.decode_config.num_beams - len(completed_exec_reqs)
+
+        global_min_log_prob = 0.0
+
+        selections = []
+        for exec_req in active_exec_reqs:
+            # Take `log_softmax` of the logits.
+            log_softmax_logits = sfnp.log_softmax(exec_req.result_logits)
```
rsuderman, line 97: Add a TODO to conditionally do this depending on the model configuration.

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/beam_search_token_selection_strategy.py
```diff
@@ -0,0 +1,220 @@
...
+            completed_reqs (Set[LlmInferenceExecRequest]): Requests that have completed.
+
+        Returns:
+            LlmInferenceExecRequest: Highest scoring request.
+        """
+        reqs = list(completed_reqs) if completed_reqs else active_reqs
+
+        max_score = self._final_score(reqs[0])
+        selected_req = reqs[0]
+        for req in reqs[1:]:
```
rsuderman, line 163: Use python list semantics
```
def argmax(iterable):
    return max(range(len(iterable)), key=lambda i: iterable[i])
```

shortfin/python/shortfin_apps/utils.py
```diff
@@ -105,6 +105,60 @@ async def run(self):
         self.logger.info("System manager command processor stopped")
 
 
+def convert_int_to_float(value: int, dtype: sfnp.DType) -> float:
```
rsuderman, line 108: Can't you just use `struct.pack` and `struct.unpack`? This should just be a bits reinterpretation.

https://stackoverflow.com/questions/76248276/how-to-reinterpret-bits-of-float-as-int-in-python

shortfin/tests/apps/llm/components/token_selection_strategy/beam_search_token_selection_strategy_test.py
```diff
@@ -0,0 +1,627 @@
...
+        self.submit = submit_cb
+        self.reserve_workitem = workitem_cb
+        self.complete_workitem = workitem_cb
+
+
+def _batcher_workitem_callback():
+    pass
+
+
+def float_to_float16_int(value):
```
rsuderman, line 73: See `struct.pack` and `struct.unpack` comment

rsuderman (changes requested):

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/beam_search_token_selection_strategy.py
stbaione: Updated, used a list and sorted it in case two selections have the same score. May not be as common in reality, but unit tests caught the possibility

shortfin/python/shortfin_apps/utils.py
stbaione: Updated, but will be replacing function soon

shortfin/python/shortfin_apps/utils.py
stbaione: Updated

shortfin/tests/apps/llm/components/token_selection_strategy/beam_search_token_selection_strategy_test.py
stbaione: Updated

shortfin/python/shortfin_apps/llm/components/messages.py
stbaione: Created a base `Beam` class, used to encapsulate `LlmInferenceExecRequest` within `BeamGroup`, and moved most of the logic in terms of sampling logits and updating scores to the `Beam` class.

Created a `BeamSearchBeam` and `MultiGreedyBeam`, which inherit from `Beam`

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/beam_group.py
stbaione: Refactored to track scores and normalization on `Beam`

rsuderman (approved):

PR #1162, opened 2025-03-26
[shortfin] Fix artifact names for Flux schnell pipeline and add tests

sogartar:
Add a config for the schnell variant.

Use only hip target.
amdgpu is not the same as hip for the purpose of an IREE target. See https://github.com/iree-org/iree/issues/20347

Add E2E smoke tests of the server.


shortfin/python/shortfin_apps/flux/components/builders.py
```diff
@@ -70,9 +70,7 @@ def get_params_filenames(model_params: ModelParams, model=None, splat: bool = Fa
     else:
         for idx, mod in enumerate(modnames):
             # schnell and dev weights are the same, except for sampler
-            subtype = (
-                "schnell" if model_params.is_schnell and mod == "sampler" else "dev"
-            )
+            subtype = "schnell" if model_params.is_schnell else "dev"
```
KyleHerndon, line 73: For flux, it would be nice to not have multiple copies of the same artifact. Dev/schnell have everything the same other than the transformer.

Maybe I'm wrong, is there a good reason to duplicate the artifacts--in case of a change upstream or downstream?

shortfin/python/shortfin_apps/flux/examples/flux_schnell_config.json
```diff
@@ -0,0 +1,37 @@
+{
+  "base_model_name": "flux_schnell",
+  "is_schnell": true,
+  "num_latents_channels": 16,
+  "cfg_mult": 1,
+  "clip_max_seq_len": 77,
```
IanNod, line 6: Have we looked at how much the clip_max_seq_len affects perf? In the past for sdxl we saw a pretty significant perf improvement using 64.

IanNod (commented):
Can you add some form of integration test to ensure shortfin flux schnell continues to be functional and performant?

shortfin/python/shortfin_apps/flux/examples/flux_schnell_config.json
sogartar: I will add a test in another PR as this will be a big item.
For the optimal seq len we may have to dispatch to different variants or have this dimension be dynamic.
The T5 model also needs this.
Also we probably would need to change the preparation of the arguments for the Flux transformer.

shortfin/python/shortfin_apps/flux/components/builders.py
sogartar: Now they are the same. I made the always prefix with just `flux_`. The export+upload would have to be refactored anyway.

shortfin/python/shortfin_apps/flux/examples/flux_schnell_config.json
IanNod: Sure, I was just curious if we have done any tests on that so far. I wouldn't expect a dynamic shape to be performant though. Fine if we try this later if it will be a bigger item

sogartar: It is needed to construct the names of files to download.

sharktank/sharktank/pipelines/flux/export_components.py
```diff
@@ -569,33 +572,21 @@ def get_filename(args):
...
-        args.height,
-        args.width,
-        args.precision,
-        args.max_length,
-        "mlir",
-        args.weights_directory,
-        args.external_weights,
-        args.external_weights_file,
-        args.decomp_attn,
+        hf_model_name=args.model,
```
KyleHerndon, line 580: Let's update this to only include flux/schnell if its mmdit as well.

KyleHerndon (approved):

shortfin/python/shortfin_apps/flux/examples/flux_schnell_config.json
IanNod: I was specifically referencing "clip_max_seq_len" within the config file. I don't see it being currently used at all.

sharktank/sharktank/pipelines/flux/export_components.py
sogartar: Could you elaborate?

shortfin/python/shortfin_apps/flux/examples/flux_schnell_config.json
KyleHerndon: It is used in `service.py`

sharktank/sharktank/pipelines/flux/export_components.py
KyleHerndon: Currently, when you run this file, unless I am mistaken, it will produce MLIR files with `_dev` or `_schnell` in the name when the differentiation is unnecessary. This is a result of the model name from args.model being used directly in the output name.

IanNod: is the differentiation unnecessary? I had thought there were small differences in the model, and therefore would expect the IR to be slightly different

.github/workflows/ci-shortfin-nightly.yml
```diff
@@ -0,0 +1,76 @@
...
+        if dpkg -s ninja-build &>/dev/null; then
+          echo 'ninja is installed'
+        else
+          sudo apt install ninja -y
+        fi
+
+    - name: Setup Python
+      uses: actions/setup-python@42375524e23c412d93fb67b49958b491fce71c38 # v5.4.0
+      with:
+        python-version: "3.12"
```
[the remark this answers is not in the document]
sogartar: No.

sharktank/sharktank/pipelines/flux/export_components.py
sogartar: The exact names of exported components are
```
flux_clip_bs1_77_bf16.mlir
flux_dev_sampler_bs1_512_1024x1024_bf16.mlir
flux_dev_scheduler_bs1_512_bf16.mlir
flux_schnell_sampler_bs1_512_1024x1024_bf16.mlir
flux_schnell_scheduler_bs1_512_bf16.mlir
flux_t5xxl_bs1_512_bf16.mlir
flux_vae_bs1_1024x1024_bf16.mlir

```

PR #1163, opened 2025-03-26
refactored linux-mi300-1gpu-ossci to linux-mi300-1gpu-ossci-nod-ai

Eliasj42:
refactored linux-mi300-1gpu-ossci to linux-mi300-1gpu-ossci-nod-ai

### DIFF
--- a/.github/workflows/ci-llama-large-tests.yaml
+++ b/.github/workflows/ci-llama-large-tests.yaml
@@ -29,7 +29,7 @@ jobs:
       matrix:
         version: [3.11]
       fail-fast: false
-    runs-on: linux-mi300-1gpu-ossci
+    runs-on: linux-mi300-1gpu-ossci-nod-ai
     defaults:
       run:
         shell: bash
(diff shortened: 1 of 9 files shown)


renxida (approved):
Just to make sure I understand ... using tag `linux-mi300-1gpu-ossci-nod-ai` gives us the whole cluster + the portion reserved for the shark org, right?

Eliasj42:
> Just to make sure I understand ... using tag `linux-mi300-1gpu-ossci-nod-ai` gives us the whole cluster + the portion reserved for the shark org, right?

This runner is the same as the old one.  However this was the first ossci runner I created, and the name ended up not being descriptive enough.  I wanted to clean this up, so we don't have to remember that the 1gpu runner that doesn't specify an org is linked to nod-ai

renxida:
> This runner is the same as the old one. However this was the first ossci runner I created, and the name ended up not being descriptive enough. I wanted to clean this up, so we don't have to remember that the 1gpu runner that doesn't specify an org is linked to nod-ai

Got it. Sounds good to me, will use new name in future.

PR #1166, opened 2025-03-26
Simplify __init__ for ShardedTensor and subclasses

Alex-Vasile:
Follow up from https://github.com/nod-ai/shark-ai/pull/1008#discussion_r1985012786

### DIFF
--- a/sharktank/sharktank/types/tensors.py
+++ b/sharktank/sharktank/types/tensors.py
@@ -755,6 +755,7 @@ class ShardedTensor(InferenceTensor):
     def __init__(
         self,
         *,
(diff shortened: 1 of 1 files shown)


sogartar (approved):
I recall that at some point my intention for `ShardedTensor` was to not know how exactly the tensor is sharded as potentially you may have more complicated arrangements with more than 1 tensor per device. We already have `shard_dim` as init argument, so the separation was not clean anyway.
Right now we only have 1 tensor per device. It is fine to enforce this in the base type.

Alex-Vasile:
`Requests.exceptions.ConnectionError: (ProtocolError('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer'))`

That's strange reason for test failure.

PR #1171, opened 2025-03-27
[tuner] add the option of providing starter td spec

bangtianliu:
This PR adds implementation for providing a starter td spec, enabling incremental merging of tuning specs when tuning across different dispatches in a real model. 

Testing has been done from my end and verified that it works by supplying td specs that include matchers for attention ops or identical matmul ops, demonstrating that merging behaves correctly based on matcher overlap.

Issue: #810

### DIFF
--- a/tuner/examples/simple/simple_tuner.py
+++ b/tuner/examples/simple/simple_tuner.py
@@ -70,12 +70,6 @@ def main():
         default="",
         help="Path to the flags file for iree-benchmark-module for model benchmarking.",
     )
-    client_args.add_argument(
-        "--simple-starter-td-spec",
-        type=str,
-        default="",
-        help="Path to a starter td spec file to merge with tuning spec files.",
-    )
     # Remaining arguments come from libtuner
     args = libtuner.parse_arguments(parser)
 
--- a/tuner/examples/simple/simple_tuner.py
+++ b/tuner/examples/simple/simple_tuner.py
@@ -70,9 +70,14 @@ def main():
         default="",
         help="Path to the flags file for iree-benchmark-module for model benchmarking.",
     )
+    client_args.add_argument(
+        "--simple-starter-td-spec",
+        type=str,
+        default="",
+        help="Path to a starter td spec file to merge with tuning spec files.",
+    )
     # Remaining arguments come from libtuner
     args = libtuner.parse_arguments(parser)
-
     path_config = libtuner.PathConfig()
     path_config.base_dir.mkdir(parents=True, exist_ok=True)
     # TODO(Max191): Make candidate_trackers internal to TuningClient.
--- a/tuner/examples/simple/simple_tuner.py
+++ b/tuner/examples/simple/simple_tuner.py
@@ -78,6 +78,7 @@ def main():
     )
     # Remaining arguments come from libtuner
     args = libtuner.parse_arguments(parser)
+
     path_config = libtuner.PathConfig()
     path_config.base_dir.mkdir(parents=True, exist_ok=True)
     # TODO(Max191): Make candidate_trackers internal to TuningClient.
--- a/tuner/tuner/candidate_gen.py
+++ b/tuner/tuner/candidate_gen.py
@@ -243,7 +243,7 @@ def generate_configs_and_td_specs(
         td_spec_module = dispatch_tuner.get_td_spec(input_module, config)
         assert td_spec_module, "Failed to generate transform dialect spec"
 
-        # if starter td spec is not provided, use the generated td spec directly.
+        # If starter td spec is not provided, use the generated td spec directly.
         if starter_td_spec is None:
             config_specs.append(td_spec_module)
             continue
@@ -264,7 +264,7 @@ def generate_configs_and_td_specs(
             log_duplicates=log_duplicates,
         )
 
-        td_spec_module = link_tuning_specs(tuner_context.mlir_ctx, td_specs_to_link)
+        td_spec_module = link_tuning_specs(tuner_context, td_specs_to_link)
         config_specs.append(td_spec_module)
 
     tune_logger.debug(f"Generated {len(config_specs)} tuning specs")
--- a/tuner/tuner/candidate_gen.py
+++ b/tuner/tuner/candidate_gen.py
@@ -147,6 +147,32 @@ def get_default_output_dir() -> str:
     return "tuning_" + datetime.now().strftime("%Y_%m_%d_%H_%M")
 
 
+def check_td_spec_matchers_overlap(
+    starter_matchers: set[str],
+    current_matchers: set[str],
+    warned_overlap_matchers: set[str],
+) -> tuple[bool, set[str]]:
+    """
+    Determines whether tuning specs should be linked and new overlapping matchers for warning.
+
+    Args:
+        starter_matchers: Matcher names from the starter tuning spec.
+        current_matchers: Matcher names from the current tuning spec.
+        warned_overlap_matchers: Set of matchers already warned about for overlaps.
+
+    Returns:
+        should_link: True if starter matchers contain any target operations not in the current spec.
+        new_warned_matchers: Set of oerlapping matchers that havenot' been warned about yet.
+    """
+
+    overlap_matchers = starter_matchers & current_matchers
+    unique_starter_matchers = starter_matchers - current_matchers
+    new_warned_matchers = overlap_matchers - warned_overlap_matchers
+
+    should_link = bool(unique_starter_matchers)
+    return should_link, new_warned_matchers
+
+
 def generate_configs_and_td_specs(
     input_module: ir.Module,  # Path to the mlir file to be tuned
     tuner_context: TunerContext,
@@ -182,7 +208,29 @@ def generate_configs_and_td_specs(
     assert len(variant_op_list) == 1, "Expect one executable variant op"
     variant_op = variant_op_list[0]
     mma_list = iree_codegen.query_mma_intrinsics(variant_op)
+    if starter_td_spec == None:
+        for i, config in enumerate(
+            generate_solutions(
+                tuner_context,
+                problem_size,
+                num_subgroups,
+                mma_list,
+                allowed_waves_per_eu,
+                pipeline_options_search_space,
+                codegen_pipeline,
+            )
+        ):
+            if i >= limit:
+                break
+            tune_logger.debug(f"Solution #{i+1}: {config}")
+            td_spec_module = dispatch_tuner.get_td_spec(input_module, config)
+            assert td_spec_module, "Failed to generate transform dialect spec"
+            config_specs.append(td_spec_module)
+            tune_logger.debug(f"Generated {len(config_specs)} tuning specs")
+            return config_specs
+
     warned_overlap_matchers: set[str] = set()
+    starter_matchers = get_matcher_names_from_td_spec(starter_td_spec)
     for i, config in enumerate(
         generate_solutions(
             tuner_context,
@@ -198,24 +246,25 @@ def generate_configs_and_td_specs(
             break
         tune_logger.debug(f"Solution #{i+1}: {config}")
         td_spec_module = dispatch_tuner.get_td_spec(input_module, config)
-        if starter_td_spec != None:
-            starter_matchers = get_matcher_names_from_td_spec(starter_td_spec)
-            current_matchers = get_matcher_names_from_td_spec(td_spec_module)
-            overlap_matchers = starter_matchers & current_matchers
-            unique_stater_matchers = starter_matchers - current_matchers
-            new_warnings = overlap_matchers - warned_overlap_matchers
+        current_matchers = get_matcher_names_from_td_spec(td_spec_module)
 
-            # Log warnings only for newly detected overlapping target operations.
-            if new_warnings:
-                logging.warning(
-                    f"Operations have been tuned in the starter tuning spec: {sorted(new_warnings)}"
-                )
-                warned_overlap_matchers.update(new_warnings)
-            # Only link td spec and starter spec if it adds unique target operations.
-            if unique_stater_matchers:
-                td_spec_module = link_tuning_specs(
-                    tuner_context.mlir_ctx, [starter_td_spec, td_spec_module]
-                )
+        should_link, new_warned_matchers = check_td_spec_matchers_overlap(
+            starter_matchers, current_matchers, warned_overlap_matchers
+        )
+
+        # Log warnings only for newly detected overlapping target operations.
+        if new_warned_matchers:
+            logging.warning(
+                f"Operations have been tuned in the starter tuning spec: {sorted(new_warned_matchers)}"
+            )
+            warned_overlap_matchers.update(new_warned_matchers)
+
+        # Only link td spec and starter spec if it adds unique target operations.
+        if should_link:
+            td_spec_module = link_tuning_specs(
+                tuner_context.mlir_ctx, [starter_td_spec, td_spec_module]
+            )
+
         assert td_spec_module, "Failed to generate transform dialect spec"
         config_specs.append(td_spec_module)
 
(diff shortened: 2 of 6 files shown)


tuner/tuner/candidate_gen.py
```diff
@@ -196,6 +198,24 @@ def generate_configs_and_td_specs(
...
+            if new_warnings:
+                logging.warning(
+                    f"Operations have been tuned in the starter tuning spec: {sorted(new_warnings)}"
+                )
+                warned_overlap_matchers.update(new_warnings)
+            # Only link td spec and starter spec if it adds unique target operations.
+            if unique_stater_matchers:
+                td_spec_module = link_tuning_specs(
+                    tuner_context.mlir_ctx, [starter_td_spec, td_spec_module]
+                )
```
kuhar, line 218: This should be a function that comes with its own unit tests

tuner/tuner/candidate_gen.py
```diff
@@ -196,6 +246,25 @@ def generate_configs_and_td_specs(
             break
         tune_logger.debug(f"Solution #{i+1}: {config}")
         td_spec_module = dispatch_tuner.get_td_spec(input_module, config)
+        current_matchers = get_matcher_names_from_td_spec(td_spec_module)
+
+        should_link, new_warned_matchers = check_td_spec_matchers_overlap(
+            starter_matchers, current_matchers, warned_overlap_matchers
+        )
```
kuhar, line 253: If you make `check_td_spec_matchers_overlap` accept td specs as arguments and call `get_matcher_names_from_td_spec` internally, you can make it take a list of tuning specs. For the case with no starter spec, the list will always have a single element.

tuner/tuner/candidate_gen.py
```diff
@@ -196,6 +246,25 @@ def generate_configs_and_td_specs(
...
+            logging.warning(
+                f"Operations have been tuned in the starter tuning spec: {sorted(new_warned_matchers)}"
+            )
+            warned_overlap_matchers.update(new_warned_matchers)
+
+        # Only link td spec and starter spec if it adds unique target operations.
+        if should_link:
+            td_spec_module = link_tuning_specs(
+                tuner_context.mlir_ctx, [starter_td_spec, td_spec_module]
+            )
```
kuhar, line 266: Similarly, linking should be able to handle a single tuning spec passed to it

kuhar (changes requested):

tuner/examples/simple/simple_tuner.py
```diff
@@ -70,9 +70,14 @@ def main():
...
     )
+    client_args.add_argument(
+        "--simple-starter-td-spec",
+        type=str,
+        default="",
+        help="Path to a starter td spec file to merge with tuning spec files.",
+    )
     # Remaining arguments come from libtuner
     args = libtuner.parse_arguments(parser)
-
```
kuhar, line 75: Can you undo this formatting change?

tuner/tuner/candidate_gen.py
```diff
@@ -147,6 +147,50 @@ def get_default_output_dir() -> str:
     return "tuning_" + datetime.now().strftime("%Y_%m_%d_%H_%M")
 
 
+def determine_td_specs_to_link(
+    td_specs: list[ir.Module], warned_overlap_matchers: set[str]
+) -> tuple[list[ir.Module], set[str]]:
+    """
+    Determines which tuning specs should be linked based on matcher overlap.
```
kuhar, line 154: I don't understand this sentence. All tuning specs should be linked, no?

tuner/tuner/candidate_gen.py
```diff
@@ -197,6 +244,22 @@ def generate_configs_and_td_specs(
         tune_logger.debug(f"Solution #{i+1}: {config}")
         td_spec_module = dispatch_tuner.get_td_spec(input_module, config)
         assert td_spec_module, "Failed to generate transform dialect spec"
+
+        td_specs_to_link, new_warned_matchers = determine_td_specs_to_link(
+            [starter_td_spec, td_spec_module]
+            if starter_td_spec is not None
+            else [td_spec_module],
```
kuhar, line 251: Make this list a local variable and append to it instead of having this branching logic

tuner/tuner/candidate_gen_test.py
```diff
@@ -207,3 +207,69 @@ def test_get_td_spec_convolution(tuner_ctx: common.TunerContext) -> None:
...
+
+            transform.named_sequence @match_bar(%arg0: !transform.any_op {transform.readonly}) -> (!transform.any_op) {
+                transform.yield %arg0 : !transform.any_op
+            }
+
+            transform.named_sequence @__kernel_config(%arg0: !transform.any_op ) -> !transform.any_op
+                attributes { iree_codegen.tuning_spec_entrypoint } {
+                %0 = transform.foreach_match in %arg0
+                @match_foo -> @apply_op_config
+                , @match_bar -> @apply_op_config
```
kuhar, line 232: The indentation is weird here

tuner/tuner/candidate_gen.py
```diff
@@ -147,6 +147,50 @@ def get_default_output_dir() -> str:
...
+
+    Args:
+        td_specs: A list of 1 or 2 tuning spec modules. If two are provided, the first is
+                the starter spec.
+        warned_overlap_matchers: Set of matchers already warned about for overlaps.
+
+    Returns:
+         A tuple containing:
+        - A list of td specs to link (possibly excluding the starter spec).
+        - A set of overlapping matchers that haven't been warned about yet.
```
kuhar, line 167: The indentation is weird here

tuner/tuner/candidate_gen.py
```diff
@@ -147,6 +147,50 @@ def get_default_output_dir() -> str:
...
+    starter_td_spec = td_specs[0]
+    current_td_spec = td_specs[1]
+
+    starter_matchers = get_matcher_names_from_td_spec(starter_td_spec)
+    current_matchers = get_matcher_names_from_td_spec(current_td_spec)
+
+    overlap_matchers = starter_matchers & current_matchers
+    unique_starter_matchers = starter_matchers - current_matchers
+
+    # Determine overlapping matchers which haven't been warned about yet.
```
kuhar, line 184: I can't parse this sentence, the grammar seems off.

tuner/tuner/candidate_gen.py
```diff
@@ -147,6 +147,50 @@ def get_default_output_dir() -> str:
...
+         A tuple containing:
+        - A list of td specs to link (possibly excluding the starter spec).
+        - A set of overlapping matchers that haven't been warned about yet.
+    """
+
+    if len(td_specs) == 1:
+        # No starter td spec provided, nothing to merge.
+        return td_specs, set()
+
+    assert len(td_specs) == 2, "Expected exactly two TD specs (starter and current)"
```
kuhar, line 174: nit: I'd put this assert on top and check for `len(td_specs) <= 2` and also allow for empty inputs. This way the preconditions are clear.

tuner/tuner/candidate_gen.py
```diff
@@ -147,6 +147,50 @@ def get_default_output_dir() -> str:
...
+    starter_td_spec = td_specs[0]
+    current_td_spec = td_specs[1]
```
kuhar, line 176: nit: you can unpack it like this
```suggestion
    starter_td_spec, current_td_spec = td_specs
```

tuner/tuner/candidate_gen.py
```diff
@@ -147,6 +147,50 @@ def get_default_output_dir() -> str:
...
+
+    starter_matchers = get_matcher_names_from_td_spec(starter_td_spec)
+    current_matchers = get_matcher_names_from_td_spec(current_td_spec)
+
+    overlap_matchers = starter_matchers & current_matchers
```
kuhar, line 181:
```suggestion
    overlapping_matchers = starter_matchers & current_matchers
```

tuner/tuner/candidate_gen.py
```diff
@@ -147,6 +147,50 @@ def get_default_output_dir() -> str:
...
+    unique_starter_matchers = starter_matchers - current_matchers
+
+    # Determine overlapping matchers which haven't been warned about yet.
+    new_warned_matchers = overlap_matchers - warned_overlap_matchers
```
kuhar, line 185: `new_diplicate_matchers`?

tuner/tuner/candidate_gen.py
```diff
@@ -147,6 +147,50 @@ def get_default_output_dir() -> str:
     return "tuning_" + datetime.now().strftime("%Y_%m_%d_%H_%M")
 
 
+def determine_td_specs_to_link(
+    td_specs: list[ir.Module], warned_overlap_matchers: set[str]
+) -> tuple[list[ir.Module], set[str]]:
```
kuhar, line 152: Instead of having the caller provide the known duplicate matchers and returning new duplicated, I think it would be easier to add a boolean flag that decides whether to print a warning about duplicates. This simplifies the interface. If you want to test the duplicate finding logic, you can make it a helper function and test in isolation.

Max191 (changes requested):
Would also be good to document how the starting spec works a bit better. It's unclear right now whether the starting spec or the new tuned spec will take priority in the linked spec.

tuner/tuner/candidate_gen.py
```diff
@@ -197,6 +242,21 @@ def generate_configs_and_td_specs(
         tune_logger.debug(f"Solution #{i+1}: {config}")
         td_spec_module = dispatch_tuner.get_td_spec(input_module, config)
         assert td_spec_module, "Failed to generate transform dialect spec"
+
```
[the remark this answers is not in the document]
kuhar: It would be good to measure the candidate generation time before and after and decide based on this.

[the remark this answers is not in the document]
Max191: The other benefit of doing this is that we reduce file sizes in the tuning tmp directory, and it is easier to look through any generated dispatch candidates, since there is less clutter from unnecessary linked IR.

Measurements are always good to have, though.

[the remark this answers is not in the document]
kuhar: On the other hand havin two different td script inherently increases the complexity. It may be annoying to have to deal with potential bugs where we get the expected attributes when compiling dispatches but not model caused by issues with linking etc. I'd rather learn whether the specs are working or not when dealing with candidates.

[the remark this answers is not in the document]
bangtianliu: I did a measurement, the linking time is around 0.047 seconds for linking matmul and attention specs. Maybe just skip linking when the number of td specs is 1.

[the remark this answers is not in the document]
bangtianliu: And also I did another measurement about whole candidate generation time with `--num-candidates=20`
The command I used for measurement `python -m examples.simple double_mmt.mlir mmt_benchmark.mlir --simple-compile-flags-file=examples/simple/compile_flags.txt --simple-model-benchmark-flags-file=examples/simple/model_benchmark_flags.txt --devices=hip://0 --simple-num-dispatch-candidates=5 --simple-num-model-candidates=3 --num-candidates=20 --simple-starter-td-spec=attention.mlir ` 

it is 1.38 seconds before (just original code without starter td spec);  it is 2.61 seconds after applying the change from this PR including determining td spec to link and the overhead of linking td specs.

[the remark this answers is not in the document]
Max191: So, sounds like about 0.05s per candidate. With something like 8000 candidates, this adds close to 7 minutes to the candidate generation. Personally, I think it is worth skipping the linking for dispatch candidates, but I don't have a particularly strong opinion, since most of the tuning time is still in benchmarking anyway. If it becomes annoying to deal with, we can always optimize later on.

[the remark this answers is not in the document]
kuhar: > I did a measurement, the linking time is around 0.047 seconds for linking matmul and attention specs. Maybe just skip linking when the number of td specs is 1.

@bangtianliu with how many candidates generated?

> With something like 8000 candidates, this adds close to 7 minutes to the candidate generation.

This is something we should be able to parallelize easily -- (8000 * 0.05s) / 192 = 2s, which seems more than acceptable to me.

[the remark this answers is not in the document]
bangtianliu: > > I did a measurement, the linking time is around 0.047 seconds for linking matmul and attention specs. Maybe just skip linking when the number of td specs is 1.
> 
> @bangtianliu with how many candidates generated?
> 
> > With something like 8000 candidates, this adds close to 7 minutes to the candidate generation.
> 
> This is something we should be able to parallelize easily -- (8000 * 0.05s) / 192 = 2s, which seems more than acceptable to me.

The 0.047 seconds is the overhead for a single linking operation (i.e., one candidate). The second test above measures the total time for 20 candidates.

tuner/tuner/candidate_gen.py
```diff
@@ -197,6 +242,32 @@ def generate_configs_and_td_specs(
...
+        # spec will be excluded from linking.
+        td_specs_to_link = determine_td_specs_to_link(
+            td_specs,
+            log_duplicates=log_duplicates,
+        )
+        if len(td_specs_to_link) == 2:
+            td_spec_module = link_tuning_specs(tuner_context.mlir_ctx, td_specs_to_link)
+        else:
+            # avoid unnessary link overhead.
+            td_spec_module = td_specs_to_link[0]
```
kuhar, line 270: You can put this micro-optimization into `link_tuning_specs` to simplify the logic and make this testable.

Actually, do we need this at all? The `continue` a few lines above already handles this case, doesn't it?

bangtianliu: We do need it, as this condition is caused by determine_td_specs_to_link, which will exclude the starter spec if all of its matchers are covered by a tuner-generated candidate spec.

tuner/tuner/libtuner.py
```diff
@@ -338,6 +338,21 @@ def parse_arguments(
...
+    """
+    Providing a starter td spec enables incremental merging of TD specs across different dispatches
+    when tuning real models.
+
+    If there are conflicts on target operations (e.g., duplicate matchers), the candidate spec takes
+    precedence, and a warning will be issued to notify the user. If the candidate spec fully covers all
+    operations from the starter spec, the starter td spec will be excluded from linking.
+    """
+    general_args.add_argument(
+        "--simple-starter-td-spec",
```
Max191, line 350: Since this isn't part of the simple example anymore, please rename to `starter-td-spec`

tuner/tuner/libtuner.py
```diff
@@ -338,6 +338,21 @@ def parse_arguments(
...
+        type=str,
```
Max191, line 351: Since it is a path, can you use `Path` for the type instead of `str`?

tuner/tuner/libtuner.py
```diff
@@ -338,6 +338,21 @@ def parse_arguments(
...
+        default="",
+        help="Path to a starter td spec file to merge with tuning spec files.",
```
Max191, line 353: Can you add a short statement that the newly generated candidates take precedence over starter specs to the `help` description? Users won't want to look into the code and find the above comments.

tuner/tuner/candidate_gen.py
```diff
@@ -197,6 +242,29 @@ def generate_configs_and_td_specs(
         tune_logger.debug(f"Solution #{i+1}: {config}")
         td_spec_module = dispatch_tuner.get_td_spec(input_module, config)
         assert td_spec_module, "Failed to generate transform dialect spec"
+
+        # if starter td spec is not provided, use the generated td spec directly.
```
kuhar, line 246:
```suggestion
        # If starter td spec is not provided, use the generated td spec directly.
```

tuner/tuner/common.py
```diff
@@ -272,14 +273,12 @@ class MLIRTransformation:
     embeddable: str
 
 
-def combine_tuning_specs(
-    tuner_ctx: TunerContext, td_specs: list[ir.Module]
-) -> ir.Module:
+def combine_tuning_specs(context: ir.Context, td_specs: list[ir.Module]) -> ir.Module:
```
kuhar, line 276: Why did you change this? Passing the larger context around in the tuner code seems fine to me

kuhar (approved):
LGTM % nit

kuhar (approved):

PR #1173, opened 2025-03-27
[sharktank] run unit tests job on MI300x machine as well

sogartar:
There are some tests that target not only the CPU device.


sogartar:
The MI300X machine seems quite busy. Maybe we can have the tests as not required.

.github/workflows/ci-sharktank.yml
```diff
@@ -23,25 +23,30 @@ concurrency:
...
             python-version: "3.11"
             torch-version: "2.3.0"
-          - os: windows-2022
+          - runs-on: windows-2022
             python-version: "3.12"
             torch-version: "2.4.1"
+          - runs-on: linux-mi300-1gpu-ossci-nod-ai
+            python-version: "3.12"
+            torch-version: "2.5.1"
+            iree-target-args: "--iree-hip-target=gfx942 --iree-hal-target-device=hip"
```
ScottTodd, line 43: > If the MI300X machine would need to be reliable enough since failures would block merging.
> The MI300X machine seems quite busy. Maybe we can have the tests as not required.

The policy we set in IREE is https://iree.dev/developers/general/github-actions/#using-github-hosted-and-self-hosted-runners
> Due to how difficult it can be to keep self-hosted runners operating reliably, and how access is limited to the group that maintains the hardware, any jobs using these self-hosted runners must be optional and easy to disable.

Some ideas for how to keep jobs not required:

* Add a separate workflow file that uses self hosted runners, like https://github.com/iree-org/iree-turbine/blob/main/.github/workflows/ci-tk.yaml
* Add a separate job with its own matrix in this workflow file
* Add a `continue-on-error` in the matrix like https://github.com/iree-org/iree/blob/5c7c1f2a3c1d3cab7e6c8f659a33fbf164fd929d/.github/workflows/build_package.yml#L115. Note that the matrix will not complete until all jobs have at least been scheduled and run to completion (pass/fail/cancel), so it probably isn't a good fit here.

.github/workflows/ci-sharktank.yml
```diff
@@ -23,25 +23,30 @@ concurrency:
...
     strategy:
       matrix:
         python-version: ["3.11", "3.12"]
         torch-version: ["2.3.0", "2.4.1", "2.5.1"]
-        os: [ubuntu-24.04]
+        iree-target-args: [""]
+        runs-on: [ubuntu-24.04]
         include:
-          - os: windows-2022
+          - runs-on: windows-2022
```
ScottTodd, line 34: Nice naming from `os` to `runs-on`, thanks!

.github/workflows/ci-llama-large-tests.yaml
```diff
@@ -68,7 +68,13 @@ jobs:
       - name: Run llama tests
         run: |
           source ${VENV_DIR}/bin/activate
-          pytest sharktank/tests/models/llama/benchmark_amdgpu_test.py -v -s --run-nightly-llama-tests --iree-hip-target=gfx942 --iree-device=hip://0 --html=out/llm/llama/benchmark/index.html
+          pytest sharktank/tests/models/llama/benchmark_amdgpu_test.py \
+            -v -s \
+            -m "expensive" \
```
ScottTodd, line 73: Thanks for splitting these out across multiple lines. Some of these small NFC changes would be useful to merge first in a smaller PR.

.github/workflows/ci-sharktank.yml
sogartar: I moved the tests to another job that is not required for merging.

PR #1174, opened 2025-03-27
[sharktank] Move expensive VAE tests to nightly

sogartar:
Change toy-size model config to circumvent compiler error for HIP. CPU still fails in another way.


KyleHerndon (approved):
Looks good!

PR #1179, opened 2025-03-27
Change ShardedTensor.clone to error out on wrong args

Alex-Vasile:
- Change how defaults are loaded into kwargs by .clone() so that incorrectly specified kwargs will error out.
  - Previously it silently used defaults and not the specified values. E.g. if you provided `shards=...` instead of `ts=...` it would use the tensor's existing shards rather than error out.
- Added tests to verify this behaviour

### DIFF
--- a/sharktank/sharktank/types/tensors.py
+++ b/sharktank/sharktank/types/tensors.py
@@ -1068,14 +1068,18 @@ class SplitPrimitiveTensor(ShardedTensorBase):
         )
 
     def clone(self, **kwargs) -> "SplitPrimitiveTensor":
-        new_kwargs = {
-            "shard_dim": kwargs.get("shard_dim", self.shard_dim),
-            "ts": kwargs.get("ts", self.shards),
-            "name": kwargs.get("name", self.name),
-            "shape": kwargs.get("shape", self.shape),
-            "devices": kwargs.get("devices", self.devices),
-        }
-        return SplitPrimitiveTensor(**new_kwargs)
+        kwargs["name"] = kwargs.get("name", self.name)
+        kwargs["devices"] = kwargs.get("devices", self.devices)
+        kwargs["shape"] = kwargs.get("shape", self.shape)
+        kwargs["shard_dim"] = kwargs.get("shard_dim", self.shard_dim)
+
+        if "ts" in kwargs:
+            # Only override shard_count if ts is a tensor.
+            if isinstance(kwargs["ts"], torch.Tensor):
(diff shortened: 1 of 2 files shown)


sharktank/sharktank/types/tensors.py
```diff
@@ -1063,14 +1063,18 @@ def __init__(
...
-        return SplitPrimitiveTensor(**new_kwargs)
+        kwargs["name"] = kwargs.get("name", self.name)
+        kwargs["devices"] = kwargs.get("devices", self.devices)
+        kwargs["shape"] = kwargs.get("shape", self.shape)
+        kwargs["shard_dim"] = kwargs.get("shard_dim", self.shard_dim)
+
+        if "ts" in kwargs:
+            # Only override shard_count if ts is a tensor.
+            if isinstance(kwargs["ts"], torch.Tensor):
+                kwargs["shard_count"] = kwargs.get("shard_count", self.shard_dim)
```
sogartar, line 1074: `self.shard_dim` looks wrong.

sharktank/sharktank/types/tensors.py
```diff
@@ -1201,17 +1205,17 @@ def __init__(
...
-            assert isinstance(new_kwargs["ts"], torch.Tensor)
-            new_kwargs["shard_count"] = kwargs["shard_count"]
+        kwargs["ts"] = kwargs.get("ts", self.shards)
+        kwargs["name"] = kwargs.get("name", self.name)
+        kwargs["devices"] = kwargs.get("devices", self.devices)
+
+        if "ts" in kwargs:
+            # Only override shard_count if ts is a tensor.
+            if isinstance(kwargs["ts"], torch.Tensor):
+                kwargs["shard_count"] = kwargs.get("shard_count", self.shard_dim)
```
sogartar, line 1215: `self.shard_dim` looks wrong.

sogartar (approved):
I was thinking before that `ts` is not a good argument name. `shards` is probably better.  I assume that is why you burned yourself by making the assumption and using the more natural name.
Maybe the classes needs proper overload notation to account for that you can pass a single tensor to split/replicate as well. Also have it as an unnamed argument.

Alex-Vasile:
> I was thinking before that ts is not a good argument name. shards is probably better. I assume that is why you burned yourself by making the assumption and using the more natural name.

Yes. This seems to be the only place they are referred to as `ts` instead of `shards`, e.g. especially since you access them by `ShardedTensor.shards`, not `ShardedTensor.ts`.

PR #1180, opened 2025-03-27
Expand_split fix

Alex-Vasile:
Current behaviour of ops.split for SplitPrimitiveTensor is not consistent with either torch.split nor with its own documentation.

See the added tests, specifically `testExpandSplitAlongSplit` should error since it's splitting along `shard_dim`, but it does not error out. Instead it silently does nothing.

This change attempts to 'fix' the behaviour to be in line with torch.split (i.e. not allow splitting along a non-singular dimension) and with its own doc (i.e. don't allow expanding along the `shard_dim`).

### DIFF
--- a/sharktank/sharktank/ops/sharded_impls.py
+++ b/sharktank/sharktank/ops/sharded_impls.py
@@ -537,26 +537,13 @@ def expand_split(
     tensor: SplitPrimitiveTensor, shape: List[int]
 ) -> SplitPrimitiveTensor:
     assert len(shape) == len(tensor.shape)
-    expanded_dims = [
-        i
-        for i, (old_dim, new_dim) in enumerate(zip(tensor.shape, shape))
-        if old_dim == 1 and new_dim != 1
-    ]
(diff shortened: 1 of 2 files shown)


sogartar:
Is this ready for review? It say `[WIP]` in the title?

Alex-Vasile:
> Is this ready for review? It say [WIP] in the title?

I guess ready for discussion is the better word. The code is ready to review, but tests are failing since they seem to be relying on this behaviour (splitting across shard_dim).

Alex-Vasile:
> Could you try with this https://github.com/nod-ai/shark-ai/tree/users/sogartar/expand_split_fix?
> 
> The `set_element` quirk is to account for the case if the split is uneven. We have not encountered such a case yet, but if the split dimension size is not divisible by the number of shards we would have an uneven split.

I see the mistake I had. Check out the current version. We can make use of -1 for the split dim to shorten it further.

rsuderman (approved):

PR #1183, opened 2025-03-27
Rework batcher for a reservation system

rsuderman:
With decode the majority of future steps are known a groupable. Created a basic reservation system so that steps can be reserved and the continuous batcher knows that a future groupable request is coming.

This should be adapted to include grouping multiple responses, and having a "grouping" exist. E.g. multiple parallel requests in flight and grouped together for the multiple hypothesis case.

### DIFF
--- a/shortfin/python/shortfin_apps/llm/components/batcher.py
+++ b/shortfin/python/shortfin_apps/llm/components/batcher.py
@@ -104,7 +104,10 @@ class LlmBatcherProcess(BatcherProcess):
             self.strobes = 0
             return
         target_size = min(self._current_workitems, self.ideal_batch_size)
-        if waiting_count < target_size and self.strobes < 2:
+        if waiting_count < target_size:
+            logger.info("Pending workitems to be enqueued")
+            return
+        if waiting_count < self.ideal_batch_size and self.strobes < 2:
             logger.info("Waiting a bit longer to fill flight")
             return
 
--- a/shortfin/python/shortfin_apps/llm/components/batcher.py
+++ b/shortfin/python/shortfin_apps/llm/components/batcher.py
@@ -35,6 +35,14 @@ logger = logging.getLogger(__name__)
 import math
 
 
+class NewWorkItem(sf.Message):
+    ...
+
+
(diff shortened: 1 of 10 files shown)


shortfin/python/shortfin_apps/llm/components/token_selection_strategy/greedy_token_selection_strategy.py
```diff
@@ -39,6 +39,7 @@ async def decode(
             exec_req (LlmInferenceExecRequest): Execution request that has had prefill invoked on it.
         """
         config = self.token_selection_strategy_config
+        config.decode_begin_callback()
```
stbaione, line 42: Should this update be made for `MultiGreedy` too?

shortfin/python/shortfin_apps/llm/components/batcher.py
```diff
@@ -35,6 +35,14 @@
...
 
+class NewWorkItem(sf.Message):
+    ...
+
+
+class DoneWorkItem(sf.Message):
+    ...
+
+
 class LlmBatcherProcess(BatcherProcess):
```
stbaione, line 46: Just noticing, I can't find any unit tests for `BatcherProcess`, `LlmBatcherProcess`, `PrefillBatcherProcess`, or `DecodeBatcherProcess`.

Should we make an issue to add those in a subsequent PR or add them here?

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/greedy_token_selection_strategy.py
rsuderman: Yes but we can do that in a follow up. The work for multi-requests is a bit more complex as we need to give up "workitems" as we go.

shortfin/python/shortfin_apps/llm/components/batcher.py
stbaione: Created issue to add tests: https://github.com/nod-ai/shark-ai/issues/1195

stbaione (approved):

PR #1187, opened 2025-03-28
[sharktank][kernels] restores unmasked attn functionality

dan-garvey:
previous patch erroneously rerouted the flash attention function inside of attention_impls.py directly to the custom kernel, skipping the important step of managing the scales properly. This restores the previous functionality. Something similar probably needs to be done for the masked kernel, and thats why the previous patch renamed flash_attention() to prepare_args(), with the intention of factoring out the code needed for both.


dan-garvey:
fyi @Giuseppe5.

@KyleHerndon If you could take this over if anything breaks I'd appreciate it.

KyleHerndon (approved):

PR #1189, opened 2025-03-28
[sharktank] Refactor Flux transformer export and compile to use the common ModelConfig

sogartar:
With the goal of unifying model export and compile under one interface, here the Flux transformer refactored.

Add model config preset registry.

### DIFF
--- a/sharktank/sharktank/layers/base.py
+++ b/sharktank/sharktank/layers/base.py
@@ -12,10 +12,13 @@ import torch
 import torch.nn as nn
 from os import PathLike
 import logging
+from pathlib import Path
 
 from ..types import InferenceTensor, Theta, AnyTensor, Dataset
-from ..utils import debugging
+from ..utils import debugging, chdir
 from .configs import ModelConfig, ExportFunctionConfig, DynamicBatchSize
+from ..utils.iree import flatten_for_iree_signature
+from iree.turbine.support.tools import iree_tool_prepare_input_args
 
 __all__ = [
     "BaseLayer",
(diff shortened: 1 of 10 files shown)


sharktank/sharktank/layers/base.py
```diff
@@ -218,21 +221,7 @@ def export_mlir(self, path: PathLike | None = None, /):
...
-        if self.config.export_functions is not None:
-            export_functions = self.config.export_functions
-
-        function_batch_size_pairs = {
-            export_function.function
-            or self.default_export_function: export_function.batch_sizes
-            or self.default_export_batch_sizes
-            for export_function in export_functions
-        }
+        function_batch_size_pairs = self._get_function_batch_sizes_map()
```
KyleHerndon, line 224: Nit: IMO "pairs" is misleading because the map is keyed on functions whose values are *lists* of ints. Are you fine with naming it `function_batch_size_map` or `function_batch_size_dict`?

sharktank/sharktank/models/flux/flux.py
```diff
@@ -155,19 +175,19 @@ def __init__(self, theta: Theta, params: FluxParams):
         self.pe_embedder = EmbedND(
             dim=pe_dim, theta=params.theta, axes_dim=params.axes_dim
         )
-        self.add_module("img_in", LinearLayer(theta("img_in")))
-        self.add_module("time_in", MLPEmbedder(theta("time_in")))
-        self.add_module("vector_in", MLPEmbedder(theta("vector_in")))
+        self.add_module("img_in", LinearLayer(self.theta("img_in")))
+        self.add_module("time_in", MLPEmbedder(self.theta("time_in")))
```
KyleHerndon, line 179: No need to change this, but for my edification: Does adding `self` ever matter? We should probably avoid modifying the theta once its being handed to constructors.

sharktank/tests/models/flux/flux_test.py
```diff
@@ -25,6 +26,7 @@
     make_random_theta,
 )
 from sharktank.models.flux.flux import FluxModelV1, FluxParams
+from sharktank.models.flux.compile import iree_compile_flags
```
KyleHerndon, line 29: Nit: Unused import

sharktank/sharktank/layers/base.py
sogartar: I renamed it to `function_batch_sizes_map`.

sharktank/sharktank/models/flux/flux.py
sogartar: Now it is allowed to init without a theta as the theta could be retrieved from the config in a standard way in the `ThetaLayer` base. This allows us to construct with only a config.

sharktank/tests/models/flux/flux_test.py
sogartar: This is still used in some tests. Potentially we can refactor the tests to use the config presets. I will leave this for another PR.

KyleHerndon (approved):

PR #1191, opened 2025-03-28
Fix the missing of --use-hf in llama benchmarktest

AmosLewis:
- Fix the missing of --use-hf in llama benchmarktest.
- Simplify iree-compile flag based on llama benchmark doc update.


aviator19941 (approved):

aviator19941 (commented):
Can you also update these flags on L#51 of `sharktank/tests/models/llama/benchmark_amdgpu_test.py`?

PR #1197, opened 2025-03-31
[sharktank] Add Tracy utils

sogartar:
Add functions for more convenient tracing of models.

E.g.
```
model = ...
model.export()
model.compile()
trace_model_with_tracy(model.config, function="forward_bs1")
```

```
def my_fn():
   ...

trace_with_tracy(my_fn, output_trace_path="capture.tracy")
```

Add a dummy model to the registry to server for testing.

### DIFF
--- a/sharktank/sharktank/utils/iree.py
+++ b/sharktank/sharktank/utils/iree.py
@@ -4,15 +4,19 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+from copy import deepcopy
 import iree.runtime
-from typing import Any, Callable, List, Tuple, Optional, Union, overload
+from typing import Any, Callable, List, Tuple, Optional, Union, overload, TYPE_CHECKING
 from pathlib import Path
 import torch
 import os
+import sys
+import json
 import numpy as np
 import collections.abc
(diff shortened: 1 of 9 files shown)


sharktank/sharktank/utils/iree.py
```diff
@@ -397,3 +404,114 @@ def __call__(self, key: str, buffer_views: List[iree.runtime.HalBufferView]):
...
+def run_model_with_iree_run_module(
+    config: "ModelConfig", function: str, **subprocess_run_kwargs
+):
+    """Run an already exported and compiled model with iree-run-module.
+    It is required that is exports its input arguments.
+    """
+    cmd = [
+        "iree-run-module",
+        f"--module={config.iree_module_path}",
+        f"--device={config.iree_hal_driver}",
```
Alex-Vasile, line 502: Multidevice support in a later PR?

sogartar: Yes, at some point. The config has a `tensor_parallelism` parameter, that can be utilized for that.
The meaning of the `parameters_path` would need to be expanded to be a list as well. Or maybe a directory that contains the IRPA files. We need this anyway if we are to split the weights due to for example constraints of our storage like Hugging Face.

Alex-Vasile (approved):

PR #1203, opened 2025-04-01
Revert "[shortfin] Enable building Rust deps on Linux (#619)"

bangtianliu:
This reverts commit 5f7bfa9d923549fbd2724d7540bf339a8be151ee.

Fix ci errors in PR #1171

### DIFF
--- a/.github/workflows/ci-libshortfin.yml
+++ b/.github/workflows/ci-libshortfin.yml
@@ -43,7 +43,7 @@ jobs:
           - name: Ubuntu (Clang)(full)
             runs-on: ubuntu-24.04
             cmake-options:
-              -DCMAKE_C_COMPILER=clang-18 -DCMAKE_CXX_COMPILER=clang++-18 -DCMAKE_LINKER_TYPE=LLD -DSHORTFIN_ENABLE_TOKENIZERS=ON
+              -DCMAKE_C_COMPILER=clang-18 -DCMAKE_CXX_COMPILER=clang++-18 -DCMAKE_LINKER_TYPE=LLD
             additional-packages: clang lld
             env:
               SHORTFIN_ENABLE_TRACING: OFF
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-libshortfin.yml
```diff
@@ -116,13 +116,22 @@ jobs:
...
+        ref: iree-3.1.0rc20241204
+
+    - name: Initalize IREE submodules
+      working-directory: ${{ env.IREE_REPO_DIR }}
+      run : |
+        git submodule update --init --depth 1 -- third_party/benchmark
+        git submodule update --init --depth 1 -- third_party/cpuinfo/
+        git submodule update --init --depth 1 -- third_party/flatcc
+        git submodule update --init --depth 1 -- third_party/googletest
+        git submodule update --init --depth 1 -- third_party/hip-build-deps/
```
ScottTodd, line 134: Looks like the revert was not clean. Only the "Setup Rust" step should be removed. These other checkout / submodule steps should not be added.

ScottTodd (changes requested):

.github/workflows/ci-libshortfin.yml
ScottTodd: this change should also remain

```diff
+              -DCMAKE_C_COMPILER=clang-18 -DCMAKE_CXX_COMPILER=clang++-18 -DCMAKE_LINKER_TYPE=LLD
-              -DCMAKE_C_COMPILER=clang-18 -DCMAKE_CXX_COMPILER=clang++-18 -DCMAKE_LINKER_TYPE=LLD -DSHORTFIN_ENABLE_TOKENIZERS=ON
```

ScottTodd: Technically only the CMake flag change is needed to disabling the native tokenizers component that is causing issues with CMake 4.0.0, but the native tokenizers component is the only thing using Rust, so we can also remove the "setup Rust" step.

ScottTodd (approved):

PR #1204, opened 2025-04-01
Response should be a json blob with request information

rsuderman:
Requests may expect multiple returned responses. This change bundles them into a JSON blob that can be returned to the use.

### DIFF
--- a/shortfin/python/shortfin_apps/llm/components/generate.py
+++ b/shortfin/python/shortfin_apps/llm/components/generate.py
@@ -5,6 +5,7 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 import asyncio
+import dataclasses
 import io
 import json
 import logging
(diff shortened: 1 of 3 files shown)


shortfin/python/shortfin_apps/llm/components/generate.py
```diff
@@ -196,23 +203,44 @@ def generate_response(
...
+
+        responses = []
+        for k in response_map:
+            r = PromptResponse(prompt=k, responses=response_map[k])
+            r = dataclasses.asdict(r)
+            responses.append(r)
+
+        response = json.dumps(responses)
+        out = io.BytesIO()
+        out.write(b"data: ")
```
stbaione, line 238: Should we ditch the b"data" prefix and b"\n\n" suffix? I think that would make it easier on the client side, because the data we're returning would be in a pure JSON format

stbaione:
The integration tests will need updates as well

shortfin/python/shortfin_apps/llm/components/generate.py
rsuderman: Removed.

stbaione:
Good once pre-commit is passing

stbaione (approved):

PR #1207, opened 2025-04-02
[sharktank] Refactor sharktank llm modeling

archana-ramalingam:
- Refactor various sharktank llms into a single `llm.py` script
- Remove all references to direct cache scripts and tests
- Make `--attention-kernel="torch"` default for all llms and tests, except grok as softcap+sdpa isn't supported yet.


sharktank/sharktank/layers/ffn_block.py
```diff
@@ -23,18 +25,31 @@ class FFN(ThetaLayer):
...
             self.add_module(
                 "ffn_gate", LinearLayer(theta("ffn_gate"), fake_quant=fake_quant)
             )
+        if "ffn_norm" in theta:
+            # Llama & MoE models
+            self.ffn_norm = RMSNormLayer(theta("ffn_norm"), epsilon=rms_epsilon)
+        elif "layer_norm" in theta:
+            # T5 model
+            self.ffn_norm = RMSNormLayer(
+                theta("layer_norm"), epsilon=rms_epsilon, dtype=activation_dtype
```
rsuderman, line 50: Just for my information - which architectures diverge on the RMS norm layer naming? Just curious if this is something we can fix on original import (e.g. if its deepseek we can always correct this during rename).

Either way we can leave that off for a followup.

sharktank/sharktank/layers/mixture_of_experts_block.py
```diff
@@ -96,8 +94,7 @@ def forward(
 
         expert_gate = expert_gate.to(ffn_input.dtype)
 
-        if self.route_scale is not None:
-            expert_gate = expert_gate * self.route_scale
+        expert_gate = expert_gate * self.route_scale
```
rsuderman, line 97: It is still plausible we will want to `if` this case. Just because `float * 1.0` is not actually canonicalizable because of floating point shenanigans.

sharktank/sharktank/models/deepseek/toy_deepseek.py
```diff
@@ -49,9 +48,7 @@ def generate(seed):
             v_head_dim=32,
             expert_count=expert_count,
             expert_used_count=used_experts,
-            model_arch="deepseek",
-            expert_score_func="sigmoid",
-            route_scale=2.0,
+            model_arch="deepseek2",
```
rsuderman, line 51: I know its dumb but could we do `deepseek_v2`. I think thats how they do their naming scheme.

sharktank/sharktank/models/llm/llm.py
```diff
@@ -153,6 +149,9 @@ def prefill(
 
         h = self.output_norm(h)
         logits = self.output_lm_head(h)
+
+        if self.inference_norm:
+            logits = logits / math.sqrt(3.0)
```
rsuderman, line 154: We should plumb the `inference_norm` through as an optional value independent of the normalization above.

sharktank/sharktank/models/llm/llm.py
```diff
@@ -213,6 +215,9 @@ def decode(
...
```
rsuderman, line 220: Ditto

sharktank/sharktank/models/llm/llm.py
```diff
@@ -137,6 +130,9 @@ def prefill(
         h = self.token_embedding(tokens)
         self.trace_tensor("llama.token_embedding", h)
 
+        if self.inference_norm:
+            h *= math.sqrt(h.shape[-1])
```
rsuderman, line 134: We should probably have this normalization factor come via configuration. It happens to do `sqrt(h.shape[-1])` due to  some modelling quirks but this could be configurable for sad reasons.

sharktank/sharktank/models/llm/llm.py
```diff
@@ -196,6 +195,9 @@ def decode(
...
```
rsuderman, line 199: Ditto above.

rsuderman (approved):
Changes are mostly optional in this submission as they can be post commit cleanup, but wanted to leave the thoughts anyway. Feel free to cleanup in this PR or perform in some follow up.

Great job!

sharktank/sharktank/layers/ffn_block.py
archana-ramalingam: T5 model has a layer norm, but turns out it might not be part of ffn after all. Will move this back to it's place.

archana-ramalingam: Reverted this change.

sharktank/sharktank/layers/mixture_of_experts_block.py
archana-ramalingam: Done

sharktank/sharktank/models/llm/llm.py
archana-ramalingam: Added a todo for this.

sharktank/sharktank/models/llm/llm.py
archana-ramalingam: Added a todo for this.

sharktank/sharktank/models/llm/llm.py
archana-ramalingam: Added a todo

sharktank/sharktank/models/llm/llm.py
archana-ramalingam: Added a todo

sharktank/sharktank/models/llm/llm.py
```diff
@@ -230,41 +238,83 @@ def __init__(
...
+                True,
+                False,
+            ),
+            "deepseek2": (
+                torch.nn.functional.sigmoid,
+                torch.nn.functional.silu,
+                False,
+                True,
+            ),
+        }
```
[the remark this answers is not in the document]
archana-ramalingam: Agreed, ideally this should be plumbed through from llm_configs.

sharktank/sharktank/layers/ffn_block.py
archana-ramalingam: Reverted this change in deepseek patch #1256, as all norm layers sit better outside ffn_block for other reasons.

PR #1209, opened 2025-04-02
Add llama tool to randomize weights in irpa file

aviator19941:
Adds a tool that takes in a llama irpa file, randomizes the weights using a normal distribution in order to maintain the same mean and standard deviation as the original weights, and outputs the new irpa file.

### DIFF
--- a/sharktank/sharktank/models/llama/testing.py
+++ b/sharktank/sharktank/models/llama/testing.py
@@ -8,6 +8,7 @@ from typing import List
 
 import torch
 
+from ...types import Dataset
 from ...types.tensors import *
 from ...types.theta import Theta
 from typing import Optional
@@ -162,4 +163,23 @@ def make_random_llama_theta(
         data=make_rand_torch((1, config.hp.embedding_length), dtype=dtype),
     )
 
+    return Theta(res)
+
+
+def make_random_theta_from_dataset(dataset: Dataset) -> Theta:
+    res = {}
+    for key, value in dataset.root_theta.flatten().items():
+        shape = value.shape
+        min_val = torch.min(value.flatten())
+        max_val = torch.max(value.flatten())
+        import pdb
+
+        pdb.set_trace()
+        random_tensor = torch.normal(
+            mean=value.as_torch().mean(),
+            std=value.as_torch().std(),
+            size=shape,
+            dtype=value.dtype,
+        )
+        res[key] = DefaultPrimitiveTensor(name=key, data=random_tensor)
     return Theta(res)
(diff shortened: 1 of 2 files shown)


sharktank/sharktank/models/llama/testing.py
```diff
@@ -163,3 +164,21 @@ def make_random_llama_theta(
     )
 
     return Theta(res)
+
+
+def make_random_theta_from_dataset(dataset: Dataset) -> Theta:
```
sogartar, line 169: This function seems general enough to go into
`sharktank/sharktank/layers/testing.py`
or into a new file
`sharktank/sharktank/types/testing.py`

It would also benefit from a short description. You could just move comment form below.

Also from a symmetry view point it should accept a `Theta` and return a `Theta`.

You could have two function `make_random_theta_from_theta` and `make_random_dataset_from_dataset`.
`make_random_dataset_from_dataset` would just additionally copy the properties.

sharktank/sharktank/models/llama/tools/randomize_llama_weights.py
```diff
@@ -0,0 +1,69 @@
...
+cli.add_input_dataset_options(parser)
+cli.add_model_options(parser)
+parser.add_argument("-s", "--seed", default=12345)
+parser.add_argument("-o", "--output", default="/tmp/random_llama.irpa")
+args = cli.parse(parser)
+
+
+def generate(seed):
+    torch.manual_seed(seed)
+    dtype = torch.float16
```
sogartar, line 38: This is unused.

sharktank/sharktank/models/llama/tools/randomize_llama_weights.py
```diff
@@ -0,0 +1,69 @@
...
+        hp=LlamaHParams.from_gguf_props(dataset.properties),
+        block_seq_stride=args.block_seq_stride,
+        device=device,
+        activation_dtype=args.activation_dtype,
+        attention_dtype=args.attention_dtype,
+        attention_kernel=args.attention_kernel,
+        kv_cache_dtype=args.kv_cache_dtype,
+        use_hf=args.use_hf,
+        tensor_parallelism_size=args.tensor_parallelism_size,
+    )
```
sogartar, line 52: What is the purpose of this? You could just clone the Dataset properties.

This tool can be more general to work on any dataset and it would be simpler if you have the `make_random_dataset_from_dataset` function.

sogartar (commented):
I would argue that the generalization to operate on any dataset would make things simpler.

Issue #1051, opened 2025-03-07
Un-xfail `test_compare_against_quark`


renxida:
Had to xfail this test due to a **very** minor over-tolerance during IREE bump #1003

It doesn't look like much of a difference and only *barely* exceeds the tolerances. However, I don't know this test well enough to go in and increase the tolerances.

--------

Failure log: https://gist.github.com/renxida/f1af75da6d9a82e662c6824fa1314ec7

Error message:
FAILED sharktank/tests/models/llama/quark_parity_test.py::QuarkParityTest::test_compare_against_quark - AssertionError: Tensor-likes are not equal!

Details:
Mismatched elements: 1 / 16384 (0.0%)
Greatest absolute difference: 0.00390625 at index (0, 1, 713)
Greatest relative difference: 0.004730224609375 at index (0, 1, 713)

--------

Will assign to test author after #1003 merges to take a look at this and maybe increase the tolerance and re-enable if that's appropriate.

KyleHerndon:
Discussing with Dan, this is actually a big deal. This test represents a significant effort to ensure parity with quark-we do expect it to be exact. I'm looking into this now.

PR #1210, opened 2025-04-03
Fix fake-quant attention by passing the correct dtype

KyleHerndon:


KyleHerndon:
This is a pre-requisite to fixing #1051

PR #1212, opened 2025-04-03
[sharktank] Add CLI tool for operations on models

sogartar:
To have a convenient generation of model artifacts we would benefit from a CLI tool. The tool here could be used for model, export, compile, tracy_trace. E.g.

```
sharktank list
```

```
sharktank export some-model-config-preset-name
sharktank compile some-model-config-preset-name
```

```
sharktank export /path/to/model/config.json
```

### DIFF
--- a/sharktank/pyproject.toml
+++ b/sharktank/pyproject.toml
@@ -22,7 +22,7 @@ dynamic = ["version", "dependencies", "optional-dependencies"]
 Repository = "https://github.com/nod-ai/shark-ai"
 
 [project.scripts]
-sharktank = "sharktank.tools.sharktank:main"
+shark = "sharktank.tools.sharktank:main"
 
 [tool.setuptools.packages.find]
 where = ["."]
--- a/sharktank/pyproject.toml
+++ b/sharktank/pyproject.toml
@@ -21,6 +21,9 @@ dynamic = ["version", "dependencies", "optional-dependencies"]
 [project.urls]
 Repository = "https://github.com/nod-ai/shark-ai"
 
+[project.scripts]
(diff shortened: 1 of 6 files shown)


sharktank/pyproject.toml
```diff
@@ -22,6 +22,9 @@ dynamic = ["version", "dependencies", "optional-dependencies"]
 [project.urls]
 Repository = "https://github.com/nod-ai/shark-ai"
 
+[project.scripts]
+sharktank = "sharktank.tools.sharktank:main"
```
stbaione, line 26: Would it make sense to call this just `shark`? 

Right now it's pointed to a `sharktank` script, but in the future, we can extend it to include things like the `llm servers`, and point the command to a file in `shark-ai` dir.

Then this could be the starting point of our unified `shark` cli command

sogartar: I guess we can have that, but then the commands for operations on models should be
```
shark tank ...
```

sogartar: Or
```
shark model ...
```

sogartar: We can use [command grouping](https://google.github.io/python-fire/guide/#grouping-commands) in Fire for that.

stbaione: I really like the `shark model ...` idea. What do you think?

sogartar: It is done.

sogartar: `shark model ...` it is.

stbaione (approved):
LGTM, make sure to update the PR description

PR #1220, opened 2025-04-04
[sharktank] Refactor sharktank - import cleanup

archana-ramalingam:
Refactor sharktank 

- Expand relative imports
- Cleanup unused imports
- Correct the source module of import (eg: Import `Theta` from `types` instead of `layers`)

Requires #1207 to merge


sharktank/sharktank/utils/iree.py
```diff
@@ -18,7 +17,10 @@
 from contextlib import contextmanager
 import subprocess
 import gc
-from ..types.tensors import (
+import numpy as np
```
IanNod, line 20: already imported numpy above

IanNod (approved):
One minor comment but otherwise looks good to me once dependant PR is merged

PR #1222, opened 2025-04-04
[sharktank] Add ops.clone and implement for DefaultPrimitiveTensor

sogartar:
Change sharded tensors to clone their data, but not the name to avoid name aliasing. Cloning should thought as equivalent to `b = a + 0`.

### DIFF
--- a/sharktank/sharktank/ops/default_impls.py
+++ b/sharktank/sharktank/ops/default_impls.py
@@ -8,7 +8,7 @@
 # generic primitive/quantized types.
 
 from typing import Optional, List, Sequence, Union, Tuple
-
+from copy import deepcopy
 import torch
 from torch import Tensor, dtype
 import torch.nn.functional as F
@@ -25,11 +25,31 @@ from ..types.tensors import unbox_tensor, AnyTensor
 from ._registry import AllOfType, AllOfExprs, AllOfExprsVariadic, IsOfType
 from .signatures import *
 import iree.turbine.ops.iree
+from iree.turbine.aot import DeviceTensorTrait
 
 
 @cat.override(AllOfType(Tensor, PrimitiveTensor))
 def cat_default(tensors: Sequence[Tensor | PrimitiveTensor], dim: int):
     return torch.cat([unbox_tensor(t) for t in tensors], dim)
+
+
+@clone.override(IsOfType(InferenceTensor))
+def clone_default(tensor: InferenceTensor):
+    return tensor.clone()
+
+
+@clone.override(IsOfType(Tensor))
+def clone_torch(tensor: Tensor):
+    # Clone introduces a tensor copy operation in the graph. It does not make sense to
+    # clone also the ExternalTensorTrait. The expectation is that when a user clones a
+    # tensor if they modify in-place the result this should not affect the original
+    # tensor. If we are to clone also the ExternalTensorTrait then we would have
+    # external_name aliasing, but the expectation is that there is no data aliasing.
+    res = tensor.clone()
+    device_trait = DeviceTensorTrait.get(tensor)
+    if device_trait is not None:
+        deepcopy(device_trait).set(res)
+    return res
 
 
 # conv2d
--- a/sharktank/sharktank/types/tensors.py
+++ b/sharktank/sharktank/types/tensors.py
@@ -288,6 +288,11 @@ class InferenceTensor(ABC):
             lambda d: {k: t.to(device=device) for k, t in d.items()}
         )
 
+    def clone(self) -> "InferenceTensor":
+        raise NotImplementedError(
+            f"InferenceTensor {type(self)} does not implement clone."
+        )
+
     def _clone_with_globals(
         self, new_globals: dict[str, torch.Tensor]
     ) -> "InferenceTensor":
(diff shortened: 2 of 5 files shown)


sharktank/sharktank/ops/default_impls.py
```diff
@@ -25,13 +25,33 @@
...
+    return tensor.clone()
+
+
+@clone.override(IsOfType(Tensor))
+def clone_torch(tensor: Tensor):
+    # Clone introduces a tensor copy operation in the graph. It does not make sense to
+    # clone also the ExternalTensorTrait. The expectation is that when a user clones a
+    # tensor if they modify in-place the result this should not affect the original
+    # tensor. If we are to clone also the ExternalTensorTrait then we would have
+    # external_name aliasing, but the expectation is that there is no data aliasing.
```
Alex-Vasile, line 47: For pipeline parallelism I need the external tensor trait behaviour to copy. We don't want to bake the pipeline parallelism decisions into the weights, so they are applied to the tensors when loaded from file inside an export script, e.g. `export_paged_llama_v1.py`, FYI changes aren't up yet.

.clone was added for the following behaviour:
1. Load tensor t.
2. Change .devices or move .shards to different devices with ops.transfer_to_logical_device
3. t_new = t.clone(ts=moved_shards, devices=new_devices).
4. Store back into theta.
5. Continue exporting model.

sharktank/tests/types/tensors_test.py
```diff
@@ -169,23 +169,30 @@ def testSplitTensorInsertSliceWithEllipsis(self):
...
+        sharded_tensor = ReplicatedTensor(
+            ts=tensor, shard_count=4, name="sharded_tensor"
+        )
         cloned_tensor = sharded_tensor.clone()
-        assert sharded_tensor.is_deep_equal(cloned_tensor)
+        assert cloned_tensor.name != sharded_tensor.name
+        assert sharded_tensor.is_deep_equal(cloned_tensor, compare_name=False)
         assert iterables_equal(sharded_tensor.devices, cloned_tensor.devices)
 
     def testCloneTensorTraits(self):
```
Alex-Vasile, line 198: How is this test passing if ExternalTensorTrait is not being copied.

Alex-Vasile (changes requested):

sharktank/sharktank/ops/default_impls.py
sogartar: If pipeline stages are not going to be a part of the same compute graph then you could do a deepcopy.

sharktank/sharktank/types/tensors.py
```diff
@@ -1075,7 +1086,7 @@ def __init__(
         )
 
     def clone(self, **kwargs) -> "SplitPrimitiveTensor":
-        kwargs["name"] = kwargs.get("name", self.name)
+        # We don't copy the name to not introduce name aliasing.
```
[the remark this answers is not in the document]
sogartar: I  assume that names should be unique. If the data is not the same it does not make sense that the name is the same. We are in the same compute graph.

sharktank/sharktank/types/tensors.py
```diff
@@ -523,6 +528,12 @@ def __init__(
         super().__init__(name=name, shape=list(data.shape))
         self._data = data
 
+    def clone(self) -> "InferenceTensor":
+        from .. import ops
+
+        # We don't clone due the name to not introduce name aliasing.
```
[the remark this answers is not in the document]
sogartar: Removed it.

sharktank/sharktank/types/tensors.py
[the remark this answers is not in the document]
sogartar: Ops that insert computation in the graph should not create tensors with the same name.

sharktank/sharktank/ops/default_impls.py
sogartar: I think that there is a discrepancy of expectations.

1. Usually a `clone` method does a deep copy.
2. In torch this term is loaded with more meaning as it creates a tensor that is a part of the compute graph. I think we should avoid deviating from that meaning as it will introduce confusion down the line.  If operations under `sharktank.ops` or methods of tensors have the same name as in torch then they should do the same.

Alex-Vasile: Perhaps the name is the problem then. I did not introduce this to be equivalent to torch.clone.

rsuderman:
@sogartar could I get a run through behind the motivation for the PR? Most of the intention for clone was being able to create a new meta tensor and override non-numeric information, so cloning a `torch` tensor does not really fulfill that purpose.

rsuderman (changes requested):

sharktank/sharktank/ops/default_impls.py
sogartar: Yes, another name then would be desirable as we may want something like `torch.clone` as some point.

sogartar:
@rsuderman, the main motivation behind this PR is that `clone` in `torch` has already established semantics and if we are to name something the same way it should have the same semantics.
If we want something else, which seems to be the case from the discussions here, we should name it something else.

Issue #1195, opened 2025-03-31
Write unit tests for LLM Server `BatcherProcess` classes


stbaione:
The following classes need unit testing added in `Shortfin LLM Server`:

- BatcherProcess
- LlmBatcherProcess
- PrefillBatcherProcess
- DecodeBatcherProcess

PR #1224, opened 2025-04-04
Integrate reservation system and normalization to decoding

rsuderman:
Normalization is significantly faster on device, so support whether the logits were returned in a normalized fashion when doing beam search. This allows existing support for previous decoding strategies.

Also add support for decoder reservation system to better support batched decoding.


stbaione (approved):
LGTM, the `config.py` file really cleans things up

It looks like the changes to `Batcher` are slowing down, so I can take a look at #1195 when I have `sampling` in a good spot

PR #1225, opened 2025-04-05
[tuner] deferring the link phase to multi-threaded compilation phase

### LINKED ISSUE #1200: Parallelize the process of candiate generation in tuner
Per https://github.com/nod-ai/shark-ai/pull/1171#discussion_r2021233743, candidate generation—including td spec linking—could be deferred to the candidate compilation phase, which is already multi-threaded.

CC @kuhar @Max191

bangtianliu:
This PR moves the linking phase, which links the starter transform td spec with candidate specs, to the candidate compilation phase. Since candidate compilation is multithreaded, this change reduces the linking overhead.

Below are the overhead results I observed (the time unit is seconds):

| Number of candidates| Candidate generation  + link       | Candidate compilation  | Total  |
|------------------------|-------------------------------|--------------------|------------ |
|              50                 |        4.85                              |    13.78                 |    18.63      |
|              100               |        5.31                              |    16.31                 |    21.62      |

After deferring the link phase from candidate generation to the compilation phase
| Number of candidates| Candidate generation         | Candidate link + Compilation  | Total |
|------------------------|-------------------------------|--------------------|------------ |
|              50                 |        1.62                              |    14.20                 |    15.82      |
|              100               |        1.94                              |    16.49                 |    18.44      |

The total overhead is reduced through deferring link phase, as shown by the table above.

fixes #1200

### DIFF
--- a/tuner/tuner/libtuner.py
+++ b/tuner/tuner/libtuner.py
@@ -67,7 +67,7 @@ class CandidateTracker:
     mlir_path: Optional[Path] = None
     compiled_vmfb_path: Optional[Path] = None
     spec_path: Optional[Path] = None
-    starter_td_spec_str: Optional[str] = None
+    td_spec_str: Optional[str] = None
 
 
 @dataclass()
@@ -457,12 +457,12 @@ def create_worker_context_queue(device_ids: list[str]) -> queue.Queue[tuple[int,
     return worker_contexts_queue
 
 
-def link_td_specs_with_outpath(starter_td_spec_str: str, output_path: Path) -> None:
+def flatten_nested_td_spec(td_spec_str: str, output_path: Path) -> None:
     iree_opt = ireec.binaries.find_tool("iree-opt")
     with tempfile.TemporaryDirectory() as tmpdir:
         input_path = os.path.join(tmpdir, "tmp_input.mlir")
         with open(input_path, "w") as f:
-            f.write(starter_td_spec_str)
+            f.write(td_spec_str)
 
         link_result = subprocess.run(
             [
@@ -484,9 +484,9 @@ def run_iree_compile_command(compile_pack: CompilePack) -> Optional[int]:
     candidate_tracker = compile_pack.candidate_tracker
     assert candidate_tracker.spec_path, "expected candidate spec path"
 
-    if candidate_tracker.starter_td_spec_str is not None:
-        link_td_specs_with_outpath(
-            candidate_tracker.starter_td_spec_str,
+    if candidate_tracker.td_spec_str is not None:
+        flatten_nested_td_spec(
+            candidate_tracker.td_spec_str,
             candidate_tracker.spec_path,
         )
 
@@ -750,7 +750,7 @@ def generate_candidate_specs(
                 candidate_num
             )
 
-            starter_td_spec_str: Optional[str] = None
+            td_spec_str: Optional[str] = None
             if starter_td_spec is not None:
                 td_specs: list[ir.Module] = [spec, starter_td_spec]
                 # Only log duplicate matchers during the first iteration.
@@ -760,7 +760,7 @@ def generate_candidate_specs(
                 )
 
                 if len(td_specs_to_link) != 1:
-                    starter_td_spec_str = str(
+                    td_spec_str = str(
                         combine_tuning_specs(
                             tuning_client.tuner_context, td_specs_to_link
                         )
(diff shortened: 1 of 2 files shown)


tuner/tuner/libtuner.py
```diff
@@ -458,9 +459,30 @@ def create_worker_context_queue(device_ids: list[str]) -> queue.Queue[tuple[int,
...
+                    input_path,
+                    "-o",
+                    candidate_tracker.spec_path,
+                ],
+                capture_output=True,
+                text=True,
+            )
+
+            if link_result.returncode != 0:
+                raise RuntimeError(f"iree-opt failed: {link_result.stderr}")
```
kuhar, line 483: Could we move this to a helper function?

tuner/tuner/libtuner.py
```diff
@@ -720,6 +741,26 @@ def generate_candidate_specs(
             spec_path = path_config.specs_dir / path_config.get_candidate_spec_filename(
                 candidate_num
             )
+
+            starter_td_spec_str: Optional[str] = None
+            if starter_td_spec is not None:
+                td_specs: list[ir.Module] = []
+                td_specs.append(spec)
+                td_specs.append(starter_td_spec)
```
kuhar, line 749:
```suggestion
                td_specs: list[ir.Module] = [spec, starter_td_spec]
```

tuner/tuner/libtuner.py
```diff
@@ -720,6 +741,26 @@ def generate_candidate_specs(
...
+                # Only log duplicate matchers during the first iteration.
+                log_duplicates = candidate_num == 0
+                td_specs_to_link = determine_td_specs_to_link(
+                    td_specs,
+                    log_duplicates=log_duplicates,
```
kuhar, line 754:
```suggestion
                td_specs_to_link = determine_td_specs_to_link(
                    td_specs,
                    log_duplicates=(candidate_num == 0),
```

kuhar (changes requested):

tuner/tuner/libtuner.py
```diff
@@ -456,11 +457,40 @@ def create_worker_context_queue(device_ids: list[str]) -> queue.Queue[tuple[int,
     return worker_contexts_queue
 
 
+def link_td_specs_with_outpath(starter_td_spec_str: str, output_path: Path) -> None:
+    iree_opt = ireec.binaries.find_tool("iree-opt")
+    with tempfile.TemporaryDirectory() as tmpdir:
+        input_path = os.path.join(tmpdir, "tmp_input.mlir")
+        with open(input_path, "w") as f:
+            f.write(starter_td_spec_str)
```
kuhar, line 465: Why are we writing the td spec to a file instead of using the path provided through CLI arguments?

tuner/tuner/libtuner.py
```diff
@@ -456,11 +457,40 @@ def create_worker_context_queue(device_ids: list[str]) -> queue.Queue[tuple[int,
...
+
+        link_result = subprocess.run(
+            [
+                iree_opt,
+                "--iree-codegen-link-tuning-specs",
+                input_path,
```
kuhar, line 471: Are we linking a single input only? I'm confused by this code.

tuner/tuner/libtuner.py
bangtianliu: Because linking has two steps, firstly it is about combining starter_td_spec and candidate spec into one module (single input) and then link through calling `--iree-codegen-link-tuning-specs`.

tuner/tuner/libtuner.py
bangtianliu: yes, it is about serializing the td specs into one module and then link,  refer to examples in https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/Common/test/link_tuning_specs.mlir#L1.

tuner/tuner/libtuner.py
kuhar: Oh, I see what you mean. In this case, I think we'd need a different argument name. This is no longer just a `starter_td_spec_str`, it's a combined/nested spec that needs linking.

kuhar: Maybe something like:
```py
def flatten_nested_td_spec(td_spec_str: str, output_path: Path) -> None:
```

kuhar (approved):
LGTM but please wait for an approval from @Max191 before merging

Max191 (approved):
Cool! This looks good to me

PR #1229, opened 2025-04-07
[sharktank] make tests optionally more verbose to log start/finish

sogartar:
Occasionally have hanging tests in CI and it would be nice to know which test is the culprit.

Here all tests are made to inherit from
`sharktank.utils.testing.TestCase`, which has a test fixture to log the start and end of all each tests.
This will enable use to see which test started and did not finish even when running multiple tests in parallel.

The fixture is required to be imported even when defining the derived class so we import with
```
from sharktank.utils.testing import *
```
This will allow us to add more fixtures to the base class if required without changing all import statements.

### DIFF
--- a/.github/workflows/ci-sharktank.yml
+++ b/.github/workflows/ci-sharktank.yml
@@ -85,7 +85,11 @@ jobs:
       - name: Run sharktank tests
         if: ${{ !cancelled() }}
         run: |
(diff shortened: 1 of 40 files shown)


sogartar:
I am closing this due to actually `pytest-xdist` not being compatible with `--capture=no`. See https://github.com/pytest-dev/pytest-xdist/issues/296

PR #1234, opened 2025-04-08
Shortfin llm top k sampling

stbaione:
# Description

This PR implements the [top_k](https://huggingface.co/blog/how-to-generate#top-k-sampling) sampling method on top of `greedy/multi_greedy` and `beam_search`.

Conceptually this PR draws a line of abstraction between `Sampling` and `Selecting`. The `sampler` implements the sampling methods: `select_top_k` and `sample_top_k`, with `sample_top_p` in a follow up PR. The `TokenSelectionStrategies` & `Beams` contain the logic for selecting and outputting the tokens for a given decode iteration.

## Explanation

`Greedy` - Single independent beam
`MultiGreedy` - Multiple independent beams
`BeamSearch` - Multiple dependent beams (dependency being the fact that we select our next set of candidates based on cumulative log_prob across all beams)

We add a sampler on top of the selection strategies:

`top_k` - Take the `top_k` highest probability tokens, and randomly sample from then based on their probabilities.

This sits on top of the `selection strategy`, which works like this:

`GreedyBeam` (Greedy/MultiGreedy) - Randomly sample `top_k` times from the `top_k` highest probability tokens. Then, select and return the highest probability token.

`BeamSearchBeam` (BeamSeach) - Randomly sample `top_k` times from the `top_k` highest probability tokens. From there, select the `num_beams` beams based on log_probability from our distribution. It's very similar to how `greedy` implements it, but the beam outputs multiple potential token selections, instead of just 1.

*NOTE*: I've noticed that the diversity of `BeamSearch` outputs decrease when used in conjunction with `top_k`, often times producing the same `N` beams. This makes since, because if there is a token with dominant probability, we will end up selecting it multiple times. This decreases the variability of the selections space, when we choose our beams based on cumulative_log_prob. `MultiGreedy` with `top_k` is much better in terms of diversity in outputs. I believe the logic is sound and tested, but would like extra attention in that section of the code when reviewing.

We create a simple `Sampler` class, which currently has the logic for `sample_top_k` and `select_top_k`. In a follow up PR, this will also have the `sample_top_p` function. So most of the logic involving selecting tokens based on probability is moved there, while the orchestration logic stays in the `Beam` classes.


shortfin/python/shortfin_apps/llm/components/io_struct.py
```diff
@@ -36,6 +36,33 @@ def __post_init__(self):
...
+@dataclass
+class SamplingParams:
+    # Number of parallel samples
+    n: int = 1
+    # Max tokens to generate during decode loop
+    max_completion_tokens: int = DEFAULT_MAX_COMPLETION_TOKENS
+    # Temperature to use during generation
+    temperature: float = DEFAULT_TEMPERATURE
+    # Use `top_k` sampling during token selection process
+    top_k: int | str = NOT_PROVIDED
```
rsuderman, line 59: Typically you do a `| None` instead, then you just check for `None`

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/beam_search_token_selection_strategy.py
```diff
@@ -59,17 +97,16 @@ def sample_logits(self, k: int):
...
-        elif self.logits_normalization == LogitsNormalization.NONE:
-            log_softmax_logits = sfnp.log_softmax(self.exec_req.result_logits)
-        elif self.logits_normalization == LogitsNormalization.SOFTMAX:
-            log_softmax_logits = sfnp.log(self.exec_req.result_logits)
-        else:
-            raise ValueError(
-                f"Unknown logits normalization {self.logits_normalization}"
-            )
-        return self._top_k(log_softmax_logits, -k)
+        if self.decode_config.top_k != NOT_PROVIDED:
```
rsuderman, line 100: Check for `none` instead.

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/greedy_token_selection_strategy.py
```diff
@@ -8,18 +8,33 @@
...
+    def _sample_logits_top_k(self):
+        top_k = self.decode_config.top_k
+        softmax_logits = self.convert_logits_normalization(
+            self.decode_config.logits_normalization,
+            LogitsNormalization.SOFTMAX,
+            self.exec_req.result_logits,
+        )
+
+        # Sample one token from the `top_k` highest probability tokens
+        tokens, probs = self.sampler.select_top_k(softmax_logits, -top_k)
```
rsuderman, line 33: Something important to note - you can do `top_k` before the `convert_logits_normalization` as softmax and log softmax are both monotonically increasing functions. This means the work to perform normalization can be decreased overall.

This is something worth exploring, it will tweak the logits a bit but not noticeably.

rsuderman (dismissed):

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/greedy_token_selection_strategy.py
stbaione: Yeah, that would make the process much faster, because I would be able to run softmax on a much smaller set of logits. Awesome, will look into that. It makes sense that if:

`argmax(raw) == argmax(softmax) == argmax(log_softmax)`, then `argpartition(raw, k) ~= argpartition(softmax, k) ~= argpartition(log_softmax, k)`

shortfin/python/shortfin_apps/llm/components/io_struct.py
stbaione: Updated

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/beam_search_token_selection_strategy.py
stbaione: Updated

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/greedy_token_selection_strategy.py
stbaione: Updated! Much faster

stbaione:
Closed in favor of #1259

PR #1237, opened 2025-04-08
Pipeline Parallelism Support for KVCache

Alex-Vasile:
Implementation and tests for pipeline parallelism support in KVCache. Instead of one tensor (torch or split) for the whole model, there is now a list of tensors (replicated for TP=1 and split for TP>1) one for each stage of the pipeline.

#1238 must be merged immediately afterwards.

PR #1238, opened 2025-04-08
Pipeline Parallellism for LLama

Alex-Vasile:
Requires #1237, developed together, split up to make reviewing easier. They **must** be merged together as both PRs change interfaces for models and the KVCache (PagedAttention).

Parallelism is implemented with an inter-layer callback which checks if transfers are required and performs them there.
Model weights are not changed or tagged when written out, instead they are modified in-place before tracing is performed by `pipeline_parallelize_theta`.

Decode for PP>1 currently does not work, blocked by https://github.com/iree-org/iree/issues/20551

### DIFF
--- a/sharktank/sharktank/examples/export_paged_llm_v1.py
+++ b/sharktank/sharktank/examples/export_paged_llm_v1.py
@@ -9,84 +9,20 @@
 import os
 import logging
 import json
-from typing import Any, Dict, Tuple
+from typing import Any, Dict
 import torch
 
 from iree.turbine.aot import *
 
 from sharktank.layers import *
 from sharktank.types import *
+from sharktank.types.pipelining import pipeline_parallelize_theta
 from sharktank.utils.math import ceildiv
 from sharktank import ops
 from sharktank.utils import cli
 
 # TODO: Should be using a base class with the protocol supported.
 from sharktank.models.llm import *
-
-
-def pipeline_parallelize_theta(
-    theta: Theta, pipeline_parallelism_size: int
-) -> tuple[tuple[int, ...], ...]:
-    """Pipeline parallelize theta."""
-    # TODO: Still modifies the shards, but the signature doesn't imply this
-    def parallelize_weight(
-        weight: ShardedTensor, new_devices: Tuple[int, ...]
-    ) -> ShardedTensor:
-        (old_shards, old_devices) = (
-            ([weight], (0,))
-            if isinstance(weight, PrimitiveTensor)
-            else (weight.shards, weight.devices)
-        )
-        new_shards = ShardedTensor.move_shards_to_new_devices(
-            old_shards, old_devices=old_devices, new_devices=new_devices
-        )
-
-        for i, (old_shard, new_shard) in enumerate(zip(old_shards, new_shards)):
-            DeviceTensorTrait(new_devices[i]).set(new_shard._data)
-            if old_tensor_trait := ExternalTensorTrait.get(old_shard._data):
-                ExternalTensorTrait(
-                    old_tensor_trait.external_scope,
-                    old_tensor_trait.external_name,
-                ).set(new_shard._data)
-
-        return (
-            ReplicatedTensor(ts=new_shards, name=weight.name, devices=new_devices)
-            if isinstance(weight, PrimitiveTensor)
-            else weight.clone(ts=new_shards, devices=new_devices)
-        )
-
-    _t = theta.tensor("token_embd")["weight"]
-    shard_count = 1 if isinstance(_t, DefaultPrimitiveTensor) else _t.shard_count
-    num_blocks = len(theta.tensor("blk"))
-
-    block_to_device_lookup = []
-    block_indices = sorted(theta.tensor("blk").keys(), key=lambda item: int(item))
-    assert (
-        bi == i for i, bi in enumerate(block_indices)
-    ), "Blocks assumed to be numbered contiguously from [0, N-1]"
-    for blk_idx in block_indices:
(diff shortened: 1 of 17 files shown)


sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -24,6 +24,67 @@
 from sharktank.models.llm import *
 
 
+def pipeline_parallelize_theta(
+    theta: Theta, pipeline_parallelism_size: int
+) -> tuple[tuple[int, ...], ...]:
+    """Pipeline parallelize theta."""
+    # TODO: Still modifies the shards, but the signature doesn't imply this
+    def f(weight: ShardedTensor, new_devices: Tuple[int, ...]) -> ShardedTensor:
```
archana-ramalingam, line 32: Can we rename the function to specify what it does?

sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -24,6 +24,67 @@
...
+
+        return (
+            ReplicatedTensor(ts=new_shards, name=weight.name, devices=new_devices)
+            if isinstance(weight, PrimitiveTensor)
+            else weight.clone(ts=new_shards, devices=new_devices)
+        )
+
+    _t = theta.tensor("token_embd")["weight"]
+    shard_count = 1 if isinstance(_t, DefaultPrimitiveTensor) else _t.shard_count
+    num_blocks = len(theta.tensor("blk"))
```
archana-ramalingam, line 58: Is there a reason we don't get this value from `config.hp.block_count` from the irpa file?

sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -24,6 +24,67 @@
...
+        pp_group = int(int(blk_idx) * pipeline_parallelism_size / num_blocks)
+        zero_4_group = shard_count * pp_group
+        devices = tuple(i + zero_4_group for i in range(shard_count))
+        block_to_device_lookup.append(devices)
+
+        block_data = theta.tensor("blk", blk_idx)
+        for t_name in block_data.keys():
+            block_data[t_name]["weight"] = f(block_data[t_name]["weight"], devices)
+
+    theta.tensor("token_embd")["weight"] = f(
```
archana-ramalingam, line 75: For future PR: We require a seamless way to support all types of model arch/all layers.

sharktank/sharktank/utils/load_llm.py
```diff
@@ -220,33 +221,53 @@ def prefill(self):
...
 
             self.dump_args(phase="prefill", arg_name="token_ids", arg=token_ids)
             self.dump_args(phase="prefill", arg_name="seq_lens", arg=self.seq_lens)
+            self.dump_args(phase="prefill", arg_name="seq_block_ids", arg=seq_block_ids)
             self.dump_args(
-                phase="prefill", arg_name="seq_block_ids", arg=seq_block_ids_tensor
-            )
-            self.dump_args(
-                phase="prefill", arg_name="cache_state", arg=self.cache_state[0]
+                phase="prefill", arg_name="cache_state", arg=self.cache_state
```
archana-ramalingam, line 264: Is self.cache_state[0] not required anymore?

Alex-Vasile: No, https://github.com/nod-ai/shark-ai/pull/1288 changes the behaviour

sharktank/sharktank/examples/export_paged_llm_v1.py
Alex-Vasile: I need `shard_count` which isn't in `hp.block_count` so I would need to do the `_t` calculation anyway. I didn't want to have to pass in `hp` since it would be an extra unnecessary argument. It being Llama specific, I figured it wouldn't help with generalizing this function to different architectures either.

sharktank/sharktank/examples/export_paged_llm_v1.py
Alex-Vasile: I'm thinking something similar to how `shard_theta` works
https://github.com/nod-ai/shark-ai/blob/244a5e9a5b8b4a50513fbb30b74ca1c360884f23/sharktank/sharktank/models/deepseek/sharding.py#L54-L55

Or perhaps even work this into shard_theta itself. Though It might need a bit more thought since that function is used for sharding weights before saving to disk, and we don't want the pipeline parallelism info to be baked into the weights.

sharktank/sharktank/utils/load_llm.py
archana-ramalingam: Will you be changing #1288 to change `write_ndarray_to_bin()` rather than the removed `dump_args()`?

Alex-Vasile: yes, and I'll be changing this PR to match.

archana-ramalingam (approved):

sharktank/sharktank/examples/paged_llm_v1.py
```diff
@@ -29,6 +39,12 @@ def main():
     cli.add_save_tensor_options(parser)
 
     args = cli.parse(parser)
+
+    prompt_seq_len = args.prompt_seq_len
```
archana-ramalingam, line 43: We can remove all references of `prompt_seq_len` too. Since we don't have [this ](https://github.com/nod-ai/shark-ai/commit/b4ab597a89d37dc33ad5a700cee8dc3dfebbb7af#diff-bdd6546b53a3a9d6eebd4d80a9485f7440a5070ab7db4133f01234e986c94942R48-R67) change, we don't use `prompt_seq_len` to gen random inputs of length `prompt_seq_len`.

PR #1241, opened 2025-04-09
Add all replicated version of index_put_

Alex-Vasile:
Needed for pipeline parallelism.

### DIFF
--- a/sharktank/sharktank/ops/sharded_impls.py
+++ b/sharktank/sharktank/ops/sharded_impls.py
@@ -669,6 +669,19 @@ def index_copy__split_replicated_split(
     return inout
 
 
+@index_put_.override(AllOfType(ReplicatedTensor))
+def index_put__replicated(
+    inout: ReplicatedTensor,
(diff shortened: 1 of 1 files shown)


sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -669,6 +669,19 @@ def index_copy__split_replicated_split(
     return inout
 
 
+@index_put_.override(AllOfType(ReplicatedTensor))
+def index_put__replicated(
+    inout: ReplicatedTensor,
+    indices: Tuple[ReplicatedTensor],
```
sogartar, line 675: I am not sure it would look into the tuple when dispatching to see that all tensors are replicated.

sogartar (approved):

sharktank/sharktank/ops/sharded_impls.py
sogartar: Actually the trampoline would take the tensors from the tuple.

PR #1254, opened 2025-04-10
Use complex instead of extract

IanWood1:
The `tensor.extract` operations cause IREE to misidentify this kernel as a "gather".

`d3 floordiv 2` is used to broadcast along the innermost pairs without introducing a `x2` dimension in the output which will cause issues with attention if this is fused.

TODO: move the bit-extend of `%input` into the kernel so that the `flow.tensor.bitcast` doesn't block fusion.


MaheshRavishankar:
Nice! Do we need tests here?

IanWood1:
> Nice! Do we need tests here?

Yep, fixed a test that wasn't running and added one for handling bit-extend

IanWood1:
Waiting for https://github.com/iree-org/iree/pull/20580, full model llama tests fail without it.

rsuderman:
Have we benchmarked to guarantee this improves performance?

IanWood1:
> Have we benchmarked to guarantee this improves performance?

Nothing substantial, just the small benchmark mentioned above

MaheshRavishankar:
More than performance this is more about a better representation. I would expect that the performance stays the same.

PR #1259, opened 2025-04-11
Shortfin llm `top_k` + `top_p` sampling

stbaione:
Stacked on #1234 

# Description

This PR implements the [top_p](https://huggingface.co/blog/how-to-generate#top-p-nucleus-sampling) (or `nucleas`) sampling strategy for the Shortfin LLM server. It also allows the use of `top_k + top_p` in series with each other, which I found to be supported by most LLM servers.

## Explanation

`top_p` is a sampling algorithm, where per-beam we sort the logits of a decode invocation. We then select tokens from those logits, until the cumulative probability of those tokens are equal to `top_p`.

We then randomly sample from that distribution by weight, where the number of samples we take depends on the `token_selection_strategy`.

If `top_k` and `top_p` are provided, we first do `top_k` sampling, then `top_p` sampling on the results from `top_k`. The idea is that you can control the subset of tokens that `top_p` is going to sample from. So, a low `top_k` can restrict the subset of tokens to select from, making `top_p` still diverse, but less random overall. A high `top_k` value can increase the size of the subset that `top_p` samples from, and potentially introduce more variation, depending on the tokens that `top_k` randomly selects.

This also includes some simplifications on the overall interface, allowing beams to share the sample `sample_top_k` and `sample_top_p` functions.


shortfin/python/shortfin_apps/llm/components/token_selection_strategy/greedy_token_selection_strategy.py
```diff
@@ -28,9 +28,36 @@ def sample_logits(self) -> int:
...
+
+        # Convert to softmax to obtain probabilities
+        softmax_logits = self.convert_logits_normalization(
+            decode_config.logits_normalization,
+            LogitsNormalization.SOFTMAX,
+            exec_req.result_logits,
+        )
+
+        tokens, probs = softmax_logits, None
+        if top_k != NOT_PROVIDED:
```
rsuderman, line 47: See comment about `NOT_PROVIDED`

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/beam_group.py
```diff
@@ -28,30 +28,84 @@
...
+        logits: sfnp.device_array,
+        **kwargs,
+    ) -> sfnp.device_array:
+        logits_conversion_map = {
+            LogitsNormalization.NONE: {
+                LogitsNormalization.LOG_SOFTMAX: sfnp.log_softmax,
+                LogitsNormalization.SOFTMAX: sfnp.softmax,
+            },
+            LogitsNormalization.SOFTMAX: {
+                LogitsNormalization.LOG_SOFTMAX: sfnp.log,
```
rsuderman, line 64: Add a no-op conversion function, just a simple lambda that returns the input is fine. It avoids the check if the function exists and assuming its a no-op. This way if it does not exist it can throw an error.

rsuderman (dismissed):

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/greedy_token_selection_strategy.py
stbaione: Updated

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/beam_group.py
stbaione: Updated

shortfin/python/array_host_ops.cc
```diff
@@ -942,6 +1068,59 @@ void BindArrayHostOps(py::module_ &m) {
...
+              fmt::format("Axis out of range: Must be [0, {}) but got {}",
+                          input.shape().size(), axis));
+        }
+        if (out && (out->dtype() != input.dtype())) {
+          throw std::invalid_argument(
+              fmt::format("out array must have dtype={} but got {}",
+                          input.dtype().name(), out->dtype().name()));
+        }
+        auto compute = [&]<typename EltTy>() {
+          auto input_t = input.map_xtensor_rw<EltTy>();
```
IanNod, line 1088: same as above, read only?

shortfin/tests/apps/llm/components/token_selection_strategy/beam_search_token_selection_strategy_test.py
```diff
@@ -80,62 +84,84 @@ def _batcher_workitem_callback(_: int):
     pass
 
 
+def approximately_equal(a: Any, b: Any, rel_tol=1e-2, abs_tol=0.0) -> bool:
```
IanNod, line 87: Doesn't this exist already in beam_group_test.py? Can we combine and not use this duplicated code?

shortfin/python/array_host_ops.cc
```diff
@@ -177,6 +177,38 @@ Implemented for dtypes: float16, float32.
   A device_array of dtype=int64, allocated on the host and not visible to the device.
 )";
 
+static const char DOCSTRING_EXP[] =
+    R"(Return the exp of the `input` array.
+
+Implemented for dtypes: float16, float32.
```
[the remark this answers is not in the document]
stbaione: I matched the `supported types` of `argmax`, which is the oldest op we had. It should be extended for fp8 support, and that is doable, but I'm wondering if that should be included in this PR, or done as a separate effort

shortfin/python/array_host_ops.cc
```diff
@@ -889,6 +939,82 @@ void BindArrayHostOps(py::module_ &m) {
...
+      [](device_array &input, std::optional<device_array> out,
+         bool device_visible) {
+        SHORTFIN_TRACE_SCOPE_NAMED("PyHostOp::log");
+        if (out && (out->dtype() != input.dtype())) {
+          throw std::invalid_argument(
+              fmt::format("out array must have dtype={} but got {}",
+                          input.dtype().name(), out->dtype().name()));
+        }
+        auto compute = [&]<typename EltTy>() {
+          auto input_t = input.map_xtensor_rw<EltTy>();
```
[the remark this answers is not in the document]
stbaione: Updated

shortfin/python/array_host_ops.cc
stbaione: Updated

shortfin/python/array_host_ops.cc
```diff
@@ -889,6 +939,82 @@ void BindArrayHostOps(py::module_ &m) {
...
+          }
+
+          auto out_t = out->map_xtensor_w<EltTy>();
+          *out_t = result;
+
+          return *out;
+        };
+
+        switch (input.dtype()) {
+          SF_UNARY_FUNCTION_CASE(float16, half_float::half);
```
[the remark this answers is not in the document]
stbaione: I agree, since all `ops` (old and new) currently only support `float16`, `bfloat16`, and `float32` should that be done as a separate PR/effort?

This ones already getting quite large

shortfin/python/array_host_ops.cc
```diff
@@ -942,6 +1068,59 @@ void BindArrayHostOps(py::module_ &m) {
...
```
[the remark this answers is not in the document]
stbaione: See above reply

shortfin/python/array_host_ops.cc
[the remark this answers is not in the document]
IanNod: Fine to do in a follow up, I would just suggest adding a TODO comment with an assignee

IanNod (approved):

shortfin/python/array_host_ops.cc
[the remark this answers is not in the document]
stbaione: Added a comment at the top of where we define our `array_host_ops`, currently with myself assigned. 

Also created and reference [an issue](https://github.com/nod-ai/shark-ai/issues/1266) in `shark-ai`

PR #1265, opened 2025-04-15
[tuner] Convert from parallel baseline benchmark to serial run

amd-vivekag:
convert from parallel baseline benchmark to serial run


tuner/tuner/libtuner.py
```diff
@@ -851,22 +851,29 @@ def benchmark_baseline(
...
 
-    worker_context_queue = create_worker_context_queue(devices)
-    baseline_results = multiprocess_progress_wrapper(
-        num_worker=len(devices),
-        task_list=task_list,
-        function=run_iree_benchmark_module_command,
-        initializer=init_worker_context,
-        initializer_inputs=(worker_context_queue,),
-    )
+    global worker_id, device_id
```
kuhar, line 855: Why do we need to make this global?

tuner/tuner/libtuner.py
```diff
@@ -851,22 +851,29 @@ def benchmark_baseline(
...
+                        benchmark_timeout=tuning_client.get_benchmark_timeout_s(),
+                        candidate_tracker=candidate_tracker,
+                    )
+                )
+
+                baseline_results.append(result)
+                pbar.update(1)  # Update progress bar
+        except KeyboardInterrupt:
+            # If Ctrl+C is pressed, terminate all child processes
+            sys.exit(1)  # Exit the script
```
kuhar, line 875: nit: Use proper punctuation in comments: https://llvm.org/docs/CodingStandards.html#commenting

kuhar (commented):
Did you notice a measurable improvement in how stable the benchmark results are after switching to serial?

tuner/tuner/libtuner.py
amd-vivekag: because run_iree_benchmark_module_command() is already referring global device_id. So, I had to update the same variable for device_id and worker_id.

tuner/tuner/libtuner.py
amd-vivekag: Full stop is missing. Right? Anything else?

amd-vivekag:
> Did you notice a measurable improvement in how stable the benchmark results are after switching to serial?

Yes, when running the baseline benchmark parallel on 32 devices, there was a huge variance, from 5700 ms to 6700 ms. But when running serially, I could see the numbers were from 5700 to 6000 ms.

tuner/tuner/libtuner.py
kuhar: > Full stop is missing. Right? Anything else?

yes

kuhar:
> Yes, when running the baseline benchmark parallel on 32 devices

We don't have any machines with 32 GPUs, so this seems like an unrealistic setup

amd-vivekag:
> > Yes, when running the baseline benchmark parallel on 32 devices
> 
> We don't have any machines with 32 GPUs, so this seems like an unrealistic setup. The most we should run is one benchmark per GPU ==> 8 benchmarks running in parallel

In CPX mode, we have 64 devices in MI300. I was able to run it using 32 devices.

kuhar:
> In CPX mode, we have 64 devices in MI300. I was able to run it using 32 devices.

These are 64 logical devices but still 8 GPUs. Running more than one benchmark per physical GPU is a bad idea.

amd-vivekag:
> > In CPX mode, we have 64 devices in MI300. I was able to run it using 32 devices.
> 
> These are 64 logical devices but still 8 GPUs. Running more than one benchmark per physical GPU is a bad idea.

So, should I convert it into SPX mode? or is there a way to know which hip:// device is part of same GPU in CPX mode?

kuhar:
CPX is fine but you have to be mindful of the affinity. Logical device IDs are consecutive per physical GPU, so for example GPU0..7 is one physical gpu and GPU8..15 is another one. To distribute across all physical GPUs, you'd select devices 0, 8, 16, ..., 56

tuner/tuner/libtuner.py
amd-vivekag: @kuhar I checked that for the use-case in the tuning code, global variable is the recommended approach. Since all the processes will create their own copy, so initializer will initialize global variable for each process and the same can be used by all the tasks. This is also helping the code to be reusable because `multiprocess_progress_wrapper()` API has been written to be generic and is being used for both compilation and benchmarking.
IMHO, the global variable here is not an issue, instead helping in multi processing. Let me know your thoughts.

tuner/tuner/libtuner.py
```diff
@@ -851,22 +851,31 @@ def benchmark_baseline(
...
+                        benchmark_timeout=tuning_client.get_benchmark_timeout_s(),
+                        candidate_tracker=candidate_tracker,
+                    )
+                )
+
+                baseline_results.append(result)
+                pbar.update(1)  # Update progress bar
+        except KeyboardInterrupt:
+            # If Ctrl+C is pressed, terminate all child processes.
+            sys.exit(1)  # Exit the script
```
kuhar, line 877:
```suggestion
            sys.exit(1)  # Exit the script.
```

tuner/tuner/libtuner.py
kuhar: @Max191 could you review this part?

Max191: I think this seems okay to me. The reason the globals are there is that, since the tasks are picked up from a pool, the device_id belongs to the worker, not the task. AFAIK there isn't a nice way around having the globals, and if there is, then it makes sense to me to leave that for a separate PR.

tuner/tuner/libtuner.py
```diff
@@ -851,22 +851,31 @@ def benchmark_baseline(
...
+    global worker_id, device_id
+
+    baseline_results = list()
+
+    # Use tqdm to create a progress bar.
+    with tqdm(total=len(devices)) as pbar:
+        try:
+            for worker_id_, device_id_ in enumerate(devices):
+                worker_id = worker_id_
+                device_id = device_id_
```
Max191, line 864: Since this is really just a single worker, it makes more sense to set the worker_id to `0` for all iterations (and then you can get rid of the `enumerate()` too).

Max191 (approved):
This seems fine to me. In general we should be tuning with just one logical device per physical device, and this will likely not make much difference in stability, but I don't think we have a great way of warning the user if they are using too many logical devices per physical device. Since the baseline benchmarks are a small percentage of overall time, I do think this will be helpful for the average user, who may not be aware that they are using too many logical devices.

In the future we could revisit this if we add some logic in the tuner to tell how many logical devices per physical device we have.

Please address the comment, and also update the description to explain the reasoning for running baselines in parallel before landing.

kuhar:
@amd-vivekag have you address the comments from Max before landing?

amd-vivekag:
> @amd-vivekag did you address the comments from Max before landing?

Hi @kuhar , I missed that comment and merged the changes. I saw that message later so I thought to add that along with another task on which I'm working right now. I hope it is fine.

tuner/tuner/libtuner.py
amd-vivekag: Hi @Max191 , I missed this comment earlier, so I have included your suggestion in another task I'm working on currently. I hope it is fine.

kuhar:
Please send a separate PR addressing the comment

PR #1269, opened 2025-04-16
Add basic overview flowchart of tuning algorithm

amd-vivekag:
Add basic overview flowchart of tuning algorithm


kuhar (commented):
Thanks for improving the documentation!

In the current form, this should be a numbered list instead of a flow chart. Alternatively, I think what is missing are loops to show where the loops are within the tuning flow.

amd-vivekag:
> Thanks for improving the documentation!
> 
> In the current form, this should be a numbered list instead of a flow chart. Alternatively, I think what is missing are loops to show where the loops are within the tuning flow.

You are right, the basic flowchart was not giving good details. Adding loops information through flowchart will definitely give good idea. Since it will require some time to create full flowchart with such details, so I've converted it into numbered list for now. 

If I will get chance, I will update it further with more details.

kuhar (approved):

PR #1273, opened 2025-04-16
Fix processing of `sampling_params` within a batched request

stbaione:
Introduced a bug in #1214 where when sending a batch request with `sampling_params` defined, we were improperly attempting to update the `DecodeConfig` with the `List[SamplingParams]`, instead of updating with the `SamplingParams` object that corresponded to the specific request within the batch.

This implements a fix for it.

### DIFF
--- a/shortfin/python/shortfin_apps/llm/components/generate.py
+++ b/shortfin/python/shortfin_apps/llm/components/generate.py
@@ -9,6 +9,8 @@ import dataclasses
 import io
 import json
 import logging
+
+from copy import copy
 from typing import List
 
 import shortfin as sf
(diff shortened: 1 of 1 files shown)


shortfin/python/shortfin_apps/llm/components/generate.py
```diff
@@ -166,6 +167,12 @@ async def run(self):
             else:
                 input_batch = self.tokenize()
             for index, input_tokens in enumerate(input_batch):
+                decode_config = copy(self.decode_config)
+                decode_config.update_from_sampling_params(
+                    self.gen_req.sampling_params
+                    if self.gen_req.is_single
+                    else self.gen_req.sampling_params[index]
+                )
```
zphoenixrises, line 175: Instead of copying every time, you could possibly only copy if there is a change in the sampling params:

```suggestion
                sampling_params = (
                    self.gen_req.sampling_params if self.gen_req.is_single 
                    else self.gen_req.sampling_params[index]
                )
                decode_config = (
                    copy(self.decode_config) if sampling_params 
                    else self.decode_config
                )
                if sampling_params:
                    decode_config.update_from_sampling_params(sampling_params)
```

zphoenixrises (approved):
Not a blocking comment. Just a slight optimization.

PR #1275, opened 2025-04-17
Reject requests to protect server

zphoenixrises:
This PR monitors the request queue within the shortfin server and rejects requests with a 503 error when the queue is filled with more than max batch size + 2

TODO: Add load testing to integration tests


stbaione (approved):
LGTM, can extend integration tests for this use case in a follow up PR

shortfin/python/shortfin_apps/llm/components/generate.py
```diff
@@ -151,12 +155,29 @@ def __init__(
 
     async def run(self):
         logger.debug("Started ClientBatchGenerateProcess: %r", self)
-        streaming = self.gen_req.stream
-        self.responder.start_response()
-        if streaming:
-            self.responder.stream_start()
+
+        # Try to add request to queue
+        # TODO(@zphoenixrises): Add load testing and integration tests for this.
```
IanNod, line 160: Can you file an issue and link it here?

IanNod (approved):
Minor nit comment but otherwise looks good to me

PR #1276, opened 2025-04-17
Remove `temperature` bottleneck

stbaione:
Previously, we were taking temperature on the entire `result_logits` from a decode invocation.

Our `mistral` logits are of shape `[1, 1, 128256]`. When I ran a local benchmark on how long the `sfnp.divide` function was taking for this size of an array, it was ~14ms. 

So, `(14ms * 16 parallel_reqs * 64 decode_steps) / 1000 == 14 s`!

In local benchmarks, I was able to decrease the latency of 16 concurrent requests from `~25s` to `~9s`, which also increased our throughput quite a bit.

The idea is pretty simple:

For greedy selection, or if already in a `softmax`/`log_softmax` form don't take temperature. Dividing by a scalar won't impact which token is the highest scoring token or which ones are the `top_k` highest scoring tokens. It doesn't make sense to apply temperature if already in `softmax` or `log_softmax` form.

Otherwise only take temperature on the values that are getting converted to `softmax`. This reduces the size of the array we are dividing drastically.


shortfin/tests/apps/llm/components/token_selection_strategy/beam_group_test.py
zphoenixrises, line 1: Can we modify the tests to keep track of the time as well? Say the time to run the test needs to be below a certain threshold. So that we are no regressing this in future commits. This would technically be part of benchmark tests. But we can refactor that later.

zphoenixrises (commented):
Awesome fix! Let's find some more gems like this!

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/beam_search_token_selection_strategy.py
```diff
@@ -69,22 +72,40 @@ def sample_logits(self, k: int):
...
+        logits = self.exec_req.result_logits
         decode_config = self.decode_config
         num_beams = decode_config.num_beams
         top_k = decode_config.top_k
         top_p = decode_config.top_p
 
         if (top_k, top_p) == (None, None):
-            log_softmax_logits = self.convert_logits_normalization(
+            tokens, probs = self.sampler.select_top_k(logits, -k)
+            if logits.dtype in [sfnp.float16]:
```
IanNod, line 83: Why do we need this? And does this need to be fp16 specific or should this be done for any float type?

stbaione: This is `fp16` specific. I assume it'll apply to other partial `float` types, but wanted to keep the scope smaller and add them when they became relevant. This isn't needed for `fp32`.

If you attempt to set the items of the `fp16` device_array, before converting them to their int equivalent, an error is thrown due to the bit-width being too large.

It's faster to build the device_array once, then do the `temperature` and `logit_conversion` on that device_array, then to do it on a list.

When you access the `items` of the fp16 device_array, you get them back as ints, so that's why they need to be converted back to float, before the log_prob values are returned

shortfin/tests/apps/llm/components/token_selection_strategy/beam_group_test.py
stbaione: We should add a test like that, but I don't think it would make a ton of sense on these tests specifically. Temperature itself didn't really get faster, we're just using it in a more intelligent way. I can look around for another spot where a benchmark measurement like that would make sense.

Would you be cool with us treating that as a separate PR/work item, so that we can get this functionality in?

zphoenixrises (approved):

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/beam_search_token_selection_strategy.py
IanNod: Can you link this issue in this todo? So would look something like this
```
# TODO: https://github.com/nod-ai/shark-ai/issues/1278 find cleaner way to do these conversions.
```

stbaione: Updated to link the issue, instead of just including issue number

IanNod (approved):

PR #1280, opened 2025-04-17
Shortfin llm multi workers and multi fibers

stbaione:
Enable spinning up the server with multiple workers and multiple fibers per worker to facilitate higher parallelism and concurrency in prefill/decode invocations.

## TODO:
- Fix `per_fiber` isolation
    - I'm also attempting to enable `per_fiber` program isolation. Currently, we've just been using `per_call`. As I understand, `per_fiber` may end up being faster. I'm getting a `std::bad_cast` error when attempting to create the input device_arrays in `get_args`. Still looking into this... Otherwise good to use with `per_call` isolation.
    - Right now, we're in a performance sprint, and get more benefit by focusing on fixing this another time. Created [an issue](https://github.com/nod-ai/shark-ai/issues/1284) for it, and will add a `NotImplemented` error in code for this case.


monorimet (approved):
LGTM. May need FiberPool to use a Queue instead of list for GIL-free when we get to that point. Have one question about device selection for workers.

shortfin/python/shortfin_apps/llm/components/service.py
```diff
@@ -77,11 +79,24 @@ def remove_from_queue(self):
...
-        self.main_worker = self.sysman.ls.create_worker(f"{self.name}-inference")
-        self.main_fiber = self.sysman.ls.create_fiber(self.main_worker)
-        self.prefill_fiber = self.sysman.ls.create_fiber(self.main_worker)
-        self.decode_fiber = self.sysman.ls.create_fiber(self.main_worker)
+        logger.info(
+            f"Creating {num_workers} workers, with {fibers_per_worker} fibers per worker..."
+        )
+        fibers = []
+        for i in range(num_workers):
+            worker = self.sysman.ls.create_worker(f"{self.name}-inference-{i}")
```
monorimet, line 90: Does this do what we want in terms of device selection?

stbaione: I think so. Currently, the LLM server doesn't have the same implementation as the SDXL server, where it can deploy multiple instances of the model to multiple devices and round robin between those devices.

Currently, if multiple devices are specified, the server code is written to assume that you're sharding a model across those devices. So, it either runs an unsharded model on one device, or a single `tpN` (sharded) model on `N` devices.

We SHOULD implement that as a future benefit, but currently for this performance sprint, all of our benchmarks are for a single device.

So, IIUC, until we have that added, we assume that all devices specified are being used to run a single instance of the model, which I believe matches the current setup. Does that make sense?

monorimet: Yeah, sounds good to me.

PR #1288, opened 2025-04-21
Change dump_args to support TP, PP, and list of tensors

Alex-Vasile:
The current implementation does not support dumping `cache_state` when running with tensor or pipeline parallelism since it doesn't handle ShardedTensors or list of tensors. Changes to allow that to work.

### DIFF
--- a/sharktank/sharktank/utils/load_llm.py
+++ b/sharktank/sharktank/utils/load_llm.py
@@ -4,6 +4,7 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+import collections
 import math
 from pathlib import Path
 from typing import Optional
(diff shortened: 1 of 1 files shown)


sharktank/sharktank/utils/load_llm.py
```diff
@@ -203,17 +204,28 @@ def dump_args(
...
+                )
+        else:
+            if arg.dtype in [torch.float8_e4m3fnuz, torch.bfloat16]:
+                arg = arg.to(torch.uint8)
+            if phase == "decode":
+                arg_name = f"{arg_name}_{decode_step}"
+            rank_str = f".rank{rank}" if rank else ""
+
+            file_path = Path(self.dump_path, f"{phase}_{arg_name}{rank_str}.npy")
+            np.save(file_path, arg.cpu().numpy())
```
archana-ramalingam, line 228: Update `write_ndarray_to_bin` as discussed.

Alex-Vasile:
Folded in to https://github.com/nod-ai/shark-ai/pull/1238

PR #1293, opened 2025-04-23
[sharktank] Fix issues from recent refactor

archana-ramalingam:
Fix issues from recent refactor:

- Eager mode ppl being skipped in nightly
- torch.view issue
- FP8 issue

### DIFF
--- a/.github/workflows/ci-sharktank-nightly.yml
+++ b/.github/workflows/ci-sharktank-nightly.yml
@@ -7,7 +7,6 @@
 name: CI - sharktank nightly
 
 on:
-  pull_request:
   workflow_dispatch:
   schedule:
     # Weekdays at 11:00 AM UTC = 03:00 AM PST / 04:00 AM PDT
--- a/.github/workflows/ci-sharktank-nightly.yml
+++ b/.github/workflows/ci-sharktank-nightly.yml
@@ -7,6 +7,7 @@
 name: CI - sharktank nightly
 
 on:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-sharktank-nightly.yml
```diff
@@ -7,6 +7,7 @@
 name: CI - sharktank nightly
 
 on:
+  pull_request:
```
IanNod, line 10: Why are we moving the nightly sharktank tests to on pull_request?

sharktank/tests/evaluate/perplexity_torch_test.py
```diff
@@ -24,7 +23,6 @@
     "batch_size",
     "device",
 )
-@is_mi300x
```
IanNod, line 27: How long does this test take on CPU?

archana-ramalingam: `@is_mi300x` is a IREE specific device specification flag and not needed in eager mode. We use `--device:'cuda:0'` instead. This flag was skipping all eager mode nightly tests, so removing it now.

.github/workflows/ci-sharktank-nightly.yml
archana-ramalingam: Was testing nightly issues, removed it now.

IanNod (approved):

PR #1294, opened 2025-04-23
[SharkTuner][WIP] Packaging tuner with next release

Muzammiluddin-Syed-ECE:
[WIP] Various changes to include tuner in next release.

- Renamed tuner to SharkTuner 
- Add to meta package


sharktuner/examples/simple/shark_tuner.py
```diff
@@ -10,7 +10,7 @@
 from tuner.common import *
 
 
-class SimpleTuner(libtuner.TuningClient):
+class SharkTuner(libtuner.TuningClient):
```
kuhar, line 13: Can you put up a separate PR that does the renaming? It should be easier to review when we reduce the scope.

build_tools/python_deploy/pypi_deploy.sh
```diff
@@ -60,6 +60,9 @@ function download_wheels() {
   #   * https://pip.pypa.io/en/stable/installation/
   python3.13t -m pip download shortfin==${RELEASE} --no-deps -f ${ASSETS_PAGE}
 
+  # tuner
```
Muzammiluddin-Syed-ECE, line 63: IIUC, currently, we have no tuner releases so I imagine this command would fail. So, I was wondering what the best call would be here. 
1) Uncomment these lines unchanged, manually set up a single release of the tuner to be consumed here.
2) Add a failure guard like an if statement that fails with a warning message if this package is not found which will be removed after the first release in a future PR

sharktuner/examples/simple/shark_tuner.py
Muzammiluddin-Syed-ECE: I'll create that PR after this one goes through, if that's ok

kuhar: If think the reverse order would be better -- first change the name, then package. This way we won't have to deal with two packages of the same thing under different names

Muzammiluddin-Syed-ECE: [Renaming PR](https://github.com/nod-ai/shark-ai/pull/1297)

marbre (commented):
Some early feedback without looking at the name changes for now.

requirements.txt
```diff
@@ -1,4 +1,6 @@
 -r sharktank/requirements.txt
 -r sharktank/requirements-tests.txt
 -r shortfin/requirements-tests.txt
+-r tuner/requirements-dev.txt
+-r tuner/requirements-tuner.txt
```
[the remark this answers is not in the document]
Muzammiluddin-Syed-ECE: I might be misunderstanding the purpose of this file, but doesn't it attempt to install the dependencies of all the subpackages in this meta package, so sharktuner's dependencies would also need to be included?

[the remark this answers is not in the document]
marbre: No, this is for pure CI / development use. The meta-package `shark-ai` only pulls in

* `shortfin`
* `sharktank`
* `iree-base-compiler`
* `iree-base-runtime`
* `iree-turbine`

via https://github.com/nod-ai/shark-ai/blob/d54016c487cce2e132e2e1f992da389bbda297f7/shark-ai/pyproject.toml#L36 which is created during the build process.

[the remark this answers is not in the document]
marbre: The packages themselves need to specify the deps, _not_ the meta-package.

sharktuner/pyproject.toml
```diff
@@ -22,3 +22,8 @@ dynamic = ["version"]
 
 [project.urls]
 Repository = "https://github.com/nod-ai/shark-ai"
+
+[tool.setuptools.packages.find]
+where = ["."]
+include = ["sharktuner*"]
```
marbre, line 28: With this `model_tuner` wont be packages which is, how I understood`, not the intention.

marbre (changes requested):
Please make sure the package builds. Right now the `pyproject.toml` needs improvement, like fixing the `name` (needs to be `sharktuner`), dependencies are not define, not all contents get packages. You may want to take a look at the other packages in this repository.

build_tools/python_deploy/README.md
ScottTodd, line 1: Can you split the `build_tools/python_deploy/` changes into a second PR? That way we can start using/testing the nightly releases without fully committing to including them in stable (until they meet whatever release bar we set).

This first PR:

1. Get the new package building
2. Publishing nightly releases of the new package

The second PR:

1. Wire up pypi deployment of the new package
2. Include the new package in the shark-ai meta package

ScottTodd (changes requested):

build_tools/python_deploy/README.md
Muzammiluddin-Syed-ECE: I have this pull requestion up: https://github.com/nod-ai/shark-ai/pull/1334

Until this first pr is merged in, I will continue rebasing this second pr.

Muzammiluddin-Syed-ECE: Let me know if this is the split you were looking for

marbre: I think PR #1334 has some things that should only part of a PR that adds nightly releases (without adding it to the meta-package). I've commented to #1334 as well.

.github/actions/pkgci-setup/action.yml
```diff
@@ -36,7 +36,7 @@ runs:
       uses: actions/cache@1bd1e32a3bdc45362d1e726936510720a7c30a57 # v4.2.0
       with:
         path: .uv-cache
-        key: ${{ runner.os }}-uv-py${{ inputs.python-version }}-${{ hashFiles('requirements-iree-pinned.txt', 'pytorch-cpu-requirements.txt', 'sharktank/requirements.txt', 'sharktank/requirements-tests.txt', 'shortfin/requirements-tests.txt') }}
+        key: ${{ runner.os }}-uv-py${{ inputs.python-version }}-${{ hashFiles('requirements-iree-pinned.txt', 'pytorch-cpu-requirements.txt', 'sharktank/requirements.txt', 'sharktank/requirements-tests.txt', 'shortfin/requirements-tests.txt', 'sharktuner/requirements-dev.txt', 'sharktuner/requirements-tuner.txt') }}
```
marbre, line 39: There should be no need to add `sharktuner/requirements-dev.txt'` here at all.

sharktuner/pyproject.toml
```diff
@@ -18,7 +22,21 @@ classifiers = [
...
+[tool.setuptools.packages.find]
+where = ["."]
+include = ["sharktank*"]
+namespaces = true
+
+[tool.setuptools.package-data]
+sharktank = ["py.typed"]
+
+[tool.setuptools.dynamic.dependencies]
+file = ["requirements-tuner.txt"]
```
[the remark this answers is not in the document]
marbre: The requirements file has some issues as it pins explicit versions. This shouldn't be need or rather define a minimum version or a version range instead of a single version. This needs to be addressed before the sharktuner can be added to the meta-package.

marbre (changes requested):

.github/actions/pkgci-setup/action.yml
```diff
@@ -36,7 +36,7 @@ runs:
       uses: actions/cache@1bd1e32a3bdc45362d1e726936510720a7c30a57 # v4.2.0
       with:
         path: .uv-cache
-        key: ${{ runner.os }}-uv-py${{ inputs.python-version }}-${{ hashFiles('requirements-iree-pinned.txt', 'pytorch-cpu-requirements.txt', 'sharktank/requirements.txt', 'sharktank/requirements-tests.txt', 'shortfin/requirements-tests.txt') }}
+        key: ${{ runner.os }}-uv-py${{ inputs.python-version }}-${{ hashFiles('requirements-iree-pinned.txt', 'pytorch-cpu-requirements.txt', 'sharktank/requirements.txt', 'sharktank/requirements-tests.txt', 'shortfin/requirements-tests.txt', 'sharktuner/requirements.txt') }}
```
marbre, line 39: I took another look: This composite action is only used in `.github/workflows/pkgci_shark_ai.yml` and that workflow only tests shortfin and sharktank. Thus there is no need to install sharktuner deps. Please revert to the previous state.

marbre (changes requested):
This is in a way better shape now and should be close to land. Please see my request to revert some of the changes. I further test to build the package locally later today and see how it goes.

build_tools/python_deploy/README.md
Muzammiluddin-Syed-ECE: Removed those changes, and have retained them in this PR.

sharktuner/examples/simple/shark_tuner.py
Muzammiluddin-Syed-ECE: Merged

sharktuner/requirements.txt
```diff
@@ -0,0 +1,5 @@
+numpy
+pytest
```
ScottTodd, line 2: pytest should be a test only requirement

The tqdm, types-tqdm, and z3_solver requirements here might also be package build/develop requirements, not install requirements? Can you check?

The end state should be:

```bash
pip install sharktuner

pip freeze
# see a minimal list of installed packages here

# ... test the package to make sure it is functional
```

ScottTodd (commented):
Thanks, the structure here looks good now. Just wondering about the exact requirements needed.

sharktuner/build_tools/build_linux_package.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
marbre, line 1: Seems this needs execute permissions (`+x+).

sharktuner/pyproject.toml
```diff
@@ -18,7 +22,21 @@ classifiers = [
...
 # Version is set via the `setup.py`.
-dynamic = ["version"]
+dynamic = ["version", "dependencies", "optional-dependencies"]
 
 [project.urls]
 Repository = "https://github.com/nod-ai/shark-ai"
+
+[tool.setuptools.packages.find]
+where = ["."]
+include = ["sharktank*"]
```
marbre, line 32:
```suggestion
include = ["sharktuner*"]
```

sharktuner/pyproject.toml
```diff
@@ -18,7 +22,21 @@ classifiers = [
...
+namespaces = true
+
+[tool.setuptools.package-data]
+sharktank = ["py.typed"]
```
marbre, line 36:
```suggestion
sharktuner = ["py.typed"]
```

marbre (changes requested):

sharktuner/pyproject.toml
Muzammiluddin-Syed-ECE: woops this was something I fixed in a different commit which got lost in the rebasing

sharktuner/requirements.txt
Muzammiluddin-Syed-ECE: FYI, after making a fresh `venv` and trying to run this package alone I discovered a dependency on `PyYAML` which is a requirement of `iree-compile` and its python bindings. I assume that this would be installed when installing `iree-compile` so I should be okay to not add it to `sharktuner/requirements.txt` file

sharktuner/pyproject.toml
```diff
@@ -18,7 +22,21 @@ classifiers = [
...
 # Version is set via the `setup.py`.
-dynamic = ["version"]
+dynamic = ["version", "dependencies", "optional-dependencies"]
 
 [project.urls]
 Repository = "https://github.com/nod-ai/shark-ai"
+
+[tool.setuptools.packages.find]
+where = ["."]
+include = ["sharktuner*"]
```
marbre, line 32:
```suggestion
include = ["sharktuner", "model_tuner"]
```
To package `model_tuner` this needs to be included here. Furthermore, I think we can drop the asterisk from `sharktuner`.

marbre (approved):
Test failures are unrelated. Changes lgtm, thanks for incorporating all the feedback!

PR #1302, opened 2025-04-24
use number of workers as number of logical devices per physical device

dezhiAmd:
**Why**
    Without a logical device for each worker, workload/papalism on each physical gpu will not change
**How**
   By setting environment variable "SHORTFIN_AMDGPU_LOGICAL_DEVICES_PER_PHYSICAL_DEVICE" using number of workers, system can create one logical device for each worker


vinayakdsci (commented):
Could we not just add this as a flag? Imagine a scenario where I did not really want a worker to queue work on the device, but rather parallelize work on the host.

This would be tying the concept of a logical device to a worker too closely. Your thoughts?

stellaraccident:
Is there a PR for how this is implemented in the AMDGPUSystemBuilder? Need to make sure that it is creating the Devices with proper queue affinities. And that should be a kwarg on the system builder.

Better to have flags vs env vars like Vinayak says.

dezhiAmd:
Thanks for the feedback. I will rework the whole picture.

dezhiAmd:
> Is there a PR for how this is implemented in the AMDGPUSystemBuilder? Need to make sure that it is creating the Devices with proper queue affinities. And that should be a kwarg on the system builder.
> 
> Better to have flags vs env vars like Vinayak says.

@stellaraccident,

This is implemented here in the existed code ([amdgpu.cc](https://github.com/nod-ai/shark-ai/blob/1a0c0ccaf461076c0fefac0a3252087cfa74df38/shortfin/src/shortfin/local/systems/amdgpu.cc#L73)

dezhiAmd:
> Could we not just add this as a flag? Imagine a scenario where I did not really want a worker to queue work on the device, but rather parallelize work on the host.
> 
> This would be tying the concept of a logical device to a worker too closely. Your thoughts?

From the [PR ](https://github.com/nod-ai/shark-ai/pull/1391#discussion_r2076294391)you authored, I think now you agree with what I have done here. Can you please approve this change since you are using the same idea?

dezhiAmd:
This PR has everything needed to create multiple logical devices on one physical device:
https://github.com/nod-ai/shark-ai/pull/1403/files

dezhiAmd:
Close this PR since FiberPool dynamically create worker and fiber

PR #1308, opened 2025-04-25
Add ops.mean for SplitPrimitiveTensor

Alex-Vasile:
Only strange thing about this op that I haven't seen in others is that doing the mean across the `shard_dim` requires us to return a `ReplicatedTensor` rather than a `SplitPrimitiveTensor`, especially when `keepdim == False` since the `shard_dim` stops existing.

### DIFF
--- a/sharktank/sharktank/ops/sharded_impls.py
+++ b/sharktank/sharktank/ops/sharded_impls.py
@@ -1063,15 +1063,32 @@ def mean_split(
     keepdim: bool,
     *,
     dtype: torch.dtype,
-) -> None:
-    assert keepdim, "Needs special care"
+) -> SplitPrimitiveTensor | ReplicatedTensor:
     if dim != x.shard_dim:
+        # If keepdim == False dim is smaller than shard_dim we need to offset
+        # shard_dim_new by -1 to have it point to the same dimension.
+        shard_dim_new = x.shard_dim
+        if (not keepdim) and dim < x.shard_dim:
+            shard_dim_new -= 1
+
         shards = [
             mean(shard, dim=dim, keepdim=keepdim, dtype=dtype) for shard in x.shards
         ]
+        return SplitPrimitiveTensor(ts=shards, shard_dim=shard_dim_new)
     else:
-        shards = x.shards  # TODO
-    return SplitPrimitiveTensor(ts=shards, shard_dim=x.shard_dim)
+        gathered = cat(
+            [
+                (
+                    transfer_to_logical_device(shard, x.devices[0])
+                    if i != 0
+                    else barrier_on_logical_device(shard, x.devices[0])
+                )
+                for i, shard in enumerate(x.shards)
+            ],
+            dim=dim,
+        )
+        meaned = mean(gathered, dim=dim, keepdim=keepdim, dtype=dtype)
+        return ReplicatedTensor(ts=meaned, shard_count=x.shard_count, devices=x.devices)
 
 
 @module_register_buffer.override(torch.nn.Module, ShardedTensor)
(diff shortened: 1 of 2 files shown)


sogartar:
> Only strange thing about this op that I haven't seen in others is that doing the mean across the `shard_dim` requires us to return a `ReplicatedTensor` rather than a `SplitPrimitiveTensor`, especially when `keepdim == False` since the `shard_dim` stops existing.

I think the sensible thing is to return a replicated tensor when a reduction dimension is sharded.

Alex-Vasile:
> > Only strange thing about this op that I haven't seen in others is that doing the mean across the `shard_dim` requires us to return a `ReplicatedTensor` rather than a `SplitPrimitiveTensor`, especially when `keepdim == False` since the `shard_dim` stops existing.
> 
> I think the sensible thing is to return a replicated tensor when a reduction dimension is sharded.

I agree, nothing else makes sense. I just haven't seen another op when this was required. That's what I've implemented here.

sharktank/tests/ops/sharded_test.py
```diff
@@ -887,6 +888,38 @@ def testReplicatedLhsAndRhs(self):
...
+)
+class MeanTest(unittest.TestCase):
+    def setUp(self):
+        self.shape = (2, 4, 6)
+        self.shard_dim = 1
+        self.mean_dim = self.shard_dim + self.mean_dim_delta
+        self.shard_count = 2
+
+    def testMeanReplicated(self):
+        tensor = torch.rand(self.shape, dtype=torch.float32)
```
sogartar, line 902: I know other tests are also guilty of this, but this will introduce nondeterminism in the test as the torch RNG was not manually seeded.

sharktank/tests/ops/sharded_test.py
```diff
@@ -887,6 +888,38 @@ def testReplicatedLhsAndRhs(self):
...
+        expected_result = ops.mean(tensor, dim=self.mean_dim, keepdim=self.keepdim)
+        actual_result = ops.mean(
+            ops.replicate(tensor, count=self.shard_count),
+            dim=self.mean_dim,
+            keepdim=self.keepdim,
+        )
+        assert ops.equal(expected_result, actual_result)
```
sogartar, line 909: If for some reason we ever change the order of additions compared to the reference this will fail due to rounding loss of precision. It is better to do `torch.testing.assert_close` to allow for some error tolerance.
The same holds for the other test.

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -1051,11 +1051,46 @@ def mean_replicated(
...
+                    transfer_to_logical_device(shard, x.devices[0])
+                    if i != 0
+                    else barrier_on_logical_device(shard, x.devices[0])
+                )
+                for i, shard in enumerate(x.shards)
+            ],
+            dim=dim,
+        )
+        meaned = mean(gathered, dim=dim, keepdim=keepdim, dtype=dtype)
+        return ReplicatedTensor(ts=meaned, shard_count=x.shard_count, devices=x.devices)
```
sogartar, line 1091: It is not critical for correctness, but for performance reasons I think we could change the implementation to
1. Each device reduces along dim.
2. All-reduce across the devices.
3. Divide by the total number of elements in the above reductions.

For most cases this should bring down communication a lot.

sogartar (approved):

sharktank/sharktank/ops/sharded_impls.py
sogartar: This can be in a later PR.

Alex-Vasile: I considered this, my worry was about potential under/over flow. And that underlying floating point operations would no longer be the same leading to slight differences between the results.

sogartar: The operation does not guarantee in what order are additions performed. The underlying implementation may tile on a reduction dimension even on a GPU. Actually, tiling would yield better numerical stability. The accumulator would not get that big and lose precision when adding small and big numbers together.

PR #1310, opened 2025-04-25
[tuner] padding support along TileAndFuse pipeline

bangtianliu:
Motivated by bug reported from this [input](https://gist.github.com/rkayaith/0c537b5b917488a890273f7d886791c7) 

**Problem:**
In the current tuner, tile sizes are required to both evenly divide the problem size and be multiples of the intrinsic size. When the problem size is not a multiple of the intrinsic size, the solver fails to find any valid solutions.

**Solution:**
The simplest fix is to round the problem size up to the nearest multiple of the intrinsic size when generating constraints, and then account for the necessary padding in the lowering configuration.

This PR introduces two main changes from Max's local experimental tuning branch into the main tuner branch:

- Change 1: modify the constraint generation to support unaligned problem shapes.
- Change 2: add padding information to the lowering configuration.
- Include corresponding tests for the above changes.

### DIFF
--- a/sharktuner/sharktuner/dispatch_constraints.py
+++ b/sharktuner/sharktuner/dispatch_constraints.py
@@ -307,7 +307,7 @@ def generate_compilation_infos(
     codegen_pipeline: iree_codegen.DispatchLoweringPassPipeline,
     pipeline_options_search_space: PipelineOptionsSearchSpace,
     allowed_waves_per_eu: list[int],
-    required_padding: bool,
+    padding: Optional[list[int]] = None,
 ) -> list[iree_codegen.CompilationInfoAttr]:
     # Create the LoweringConfigAttr.
     lowering_config_args = {
@@ -320,15 +320,8 @@ def generate_compilation_infos(
         "promote_operands": promote_operands,
     }
 
-    if required_padding:
-        workgroup_tile_m, workgroup_tile_n, _ = workgroup_tile_sizes
-        _, _, reduction_tile_k = reduction_tile_sizes
-        _, _, intrinsic_k = mma_attr.mnk_shape
-        lowering_config_args["padding"] = (
-            workgroup_tile_m,
-            workgroup_tile_n,
-            reduction_tile_k * intrinsic_k,
-        )
+    if padding is not None:
+        lowering_config_args["padding"] = padding
 
     if codegen_pipeline == iree_codegen.DispatchLoweringPassPipeline.LLVMGPUTileAndFuse:
         lowering_config_args["subgroup"] = subgroup_tile_sizes
@@ -527,7 +520,18 @@ def generate_solutions(
         )
         promote_operands = [0, 1]
         if required_padding:
+            # TODO: Remove promotion of operand 2 once codegen supports handling padded outputs without promotion.
             promote_operands = [0, 1, 2]
+            workgroup_tile_m, workgroup_tile_n, _ = workgroup_tile_sizes
+            _, _, reduction_tile_k = reduction_tile_sizes
+            _, _, mma_intrinsic_k = mma_attr.mnk_shape
+            padding = [
+                workgroup_tile_m,
+                workgroup_tile_n,
+                reduction_tile_k * mma_intrinsic_k,
+            ]
+        else:
+            padding = None
 
         compilation_infos = generate_compilation_infos(
             tuner_ctx,
@@ -543,7 +547,7 @@ def generate_solutions(
             codegen_pipeline,
             pipeline_options_search_space,
             allowed_waves_per_eu,
-            required_padding=required_padding,
+            padding=padding,
         )
 
         solver.add(z3.simplify(z3.Not(z3.And(list(x == model[x] for x in all_vars)))))
--- a/sharktuner/sharktuner/dispatch_constraints.py
+++ b/sharktuner/sharktuner/dispatch_constraints.py
@@ -160,9 +160,12 @@ def generate_tile_and_fuse_constraints(
     subgroup_n_count: z3.ArithRef,
     mma_intrinsics: list[iree_gpu.MMAIntrinsic],
 ):
-    M, N, K = problem_size.MNK
+    M, N, K = map(list, problem_size.MNK)
     m_tiles, n_tiles, k_tiles, subgroup_m_tiles, subgroup_n_tiles = tile_sizes
     intrinsic_mn, intrinsic_k = intrinsic_size
+    M[-1] = ((M[-1] + intrinsic_mn - 1) / intrinsic_mn) * intrinsic_mn
+    N[-1] = ((N[-1] + intrinsic_mn - 1) / intrinsic_mn) * intrinsic_mn
+    K[-1] = ((K[-1] + intrinsic_k - 1) / intrinsic_k) * intrinsic_k
     wg_x, wg_y, wg_z = workgroup_size
     wg_threads = wg_x
     constraints = [wg_y == 1, wg_z == 1]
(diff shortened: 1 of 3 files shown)


sharktuner/sharktuner/dispatch_constraints.py
```diff
@@ -502,6 +521,14 @@ def set_cdim_tile_sizes(tile_sizes, contraction_dims, csizes):
             [lookup(v) for v in k_vars],
         )
 
+        required_padding = any(
+            p[-1] % i != 0
+            for p, i in zip(problem_size.MNK, intrinsic_mnk_shape, strict=True)
+        )
+        promote_operands = [0, 1]
+        if required_padding:
+            promote_operands = [0, 1, 2]
```
Max191, line 530: Add a TODO here for removing operand `2` promotion when codegen can support it. We currently need to promote the C matrix when we have padding, but Nirvedh and Jerry are working on getting rid of this restriction. Let's add the TODO so we don't forget about this later.

sharktuner/sharktuner/dispatch_constraints.py
```diff
@@ -300,9 +303,11 @@ def generate_compilation_infos(
     subgroup_size: int,
     subgroup_m_count: int,
     subgroup_n_count: int,
+    promote_operands: list[int],
     codegen_pipeline: iree_codegen.DispatchLoweringPassPipeline,
     pipeline_options_search_space: PipelineOptionsSearchSpace,
     allowed_waves_per_eu: list[int],
+    required_padding: bool,
```
Max191, line 310: Let's refactor this a little and instead of passing a bool for required padding, we can pass the list of actual padding to add. This way, we could even make the amount of padding a tunable parameter, and include it as part of the constraints if we want in the future.

sharktuner/sharktuner/dispatch_constraints_test.py
```diff
@@ -51,6 +51,44 @@ def test_generate_solutions(tuner_ctx: common.TunerContext) -> None:
...
+            pipeline_options_search_space=dispatch_constraints.PipelineOptionsSearchSpace(),
+            codegen_pipeline=iree_codegen.DispatchLoweringPassPipeline.LLVMGPUTileAndFuse,
+        )
+    )
+
+    assert len(solutions) > 0, "No solutions generated with TileAndFuse pipeline."
+    assert all(isinstance(sol, iree_codegen.CompilationInfoAttr) for sol in solutions)
+    assert all(
+        "padding =" in str(sol.lowering_config) for sol in solutions
+    ), "Not all lowering configs have padding option"
```
Max191, line 89: Also add an assert that the `promote_operands` is `[0, 1, 2]`.

Max191 (commented):
Overall looks good! Just a couple of small comments

rkayaith:
I tried this out (on the same input IR linked in the description), the config that the tuner finds as best is:
```mlir
#iree_codegen.compilation_info<
  lowering_config = #iree_gpu.lowering_config<{
    mma_kind = #iree_gpu.mma_layout<MFMA_F32_16x16x16_BF16>, padding = [48, 128, 32], promote_operands = [0, 1, 2],
    reduction = [0, 0, 2], subgroup = [3, 1, 0],
    subgroup_m_count = 1 : i64, subgroup_n_count = 8 : i64,
    workgroup = [48, 128, 0]
  }>,
  translation_info = <
    pipeline = LLVMGPUTileAndFuse workgroup_size = [512, 1, 1] subgroup_size = 64,
    {gpu_pipeline_options = #iree_gpu.pipeline_options<prefetch_shared_memory = true>, llvm_func_attrs = {"amdgpu-waves-per-eu" = "2"}}
  >
>
```
Experimentally, @jerryyin found a different config which runs ~30% faster than what the tuner found:
```mlir
#iree_codegen.compilation_info<
  lowering_config = #iree_gpu.lowering_config<{
    mma_kind = #iree_gpu.mma_layout<MFMA_F32_16x16x16_BF16>, padding = [64, 128, 32], promote_operands = [0, 1, 2],
    reduction = [0, 0, 2], subgroup = [2, 4, 0],
    workgroup = [64, 128, 0]
  }>,
  translation_info = <
    pipeline = LLVMGPUTileAndFuse workgroup_size = [256, 1, 1] subgroup_size = 64,
    {gpu_pipeline_options = #iree_gpu.pipeline_options<prefetch_shared_memory = true, no_reduce_shared_memory_bank_conflicts = false, use_igemm_convolution = false>}
  >
>
```
I looked at the generated spec files and didn't see the faster one as a candidate, any idea why that would be?

Max191:
> I looked at the generated spec files and didn't see the faster one as a candidate, any idea why that would be?

This is because the current constraints are fairly basic and they only set the padding to the next multiple of the intrinsic (48 along the outermost dimension in this case). Jerry added a bit of extra padding, which allowed for nice Po2 sizes, and was possible responsible for the performance improvement.

Eventually it would be good to add the amount of padding as a tunable parameter, but that can be a followup.

Max191:
Main difference to notice being that the subgroup tile sizes in the first config (from the tuner) are `subgroup = [3, 1, 0]`, which doesn't seem great.

rkayaith:
I tried increasing `--num-candidates` to 2000 and it ended up generating a candidate that's _almost_ the same as Jerry's:
```
   %0 = transform.param.constant #iree_codegen.compilation_info<lowering_config = #iree_gpu.lowering_config<{mma_kind = #iree_gpu.mma_layout<MFMA_F32_16x16x16_BF16>, padding = [64, 128, 32], promote_operands = [0, 1, 2], reduction = [0, 0, 2], subgroup = [2, 4, 0], subgroup_m_count = 2 : i64, subgroup_n_count = 2 : i64, workgroup = [64, 128, 0]}>, translation_info = <pipeline = LLVMGPUTileAndFuse workgroup_size = [256, 1, 1] subgroup_size = 64, {gpu_pipeline_options = #iree_gpu.pipeline_options<prefetch_shared_memory = true>, llvm_func_attrs = {"amdgpu-waves-per-eu" = "2"}}>> -> !transform.any_param
```
If I remove ` llvm_func_attrs = {"amdgpu-waves-per-eu" = "2"}` from that, it matches the performance of Jerry's config.

MaheshRavishankar:
actually we probably want to remove the `waves-per-eu` thing. Or at least allow tuner to drop it. That one is a bit of a hack.

bangtianliu:
> actually we probably want to remove the `waves-per-eu` thing. Or at least allow tuner to drop it. That one is a bit of a hack.

Sure, I will add this to my todo list: retire `waves-per-eu` thing in tuner. CC @kuhar @Max191 .

kuhar:
> > actually we probably want to remove the `waves-per-eu` thing. Or at least allow tuner to drop it. That one is a bit of a hack.
> 
> Sure, I will add this to my todo list: retire `waves-per-eu` thing in tuner. CC @kuhar @Max191 .

We don't want to retire it. We can make it optional thought.

sharktuner/sharktuner/dispatch_constraints.py
```diff
@@ -502,6 +514,25 @@ def set_cdim_tile_sizes(tile_sizes, contraction_dims, csizes):
...
+            workgroup_tile_m, workgroup_tile_n, _ = workgroup_tile_sizes
+            _, _, reduction_tile_k = reduction_tile_sizes
+            _, _, mma_intrinsic_k = mma_attr.mnk_shape
+            padding = [
+                workgroup_tile_m,
+                workgroup_tile_n,
+                reduction_tile_k * mma_intrinsic_k,
+            ]
+        else:
+            padding = None
```
Max191, line 534: nit: It is a bit more clear and concise if `padding` is defined outside the if.
```suggestion
        padding = None
        if required_padding:
            # TODO: Remove promotion of operand 2 once codegen supports handling padded outputs without promotion.
            promote_operands = [0, 1, 2]
            workgroup_tile_m, workgroup_tile_n, _ = workgroup_tile_sizes
            _, _, reduction_tile_k = reduction_tile_sizes
            _, _, mma_intrinsic_k = mma_attr.mnk_shape
            padding = [
                workgroup_tile_m,
                workgroup_tile_n,
                reduction_tile_k * mma_intrinsic_k,
            ]
```

Max191 (approved):
Looks good! Just one nit

PR #1316, opened 2025-04-26
[Shortfin] Implement preliminary support for the f8e4m3fnuz data type

vinayakdsci:
Implements support for the `f8e4m3fnuz` data type, or `float8`.

Note that while LLM accuracy is correct, the elementwise correctness tests show failures (which might be coming from out-of-range issues rather than implementation bugs).

The patch unconditionally uses float8 format as the default for the KV-cache at the moment. It might be better to support the ability to dump the dtype in `config.json` generated by sharktank so that no such assumptions have to be made.


vinayakdsci:
Closing as @stbaione has a PR that already includes the changes.

PR #1326, opened 2025-04-28
[model_tuner] Add output of best spec file.

rkayaith:
`model_tuner` now outputs the spec file for the best configuration, with the path configurable using `--output-td-spec`.

This is adapted from changes in https://github.com/Max191/shark-ai/commit/6de0af4971ed1daa016dcbe047a9f4069ef52a73

### DIFF
--- a/sharktuner/model_tuner/model_tuner.py
+++ b/sharktuner/model_tuner/model_tuner.py
@@ -70,6 +70,12 @@ def arg_parse() -> argparse.Namespace:
         default="",
         help="Path to the flags file for iree-benchmark-module for model benchmarking.",
     )
+    client_args.add_argument(
+        "--spec-output-path",
+        type=Path,
+        help="Path to write the best tuned spec after",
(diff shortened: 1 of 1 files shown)


sharktuner/model_tuner/model_tuner.py
```diff
@@ -70,6 +70,12 @@ def arg_parse() -> argparse.Namespace:
         default="",
         help="Path to the flags file for iree-benchmark-module for model benchmarking.",
     )
+    client_args.add_argument(
+        "--spec-output-path",
+        type=Path,
+        help="Path to write the best tuned spec after",
```
Max191, line 76: Can you add a note to the help description like:
```
Dumps the best tuned model spec by default, and the best tuned dispatch spec when --stop-after is set to benchmark-dispatches.
```

bangtianliu (approved):

Max191 (approved):
Looks good, thanks!

PR #1327, opened 2025-04-28
Implement ops.sum

Alex-Vasile:
Implementing this will clobber the builtin ops function within the ops namespace and anywhere we do `from sharktank.ops import *`.

### DIFF
--- a/sharktank/tests/ops/sharded_test.py
+++ b/sharktank/tests/ops/sharded_test.py
@@ -1460,6 +1460,27 @@ class SoftmaxTest(unittest.TestCase):
         ops.equal(expected_result, ops.softmax(sharded_tensor, dim=dim + 1))
 
 
+class SumTest(unittest.TestCase):
+    def setUp(self):
+        torch.random.manual_seed(12345)
+
+    @parameterized.expand(((0,), ([0, 1],), ([2, 0],)))
+    def testSumReplicated(self, sum_dim: int | list[int]):
+        tensor = torch.rand(4, 6, 5, dtype=torch.float32)
+        expected_result = ops.sum(tensor, dim=sum_dim)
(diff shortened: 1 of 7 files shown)


sogartar:
> Implementing this will clobber the builtin ops function within the ops namespace and anywhere we do `from sharktank.ops import *`.
> 
> How to handle this? Should I add another override if an Interable is passed and call python's builtin `sum`?

I think it would be better to change to explicit imports and handle `sum` differently to use `ops.sum`.
```
from sharktank import ops
from sharktank.ops import <list-all-ops-that-are-used-explicitly>
```

Another option is to change all usages to be prefixed by `ops.`, but that my touch a lot of lines.

sogartar:
The user is responsible for doing
```
from sharktank.ops import *
```

In general we should discourage this pattern and favor listing everything explicitly.

sharktank/tests/ops/sharded_test.py
```diff
@@ -1460,6 +1460,27 @@ def testSoftmaxSplit(self):
         ops.equal(expected_result, ops.softmax(sharded_tensor, dim=dim + 1))
 
 
+class SumTest(unittest.TestCase):
```
sogartar, line 1463: There is no test for the `keepdim=True` case. It is not critical but it would be nice to have one.

sogartar (approved):

sogartar:
I see on our codebase we don't use the `*` to import ops, so we should be fine.

sharktank/tests/ops/sharded_test.py
Alex-Vasile: Added

Alex-Vasile:
> I see in our codebase we don't use the * pattern to import ops, so we should be fine.

The problem is that within the ops folder this implementation blocks access to the built-in sum. I've added a version to custom_impl to bring back access to the built-in version.

PR #1330, opened 2025-04-28
Minor fixes to tests

Alex-Vasile:
- Certain passing tests didn't have required asserts in them.
- `testSoftmaxSplit` was not coded correctly.

### DIFF
--- a/sharktank/tests/ops/sharded_test.py
+++ b/sharktank/tests/ops/sharded_test.py
@@ -1546,7 +1546,7 @@ class ViewTest(unittest.TestCase):
 
         expected_result = ops.view(tensor, new_shape)
         actual_result = tensor_rep.view(new_shape)
-        ops.equal(expected_result, actual_result)
+        assert ops.equal(expected_result, actual_result)
 
     @parameterized.expand((([8, 5, 3, 2],), ([4, 2, 5, 3, 2],)))
     def testViewReplicatedExpand(self, new_shape: list[int]):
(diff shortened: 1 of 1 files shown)


sharktank/tests/ops/sharded_test.py
```diff
@@ -1445,19 +1445,24 @@ def setUp(self):
...
+        tensor = torch.rand(2, 2, 2, dtype=torch.float32)
         dim = 1
-        expected_result = ops.softmax(tensor, dim=dim)
         sharded_tensor = ops.reshard_split(tensor, dim=dim, count=2)
-        ops.equal(expected_result, ops.softmax(sharded_tensor, dim=dim - 1))
-        ops.equal(expected_result, ops.softmax(sharded_tensor, dim=dim + 1))
+
+        expected_result = ops.softmax(tensor, dim=dim - 1)
+        actual_result = ops.softmax(sharded_tensor, dim=dim - 1)
+        assert ops.equal(expected_result, actual_result)
```
sogartar, line 1461: It is better to do an approximate check with `torch.testing.assert_close`.

There is one more such assert in this function.

sogartar (approved):

sharktank/tests/ops/sharded_test.py
Alex-Vasile: Addressing in https://github.com/nod-ai/shark-ai/pull/1333

PR #1331, opened 2025-04-28
Cleaner kernel invocations

stbaione:
Opening a draft PR to get feedback on this implementation of using the `argmax` kernel, instead of `xtensor`/`numpy`


stbaione:
Stale

PR #1333, opened 2025-04-28
Ops.zero_like and ops.scatter_

Alex-Vasile:
`scatter_split_split` has a naive implementation and a more efficient version may be possible.

### DIFF
--- a/sharktank/sharktank/ops/sharded_impls.py
+++ b/sharktank/sharktank/ops/sharded_impls.py
@@ -1940,5 +1940,29 @@ def view_as_real_rep(tensor: ReplicatedTensor) -> ReplicatedTensor:
     return ReplicatedTensor(ts=shards)
 
 
+@zeros_like.override(AllOfType(ReplicatedTensor, SplitPrimitiveTensor))
+def zeros_like_replicated(
+    tensor: ReplicatedTensor | SplitPrimitiveTensor,
+    *,
+    dtype: torch.dtype | None,
+    layout: torch.layout | None,
+    device: torch.device | None,
+    requires_grad: bool,
+    memory_format: torch.memory_format,
+) -> ReplicatedTensor | SplitPrimitiveTensor:
+    shards = [
+        zeros_like(
+            unbox_tensor(shard),
+            dtype=dtype,
+            layout=layout,
+            device=device,
+            requires_grad=requires_grad,
+            memory_format=memory_format,
+        )
+        for shard in tensor.shards
+    ]
+    return tensor.clone(ts=shards)
+
+
 # Note: Must be last thing in file
 sharded_unwrap_override()
--- a/sharktank/sharktank/ops/sharded_impls.py
+++ b/sharktank/sharktank/ops/sharded_impls.py
@@ -2031,7 +2031,7 @@ def zeros_like_replicated(
 ) -> ReplicatedTensor | SplitPrimitiveTensor:
     shards = [
         zeros_like(
-            unbox_tensor(shard),
+            shard,
             dtype=dtype,
             layout=layout,
             device=device,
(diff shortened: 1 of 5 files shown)


sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -1407,6 +1407,85 @@ def reshard_like_unreduced_to_replicated(
...
+        for inout_shard, tmp_shard in zip(inout.shards, tmp_tensor.shards):
+            inout_shard.as_torch().copy_(tmp_shard.as_torch())
+        return inout
+
+    shard_dim = inout.shard_dim
+    if index.shape[shard_dim] == inout.shape[shard_dim]:
+        assert index.shard_dim == inout.shard_dim
+        index_shards = index.shards
+        last_shard_idx = inout.shard_count - 1
+    else:
```
sogartar, line 1451: I don't understand that case.
After all-gathering the `index`, shouldn't we reshard it like the `inout`?
Then we can scatter locally for each pair of corresponding `inout` and index shards.

sharktank/tests/ops/sharded_test.py
```diff
@@ -1366,6 +1366,44 @@ def testGroupNormSplitChannelSharding(self):
...
+        ops.scatter_(sharded_tensor, 0, index_sharded, value)
+        ops.scatter_(tensor, 0, index, value)
+        assert ops.equal(tensor, sharded_tensor)
+
+    @parameterized.expand((([3, 1],), ([3, 2],), ([9, 1],), ([9, 6],)))
+    def testScatterSplitSplitNonShardDim(self, index_shape: list[int]):
+        scatter_dim = 1
+        value = 1
+        tensor = torch.zeros(9, 6, dtype=torch.float32)
+        index = torch.randint(0, tensor.shape[scatter_dim], index_shape)
```
sogartar, line 1398: It seems we can get a conflicting indexing if we have nonunique destination along the scatter axis. Then who wins and gets written last is nondeterministic.
The other 2 tests also do a random index.

You could use [numpy.random.Generator.choice](https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.choice.html) with `replace=False` to do a non-repeating random sample.

sharktank/sharktank/ops/signatures.py
```diff
@@ -1363,3 +1393,44 @@ def _view_as_real_trampoline(d: SignatureDispatcher, tensor: AnyTensor) -> AnyTe
...
+
+@overridable
+def zeros_like(
+    tensor: AnyTensor,
+    *,
+    dtype: torch.dtype | None = None,
+    layout: torch.layout | None = None,
+    device: torch.device | None = None,
+    requires_grad: bool = False,
+    memory_format: torch.memory_format = torch.preserve_format,
```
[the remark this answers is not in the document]
Alex-Vasile: What are you suggesting I do differently here?

sharktank/sharktank/ops/sharded_impls.py
Alex-Vasile: By default we are sharding both `inout` and `index` to be balanced. However, if `index` has fewer entries along the shard_dim that `inout` then entry `i` along `shard_dim` that's in shard `j` of `inout` may end up in shard `k` of index. We need to reshard so that slices along `shard_dim` within each shard line up.

sharktank/tests/ops/sharded_test.py
```diff
@@ -1366,6 +1366,55 @@ def testGroupNormSplitChannelSharding(self):
         assert expected_bias.is_deep_equal(sharded_theta("bias"), compare_name=False)
 
 
+class Scatter_Test(unittest.TestCase):
+    def setUp(self):
+        import numpy as np
+
+        np.random.seed(12345)
+        self.rng = np.random.default_rng()
```
sogartar, line 1374: You would need to seed the RNG to avoid a non-deterministic test?
I think we need to refactor at some point a introduce a base test case class that does all this by RNG seeding default.

sogartar (approved):

PR #1335, opened 2025-04-29
Enable support for `fp8` kvcache in shortfin LLM server

stbaione:
Co-Authored by: @vinayakdsci, who did most of the heavy lifting by adding the `float8_e4m3fnuz` datatype to `sfnp.array`.

This enables support for `float8_e4m3fnuz` KVCache data type in Shortfin LLM Server.

We add a `--to-fp16` option in the export script to cast the logits to fp16. This is both for simplicity and to use optimized optimizations as fp8 ops across popular packages are typically not supported or very slow.

We may be able to remove this when we enable the use of `post-processing kernels` in the Shortfin LLM server. Regardless, this ensures that we are able to easily support fp8 models with our current structure.


vinayakdsci (approved):
LGTM, thanks! Looks good to merge when the CI is green.

rsuderman (approved):

PR #1341, opened 2025-04-29
Sharktank initial post op support

stbaione:
Stacked on #1335

This allows us to use Sharktank's LLM export script to include post-processing-ops, in this case specifically argmax.

You can export `argmax` to the MLIR by appending the following arg:

```bash
...
--use-post-processing-ops
```

Minimal change on the shortfin side, just adding the names of the `post_processing_ops` that we export to `ModelParams`.

Follow Ups, in order:

- Split argmax
- Add support for argmax kernel in Shortfin
- top_k/Split top_k
- Add support for top_k kernel in Shortfin


sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -86,6 +86,7 @@ def main():
         activation_dtype=args.activation_dtype,
         attention_dtype=args.attention_dtype,
         kv_cache_dtype=args.kv_cache_dtype,
+        post_processing_ops=[] if args.use_post_processing_ops else None,
```
rsuderman, line 89: This should just be an enum like `log_softmax` and `softmax` for post score selection. This will be relatively finite and general "list of ops to run" is unlikely to be good in the longterm.

sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -407,6 +417,7 @@ def _(
                 start_positions=start_positions,
                 seq_block_ids=seq_block_ids,
                 cache_state=cache_state,
+                to_fp16=args.to_fp16,
```
rsuderman, line 420: Don't worry about making this configurable. We honestly should never be returning `fp8` values. Just check before return that if the value is fp8 we should cast to fp16.

rsuderman (changes requested):

sharktank/sharktank/examples/export_paged_llm_v1.py
stbaione: Made this change in the [original PR](https://github.com/nod-ai/shark-ai/pull/1335), will merge it into here as well

sharktank/sharktank/examples/export_paged_llm_v1.py
stbaione: Gotcha, I'm thinking just adding a boolean flag in the config that says that we exported with post_processing ops. 

I don't think I want to just make a string field and restrict it to a singular scoring function, like we do with logits_normalization.

We already are able to pretty easily switch between pure `greedy` token selection and `top_k/top_p` selection through request args from the client, and don't wanna break that.

I could make an `enum` on the `Shortfin` side to track the ops that we have available to us, and then I can adjust the `Sampler` class, so that we use the GPU kernel, if that function is present.

What do you think?

rsuderman: Its true that a simple enum likely isn't enough (e.g. for top-k we need a k value). Overall we just want to have some simple way of selecting / configuring the export via flags.  This could use be `--top-k=4` or `--top-k=1` for argmax. Just a boolean probably is not quite what we are doing.

stbaione: Updated to use `top_k` int. I think there's still a future opportunity to export both argmax and a generic `top_k` for use in Shortfin, but this should allow a simple and clear path of iterating to that point.

rsuderman (approved):

PR #1343, opened 2025-04-29
Refactor InferenceTensor.to to be close to Tensor.to

Alex-Vasile:
- Supports all 3 overloaded versions from tensor.to
- Added tests
- Fix bugs in `_clone_with_globals`


sharktank/tests/ops/sharded_test.py
```diff
@@ -1499,6 +1499,44 @@ def testSumBuiltinFunction(
         assert expected_result == actual_result
 
 
+class ToTest(unittest.TestCase):
```
Alex-Vasile, line 1502: These pass locally but fail because of the "cuda:0". How to properly tests?

sharktank/sharktank/types/tensors.py
```diff
@@ -274,18 +275,64 @@ def transform_globals(
...
+        device_overload = ("device" in kwargs) or isinstance(arg0, (str, torch.device))
+        other_overload = ("other" in kwargs) or isinstance(arg0, AnyTensor)
+        memory_overload = ("memory_format" in kwargs) or isinstance(arg0, torch.dtype)
+
+        if device_overload:
+            # Do we always want to clone with globals?
+            # This makes our type inconsistent with torch tensors.
+            # If we use this to transform a theta we want to change the theta.
+            # If we want to use this in a computation we don't want to change the theta.
+            return self.transform_globals(
```
sogartar, line 323: The problem here is that `transform_globals` would change to what the external external tensor maps (`ExternalTensorTrait`). Which will be wrong if you use this mid computation in the model and not when doing transformations when preparing a Theta. It will probably manifest rarely as you would need to call this on a Theta tensor that has the external tensor trait.

`transform_globals` would also return a tensor with the same name. The name duplication is maybe fine. In Thetas it is not.

I think we should rename the function when we want to transform thetas and make `to` not be loaded with this additional semantics so it match Torch. This is best to be in another PR.

sharktank/tests/ops/sharded_test.py
sogartar: I don't see what you could do. I don't think PyTorch support multiple logical CPU devices and we don't want to go into that rabbit hole. Maybe put a conditional skip where you check for the availability of the device.
I recall that `parameterized` supported per fixture skipping.

sogartar (approved):

PR #1348, opened 2025-04-30
Ops.reshape for ReplicatedTensor

Alex-Vasile:
- reshape implementation
- bug fix for InferenceTensor.reshape


sogartar (approved):
Maybe the correct is to check if the first arg is a sequence and if it is to pass it as the shape.
If you do
```
torch.rand([2, 6]).reshape([2, 3, 2], 5).shape
```
Torch would not complain for the trailing `5` and return
```
torch.Size([2, 3, 2])
```

PR #1350, opened 2025-04-30
Use newly implemented ops

Alex-Vasile:


archana-ramalingam (approved):

archana-ramalingam:
Looks like a lot more tests are using `zeros_like` op.

PR #1356, opened 2025-04-30
Toy model fixes

Alex-Vasile:
- Fixes for toy deepseek
- Read more data from gguf_props, using default values if unavailable.

### DIFF
--- a/sharktank/sharktank/layers/configs/llm_configs.py
+++ b/sharktank/sharktank/layers/configs/llm_configs.py
@@ -85,7 +85,14 @@ class LlamaHParams:
         default_expert_used_count = 0
         default_rope_freq_base = 500000.0
         default_rope_dimension_count = 128
+        default_n_expert_groups = 8
         defaut_n_dense_layers = 0
+        default_n_limited_groups = 4
+
+        default_qk_rope_head_dim = 64
+        default_qk_nope_head_dim = 128
+        default_v_head_dim = 128
+
         attention_head_count = _int_prop(p, f"{name_prefix}.attention.head_count")
         rope_dimension_count = _optional_int_prop(
             p, f"{name_prefix}.rope.dimension_count", default_rope_dimension_count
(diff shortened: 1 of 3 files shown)


sharktank/sharktank/layers/testing.py
```diff
@@ -124,7 +124,7 @@ def make_ffn_block_theta(
         {
             "ffn_norm.weight": DefaultPrimitiveTensor(
                 name=f"blk.{block_idx}.ffn_norm.weight",
-                data=make_rand_torch((head_count * head_dim), dtype=dtype),
+                data=make_rand_torch((embedding_length), dtype=dtype),
```
archana-ramalingam, line 127: Toy FFN changes might affect llama too.

sharktank/sharktank/models/deepseek/toy_deepseek.py
```diff
@@ -25,7 +25,7 @@ def generate(seed):
     block_seq_stride = 32
     max_blocks = 8
 
-    rope_dimension_count = 16
+    rope_dimension_count = 64
```
archana-ramalingam, line 28: We need to create a distinction in all toy models as to what can be changed and what needs to remain constant.

sharktank/sharktank/layers/configs/llm_configs.py
```diff
@@ -94,10 +101,25 @@ def from_gguf_props(p: dict[str, Any]):
         attention_softcap = 30.0 if name_prefix == "grok" else None
 
         if name_prefix == "deepseek2":
+            qk_rope_head_dim = _optional_int_prop(
+                p, f"{name_prefix}.attention.qk_rope_head_dim", default_qk_rope_head_dim
+            )
+            qk_nope_head_dim = _optional_int_prop(
+                p, f"{name_prefix}.attention.qk_nope_head_dim", default_qk_nope_head_dim
+            )
+            v_head_dim = _optional_int_prop(
```
archana-ramalingam, line 110: Can be changed after, but by default these values are set to None at the class beginning and llm logic flows better with None. We can remove default values and switch to `_int_prop` instead of `_optional_int_prop`

archana-ramalingam (approved):

sharktank/sharktank/models/deepseek/toy_deepseek.py
Alex-Vasile: Agreed!

PR #1359, opened 2025-05-01
[sharktank] Dense FFN expert block in MoE

sogartar:
The current PreGatherFFNMOE would index into the expert weights with the selected top experts. Resulting in a large tensor that IREE does not have a way to not materialize it and this will blow up the memory.
```
        weights = weights[experts, :, :]
```
The shape of experts is `(batch_size * sequence_length, num_top_experts)`. For large `batch_size * sequence_length` the resulting matrix would become quite big. For Deepseek V3 for example we would get `(batch_size * sequence_length, num_top_experts, 2048, 7168)`.

Per the suggestion of Rob Suderman and when looking how Hugging Face's Llama 4 implements MoE, here is introduced a Dense FFN expert block. It routes all tokens to all experts no matter the number of top experts. I don't know how this could be faster than a sparse block as it does a bunch of useless computations, but it reduces the number of high-level ops as it does not loop through the experts. We could at least try it. It is supposed to perform a bit better with expert-parallel sharding.

### DIFF
--- a/sharktank/sharktank/layers/ffn_block.py
+++ b/sharktank/sharktank/layers/ffn_block.py
@@ -27,7 +27,7 @@ class FFN(ThetaLayer):
         theta: Theta,
         rms_epsilon: float | None = None,
         is_gated: bool = True,
-        activation_fn: Callable[[AnyTensor], AnyTensor] = F.silu,
+        activation_fn: Callable[[torch.Tensor], torch.Tensor] = F.silu,
         activation_dtype: Optional[torch.dtype] = torch.float16,
         fake_quant: bool = False,
         add_residual: bool = True,
--- a/sharktank/sharktank/layers/ffn_moe_block.py
+++ b/sharktank/sharktank/layers/ffn_moe_block.py
@@ -14,6 +14,7 @@ from .linear import LinearLayer
 from . import FFN
 from sharktank.types import AnyTensor, DefaultPrimitiveTensor, Theta
 from sharktank.ops import einsum_2args, elementwise
+from sharktank import ops
 
 __all__ = [
     "DenseFFNMOE",
@@ -91,7 +92,7 @@ class DenseFFNMOE(ThetaLayer):
         theta: Theta,
         rms_epsilon: float | None = None,
         is_gated: bool = True,
-        activation_fn: Callable[[AnyTensor], AnyTensor] = F.silu,
+        activation_fn: Callable[[torch.Tensor], torch.Tensor] = F.silu,
         activation_dtype: Optional[torch.dtype] = None,
         fake_quant: bool = False,
     ):
@@ -154,7 +155,7 @@ class DenseFFNMOE(ThetaLayer):
         router_indices = router_indices.reshape(-1, 1).expand(-1, input_feature_dim)
 
         # (self.num_experts * num_tokens, input_feature_dim)
-        routed_in = torch.gather(
+        routed_in = ops.gather(
             input=h,
             dim=0,
             index=router_indices,
@@ -169,7 +170,7 @@ class DenseFFNMOE(ThetaLayer):
         routed_out = routed_out * router_scores.reshape(-1, 1)
 
         # (num_tokens, input_feature_dim)
-        return torch.zeros_like(h).scatter_add(
+        return ops.zeros_like(h).scatter_add(
             dim=0, index=router_indices, src=routed_out
         )
 
--- a/sharktank/sharktank/layers/mixture_of_experts_block.py
+++ b/sharktank/sharktank/layers/mixture_of_experts_block.py
@@ -51,7 +51,9 @@ class MoeBlock(ThetaLayer):
         self.ffn_norm = torch.nn.Identity()
         self.layer_output_norm = torch.nn.Identity()
         self.shared_experts = None
-        self.route_scale = route_scale if route_scale > 1 else None
+        self.route_scale = None
+        if route_scale is not None and route_scale != 1:
+            self.route_scale = route_scale
 
         # Add FFN norm
         if "ffn_norm" in theta:
--- a/sharktank/tests/layers/mixture_of_experts_block_test.py
+++ b/sharktank/tests/layers/mixture_of_experts_block_test.py
@@ -86,8 +86,8 @@ class MoeBlockTest(unittest.TestCase):
             ),
             param(
                 dtype=torch.float32,
-                feature_dim=1,
-                expert_hidden_dim=1,
+                feature_dim=2,
+                expert_hidden_dim=3,
                 num_experts=3,
                 expert_used_count=2,
                 num_shared_experts=2,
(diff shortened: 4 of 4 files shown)


sharktank/sharktank/layers/ffn_moe_block.py
```diff
@@ -74,7 +76,105 @@ def forward(
...
+    What approach is better would depend on the number experts and the number of top
+    experts.
+    """
+
+    def __init__(
+        self,
+        theta: Theta,
+        rms_epsilon: float | None = None,
+        is_gated: bool = True,
+        activation_fn: Callable[[AnyTensor], AnyTensor] = F.silu,
```
Alex-Vasile, line 94: This won't work with AnyTensor, it will only work for torch.Tensor.

sogartar: That is strange, because I copy pasted it from
The [FFN](https://github.com/nod-ai/shark-ai/blob/8f299d354c2a3b64c844671df26d1398fcb87988/sharktank/sharktank/layers/ffn_block.py#L30) block.
And it is used with a default value in the [LLM](https://github.com/nod-ai/shark-ai/blob/8f299d354c2a3b64c844671df26d1398fcb87988/sharktank/sharktank/models/llm/llm.py#L374) models.
To be honest I don't know how this works with the sharded tensors at all.

sogartar: This actually propagated to `ops.elementwise`.

sharktank/sharktank/layers/ffn_moe_block.py
```diff
@@ -74,7 +76,105 @@ def forward(
...
+        routed_in = routed_in.view(self.num_experts, num_tokens, input_feature_dim)
+
+        # (self.num_experts * num_tokens, input_feature_dim)
+        routed_out = self.ffn(routed_in).view(
+            self.num_experts * num_tokens, input_feature_dim
+        )
+        routed_out = routed_out * router_scores.reshape(-1, 1)
+
+        # (num_tokens, input_feature_dim)
+        return torch.zeros_like(h).scatter_add(
```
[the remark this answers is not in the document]
sogartar: I changed it.

sharktank/sharktank/layers/ffn_moe_block.py
sogartar: I changed the type hint on both places.

Alex-Vasile (approved):

sharktank/sharktank/layers/ffn_block.py
```diff
@@ -70,4 +76,6 @@ def forward(
             ffn_act = ops.elementwise(self.activation_fn, ffn_up)
             ffn_down = self.ffn_down(ffn_act)
 
-        return h + ffn_down
+        if self.add_residual:
```
archana-ramalingam, line 79: ffn_norm and add_residual needs to be moved out of ffn/ffn_moe blocks. For llms with shared experts, we end up doing add_residual and ffn_norm twice. This is fine to land on main, can fix this on deepseek branch before landing.

sharktank/tests/layers/mixture_of_experts_block_test.py
```diff
@@ -0,0 +1,187 @@
...
+from parameterized import parameterized, param
+from typing import Callable
+
+import torch
+from iree.turbine.aot import *
+from sharktank.models.llama.testing import make_moe_block_theta, make_rand_torch
+from sharktank.layers.mixture_of_experts_block import MoeBlock
+
+
+class MoeBlockTest(unittest.TestCase):
```
archana-ramalingam, line 17: We already had a moe export test [here](https://github.com/nod-ai/shark-ai/blob/main/sharktank/tests/models/llama/moe_block_test.py). If this test covers all cases, should we remove that?

archana-ramalingam: Noticed it's been removed already.

sharktank/sharktank/layers/ffn_block.py
sogartar: I will rebase the deepseek branch.

sharktank/sharktank/layers/ffn_moe_block.py
```diff
@@ -74,7 +77,105 @@ def forward(
...
+        # (self.num_experts, num_tokens)
+        router_scores = (
+            torch.full([num_tokens, self.num_experts], 0, dtype=h.dtype)
+            .scatter_(1, top_experts_index, expert_gate)
+            .transpose(0, 1)
+        )
+
+        # (self.num_experts, num_tokens)
+        router_indices = (
+            torch.arange(num_tokens, device=h.device)
```
rsuderman, line 150: We should be careful about using `torch` operators explicitly. It causes problems when we try to do sharding work. Long term we should have `range` as an operation in sharktank's ops.

sharktank/sharktank/layers/ffn_moe_block.py
```diff
@@ -74,7 +77,105 @@ def forward(
...
+
+        Returns:
+            Tensor of shape (batch_size * sequence_length, expert_feature_dim)
+        """
+        num_tokens, input_feature_dim = h.shape
+
+        # (self.num_experts, num_tokens)
+        router_scores = (
+            torch.full([num_tokens, self.num_experts], 0, dtype=h.dtype)
+            .scatter_(1, top_experts_index, expert_gate)
```
rsuderman, line 144: Scattering into a full feels weird. There should be a better way to do this computationally.

rsuderman: Ditto comment about `torch` operators below. We should try to stay to consistent operators compatible with sharding.

sharktank/sharktank/layers/ffn_moe_block.py
```diff
@@ -74,7 +77,105 @@ def forward(
...
+        routed_in = routed_in.view(self.num_experts, num_tokens, input_feature_dim)
+
+        # (self.num_experts * num_tokens, input_feature_dim)
+        routed_out = self.ffn(routed_in).view(
+            self.num_experts * num_tokens, input_feature_dim
+        )
+        routed_out = routed_out * router_scores.reshape(-1, 1)
+
+        # (num_tokens, input_feature_dim)
+        return ops.zeros_like(h).scatter_add(
```
rsuderman, line 173: Against about specialized operations. `scatter_add` feels weird even because if the input is zero this would just be scatter anyway.

sharktank/sharktank/layers/mixture_of_experts_block.py
```diff
@@ -33,6 +33,7 @@ def __init__(
         rms_epsilon: float,
         moe_activation=torch.nn.functional.silu,
         *,
+        experts_ffn_moe_block: PreGatherFFNMOE | DenseFFNMOE | str = "DenseFFNMOE",
```
rsuderman, line 36: We should either stick to dependency injection OR enum configuration. Supporting both is a recipe for difficult maintenance.

sharktank/sharktank/layers/ffn_moe_block.py
sogartar: The mapping is many-to-one. Meaning that multiple addition may occur for the same destination location.

PR #1360, opened 2025-05-01
Sharktank LLM split argmax op

stbaione:
Putting up a draft PR to get feedback on my attempt at manual `split_argmax`.

I was seeing worse performance metrics when benchmarking it:

- Argmax: `~4ms`
- Split Argmax: `~4.5ms`

So, I'm assuming that the parallelization isn't happening the way that we want it to. Wanted to get some feedback on the implementation, there's a pretty high likelihood that I missed a concept that would allow us to actually achieve the parallelization.


vinayakdsci (commented):
cc @rsuderman 

From the conversation I had with @MaheshRavishankar when the kernel performance metric initially came up, this is equivalent to how the compiler would handle it eventually by split reducing the kernel.

```python
def argmax(result_logits: sfnp.device_array):
    # Reshape the result logits first into (current_ndim + 1) dimensions.
    # The reshaped innermost dimension will need choosing based on the vocab size
    # and the way you would want to distribute.
    result_logits_reshaped = result_logits.reshape(*result_logits.shape[:-1], result_logits.shape[-1] // 1024, 1024) 
    argmax_1 = torch.argmax(result_logits_reshaped, dim=-1)
    # The innermost dim should now be reduced. Do another reshape (this can possibly be removed if that dimension does not show up).
    reshape_2 = argmax_1.reshape(*result_logits.shape[:-1], result_logits.shape[-1] // 1024)
    # This reduces the innermost dimension again.
    final_argmax = torch.argmax(reshape_2, dim=-1)
    return final_argmax
```
This implementation ran, but had numerics issues. The implementation does follow the idea of split reduction correctly, and this would be the correct way if the results are accurate.

sharktank/sharktank/ops/signatures.py
```diff
@@ -749,6 +758,32 @@ def _matmul_trampoline(
         d.fail(tensors)
 
 
+@overridable
+def max(
```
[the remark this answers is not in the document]
stbaione: Removed those ops

stbaione:
> cc @rsuderman
> 
> From the conversation I had with @MaheshRavishankar when the kernel performance metric initially came up, this is equivalent to how the compiler would handle it eventually by split reducing the kernel.
> 
> ```python
> def argmax(result_logits: sfnp.device_array):
>     # Reshape the result logits first into (current_ndim + 1) dimensions.
>     # The reshaped innermost dimension will need choosing based on the vocab size
>     # and the way you would want to distribute.
>     result_logits_reshaped = result_logits.reshape(*result_logits.shape[:-1], result_logits.shape[-1] // 1024, 1024) 
>     argmax_1 = torch.argmax(result_logits_reshaped, dim=-1)
>     # The innermost dim should now be reduced. Do another reshape (this can possibly be removed if that dimension does not show up).
>     reshape_2 = argmax_1.reshape(*result_logits.shape[:-1], result_logits.shape[-1] // 1024)
>     # This reduces the innermost dimension again.
>     final_argmax = torch.argmax(reshape_2, dim=-1)
>     return final_argmax
> ```
> 
> This implementation ran, but had numerics issues. The implementation does follow the idea of split reduction correctly, and this would be the correct way if the results are accurate.

Pushed a new implementation that follows this idea and passes the tests numerically

sharktank/tests/ops/ops_test.py
```diff
@@ -42,6 +42,39 @@ def testArgmaxDim0(self):
...
+            a[1][2] = 99
+            a[0][1] = 1
+            a[0][3] = 1
+            result = ops.argmax(a, dim=0, keepdim=True)
+            expected = torch.tensor([[1, 0, 1, 0]], dtype=torch.int64)
+            assert result.shape == (1, 4)
+            assert torch.equal(result, expected)
+
+
+class SplitArgmaxTest(unittest.TestCase):
```
Alex-Vasile, line 58: I think these tests should go under `ArgmaxTest` since this is a still argmax. See for example `MatmulTest` which has the custom and default versions in the same class.

sharktank/tests/ops/ops_test.py
```diff
@@ -42,6 +42,39 @@ def testArgmaxDim0(self):
...
+    def testSplitArgmax(self):
+        for dtype in [torch.float16, torch.float32]:
+            a = torch.zeros(1, 1, 256, dtype=dtype)
+            a[0][0][42] = 42
+            assert ops.custom_impls.split_argmax(a, -1, 16) == 42
```
Alex-Vasile, line 63: Use a @parameterized instead of a for-loop inside the test. Makes it easier to debug, and lets them run in parallel.

Alex-Vasile: Same comment for all of the other tests with the for loop.

Alex-Vasile (changes requested):

sharktank/sharktank/ops/custom_impls.py
```diff
@@ -48,6 +48,33 @@
 #     return mmtfp(lhs, rhs)
 
 
+# Argmax
+
+
+@argmax.override(Tensor, QuantizedTensor)
```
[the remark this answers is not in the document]
stbaione: Fix override to just say Tensor, was under the impression that this just denoted the types of Tensors supported for the op, as opposed to a tuple of expected arg types

sharktank/sharktank/ops/signatures.py
```diff
@@ -125,16 +125,23 @@ def _all_reduce_trampoline(d: SignatureDispatcher, tensor: AnyTensor):
 
 
 @overridable
-def argmax(tensor: AnyTensor, axis: int) -> AnyTensor:
+def argmax(
+    tensor: AnyTensor, dim: Optional[int] = None, keepdim: bool = False
+) -> AnyTensor:
```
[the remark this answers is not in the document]
stbaione: Fixed

sharktank/tests/ops/ops_test.py
stbaione: Combined tests into one class, but `pytest.mark.parametrize` decorator is incompatible with `unittest.TestCase`, which is what the rest of the tests are written with:

[source](https://stackoverflow.com/questions/18182251/does-pytest-parametrized-test-work-with-unittest-class-based-tests)

sharktank/sharktank/ops/signatures.py
```diff
@@ -125,16 +125,23 @@ def _all_reduce_trampoline(d: SignatureDispatcher, tensor: AnyTensor):
...
     ...
 
 
 @argmax.trampoline
-def _argmax_trampoline(d: SignatureDispatcher, tensor: AnyTensor, axis: int):
+def _argmax_trampoline(
+    d: SignatureDispatcher,
+    tensor: AnyTensor,
+    dim: Optional[int] = None,
+    keepdim: bool = False,
```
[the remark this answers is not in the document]
stbaione: Fixed

sharktank/tests/ops/ops_test.py
stbaione: Combined into a single test class

sharktank/sharktank/ops/default_impls.py
```diff
@@ -27,6 +27,15 @@
 import iree.turbine.ops.iree
 
 
+@argmax.override(Tensor)
+def argmax_default(
+    x: Tensor,
+    dim: Optional[int] = None,
+    keepdim: bool = False,
+) -> None:
```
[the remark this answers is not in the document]
stbaione: Moving switch from `signatures` to `default_impl`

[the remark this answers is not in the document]
stbaione: Moved over to `default_impl`

sharktank/sharktank/ops/custom_impls.py
```diff
@@ -48,6 +48,33 @@
...
+def split_argmax(input_tensor, dim, chunk_size: int = 128):
```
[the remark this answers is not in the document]
stbaione: Added `keepdim`

sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -435,11 +435,17 @@ def _(
             return logits
 
     def generate_argmax():
+        # TODO: Remove this when the corresponding `dtype` conversion is
+        # removed in `PagedLlmModelV1.prefill/decode`
```
Alex-Vasile, line 439: Is someone actively working on that? If not, it may be worth creating an issue to track this.

sharktank/sharktank/ops/default_impls.py
```diff
@@ -27,6 +27,47 @@
 import iree.turbine.ops.iree
 
 
+@argmax.override(Tensor)
+def argmax_default(
+    x: Tensor,
+    dim: Optional[int] = None,
+    chunk_size: Optional[int] = None,
+    keepdim: bool = False,
```
Alex-Vasile, line 35: These two should be flipped. `chunk_size` is not an argument for `torch.argmax`, and may lead to silent bugs if people call `ops.argmax` expecting positional arguments to be in the same order as `torch.argmax` if torch will accept a bool instead of an int. Same idea for other signatures for this function.

sharktank/sharktank/ops/default_impls.py
```diff
@@ -27,6 +27,47 @@
...
+    argmax_1 = argmax(tensor_split, dim + 1)
+    argmax_expanded = unsqueeze(argmax_1, dim + 1)
+
+    max_vals = gather(tensor_split, dim + 1, argmax_expanded)
+    max_vals = squeeze(max_vals, dim + 1)
+
+    argmax_2 = argmax(max_vals, dim)
+    argmax_2_expanded = unsqueeze(argmax_2, 0)
+
+    final_index_in_chunk = gather(argmax_1, dim, argmax_2_expanded)
```
Alex-Vasile, line 60: Shouldn't this use argmax_1_expanded?

Alex-Vasile (changes requested):

sharktank/sharktank/examples/export_paged_llm_v1.py
stbaione: It's a potential result of us moving the Shortfin server over to GPU kernels. Right now, we cast the logits in the model to fp16 for those formats, so that we have better support and performance in things like `xtensor, numpy, etc.`.

If we move the LLM server over the GPU kernels completely, then we may be able to remove that, should we see a potential perf benefit from it.

It may remain though, because we might want to be able to swap between CPU and GPU invocations (i.e. if a user doesn't have a GPU).

So, I'm not sure if it's an issue per-say, but something I wanted to make sure was documented in case we remove that in the future

sharktank/sharktank/ops/default_impls.py
rsuderman: Nope. We are just gathering the values, having the additional dimension doesn't matter.

sharktank/sharktank/ops/signatures.py
```diff
@@ -125,16 +125,27 @@ def _all_reduce_trampoline(d: SignatureDispatcher, tensor: AnyTensor):
 
 
 @overridable
-def argmax(tensor: AnyTensor, axis: int) -> AnyTensor:
+def argmax(
+    tensor: AnyTensor,
+    dim: Optional[int] = None,
+    chunk_size: Optional[int] = None,
```
rsuderman, line 131: You will need to update the `sharded_impls.py` version as well so that it has `chunk_size`. It should just involve proxy-ing the value to the sub calls.

rsuderman (changes requested):

rsuderman (approved):
Wait for presubmits but this should be good to land.

Alex-Vasile (approved):

PR #1361, opened 2025-05-01
Fail out early if mismatch in tensor_parallelism_size

Alex-Vasile:
Currently if you have a mismatch between `--tensor-parallelism-size` and the sharding of the provided IRPA file you will get a hard to understand error message.
E.g. if you provide an unsharded IRPA, but specified to `export_paged_llm_v1` that the data is sharded with TP>1 you will get the following hard to understand error.
```
  File "/home/alvasile/repos/shark-ai/sharktank/sharktank/ops/signatures.py", line 292, in _embedding_lookup_trampoline
    d.fail(tensors)
  File "/home/alvasile/repos/shark-ai/sharktank/sharktank/ops/_registry.py", line 270, in fail
    raise NotImplementedError(
NotImplementedError: Overridable operator sharktank.ops.signatures.embedding_lookup does not have an implementation for argument types: [<class 'sharktank.types.tensors.ReplicatedTensor'>, <class 'sharktank.types.tensors.DefaultPrimitiveTensor'>]
```

### DIFF
--- a/sharktank/sharktank/examples/export_paged_llm_v1.py
+++ b/sharktank/sharktank/examples/export_paged_llm_v1.py
@@ -59,11 +59,18 @@ def main():
     dataset_type = "irpa" if "irpa" in dataset_type else "gguf"
     dataset = cli.get_input_dataset(args)
     hp = configs.LlamaHParams.from_gguf_props(dataset.properties)
-    tensor_parallelism_size = (
-        dataset.properties["tensor_parallelism_size"]
-        if "tensor_parallelism_size" in dataset.properties
-        else args.tensor_parallelism_size
-    )
+    if "tensor_parallelism_size" in dataset.properties:
+        dataset_tensor_parallelism_size = dataset.properties["tensor_parallelism_size"]
+        if dataset_tensor_parallelism_size != args.tensor_parallelism_size:
+            raise ValueError(
+                f"Tensor parallelism size mismatch: dataset={dataset_tensor_parallelism_size} while arg={args.tensor_parallelism_size}. Wrong value for --tensor-parallelism-size."
+            )
+    else:
+        if args.tensor_parallelism_size != 1:
+            raise ValueError(
+                f"Unsharded dataset file provided, but specified --tensor-parallelism-size={args.tensor_parallelism_size}. Likely wrong dataset provided."
+            )
+        tensor_parallelism_size = 1
 
     if args.pipeline_parallelism_size > 1:
         block_to_device_lookup = pipeline_parallelize_theta(


sogartar (approved):
I think at some point we did not record the tensor parallelism in the IRPA or maybe we where sharding the dataset on-the-fly.
I think this use cases are not present anymore, so this check is good.

Alex-Vasile:
> I think at some point we did not recording the tensor parallelism size in the IRPA or maybe we were sharding the dataset on-the-fly. I think this use cases are not present anymore, so this check is good.

If we have datasets that have this problem we can change the check to inspect the contents of root_theta, but this seemed more elegant.

AmosLewis:
This patch led to new error for both 70b and 405b f16 test:
```
python3 -m sharktank.examples.export_paged_llm_v1 --irpa-file=/shark-dev/data/llama3.1/weights/405b/fp16/tp8/llama3.1_405b_fp16_tp8_parameters.irpa --output-mlir=/home/chi/src/shark-ai/2025-05-01/llama-405b/f16_torch_128.mlir --output-config=/home/chi/src/shark-ai/2025-05-01/llama-405b/f16_torch_128.json --bs-prefill=4 --bs-decode=4 --block-seq-stride=32 --attention-dtype=float16 --activation-dtype=float16 --tensor-parallelism-size=8 --pipeline-parallelism-size=1 --attention-kernel=torch
/home/chi/src/shark-ai/.venv/lib/python3.12/site-packages/iree/turbine/aot/params.py:163: UserWarning: The given NumPy array is not writable, and PyTorch does not support non-writable tensors. This means writing to this tensor will result in undefined behavior. You may want to copy the array to protect its data or make it writable before converting it to a tensor. This type of warning will be suppressed for the rest of this program. (Triggered internally at /pytorch/torch/csrc/utils/tensor_numpy.cpp:203.)
  return torch.from_numpy(wrapper)
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/chi/src/shark-ai/sharktank/sharktank/examples/export_paged_llm_v1.py", line 490, in <module>
    main()
  File "/home/chi/src/shark-ai/sharktank/sharktank/examples/export_paged_llm_v1.py", line 80, in main
    block_to_device_lookup = tuple(
                             ^^^^^^
  File "/home/chi/src/shark-ai/sharktank/sharktank/examples/export_paged_llm_v1.py", line 81, in <genexpr>
    tuple(range(tensor_parallelism_size)) for _ in range(hp.block_count)
                ^^^^^^^^^^^^^^^^^^^^^^^
NameError: cannot access free variable 'tensor_parallelism_size' where it is not associated with a value in enclosing scope
```

PR #1363, opened 2025-05-01
[tuner] Consider baseline config when all generated configs are slower

rkayaith:
Sometimes the tuner isn't able to generate a config faster than the baseline. Adding the baseline to the list of candidates in this case prevents the tuner from producing a spec file that regresses performance.

### DIFF
--- a/sharktuner/sharktuner/libtuner.py
+++ b/sharktuner/sharktuner/libtuner.py
@@ -656,7 +656,7 @@ def calculate_md5(file_path: Path) -> str:
 
 
 def find_collisions(
-    hash_list: list[tuple[int, str]]
+    hash_list: list[tuple[int, str]],
 ) -> tuple[bool, list[tuple[str, list[int]]]]:
     """
     Detect hash value collisions
@@ -1119,6 +1119,13 @@ def benchmark(
     all_candidates_with_speedup = baseline_handler.get_candidates_ordered_by_speedup(
         candidate_results
     )
+    if all(speedup >= 1 for _, speedup in all_candidates_with_speedup):
+        logging.warning(
+            "No tuner configs provided a speedup, including baseline config in results"
+        )
+        all_candidates_with_speedup = [
+            (first_baseline_result[0], 1.0)
+        ] + all_candidates_with_speedup
     top_candidates_with_speedup = all_candidates_with_speedup[:num_candidates]
     top_candidate_ids = []
     if baseline_handler.is_valid():


sharktuner/sharktuner/libtuner.py
```diff
@@ -1119,6 +1119,13 @@ def benchmark(
     all_candidates_with_speedup = baseline_handler.get_candidates_ordered_by_speedup(
         candidate_results
     )
+    if all(speedup >= 1 for _, speedup in all_candidates_with_speedup):
+        logging.warning(
+            "No tuner configs provided a speedup, including baseline config in results"
+        )
+        all_candidates_with_speedup = [
+            (first_baseline_result[0], 1.0)
```
bangtianliu, line 1127: Why choose `first_baseline_result[0]`?

I recommend reviewing how `BaselineResultHandler` is implemented and how the speedup is calculated. Based on that, you should return the correct result here or even add one function to `BaselineResultHandler` to handle the case in this PR.

rkayaith: From reading the code, the only difference seemed to be the device that the baseline was benchmarked on. I'm not sure how the device is relevant to `all_candidates_with_speedup`, so one is just picked arbitrarily.

bangtianliu: Benchmarking is performed using multiple processes, with each `BenchmarkResult` associated with a specific `device_id`. The speedup is typically calculated as the candidate's runtime divided by the average baseline runtime on the same `device_id`.
https://github.com/nod-ai/shark-ai/blob/4a120712e4c2d421c3b4dc5eace9d5cb0ef84e8b/sharktuner/sharktuner/libtuner.py#L969-L984

You should either use average time across all the devicd ids or on one specific device id.

rkayaith: the speedup here on line `1127` is for a baseline result, shouldn't it by definition be `1.0`, regardless of the device?

bangtianliu: Yes, but you need to get correct baseline runtime.

Max191: Is there any reason to log the baseline times when no candidates are better than the baseline? It seems like it could be misleading because it will just say:
```
Candidate 0 time: __ ms (100.0% of baseline)
```

Maybe instead this should just print a warning saying that no candidates were faster than the baseline, print the list of baseline runtimes (or the average), and then return early with a return value of `[first_baseline_result[0].candidate_id]`. It's better this way IMO because it will make it clear to the user that something may have gone wrong, instead of just doing the same thing it usually does but with some different candidates.

Max191 (changes requested):
This also needs a test

sharktuner/sharktuner/libtuner.py
bangtianliu: @Max191 +1 Yeah, I agree. Do not need to reuse existing code for print candidates since it is for normal cases.

rkayaith:
Any suggestions for how to test this?

PR #1364, opened 2025-05-01
View and transpose operations for TensorScaledLayout quantized tensors

KyleHerndon:
This gets Mistral Nemo to the point of rotary embedding


dan-garvey (approved):
Lol, I like how we had blockscaledi4 layout support but not tensor scaled support for these.

PR #1366, opened 2025-05-01
Improve performance of ops.mean 

### LINKED ISSUE #1311: Improve performance of ops.mean
_Originally posted by @sogartar in https://github.com/nod-ai/shark-ai/pull/1308#discussion_r2060625277_

oyazdanb:
Fixes #1311 

changing ops.mean based on the suggestions here https://github.com/nod-ai/shark-ai/pull/1308#discussion_r2060625277


sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -1126,19 +1126,33 @@ def mean_split(
...
-                    if i != 0
-                    else barrier_on_logical_device(shard, x.devices[0])
-                )
-                for i, shard in enumerate(x.shards)
-            ],
-            dim=x.shard_dim,
+        partial_sums = [
+            torch.sum(unbox_tensor(sh), dim=dim, keepdim=keepdim, dtype=dtype)
+            for sh in x.shards
+        ]
```
Alex-Vasile, line 1132:
```suggestion
        partial_sums = [
            sum(shard, dim=dim, keepdim=keepdim, dtype=dtype)
            for shard in x.shards
        ]
```

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -1126,19 +1126,33 @@ def mean_split(
...
+        partial_cnts = [
+            torch.tensor(
+                math.prod(unbox_tensor(sh).shape[d] for d in dim),
+                device=unbox_tensor(sh).device,
+                dtype=dtype,
+            )
+            for sh in x.shards
+        ]
```
Alex-Vasile, line 1140:
```suggestion
            torch.tensor(
                math.prod(shard.shape[d] for d in dim),
                device=unbox_tensor(sh).device,
                dtype=dtype,
            )
            for shard in x.shards
        ]
```

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -1126,19 +1126,33 @@ def mean_split(
...
+        sums_on_hub = [barrier_on_logical_device(partial_sums[0], x.devices[0])] + [
+            transfer_to_logical_device(ps, x.devices[0]) for ps in partial_sums[1:]
+        ]
+        cnts_on_hub = [barrier_on_logical_device(partial_cnts[0], x.devices[0])] + [
+            transfer_to_logical_device(ct, x.devices[0]) for ct in partial_cnts[1:]
+        ]
+
+        global_sum = torch.stack(sums_on_hub).sum(dim=0)
+        global_cnt = torch.stack(cnts_on_hub).sum(dim=0)
+        global_mean = global_sum / global_cnt
```
Alex-Vasile, line 1151: I think this would be much cleaner if we had something like `sharded_sum` implemented, similarly to `sharded_cat`.  Something very similar to this has already come up in several ops so far, e.g. top of `all_reduce_split_or_unreduced`.

sharktank/sharktank/ops/sharded_impls.py
Alex-Vasile: Same as previous one.

sharktank/sharktank/ops/sharded_impls.py
Alex-Vasile: the unbox is unneeded, as is the torch.

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -1126,19 +1126,30 @@ def mean_split(
...
+            )
+            for shard in x.shards
+        ]
+        sums_spt = SplitPrimitiveTensor(ts=partial_sums, shard_dim=0, devices=x.devices)
+        cnts_on_hub = [barrier_on_logical_device(partial_cnts[0], x.devices[0])] + [
+            transfer_to_logical_device(ct, x.devices[0]) for ct in partial_cnts[1:]
+        ]
+
+        global_sum = sharded_sum(sums_spt)
+        global_cnt = torch.stack(cnts_on_hub).sum()
```
rsuderman, line 1147: I would manually iterate and sum rather than stack and sum. It is counter intuitive but stacking then summing will likely be slower.

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -1126,19 +1126,30 @@ def mean_split(
...
+        global_mean = global_sum / global_cnt
```
rsuderman, line 1148: We should consider doing the normalization before the transfer. We can normalize by the number of elements as they are known.

rsuderman (changes requested):

sogartar (approved):

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -1126,19 +1126,21 @@ def mean_split(
...
+            SplitPrimitiveTensor(ts=partial_sums, shard_dim=0, devices=x.devices)
+        )
+
+        total_cnt = math.prod(x.shape[d] for d in dim)
+
+        global_mean = total_sum / total_cnt
+
+        return ReplicatedTensor(
+            ts=global_mean, shard_count=x.shard_count, devices=x.devices
         )
```
sogartar, line 1143: `sharded_sum` would return a tensor on device 0. Actually we would like to do an `all_reduce`. I understand that due to performance limitation of parallel transfers in IREE this actually does a bit less work as it does the normalization locally on only one device. But ultimately we want an all-reduce. You could at least leave a TODO comment that once IREE can issue parallel transfers on one queue we should do an all-reduce.

sharktank/sharktank/ops/sharded_impls.py
sogartar: This will not work correctly for int types. Actually, I am not sure what we have now would be correct for int due to rounding.

sogartar (approved):

PR #1375, opened 2025-05-02
Update builders.py

PhaneeshB:
- updated vmfbs with tuned flux transformer


amd-vivekag (approved):
LGTM, tested and worked for me

PR #1378, opened 2025-05-02
Integrate wave attention

aviator19941:
This PR adds a standard wave sdpa kernel invocation and uses the existing sharktank kernel logic to create the IR that calls the wave kernel's stream.executable. The tests passing depend on https://github.com/iree-org/iree-turbine/pull/787 being merged first.

### DIFF
--- a/sharktank/sharktank/kernels/base.py
+++ b/sharktank/sharktank/kernels/base.py
@@ -108,6 +108,7 @@ def call_function(target_function: Operation, *operands: Value) -> Sequence[Valu
     op_name = (
         "func.call" if target_function.operation.name == "func.func" else "util.call"
     )
+    target_function.attributes["sym_visibility"] = StringAttr.get("private")
 
     return Operation.create(
         op_name,


sharktank/sharktank/kernels/attention.py
```diff
@@ -208,3 +222,133 @@ def generate(self, ksel: KernelSelection, kb: KernelBuilder):
...
+            dynamic_symbols=dynamic_symbols,
+            dynamic_symbols_map=dynamic_symbols_map,
+            waves_per_eu=2,
+            denorm_fp_math_f32="preserve-sign",
+            func_name=target_function_name,
+        )
+        options = set_default_run_config(options)
+        base_attention = wave_compile(options, base_attention_func)
+
+        asm = base_attention.asm
```
zjgarvey, line 346: I'm not sure I understand the purpose of compiling the kernel to vmfb just to get the mlir asm. Especially since the executable being generated isn't going to be used for eager execution anyway. 

I think you could add the `WaveCompileOptions` option `compile_to_mlir=True` to avoid compiling to a vmfb with `wave_compile`.

sharktank/sharktank/kernels/base.py
```diff
@@ -73,7 +75,8 @@ def unpack_tensor_type(tensor_type: IrType) -> Tuple[str, str, IrType]:
     """
     rtt = RankedTensorType(tensor_type)
     ident = f"{'x'.join([str(dim) if dim >= 0 else 'D' for dim in rtt.shape])}x{rtt.element_type}"
-    return str(rtt), ident, rtt.element_type
+    dims = tuple(rtt.shape)
+    return str(rtt), ident, dims, rtt.element_type
```
zjgarvey, line 79: I don't think this is going to be a good idea. This util function is used in many places, and this change is likely to break all those other uses.

If you need the dims from a `RankedTensorType`, I'd recommend just doing `tuple(rtt.shape)` directly. `RankedTensorType` is also an `IrType`, so you can pass it into this function if you also need the identifier or other info.

sharktank/sharktank/kernels/base.py
```diff
@@ -105,8 +108,12 @@ def call_function(target_function: Operation, *operands: Value) -> Sequence[Valu
     target_symbol = FlatSymbolRefAttr.get(target_function.attributes["sym_name"].value)
     ftype = FunctionType(TypeAttr(target_function.attributes["function_type"]).value)
     operands = [i for i in operands if i is not None]
+    op_name = (
+        "func.call" if target_function.operation.name == "func.func" else "util.call"
+    )
+
     return Operation.create(
-        "util.call",
+        op_name,
```
zjgarvey, line 116: I'd like this change to be made upstream. I'm not sure what the update cadence is for the turbine version in shark-ai, so if this is urgent to get in, then we should do the change in turbine in parallel and make sure to remove the duplicated code once turbine is up to date in shark-ai.

sharktank/sharktank/kernels/base.py
```diff
@@ -163,3 +170,53 @@ def load_jinja_template(kb: KernelBuilder, template_file: str, **kwargs) -> Oper
...
+    source_module_op = load_jinja_asm(kb, asm)
+    if logger.isEnabledFor(logging.DEBUG):
+        logger.debug(
+            "Generated kernel IR %s:\n%s", function_name, str(source_module_op)
+        )
+    merger = Merger(
+        source_module_op, kb.module_body.owner, target_symbol_table=kb.symbol_table
+    )
+    merger.merge()
+    return kb.symbol_table[function_name]
```
zjgarvey, line 222: This whole code addition could be implemented by 

```python
from iree.turbine.runtime.op_reg.impl_helper import TemplateLoader

class ASMLoader(TemplateLoader):
    """Performs the same tasks as other template loaders, but uses a pre-loaded python string asm."""
    def __init__(self, asm : str):
        self.asm = asm

    def load_template(self, kb: KernelBuilder, name: str, **kwargs) -> Operation:
        return self._parse_module_asm(kb, self.asm)
```

Then you can initialize `_loader = ASMLoader(asm)`, and `_loader.inline_template_function` will do the same as `inline_wave_function` here.

zjgarvey (commented):
Nice. For now, I just have a few preliminary comments to trim out some code duplication.

sharktank/sharktank/kernels/base.py
```diff
@@ -163,3 +170,53 @@ def load_jinja_template(kb: KernelBuilder, template_file: str, **kwargs) -> Oper
             f"\n{lines_numbered}"
         )
     return module_op.operation
+
+
+def load_jinja_asm(kb: KernelBuilder, asm: str, **kwargs) -> Operation:
```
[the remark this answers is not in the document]
zjgarvey: Actually, see the other comment on this whole section.

sharktank/sharktank/kernels/base.py
aviator19941: yeah I realized that after the CI started blowing up because of it haha

sharktank/sharktank/kernels/attention.py
aviator19941: Done

sharktank/sharktank/kernels/base.py
aviator19941: Done

sharktank/sharktank/kernels/base.py
aviator19941: Done

sharktank/sharktank/kernels/attention.py
```diff
@@ -208,3 +223,133 @@ def generate(self, ksel: KernelSelection, kb: KernelBuilder):
         )
         kb.yield_results(*call_function(target_function, q, k, v, scale))
         pass
+
+
+@CustomOp.register(library=LIBRARY)
+class wave_flash_attention(CustomOp):
```
Groverkss, line 229: Writing a custom op every time you want to use a Wave kernel is a lot of work.  You realistically want to do something that we did for TK https://github.com/iree-org/iree-turbine/blob/main/iree/turbine/kernel/gen/kernel.py#L52 and generate the CustomOp automatically from the function signature. I don't think this is a good interface. You can also use `mlir_kernel` to build a `wave_kernel` class https://github.com/nod-ai/shark-ai/blob/main/sharktank/sharktank/kernels/mlir_kernel.py (Example usage: https://github.com/nod-ai/shark-ai/blob/main/sharktank/tests/kernels/mlir_kernel_test.py#L50) but exposing the entire pytorch registeration system to call a wave op is too much.

Groverkss (changes requested):

sharktank/sharktank/kernels/attention.py
aviator19941: Actually I'm going to do something similar to https://github.com/nod-ai/shark-ai/pull/1481/files, except for the wave BHSD sdpa kernel. I think that is probably the best way to integrate the wave kernels without having to create a CustomOp each time for a new Wave kernel. I would only need to create a new `mlir_kernel` decorator method for each new Wave kernel.

PR #1384, opened 2025-05-05
[sharktank] General implementation for trivially replicable ops

sogartar:
The majority of operations have essentially the same replication mechanism across device. We call the underlying unsharded op with the corresponding shards and then construct a replicated tensor form the resulting shards.

This change solves this in general way to not have to write the same replication pattern for all ops.

### DIFF
--- a/sharktank/sharktank/ops/utils.py
+++ b/sharktank/sharktank/ops/utils.py
@@ -33,15 +33,12 @@ def call_trivially_replicable(
         is_any_tensor,
     )
 
-    def is_leaf(v: Any) -> bool:
-        return is_any_tensor(v) or tree.is_leaf_default(v)
-
     flat_args = tree.flatten(
         (
             args,
             kwargs,
         ),
-        is_leaf=is_leaf,
+        is_leaf=_is_leaf,
     )
     first_replicated_tensor_arg = [
         arg for arg in flat_args if isinstance(arg, ReplicatedTensor)
@@ -71,26 +68,24 @@ def call_trivially_replicable(
         a.append(b)
         return a
 
-    def tensor_to_empty_list(x: Any) -> list:
-        if is_any_tensor(x):
-            return []
-        return x
+    def tensor_to_empty_list_if_tensor(x: Any) -> Any:
+        return [] if is_any_tensor(x) else x
 
     reduce_initial = tree.map_leaves(
-        per_shard_results[0], f=tensor_to_empty_list, is_leaf=is_leaf
+        per_shard_results[0], f=tensor_to_empty_list_if_tensor, is_leaf=_is_leaf
     )
 
     # Make a lists of tensor shards that correspond to the same replicated tensor.
     reduced_results = tree.reduce_horizontal(
         fn=reduce_to_list_of_tensors_fn,
         trees=per_shard_results,
-        is_leaf=is_leaf,
+        is_leaf=_is_leaf,
         initial=reduce_initial,
     )
 
     def make_replicated_tensor_if_tensor_collection(x: Any) -> Any:
         if _is_tensor_collection(x):
-            return ReplicatedTensor(ts=x, devices=first_replicated_tensor_arg.devices)
+            return ReplicatedTensor(ts=x)
         return x
 
     result_with_replicated_tensor = tree.map_leaves(
@@ -156,9 +151,7 @@ def _extract_tensor_shard_if_replicate(v: Any, shard_index: int) -> Any:
 
 
 def _is_leaf(v: Any) -> bool:
-    from sharktank.types import (
-        is_any_tensor,
-    )
+    from sharktank.types import is_any_tensor
 
     return is_any_tensor(v) or tree.is_leaf_default(v)
 
@@ -168,8 +161,6 @@ def _is_leaf_or_tensor_collection(x: Any) -> bool:
 
 
 def _is_tensor_collection(x: Any) -> bool:
-    from sharktank.types import (
-        is_any_tensor,
-    )
+    from sharktank.types import is_any_tensor
 
     return isinstance(x, Sequence) and all(is_any_tensor(v) for v in x)
--- a/sharktank/sharktank/ops/utils.py
+++ b/sharktank/sharktank/ops/utils.py
@@ -1,0 +1,175 @@
+# Copyright 2025 Advanced Micro Devices, Inc.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import functools
+
+from typing import Callable, Sequence, Any
+from sharktank.utils import tree
+
+
+def call_trivially_replicable(
+    fn: Callable[..., Any], args: tuple[Any, ...], kwargs: dict[str, Any]
+) -> Any:
+    """Call an function with replicated tensor arguments that can be called with
+    unsharded tensors.
+
+    It is expected that all replicated tensors have the same number of shards.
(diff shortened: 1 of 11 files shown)


sogartar:
This causes a failure when tracing with torch 2.3, but it seems we are going to drop 2.3. I will open a PR for that shortly, but other than that the PR is good for review.

sharktank/sharktank/ops/utils.py
```diff
@@ -0,0 +1,175 @@
...
+        fn=reduce_to_list_of_tensors_fn,
+        trees=per_shard_results,
+        is_leaf=is_leaf,
+        initial=reduce_initial,
+    )
+
+    def make_replicated_tensor_if_tensor_collection(x: Any) -> Any:
+        if _is_tensor_collection(x):
+            return ReplicatedTensor(ts=x, devices=first_replicated_tensor_arg.devices)
+        return x
```
Alex-Vasile, line 94: This is only doing some of the work in `sharded_wrap_override`. And now some ops are going through `sharded_wrap_override` and others are not (even for ReplicatedTensors). You shouldn't need to specify `devices` here, and instead be wrapping so that we call `sharded_wrap_override`.

sharktank/sharktank/ops/utils.py
```diff
@@ -0,0 +1,175 @@
...
+    def reduce_to_list_of_tensors_fn(a: Any, b: Any) -> Any:
+        if not is_any_tensor(b):
+            return a
+        a.append(b)
+        return a
+
+    def tensor_to_empty_list(x: Any) -> list:
+        if is_any_tensor(x):
+            return []
+        return x
```
Alex-Vasile, line 77:
```suggestion
        return [] if is_any_tensor(x) else x
```

sharktank/sharktank/ops/utils.py
```diff
@@ -0,0 +1,175 @@
...
+        for per_shard_args, per_shard_kwargs in per_shard_args_and_kwargs
+    ]
+
+    def reduce_to_list_of_tensors_fn(a: Any, b: Any) -> Any:
+        if not is_any_tensor(b):
+            return a
+        a.append(b)
+        return a
+
+    def tensor_to_empty_list(x: Any) -> list:
```
Alex-Vasile, line 74: The type hint for the return isn't correct. It should be either `list | Any` or just `Any`.

sharktank/sharktank/ops/utils.py
```diff
@@ -0,0 +1,175 @@
...
+
+    if isinstance(v, ReplicatedTensor):
+        return v.shards[shard_index]
+    return v
+
+
+def _is_leaf(v: Any) -> bool:
+    from sharktank.types import (
+        is_any_tensor,
+    )
```
Alex-Vasile, line 161: Is the comma after `is_any_tensor` needed? We don't do this with other single imports, and it's causing black to format this over 3 lines instead of 1, making both this and the outer function much longer than necessary. Same comment for many of the other one-line functions

sharktank/sharktank/utils/tree.py
```diff
@@ -6,6 +6,7 @@
 
 from typing import Dict, Any, Callable, Mapping, Iterable, Sequence, Union
 from collections.abc import Mapping as MappingABC, Iterable as IterableABC
+import functools
```
Alex-Vasile, line 9: You'll want someone else to review this file, since I haven't worked with the tree functionality before.

sharktank/sharktank/ops/utils.py
Alex-Vasile: That this isn't failing anything means that we need more tests, including a few negative ones, inside `pipeline_parallized_test.py`.

For example, adding the following test to `pipeline_parallized_test.py::CatTest`, will throw an error since the devices of the args aren't being checked:
```python
    def testCatReplicatedShouldFail(self):
        """Concatenation of replicated tensors."""
        shard_count = 2
        cat_dim = 0
        # Done to ensure no overlap with default devices so incorrect placement is caught
        devices = tuple(range(shard_count, 2 * shard_count))

        a = torch.rand(5, 4, dtype=torch.float32)
        b = torch.rand(3, 4, dtype=torch.float32)

        sharded_a = ReplicatedTensor(ts=a, shard_count=shard_count, devices=devices)
        sharded_b = ReplicatedTensor(ts=b, shard_count=shard_count, devices=tuple(range(shard_count)))

        try:
            ops.cat([sharded_a, sharded_b], dim=cat_dim)
        except ValueError:
            return  # Should throw an error since the tensors are not on the same devices
        raise AssertionError("Should have thrown an error since the tensors are not on the same devices")
```

Alex-Vasile (changes requested):

sharktank/sharktank/utils/tree.py
sogartar: I don't think anyone else is. This is some general functionality for tree manipulation. If it is not clear I can add more documentation.

sharktank/sharktank/ops/utils.py
```diff
@@ -0,0 +1,175 @@
...
+    When collecting the results all non-tensor results are taken from the first device.
+    """
+
+    from sharktank.types import (
+        ReplicatedTensor,
+        is_any_tensor,
+    )
+
+    def is_leaf(v: Any) -> bool:
+        return is_any_tensor(v) or tree.is_leaf_default(v)
```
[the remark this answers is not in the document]
sogartar: Removed it.

sharktank/sharktank/types/tensors.py
```diff
@@ -1459,6 +1460,10 @@ def clone(self, **kwargs) -> "UnreducedTensor":
         return UnreducedTensor(**kwargs)
 
 
+def is_any_tensor(x: Any) -> bool:
+    return isinstance(x, InferenceTensor) or isinstance(x, torch.Tensor)
```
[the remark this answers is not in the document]
sogartar: Done.

sharktank/sharktank/ops/utils.py
sogartar: I refactored a bit such that the trivially replicated overrides happen inside `sharded_impls.py`. I may come back at some point and refactor this composition. We can't declare right now ops outside of `signatures.py`. For example we can't have tests declare ops.

sharktank/sharktank/ops/utils.py
sogartar: Done.

sharktank/tests/ops/sharded_test.py
```diff
@@ -295,6 +299,26 @@ def testCatNonSplitDim(self):
         actual_result = ops.cat([sharded_a, sharded_b], dim=cat_dim)
         assert expected_result.is_deep_equal(actual_result, compare_name=False)
 
+    def testCatReplicatedShouldFail(self):
```
Alex-Vasile, line 302: This probably belongs in `pipeline_parallelized_test.py` since that's where all the pp-related ops tests currently are.

sharktank/sharktank/utils/tree.py
```diff
@@ -14,59 +15,157 @@
...
+    dict_type: type = dict,
+    sequence_type: type = tuple,
+) -> Tree:
+    """Reduce a list of trees such that elements with the same path in the tree are
+    aligned.
+
+    E.g.
+    ```
+    {"a": a1, {"b": b1}}
+    {"a": a2, {"b": b2}}
```
Alex-Vasile, line 122: I don't understand what this example is showing. Is it complete, or what am I missing?

Alex-Vasile (approved):
Two minor comments, but good to go.

sharktank/tests/ops/sharded_test.py
sogartar: I moved it.

PR #1386, opened 2025-05-05
Update llama_serving.md doc to reflect the latest changes

pravg-amd:
1. Update the shortfin response format, which contains the entire json instead of just the response text.
2. Update the export and compilation commands for sharded models.

### DIFF
--- a/app_tests/integration_tests/llm/model_management.py
+++ b/app_tests/integration_tests/llm/model_management.py
@@ -335,7 +335,6 @@ class ModelStageManager:
                     "sharktank.tools.dump_gguf",
                     f"--gguf-file={gguf_path}",
                     f"--output-irpa={str(irpa_path)}",
-                    "--save-input-output-blocks",
                 ],
                 check=True,
             )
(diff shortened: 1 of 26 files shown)


docs/shortfin/llm/user/llama_serving.md
```diff
@@ -404,7 +409,11 @@ iree-compile /path/to/output/llama3.1-405b.mlir \
   --iree-hal-target-device=hip[5] \
   --iree-hal-target-device=hip[6] \
   --iree-hal-target-device=hip[7] \
-  --iree-hip-target=gfx942
+  --iree-hip-target=gfx942 \
+  --iree-opt-level=O3 \
```
stbaione, line 413: Do we know why these extra flags are required now for the compile?

pravg-amd: I am not aware of exactly why we need these optimizations. Without these, was running into the following error

```
ValueError: shortfin_iree-src/runtime/src/iree/vm/bytecode/verifier.c:344: RESOURCE_EXHAUSTED; register count overflow
```

These flags are used in the llama benchmark tests in sharktank. Maybe @AmosLewis can provide more insight to this ?

amd-vivekag: I faced the same issue and I also had to use these options for 405B model.
I noticed that option `--iree-opt-level=O3` is not working with iree release 0604, it is causing compilation error. The option was working fine with 0515 release though.

PR #1391, opened 2025-05-05
[shortfin][LLM] Support multiple HIP streams

vinayakdsci:
Adds support for running multiple instances of an LLM service on a distinct HIP stream.

### DIFF
--- a/shortfin/python/shortfin_apps/llm/components/lifecycle.py
+++ b/shortfin/python/shortfin_apps/llm/components/lifecycle.py
@@ -22,6 +22,7 @@ from .tokenizer import Tokenizer
 from typing import TYPE_CHECKING
 from fastapi import FastAPI
 
+import os
 
 from contextlib import asynccontextmanager
 import logging
@@ -49,6 +50,10 @@ class ShortfinLlmLifecycleManager:
 
     def __init__(self, args):
         # Load server configuration with priority: command line > config file > defaults
+        if args.enable_multiple_hip_streams:
+            os.environ["SHORTFIN_AMDGPU_LOGICAL_DEVICES_PER_PHYSICAL_DEVICE"] = str(
+                args.instances
+            )
         model_params = ModelParams.load_json(args.model_config)
         server_params = ServerParams.load(
             args.server_config if hasattr(args, "server_config") else None
@@ -73,6 +78,12 @@ class ShortfinLlmLifecycleManager:
             amdgpu_allow_device_reuse=server_params.amdgpu_allow_device_reuse,
         )
 
+        if (
+            args.enable_multiple_hip_streams
+            and len(sysman.ls.devices) != args.instances
+        ):
+            raise RuntimeError("Failed to start multiple HIP streams for execution.")
+
         # Setup each service we are hosting.
         eos_token = get_eos_from_tokenizer_config(args.tokenizer_config_json)
         tokenizer = Tokenizer.from_tokenizer_json_file(
@@ -89,6 +100,7 @@ class ShortfinLlmLifecycleManager:
                 model_params=model_params,
                 server_params=server_params,
                 program_isolation=server_params.program_isolation,
+                service_idx=i if args.enable_multiple_hip_streams else 0,
             )
             service.load_inference_module(args.vmfb)
             service.load_inference_parameters(*args.parameters, parameter_scope="model")
--- a/shortfin/python/shortfin_apps/llm/components/service.py
+++ b/shortfin/python/shortfin_apps/llm/components/service.py
@@ -54,6 +54,7 @@ class LlmGenerateService(GenerateService):
         server_params: "ServerParams",
         program_isolation: str = "per_fiber",
         max_queue_size: int = 3,  # Maximum number of requests in queue
+        service_idx: int = 0,
     ):
         super().__init__(sysman)
         self.name = name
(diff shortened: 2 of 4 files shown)


shortfin/python/shortfin_apps/llm/components/service.py
```diff
@@ -113,9 +119,9 @@ def initialize_worker_and_fiber(self):
 
         self.devices = fibers[0].devices_dict.values()
         self.main_worker = self.sysman.ls.create_worker(f"{self.name}-inference")
-        self.main_fiber = self.sysman.ls.create_fiber(self.main_worker)
-        self.prefill_fiber = self.sysman.ls.create_fiber(self.main_worker)
-        self.decode_fiber = self.sysman.ls.create_fiber(self.main_worker)
+        self.main_fiber = self.fiber_pool.fibers[0]
+        self.prefill_fiber = self.fiber_pool.fibers[0]
+        self.decode_fiber = self.fiber_pool.fibers[0]
```
rsuderman, line 124: This assigns the same fiber to `main`, `prefill`, and `decode` where previously they were on different fibers. I had a discussion with Stephen and there is a high likelihood that me want to revert the fiber pool work as its current implementation actually slows down processing.

shortfin/python/shortfin_apps/llm/components/service.py
```diff
@@ -94,11 +96,15 @@ def initialize_worker_and_fiber(self):
         logger.info(
             f"Creating {num_workers} workers, with {fibers_per_worker} fibers per worker..."
         )
+
         fibers = []
+        assigned_hal_device = self.sysman.ls.devices[self.service_idx]
```
rsuderman, line 101: This appears to grab a different device instead of a different stream. I assume this is related to the environment variables above?

shortfin/python/shortfin_apps/llm/components/lifecycle.py
```diff
@@ -49,6 +50,10 @@ class ShortfinLlmLifecycleManager:
 
     def __init__(self, args):
         # Load server configuration with priority: command line > config file > defaults
+        if args.enable_multiple_hip_streams:
+            os.environ["SHORTFIN_AMDGPU_LOGICAL_DEVICES_PER_PHYSICAL_DEVICE"] = str(
```
rsuderman, line 54: Can you include a link to where this variable is used? Having multiple logical devices adds some complexity. We should also note whether this approach fully replicates allocated buffers or if we are able to access resources across streams.

shortfin/python/shortfin_apps/llm/server.py
```diff
@@ -68,6 +69,12 @@ def parse_args(argv):
     parser.add_argument(
         "--instances", type=int, default=1, help="Number of shortfin instances to run"
     )
+    parser.add_argument(
+        "--enable_multiple_hip_streams",
+        action=BooleanOptionalAction,
```
rsuderman, line 74: Just use `argparse.BooleanOptionalAction` rather than including the additional import.

rsuderman (changes requested):

shortfin/python/shortfin_apps/llm/components/service.py
daveliddell: Yes, my understanding is that there is no stream object modeled in shortfin directly. Instead, you use multiple streams by creating multiple logical logical devices for a physical device, which you do by turning on that environment variable. Each logical device creates its own stream.

This looks like another place we're assuming we have only one physical device. I hope that's ok. In this case, you'll only see device 0 if logical devices are off, as we're assuming that there can't be a second physical device.

daveliddell (approved):
Looks good

shortfin/python/shortfin_apps/llm/components/lifecycle.py
dezhiAmd: This change come from [PR ](https://github.com/nod-ai/shark-ai/pull/1302l)authored by me. This [link](https://github.com/nod-ai/shark-ai/pull/1302#issuecomment-2855992217) here can answer your question about where this variable is used.

shortfin/python/shortfin_apps/llm/components/lifecycle.py
```diff
@@ -49,6 +50,10 @@ class ShortfinLlmLifecycleManager:
 
     def __init__(self, args):
         # Load server configuration with priority: command line > config file > defaults
+        if args.enable_multiple_hip_streams:
```
dezhiAmd, line 53: In my opinion, this if check is not needed.
When you set "SHORTFIN_AMDGPU_LOGICAL_DEVICES_PER_PHYSICAL_DEVICE" more than 1, you automatically created multiple hip streams. 

This is what I see:
logical_devices_per_physical_device is set using an env variable or a config file
([amdgpu.cc])(

[shark-ai/shortfin/src/shortfin/local/systems/amdgpu.cc](https://github.com/nod-ai/shark-ai/blob/1a0c0ccaf461076c0fefac0a3252087cfa74df38/shortfin/src/shortfin/local/systems/amdgpu.cc#L73)

Line 73 in [1a0c0cc](https://github.com/nod-ai/shark-ai/commit/1a0c0ccaf461076c0fefac0a3252087cfa74df38)

 "amdgpu_logical_devices_per_physical_device", /*non_negative=*/true); 
)
logical_devices_per_physical_device_ is used to create device, refer [here](https://github.com/nod-ai/shark-ai/blob/0ca865093c2528058b6ee2eae916d952c37e73a7/shortfin/src/shortfin/local/systems/amdgpu.cc#L212).

PR #1393, opened 2025-05-06
[sharktank] Tensor-parallel MoE block

sogartar:
Add support for expert-prallel dense mixture of experts block.

This change includes adding support for matmul where the RHS is a 3D tensor split on the batch dimension. This drives some changes like using `.mT` instead of `.T` when `transpose_b=True`.

Made `scatter_` work with non-number source arguments.

Fixed dispatching to trivially replicable to handle if some dispatch args are not tensors, but all tensor args are replicated.

### DIFF
--- a/sharktank/sharktank/layers/mixture_of_experts_block.py
+++ b/sharktank/sharktank/layers/mixture_of_experts_block.py
@@ -9,8 +9,8 @@ from typing import Optional
 import torch
 
 from sharktank.layers import *
-from sharktank.ops import replicate, softmax, topk, zeros_like
-from sharktank.types import Theta
+from sharktank.ops import softmax, topk, zeros_like, reshard_like
+from sharktank.types import ShardedTensor, Theta
 
 __all__ = [
     "MoeBlock",
@@ -36,23 +36,28 @@ class MoeBlock(ThetaLayer):
         score_experts=softmax,
         normalize_experts=True,
         add_residual=True,
-        tensor_parallel_devices: Optional[tuple[int, ...]] = None,
         expert_count: Optional[int] = None,
         n_expert_groups: Optional[int] = None,
         n_limited_groups: Optional[int] = None,
         route_scale: Optional[float] = 1.0,
     ):
-        """
-        tensor_parallel_devices:
-            The devices that would be involved in tensor parallel distribution.
-            Supports the degenerate case of 1 device, where sharded tensors would also
-            be employed. If None then no distribution would be done.
-        """
         super().__init__(theta)
-        if n_expert_groups is not None and expert_count % n_expert_groups != 0:
-            raise ValueError(
-                f"Number of experts {expert_count} must be divisible by the number of expert groups {n_expert_groups}."
-            )
+        if n_expert_groups is not None:
+            if expert_count % n_expert_groups != 0:
+                raise ValueError(
+                    (
+                        f"Number of experts {expert_count} must be divisible by the "
+                        f"number of expert groups {n_expert_groups}."
+                    )
+                )
+            n_experts_per_group = expert_count // n_expert_groups
+            if n_experts_per_group < n_limited_groups:
+                raise ValueError(
+                    (
+                        f"Number of limited expert groups {n_limited_groups} must be at "
+                        f"most the number of experts per group {n_experts_per_group}."
+                    )
+                )
         self.expert_used_count = expert_used_count
         self.expert_count = expert_count
         self.n_expert_groups = n_expert_groups
@@ -67,7 +72,6 @@ class MoeBlock(ThetaLayer):
         self.ffn_norm = torch.nn.Identity()
         self.layer_output_norm = torch.nn.Identity()
         self.shared_experts = None
-        self.tensor_parallel_devices = tensor_parallel_devices
         self.route_scale = None
         if route_scale is not None and route_scale != 1:
             self.route_scale = route_scale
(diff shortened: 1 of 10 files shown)


sharktank/sharktank/layers/mixture_of_experts_block.py
```diff
@@ -36,11 +36,18 @@ def __init__(
         score_experts=softmax,
         normalize_experts=True,
         add_residual=True,
+        tensor_parallel_devices: Optional[tuple[int, ...]] = None,
```
sogartar, line 39: I am actually thinking to drop this init parameter and get this from forward's input.

sharktank/sharktank/ops/default_impls.py
```diff
@@ -500,16 +500,33 @@ def scatter__default(
     *,
     reduce: str | None = None,
 ) -> Tensor:
-    assert isinstance(value, Number), "Tensor version of this op not implemented"
     inout = unbox_tensor(inout)
     index = unbox_tensor(index)
+    if isinstance(value, (torch.Tensor, InferenceTensor)):
+        assert isinstance(value, (torch.Tensor, PrimitiveTensor))
```
Alex-Vasile, line 506: How come this isn't being handled by the override and the trampoline?

sharktank/sharktank/ops/signatures.py
```diff
@@ -303,6 +304,47 @@ def _embedding_lookup_trampoline(
...
+    device: torch.device | None = None,
+    requires_grad: bool = False,
+    memory_format: torch.memory_format = torch.preserve_format,
+) -> AnyTensor:
+    """See torch.zeros_like"""
+    ...
+
+
+@empty_like.trampoline
+def _zeros_like_trampoline(
```
Alex-Vasile, line 322:
```suggestion
def _empty_like_trampoline(
```

sharktank/sharktank/ops/signatures.py
```diff
@@ -1098,6 +1138,34 @@ def _scatter__trampoline(
         d.fail(tensors)
 
 
+@overridable
+def scatter_add(
+    input: AnyTensor, dim: int, index: AnyTensor, src: AnyTensor | Number
+) -> AnyTensor:
+    """
+    See torch.scatter_add
+    NOTE: Does not modify the inout tensor in place for ShardedTensors, will return copy.
```
Alex-Vasile, line 1147: `scatter_add` shouldn't modify any of them in-place. Only `scatter_add_` should be in-place.

sharktank/tests/layers/mixture_of_experts_block_test.py
```diff
@@ -200,6 +203,106 @@ def testParityOfExpertPreGatherFfnAndDenseFfn(
...
+                rms_epsilon=0.01,
+                moe_activation_fn=torch.nn.functional.silu,
+                score_experts_fn=torch.nn.functional.sigmoid,
+                normalize_experts=True,
+                add_residual=False,
+                route_scale=None,
+                tensor_parallelism_size=2,
+            ),
+        ]
+    )
```
Alex-Vasile, line 229: Did you mean to add multiple tests?

sharktank/tests/ops/ops_test.py
```diff
@@ -352,6 +352,22 @@ def forward(self, a, d, qs, m):
         self.assertIn("mmt_block_scaled_offset_q4_unsigned.default", s)
 
 
+class TestScatterAdd(unittest.TestCase):
+    def setUp(self):
+        torch.random.manual_seed(0)
+
+    def test(self):
```
Alex-Vasile, line 359: You should have a test for when `src` is a number. I don't think that the override with match against it.

sharktank/sharktank/layers/mixture_of_experts_block.py
Alex-Vasile: I think that's a good idea. The extra state isn't needed.

Alex-Vasile (changes requested):

sharktank/sharktank/ops/default_impls.py
sogartar: We don't dispatch on value.

sharktank/sharktank/layers/mixture_of_experts_block.py
sogartar: I removed it.

sharktank/sharktank/ops/default_impls.py
sogartar: `torch.scatter_add` actually does not support source as number, but `torch.scatter` does. I refactored `scatter_` to dispatch on the source as well.
Had to fix the trivially replicable dispatch logic to also handle non-tensor dispatch args.

sharktank/sharktank/ops/signatures.py
sogartar: Done.

sharktank/sharktank/ops/signatures.py
sogartar: Sorry for that copy-paste remnant. Its removed.

sharktank/tests/layers/mixture_of_experts_block_test.py
sogartar: I added one more test with another set of parameters and bf16 dtype.

Alex-Vasile (approved):

PR #1395, opened 2025-05-06
Adding benchmark_client.py from demo branch

pdhirajkumarprasad:
adding benchmark client file from demo branch to main

### DIFF
--- a/shortfin/python/shortfin_apps/llm/benchmark_client.py
+++ b/shortfin/python/shortfin_apps/llm/benchmark_client.py
@@ -1,0 +1,699 @@
+import asyncio
+import time
+import requests
+import uuid
+from typing import Dict, Any, List, Tuple
+import numpy as np
+import csv
+import os
+from datetime import datetime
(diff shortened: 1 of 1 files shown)


pravg-amd (approved):
Can we run the python formatter on this file once ?

shortfin/python/shortfin_apps/llm/benchmark_client.py
```diff
@@ -0,0 +1,699 @@
...
+
+        results = {}
+
+        def process_stream():
+            start_time = time.perf_counter()
+            token_times = []
+            generated_text = []
+            if self.stream:
+                try:
+                    print(f"Sending request to {self.base_url}/generate")
```
pravg-amd, line 53: Would it be better to move these prints to logger?

pdhirajkumarprasad: @pravg-amd this help to see if the client-server is progressing or hang so we can keep it at stdout

PR #1397, opened 2025-05-06
Shortfin llm gpu argmax

stbaione:
# Description

Enable using the `gpu argmax` kernel in the Shortfin LLM server.

This creates two `samplers`, a `CPUSampler` and a `GPUSampler`.

If the `--enable_gpu_sampling` flag is passed when starting up the server, it will use the `select_greedy` function from the `GPUSampler`.

If the `--enable_gpu_sampling` flag is not passed then we'll use the `CPUSampler` select_greedy function.

Since we don't have a `top_k` kernel exporting yet, we still use the `CPUSampler` if the user sets the server to `beam_search`, or if they send a request with `top_p` or `top_k` set.

Seeing positive perf results. Also verified that `CPU` perf is maintained as compared to `main`:

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>

<div style='direction:ltr'>

CPU/GPU | BS_Prefill | BS_Decode | Concurrent Reqs. | Dtype | Latency | Throughput
-- | -- | -- | -- | -- | -- | --
CPU | 8 | 64 | 8 | Fp16 | 16.67 | 0.48
GPU | 8 | 64 | 8 | Fp16 | 10.69 | 0.75
CPU | 8 | 64 | 8 | FP8 | 22.73 | 0.35
GPU | 8 | 64 | 8 | FP8 | 16.50 | 0.48

</div>

</body>

</html>


shortfin/python/shortfin_apps/llm/components/token_selection_strategy/sampler.py
```diff
@@ -8,15 +8,45 @@
...
+import shortfin as sf
 import shortfin.array as sfnp
 
 from shortfin_apps.utils import convert_int_to_float
 
+from .config import SamplingKernels
+
+
+@dataclass
+class GPUSampler:
```
IanNod, line 24: Can we combine cpu and gpu into one sampler class or at least have one abstract base sampler class? Would help clean up a lot of the changes in this PR

stbaione:
Dropping in favor of #1454

PR #1400, opened 2025-05-06
improving topk

### LINKED ISSUE #1325: Implement more efficient version of ops.topk
The version of `ops.topk` implemented, in https://github.com/nod-ai/shark-ai/pull/1324, for `SplitPrimitiveTensor`s when performing it across the `shard_dim` could be improved to be more efficient.

Current implementation is at:
https://github.com/nod-ai/shark-ai/blob/65919669e53aadd34dc07c941c116e5f60b755af/sharktank/sharktank/ops/sharded_impls.py#L1653-L1660

oyazdanb:
fixes #1325
updating the topk ops for when dim==shard_dim

### DIFF
--- a/sharktank/sharktank/ops/sharded_impls.py
+++ b/sharktank/sharktank/ops/sharded_impls.py
@@ -1716,33 +1716,30 @@ def topk_split(
         all_v_loc = []
         all_i_glob = []
         offset = 0
-        running_vals, running_inds = None, None
         for i, shard in enumerate(input.shards):
             v_loc, i_loc = topk(shard, k=k, dim=dim, largest=largest, sorted=sorted)
 
             i_glob = i_loc + offset
             offset += shard.shape[dim]
 
-            if i != 0:
+            if i == 0:
+                v_loc = barrier_on_logical_device(v_loc, input.devices[0])
+                i_glob = barrier_on_logical_device(i_glob, input.devices[0])
+            else:
                 v_loc = transfer_to_logical_device(v_loc, input.devices[0])
                 i_glob = transfer_to_logical_device(i_glob, input.devices[0])
 
-            else:
-                v_loc = barrier_on_logical_device(v_loc, input.devices[0])
-                i_glob = barrier_on_logical_device(i_glob, input.devices[0])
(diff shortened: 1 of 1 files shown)


sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -1711,14 +1711,44 @@ def topk_split(
...
+            if running_vals is None:
+                running_vals, running_inds = v_loc, i_glob
+            else:
+                cand_vals = torch.cat([running_vals, v_loc], dim=dim)
+                cand_inds = torch.cat([running_inds, i_glob], dim=dim)
+
+                running_vals, pos = torch.topk(
+                    cand_vals, k=k, dim=dim, largest=largest, sorted=sorted
+                )
+                running_inds = torch.take_along_dim(cand_inds, pos, dim=dim)
```
Alex-Vasile, line 1742: I'm pretty sure this is going to create one large serial dependency chain between the different iterations, which I don't think IREE will break for us.

I think the following would be faster:
1. topk each shard in parallel
2. adjust the indices to be indices into full tensor rather than into each shard.
3. Transfer values and indices to devices[0].
4. Cat values together, same with indices.
5. Perform 1 last topk on values and use resulting indices to index into the adjusted index tensor from step 3.
6. Return

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -1711,14 +1711,44 @@ def topk_split(
...
+
+        top_vals = ReplicatedTensor(
+            ts=running_vals, shard_count=input.shard_count, devices=input.devices
+        )
+        top_inds = ReplicatedTensor(
+            ts=running_inds, shard_count=input.shard_count, devices=input.devices
```
Alex-Vasile, line 1748: You can omit the `devices=input.device` to make it shorter. This function is getting wrapped by `sharded_wrap_override` which will ensure `devices` is set properly.

Alex-Vasile (approved):
Minor formatting, but I think it's good to go.

Alex-Vasile (approved):

PR #1404, opened 2025-05-07
 Complete overhaul of the KVCache 

rsuderman:
Rather than a single mega KV Cache class integrated into PagedAttention
I split the three main behaviors of the KV cache into composable
components:

- KVCache
- ShardedCache
- PipelinedCache

KVCache handles the unpipelined TP-1 case with direct tensor
manipulations. This is what is relevant for codegen / underlying
implementation.

ShardedCache performs the management of the separate pieces of the
KVCache. This includes handling the split primitive tensor across each
dimension. It has no understanding of how the cache is laid out and
purely handles sharding across the head dimension instead relying on
subcaches stored in the ShardedCache.

PipelinedCache performs the per-block sharding of the KVCache and
similar to ShardedCache has no underlying understanding of how the
buffers are stored. It only has an understanding of the unsharded
version of the cache.


sharktank/sharktank/layers/paged_attention.py
```diff
@@ -283,21 +290,31 @@ def read(
...
+        selected = ops.index_select(page_table, 0, subblock_ids.flatten(0, 1))
 
         selected = selected.unflatten(0, blocked_shape[:2])
         key = selected[:, :, 0, :].flatten(1, 2)
         value = selected[:, :, 1, :].flatten(1, 2)
 
         return key, value
 
+    @staticmethod
+    def deshard_arg(arg0, arg1):
```
Alex-Vasile, line 309: It isn't clear without reading the whole function which of the args get desharded. Something like `def deshards_like(arg, like)` would be easier to understand.

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -283,21 +290,31 @@ def read(
...
+        devices = arg0.devices
+
+        if isinstance(arg1, SplitPrimitiveTensor):
+            assert all([a == b for a, b in zip(devices, arg1.devices)])
+
+        device_map = {d: i for i, d in enumerate(arg1.devices)}
```
Alex-Vasile, line 315: This will quietly fail when devices are duplicated.

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -136,19 +155,7 @@ def unflatten_page_tables(
...
-                result = SplitPrimitiveTensor(
-                    ts=shards, shard_dim=4, devices=page_slab.devices
-                )
-            elif pipeline > 1:
-                result = ReplicatedTensor(ts=shards, devices=page_slab.devices)
-
+            result = self.unflatten_page_table(page_slab, pipeline=pipeline)
             unflattened.append(result)
 
         return unflattened
```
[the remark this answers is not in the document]
rsuderman: Done.

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -315,116 +332,209 @@ def write_timestep(
...
+                and isinstance(seq_positions, ReplicatedTensor)
+                and isinstance(page_ids, ReplicatedTensor)
+            ):
+
+                page_table_split = isinstance(page_table, SplitPrimitiveTensor)
+                partition_split = isinstance(partition, SplitPrimitiveTensor)
+                assert page_table_split == partition_split
+
+                def build_map(t):
+                    return {d: i for i, d in enumerate(t.devices)}
```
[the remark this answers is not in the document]
rsuderman: Removed.

sharktank/sharktank/layers/paged_attention.py
rsuderman: renamed

sharktank/sharktank/layers/paged_attention.py
rsuderman: Added an assertion that devices are unique

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -315,116 +332,209 @@ def write_timestep(
...
+                and isinstance(partition, ShardedTensor)
+                and isinstance(page_ids, ReplicatedTensor)
+            ):
+
+                page_table_split = isinstance(page_table, SplitPrimitiveTensor)
+                partition_split = isinstance(partition, SplitPrimitiveTensor)
+                assert page_table_split == partition_split
+
+                def build_map(t):
+                    return {d: i for i, d in enumerate(t.devices)}
```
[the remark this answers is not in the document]
rsuderman: Done.

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -315,116 +332,209 @@ def write_timestep(
...
+
+                page_table_shards = page_table.shards
+                partition_shards = self.deshard_arg(page_table, partition)
+                seq_shards = self.deshard_arg(page_table, seq_positions)
+                page_ids_shards = self.deshard_arg(page_table, page_ids)
+
+                for i in range(page_table.shard_count):
+                    device = page_table.devices[i]
```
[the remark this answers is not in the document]
rsuderman: Done.

Alex-Vasile (approved):

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -126,6 +128,23 @@ def __init__(
...
+    ) -> torch.Tensor | SplitPrimitiveTensor | ReplicatedTensor:
+        """Unflattens the 2D page tables to 6D tensors."""
+        shards = [state] if self.shard_count == 1 else state.shards
+        shards = [shard.unflatten(1, self.sub_page_dims[pipeline]) for shard in shards]
+
+        if self.shard_count > 1:
+            return SplitPrimitiveTensor(ts=shards, shard_dim=4, devices=state.devices)
+
+        if pipeline > 1:
+            return ReplicatedTensor(ts=shards, devices=state.devices)
```
Groverkss, line 144: I can't really understand what `pipeline` is from this file. Can you document how this is sharded?

Groverkss, line 144: I can't really understand what `pipeline` is from this file. Can you document how this is sharded for tensor parallelism and pipeline parallelism as the class documentation?

Groverkss, line 144: What if we are doing a mix of pipeline and tensor parallelism? Is that not possible? Can we document that?

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -126,6 +128,23 @@ def __init__(
...
+
+        return shards[0]
+
     def unflatten_page_tables(
         self, state: list[torch.Tensor | SplitPrimitiveTensor | ReplicatedTensor]
     ) -> list[torch.Tensor | SplitPrimitiveTensor | ReplicatedTensor]:
```
Groverkss, line 150: Why is the state a list of tensors here?

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -126,6 +128,23 @@ def __init__(
         self.cache_dtype = cache_dtype
         self.attn_dtype = attn_dtype
 
+    def unflatten_page_table(
+        self,
+        state: torch.Tensor | SplitPrimitiveTensor | ReplicatedTensor,
+        pipeline: int,
+    ) -> torch.Tensor | SplitPrimitiveTensor | ReplicatedTensor:
```
Groverkss, line 135: Can you document what the difference between `unflatten_page_table` and `unflatten_page_tables` is? Why do we have multiple page tables? Is one for the unsharded version? This needs more documentation.

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -283,21 +288,32 @@ def read(
...
+        selected = ops.index_select(page_table, 0, subblock_ids.flatten(0, 1))
 
         selected = selected.unflatten(0, blocked_shape[:2])
         key = selected[:, :, 0, :].flatten(1, 2)
         value = selected[:, :, 1, :].flatten(1, 2)
 
         return key, value
 
+    @staticmethod
+    def deshard_like(like, arg):
```
Groverkss, line 307: Needs type annotations

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -283,21 +288,32 @@ def read(
...
+        devices = like.devices
+
+        if isinstance(arg, SplitPrimitiveTensor):
+            assert all([a == b for a, b in zip(devices, arg.devices)])
+
+        assert len(set(arg.devices)) == len(arg.devices)
+        device_map = {d: i for i, d in enumerate(arg.devices)}
+        return [arg.shards[device_map[d]] for d in devices]
```
Groverkss, line 315: Needs more documentation on how sharding is working here

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -315,116 +331,201 @@ def write_timestep(
         Note that this internally loops over the batch size, which cannot be
         dynamic.
         """
-        device = self.device
-        page_tables = self.unflatten_page_tables(state)  # 6D
         pipeline = self.block_to_pipeline_map[transformer_block_index]
         devices = self.pipeline_to_device_map[pipeline]
         transformer_block_count = self.pipeline_to_block_count[pipeline]
         block_offset = self.pipeline_to_block_offset[pipeline]
+        transformer_block_index = transformer_block_index - block_offset
```
Groverkss, line 338: What is block offset here? This needs to be documented.

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -315,116 +331,201 @@ def write_timestep(
...
 
-        # [bs, 1, atten_head_count, attn_head_dim]
-        for idx, cache_partition in enumerate(cache_partitions):
-            # [bs, 1]
-            page_index = seq_positions // self.block_seq_stride
+            assert False
 
-            page_id = ops.gather(page_ids, dim=1, index=page_index.unsqueeze(1))
-            page_offset = (seq_positions % self.block_seq_stride).unsqueeze(1)
+    def write_timestep_shard(
```
Groverkss, line 388: I'm not really sure what is happening here. There are a lot of sharding details but no documentation on what we are doing. I'm guessing you are splitting things into sharded and unsharded versions for each method. Can you document the sharding details and what parts are the unsharded versions? This is very hard to follow.

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -315,116 +331,201 @@ def write_timestep(
...
+        page_offset = (seq_positions % self.block_seq_stride).unsqueeze(1)
+
+        # [1, 1]
+        partitions = torch.tensor(partition_id, device=device).unsqueeze(0)
+        partitions = partitions.repeat(bs, 1)
+
+        index = page_id
+        index = index * transformer_block_count + transformer_block_index
+        index = index * self.cache_partition_count + partitions
+        index = index * self.block_seq_stride + page_offset
```
Groverkss, line 416: Add a comment what parts of the page table we are linearizing.

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -315,116 +331,201 @@ def write_timestep(
...
+
+                    self.write_shard(
+                        page_table=page_table_shards[i],
+                        cache_partition=partition_shards[i],
+                        cache_partition_id=p,
+                        transformer_block_index=transformer_block_index,
+                        transformer_block_count=transformer_block_count,
+                        page_ids=page_ids_shards[i],
+                    )
+                continue
```
Groverkss, line 490: Needs details on how this is sharded, still hard to understand. I'm also not sure why these sharding details are leaking into the PagedAttention layer, PagedAttention should simply work on a single shard and these details should be outside.

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -315,116 +331,201 @@ def write_timestep(
...
-        # frame.
-        # The subblock slab is organized as:
-        #   [page, attn_layer, cache_partition]
-        # Where the cache line can be 0 (k) or 1 (v).
-        subblock_table = page_table.flatten(start_dim=0, end_dim=2)
-        pipeline = self.block_to_pipeline_map[transformer_block_index]
+        index = page_ids
+        index = index * transformer_block_count + transformer_block_index
+        index = index * self.cache_partition_count + cache_partition_id
+        index = index.flatten(0, 1)
```
Groverkss, line 514: Please keep the comments, they are important here.

Groverkss (changes requested):

sharktank/sharktank/layers/paged_attention.py
Alex-Vasile, line 307:
```suggestion
    def deshard_like(arg, like):
```
I think the arguments should be flipped, similar to `reshard_like`.

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -283,21 +288,32 @@ def read(
...
         return key, value
 
+    @staticmethod
+    def deshard_like(like, arg):
+        devices = like.devices
+
+        if isinstance(arg, SplitPrimitiveTensor):
+            assert all([a == b for a, b in zip(devices, arg.devices)])
+
+        assert len(set(arg.devices)) == len(arg.devices)
```
Alex-Vasile, line 313: Rereading this, I don't know when we would want to support duplicate devices, these are logical devices not physical ones.

This needs a similar assert but for `devices` as the current implementation will silently fail when `like` has duplicate devices.

Alex-Vasile (changes requested):
Came back to this with fresh eyes, and I think there's two further issues.

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -367,64 +227,622 @@ def write_timestep(
...
+        assert len(state) == 1
+        assert state[0].shard_count == len(self.caches)
+
+        state = [
+            cache.unshard_state([shard])[0]
+            for cache, shard in zip(self.caches, state[0].shards)
+        ]
+        state = SplitPrimitiveTensor(ts=state, shard_dim=4, devices=self.devices)
+
+        return [ops.unshard(state)]
```
Alex-Vasile, line 318: Why accept and return a list of length 1 rather than a value directly?

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -367,64 +227,622 @@ def write_timestep(
...
+
+    def unflatten_page_table(
+        self, state: List[SplitPrimitiveTensor]
+    ) -> List[SplitPrimitiveTensor]:
+        assert False
```
Alex-Vasile, line 323: This could use an explanation as to why it should never be called.

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -367,64 +227,622 @@ def write_timestep(
...
+        state: List[SplitPrimitiveTensor],
+        *,
+        transformer_block_index: int,
+        page_ids: ReplicatedTensor,
+    ):
+        assert len(state) == 1
+        assert state[0].shard_count == self.shard_count
+
+        for device in state[0].devices:
+            assert device in page_ids.devices
```
Alex-Vasile, line 336: Do we not also want to make sure they're in the correct order?

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -367,64 +227,620 @@ def write_timestep(
...
+        cache_dtype: torch.dtype = torch.float32,
+        device: Optional[torch.device] = None,
+    ):
+        caches = []
+        for i in range(shard_count):
+            start = i * attn_head_count // shard_count
+            end = (i + 1) * attn_head_count // shard_count
+            sharded_attn_head_count = end - start
+
+            cache = KVCache(
```
Alex-Vasile, line 251: This should be getting passed devices[i], otherwise placement won't be correct.

Alex-Vasile (changes requested):

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -64,211 +44,81 @@ def __init__(
...
+            )
+        ]
 
-            block_min = self.pipeline_to_block_offset[pipeline]
-            block_sz = self.pipeline_to_block_count[pipeline]
-            block_max = block_min + block_sz
-            selected = page_table[:, block_min:block_max, ...]
+        # If we have explicit devices we should attach device information:
+        if self.devices is not None:
+            tensors = [ReplicatedTensor(ts=[t], devices=self.devices) for t in tensors]
```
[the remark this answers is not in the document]
rsuderman: There will be no transferring. This is purely affixing the existing placement information that would be attached. Allocation is used either during eager or tracing as a sample input. If `devices` is set this doesn't matter.

sharktank/sharktank/layers/paged_attention.py
rsuderman: We should not assume that there is only one state vector. The cache partitions are not guaranteed to be the same size. We should actually consider having two separate caches for each cache partition.
 
This was also the already defined interface so we should avoid changing interface behavior significantly during refactors.

sharktank/sharktank/layers/paged_attention.py
rsuderman: `ReplicatedTensor` has no behavioral different by device ordering as each tensor should contain identical data. We just need to grab the tensor placed on the correct device.

sharktank/sharktank/layers/paged_attention.py
rsuderman: No, `SplitPrimitiveTensor` does not rely on the `ReplicatedTensor` device placement and afixes the device placement information itself. It purely requires that it is consist with its own tracking / device placement of the tensors.

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -64,211 +44,81 @@ def __init__(
...
-                self.head_count_kv,
-                self.attn_head_dim,
-            ]
-        )
+        self.page_slab_flat_dims = math.prod(self.sub_page_dims)
 
-        flat_sharded_page_tables = []
-        for pipeline in range(self.pipeline_count):
-            devices = self.pipeline_to_device_map[pipeline]
+    def allocate(self, page_count: int) -> List[torch.Tensor]:
```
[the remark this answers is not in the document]
rsuderman: Done.

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -367,64 +227,622 @@ def write_timestep(
...
+        cache_dtype: torch.dtype = torch.float32,
+        block_to_pipeline_map: list[int],
+        pipeline_to_device_map: list[list[int]],
+        device: Optional[torch.device] = None,
+    ):
+        assert transformer_block_count == len(block_to_pipeline_map)
+
+        pipeline_count = len(pipeline_to_device_map)
+
+        transformer_block_map = {}
```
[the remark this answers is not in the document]
rsuderman: Done.

sharktank/sharktank/layers/paged_attention.py
rsuderman: Removed instead. It was leftover during development.

sharktank/sharktank/layers/paged_attention.py
Alex-Vasile: Am I missing something? It doesn't look like the downstream code performs that grabbing. It zips them in positional order, rather than based on matching device.

rsuderman: Updated for missing indirection

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -367,64 +227,614 @@ def write_timestep(
...
+            assert device in page_ids.devices
+
+        page_id_map = {d: i for i, d in enumerate(page_ids.devices)}
+        page_id_shards = [page_ids.shards[page_id_map[d]] for d in state[0].devices]
+
+        shards = []
+        for shard_state, cache, shard_page_ids in zip(
+            state[0].shards, self.caches, page_id_shards
+        ):
+            read = cache.read(
```
Alex-Vasile, line 340: `KVCache.read` returns `(key, value)`. so shards will be `[(k,v)_0, (k,v)_1, ...]`. What is like 349 below doing when it's indexing into each (k, v) based on `i in range(self.cache_partition_count)`?

Alex-Vasile (approved):

sharktank/sharktank/layers/paged_attention.py
rsuderman: Discussed offline.

PR #1407, opened 2025-05-07
[tuner] Improve support for convolutions

rkayaith:
This improves support for convolutions in the tuner by:
- generalizing `ConvolutionOpInterfaceParser` to support parsing `linalg.generic` ops, as well as additional layouts. However since constraint generation for more layouts is untested, only `nhwc_hwcf` and the generic equivalent are considered valid.
- supporting generating the padding lowering attribute for convolutions

### DIFF
--- a/sharktuner/sharktuner/dispatch_parser.py
+++ b/sharktuner/sharktuner/dispatch_parser.py
@@ -96,41 +96,50 @@ class ContractionOpInterfaceParser(DispatchParser):
         )
 
 
-# TODO(Max191): Support more convolution types. Only NHWC convs are supported.
 class ConvolutionOpInterfaceParser(DispatchParser):
     def __init__(self, root_op: ir.Operation):
         super().__init__(root_op)
-        self.supported_ops = ["linalg.conv_2d_nhwc_hwcf"]
 
     def has_valid_root_op(self) -> bool:
         root_op = self.get_root_op()
-        if not linalg.isa_convolution_op(root_op):
-            return False
-
-        return root_op.name in self.supported_ops
+        return linalg.isa_convolution_op(root_op)
 
     def get_problem_size(self) -> ProblemSize:
         root_op = self.get_root_op()
-        lhs_type = ir.RankedTensorType(root_op.operands[0].type)
-        rhs_type = ir.RankedTensorType(root_op.operands[1].type)
-        res_type = ir.RankedTensorType(root_op.operands[2].type)
-        dim_info = ConvDimInfo.from_rhs_res(rhs_type, res_type)
+
+        def find_iter_dim_size(iter_dim: int, operand: int):
+            operand_type = root_op.operands[operand].type
+            indexing_map = linalg.get_indexing_maps(root_op)[operand]
+            tensor_dim = list(indexing_map.value.results).index(
+                ir.AffineExpr.get_dim(iter_dim)
+            )
+            return operand_type.shape[tensor_dim]
+
         convolution_dims = linalg.infer_convolution_dimensions(root_op)
         assert convolution_dims, "no convolution dimensions"
+        contraction_dims = ContractionDimensions(
+            batch=list(convolution_dims.depth),
+            m=list(convolution_dims.batch) + list(convolution_dims.output_image),
+            n=list(convolution_dims.output_channel),
+            k=list(convolution_dims.filter_loop) + list(convolution_dims.input_channel),
+        )
+        # Parallel dimension sizes come from the output operand, and reduction
+        # dimension sizes come from filter.
+        matmul_size = ContractionSizes(
+            B=[find_iter_dim_size(d, operand=2) for d in contraction_dims.batch],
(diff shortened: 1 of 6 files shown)


bangtianliu (commented):
Overall, it looks good—I’ll review it more thoroughly later.

For now, I have one question: which convolution ops does the tuner support besides `conv_2d_nhwc_hwcf`?

rkayaith:
> For now, I have one question: which convolution ops does the tuner support besides `conv_2d_nhwc_hwcf`?

Currently I'm only using the tuner on `nhwc_hwcf`, but represented as `linalg.generic`, so only the parser needs changes to support that. Next I'm looking at group convolutions, which needs more changes in the tuner (I'll be asking for help with that soon 😄). And after that, probably 3D convs.

bangtianliu:
> > For now, I have one question: which convolution ops does the tuner support besides `conv_2d_nhwc_hwcf`?
> 
> Currently I'm only using the tuner on `nhwc_hwcf`, but represented as `linalg.generic`, so only the parser needs changes to support that. Next I'm looking at group convolutions, which needs more changes in the tuner (I'll be asking for help with that soon 😄). And after that, probably 3D convs.

Then how should we reflect that in the tuner? The current PR gives the impression that the tuner supports all convolution operations with different layouts.

rkayaith:
> Then how should we reflect that in the tuner? The current PR gives the impression that the tuner supports all convolution operations with different layouts.

If the configuration generator doesn't support something, I suppose it would either error out or produce no configs. I can add an error for the latter case.
(imo it would be best for any assumptions being made in the constraints generation to also be validated there, but I'm not familiar enough with the code to identify those places).

bangtianliu:
> If the configuration generator doesn't support something, I suppose it would either error out or produce no configs. I can add an error for the latter case. (imo it would be best for any assumptions being made in the constraints generation to also be validated there, but I'm not familiar enough with the code to identify those places).

The tuner might either crash or proceed with tuning unnecessarily, which wastes time (I'm not sure, as I haven't tested it with other conv ops). But ideally, we should catch this case early to terminate the tuner. The contraction dimension information can often indicate the layout, so it can be used for filtering. Alternatively, you can just add a TODO for now — I can follow up with a separate PR to handle it properly.

Max191:
Fly by comment for right now. I'll do a full review later.

> Then how should we reflect that in the tuner?

We should be testing everything that we support, so whatever we decide on, we need to also add tests for it. The simplest thing for now would be to add some additional artificial constraints in `has_valid_root_op`, and we can slowly relax those as we test and support more types of layouts. This way we can reduce the size of each PR, which makes reviewing and landing easier.

We don't need to test every single permutation of layouts (a few key layouts that we regularly see would be good), but we want a couple representative tests each for things like 2D convs, 3D convs, group convs, etc. as we support each type.

For this PR, I think let's just support the 2D layouts. A good place to start for which layouts to test is in the IREE tests for conv_to_im2col (I see you already have some tests, which is great!): https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Dialect/LinalgExt/Transforms/test/conv_to_im2col.mlir

sharktuner/tests/dispatch_parser_test.py
```diff
@@ -247,6 +250,82 @@ def test_get_conv_operation(tuner_ctx: common.TunerContext) -> None:
...
+           root_op,
+           dilations = dense<1> : tensor<2xi64>, strides = dense<1> : tensor<2xi64>
+        } ins(%arg0, %arg1: tensor<2x10x10x7x4xf32>, tensor<7x16x3x3x4xf32>)
+          outs(%arg2: tensor<2x8x8x7x16xf32>) -> tensor<2x8x8x7x16xf32>
+        return %0 : tensor<2x8x8x7x16xf32>
+      }
+    }
+    """
+    module = ir.Module.parse(module_str, context)
+    module = ir.Module.parse(module_str, context)
```
rkayaith, line 279: note to self: redundant line

sharktuner/sharktuner/dispatch_parser.py
```diff
@@ -96,41 +96,64 @@ def get_problem_size(self) -> ProblemSize:
...
 
     def has_valid_root_op(self) -> bool:
         root_op = self.get_root_op()
         if not linalg.isa_convolution_op(root_op):
             return False
-
-        return root_op.name in self.supported_ops
+        convolution_dims = linalg.infer_convolution_dimensions(root_op)
+        assert convolution_dims, "no convolution dimensions"
+        # Only allow 'nhwc_hwcf' convs.
```
bangtianliu, line 109: Maybe also add a comment like "TODO: support more conv ops" in case we miss this.

bangtianliu (approved):
Thanks for improving the tuner.
LGTM, it seems that you need to do a rebase.

bangtianliu:
nvm lol

rkayaith:
Thanks for the reviews!

PR #1409, opened 2025-05-08
Add PPL path for TP2 sharded Llama 8B

Alex-Vasile:


Alex-Vasile:
PPL tests run and pass numerics, when run on their own rather than in parallel. 

The torch tests were failing in parallel before this PR and are tracked in #1410.

archana-ramalingam (approved):

PR #1415, opened 2025-05-08
[shark Tank] Implementation of sharded_gather

### LINKED ISSUE #1396: Sharded_gather for shard collection to a device
we can add a sharded_gather based on the disscusion @sogartar  [here](https://github.com/nod-ai/shark-ai/pull/1366#discussion_r2075662435)

oyazdanb:
Fixes #1396
Implementation of sharded_gather ops
this op is not an override for torch.gather which works based on indexing. It focuses on gathering SplitPrimitiveTensor and ReplicatedTensor shards across devices into a single device.

### DIFF
--- a/sharktank/sharktank/ops/default_impls.py
+++ b/sharktank/sharktank/ops/default_impls.py
@@ -603,7 +603,7 @@ def sharded_cat_unsharded(maybe_sharded):
 
 @sharded_gather.override(Tensor)
 def sharded_gather_unsharded(input):
-    return unbox_tensor(input)
+    return input
 
 
 @sharded_sum.override(Tensor)
--- a/sharktank/sharktank/ops/default_impls.py
+++ b/sharktank/sharktank/ops/default_impls.py
@@ -585,6 +585,11 @@ def sharded_cat_unsharded(maybe_sharded):
     return unbox_tensor(maybe_sharded)
 
 
+@sharded_gather.override(Tensor)
+def sharded_gather_unsharded(input):
+    return unbox_tensor(input)
+
+
 @sharded_sum.override(Tensor)
 def sharded_sum_unsharded(maybe_sharded):
     return unbox_tensor(maybe_sharded)
--- a/sharktank/sharktank/ops/default_impls.py
+++ b/sharktank/sharktank/ops/default_impls.py
@@ -603,7 +603,7 @@ def sharded_cat_unsharded(maybe_sharded):
 
 @sharded_gather.override(Tensor)
 def sharded_gather_unsharded(input):
-    return input
+    return [input]
 
 
 @sharded_sum.override(Tensor)
--- a/sharktank/sharktank/ops/sharded_impls.py
+++ b/sharktank/sharktank/ops/sharded_impls.py
@@ -1384,9 +1384,7 @@ def sharded_cat_unsharded(tensor: SplitPrimitiveTensor):
 
 
 @sharded_gather.override(SplitPrimitiveTensor)
-def sharded_gather_split(
-    input: SplitPrimitiveTensor, root_rank: int, dim: int | None
-) -> List[Tensor]:
+def sharded_gather_split(input: SplitPrimitiveTensor, root_rank: int) -> List[Tensor]:
 
     shard_ts = [
         (
@@ -1400,9 +1398,7 @@ def sharded_gather_split(
 
 
 @sharded_gather.override(ReplicatedTensor)
-def sharded_gather_replicated(
-    input: ReplicatedTensor, root_rank: int, dim: int | None
-) -> List[Tensor]:
+def sharded_gather_replicated(input: ReplicatedTensor, root_rank: int) -> List[Tensor]:
     shard = input.shards[root_rank]
     return [shard.as_torch().clone() for _ in range(input.shard_count)]
 
--- a/sharktank/sharktank/ops/sharded_impls.py
+++ b/sharktank/sharktank/ops/sharded_impls.py
@@ -1383,24 +1383,24 @@ def sharded_cat_unsharded(tensor: SplitPrimitiveTensor):
     return torch.cat(shard_ts, dim=tensor.shard_dim)
 
 
-@sharded_gather.override(SplitPrimitiveTensor)
-def sharded_gather_split(input: SplitPrimitiveTensor, root_rank: int) -> List[Tensor]:
-
-    shard_ts = [
-        (
-            transfer_to_logical_device(shard, input.devices[root_rank])
-            if i != root_rank
-            else barrier_on_logical_device(shard, input.devices[root_rank])
-        )
-        for i, shard in enumerate(input.shards)
-    ]
-    return shard_ts
-
-
-@sharded_gather.override(ReplicatedTensor)
-def sharded_gather_replicated(input: ReplicatedTensor, root_rank: int) -> List[Tensor]:
-    shard = input.shards[root_rank]
-    return [shard.as_torch().clone() for _ in range(input.shard_count)]
+@sharded_gather.override(IsOfType(SplitPrimitiveTensor, ReplicatedTensor))
+def sharded_gather_split(
+    input: SplitPrimitiveTensor | ReplicatedTensor, root_rank: int
+) -> List[Tensor]:
+    # if input is SplitPrimitiveTensor
+    if type(input) == SplitPrimitiveTensor:
+        shard_ts = [
+            (
+                transfer_to_logical_device(shard, input.devices[root_rank])
+                if i != root_rank
+                else barrier_on_logical_device(shard, input.devices[root_rank])
+            )
+            for i, shard in enumerate(input.shards)
+        ]
+        return shard_ts
+    else:
+        shard = input.shards[root_rank]
+        return [shard.as_torch().clone() for _ in range(input.shard_count)]
 
 
 def _sharded_sum_sharded(tensor: ShardedTensor, root_rank: int) -> Tensor:
--- a/sharktank/sharktank/ops/sharded_impls.py
+++ b/sharktank/sharktank/ops/sharded_impls.py
@@ -1385,29 +1385,25 @@ def sharded_cat_unsharded(tensor: SplitPrimitiveTensor):
 
 @sharded_gather.override(SplitPrimitiveTensor)
 def sharded_gather_split(
-    input: SplitPrimitiveTensor, device_ordinal: int, concat: bool = False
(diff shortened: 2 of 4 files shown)


sharktank/tests/ops/sharded_test.py
```diff
@@ -1657,6 +1657,56 @@ def testSoftmaxSplit(self):
...
+        shards = [torch.rand(2, 5, 4) for _ in range(3)]
+        tensor_sp = SplitPrimitiveTensor(ts=shards, shard_dim=shard_dim)
+
+        # Gather without concat
+        actual = ops.sharded_gather(tensor_sp, device_ordinal=0, concat=False)
+        self.assertEqual(len(actual), 3)
+        self.assertEqual(actual[0].shape, (2, 5, 4))
+
+        for i, g in enumerate(actual):
+            torch.testing.assert_close(g, shards[i])
```
Alex-Vasile, line 1676: More descriptive variable names than `g` will help long term.

Alex-Vasile (changes requested):

sharktank/sharktank/ops/default_impls.py
```diff
@@ -601,6 +601,11 @@ def sharded_cat_unsharded(maybe_sharded):
     return unbox_tensor(maybe_sharded)
 
 
+@sharded_gather.override(Tensor)
+def sharded_gather_unsharded(input):
+    return unbox_tensor(input)
```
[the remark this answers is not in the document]
sogartar: Yes. Ultimately we should avoid working with unwrapped torch tensors, but only when calling torch functions.
This unboxing for example would be problematic if we have sharded quantized tensors as this would dequantize the tensor.

sharktank/sharktank/ops/signatures.py
```diff
@@ -1214,6 +1215,25 @@ def _sharded_cat_trampoline(d: SignatureDispatcher, maybe_sharded: AnyTensor):
         d.fail(tensors)
 
 
+@overridable(is_trivially_replicable=False)
+def sharded_gather(input: AnyTensor, device_ordinal: int, concat: bool) -> AnyTensor:
```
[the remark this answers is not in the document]
sogartar: This op should have the same parameters as the `all_gather` op.
```
def all_gather(maybe_sharded: AnyTensor, *, dim: int | None = None) -> AnyTensor:
```

For a sharded split tensor `dim` can be omitted and default to the sharded dimension of the input tensor. For unreduced tensor there is no natural dimension. This operation have similar semantics to [torch.cat](https://docs.pytorch.org/docs/stable/generated/torch.cat.html), where the default concatenation dimension is 0. But if we add such a default it would be confusing that this is different for split and unreduced tensors. I say we allow the default only for a split tensor.

Aside from these shared parameters it should have the destination as the shard rank within the sharded tensor. It is better to call this `root_rank`. At some point we may add support of doing a gather where the destination device is not one of the input tensor devices, then it would be appropriate to call this argument `device_ordinal`.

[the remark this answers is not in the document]
sogartar: Regarding the return type I am not sure whether it should return a list of tensors or a concatenated tensor. There are probably scenarios where you would want them as separate tensors and then you would have to split.
Right now underneath we are transferring the tensors one-by-one but at some point it may be more preformat to write into a single tensor directly.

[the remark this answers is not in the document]
sogartar: For the sake of symmetry `all_gather` and `sharded_gather` should differ only on the `root_rank` argument, but then in the case of doing a reduction afterwards you would want the tensors as a list. If `sharded_gather` returned a concatenation it would be less useful as a building block in that scenario.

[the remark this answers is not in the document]
sogartar: From the point of view of compiler optimization if you return a list of tensors and then concatenate it, it is safe to assume that this should be optimized by fusing and transferring the pieces into a single block, thus avoiding a useless copy. Also currently under the exported MLIR does transfer into distinct tensors, so even if the compiler is lacking such an optimization now it would not matter.
If we return a list here it should not hurt future performance.

sharktank/sharktank/ops/default_impls.py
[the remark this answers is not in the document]
oyazdanb: fixed; now it returns only list.

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -1383,7 +1383,32 @@ def sharded_cat_unsharded(tensor: SplitPrimitiveTensor):
...
 
 
-# Sharded sum.
+@sharded_gather.override(SplitPrimitiveTensor)
+def sharded_gather_split(
+    input: SplitPrimitiveTensor, device_ordinal: int, concat: bool = False
+) -> Union[List[Tensor], Tensor]:
+    shard_ts = [
+        (
+            transfer_to_logical_device(shard, device_ordinal)
```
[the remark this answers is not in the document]
oyazdanb: fixed.

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -1383,7 +1383,32 @@ def sharded_cat_unsharded(tensor: SplitPrimitiveTensor):
...
+            if i != 0
```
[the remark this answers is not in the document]
oyazdanb: fixed

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -1383,7 +1383,32 @@ def sharded_cat_unsharded(tensor: SplitPrimitiveTensor):
...
+            else barrier_on_logical_device(shard, device_ordinal)
```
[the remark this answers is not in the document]
oyazdanb: fixed.

sharktank/tests/ops/sharded_test.py
oyazdanb: fixed.

sharktank/tests/ops/sharded_test.py
```diff
@@ -1657,6 +1657,41 @@ def testSoftmaxSplit(self):
...
+        root_rank = 0
+        shards = [torch.rand(2, 5, 4) for _ in range(shard_count)]
+        tensor_sp = SplitPrimitiveTensor(ts=shards, shard_dim=shard_dim)
+
+        actual = ops.sharded_gather(tensor_sp, root_rank=root_rank)
+        self.assertEqual(len(actual), 3)
+        self.assertEqual(actual[0].shape, (2, 5, 4))
+
+        for i, shard in enumerate(actual):
+            ops.equal(shard, shards[i])
```
Alex-Vasile, line 1676: This needs an assert. ops.equal will not throw an error, it only returns true or false.

Alex-Vasile (changes requested):

sharktank/sharktank/ops/signatures.py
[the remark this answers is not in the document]
sogartar: I we will be returning a list then the `dim` argument should be removed.

sharktank/tests/ops/sharded_test.py
oyazdanb: fixed.

sharktank/sharktank/ops/signatures.py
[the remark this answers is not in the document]
oyazdanb: removed dim.

sharktank/sharktank/ops/default_impls.py
```diff
@@ -601,6 +601,11 @@ def sharded_cat_unsharded(maybe_sharded):
     return unbox_tensor(maybe_sharded)
 
 
+@sharded_gather.override(Tensor)
+def sharded_gather_unsharded(input):
+    return input
```
Alex-Vasile, line 606: Does this need a default implementation for unsharded?

If so, I think it should be returning `[input]` so the return type is consistent and can be used interchangable.

Alex-Vasile: @sogartar Thoughts?

sharktank/sharktank/ops/signatures.py
```diff
@@ -1214,6 +1215,25 @@ def _sharded_cat_trampoline(d: SignatureDispatcher, maybe_sharded: AnyTensor):
         d.fail(tensors)
 
 
+@overridable(is_trivially_replicable=False)
+def sharded_gather(input: AnyTensor, root_rank: int) -> AnyTensor:
```
Alex-Vasile, line 1219:
```suggestion
def sharded_gather(input: AnyTensor, root_rank: int) -> list[AnyTensor]:
```

Same change for the signature of the trampoline.

sharktank/tests/ops/sharded_test.py
```diff
@@ -1657,6 +1657,41 @@ def testSoftmaxSplit(self):
         torch.testing.assert_close(expected_result, ops.unbox_tensor(actual_result))
 
 
+class ShardedGatherTest(unittest.TestCase):
```
Alex-Vasile, line 1660: Can you move this up between Sigmoid and Softmax to keep things alphabetical.

Alex-Vasile (changes requested):

sharktank/sharktank/ops/default_impls.py
sogartar: Yes, lets return a list for consistency.

sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -1383,7 +1383,24 @@ def sharded_cat_unsharded(tensor: SplitPrimitiveTensor):
     return torch.cat(shard_ts, dim=tensor.shard_dim)
 
 
-# Sharded sum.
+@sharded_gather.override(SplitPrimitiveTensor)
```
sogartar, line 1386: The `SplitPrimitiveTensor` and `UnreducedTensor` can share an implementation with
```
@sharded_gather.override(IsOfType(SplitPrimitiveTensor, UnreducedTensor))
```

oyazdanb: fixed.

sharktank/tests/ops/sharded_test.py
oyazdanb: fixed.

sharktank/sharktank/ops/signatures.py
oyazdanb: fixed.

Alex-Vasile (approved):

PR #1419, opened 2025-05-09
Use numpy.argpartition instead of sfnp for select_top_k

pravg-amd:
Using numpy.argpartition results in speedup of select_top_k when compared to sfnp.

### DIFF
--- a/shortfin/python/shortfin_apps/llm/components/token_selection_strategy/sampler.py
+++ b/shortfin/python/shortfin_apps/llm/components/token_selection_strategy/sampler.py
@@ -45,10 +45,10 @@ class Sampler:
         """
         partitioned_tokens = np.argpartition(logits, k)
         # Slice off all axes except the last one
-        zero_indices = [0] * (len(partitioned_tokens.shape) - 1)
+        zero_indices = (0,) * (partitioned_tokens.ndim - 1)
 
         # Obtain tokens & values from partition
-        top_tokens = partitioned_tokens[*zero_indices, k:].tolist()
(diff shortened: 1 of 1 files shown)


stbaione:
This works as a temporary patch to make `sampler.select_top_k` faster. That does make the `top_p` path faster as well, however, I think the better move would be to replace `xtensor` with `numpy` in all paths.

If you check out the `sample_logits` functions in [beam_search_beam](https://github.com/nod-ai/shark-ai/blob/main/shortfin/python/shortfin_apps/llm/components/token_selection_strategy/beam_search_token_selection_strategy.py#L34) and [greedy beam](https://github.com/nod-ai/shark-ai/blob/main/shortfin/python/shortfin_apps/llm/components/token_selection_strategy/greedy_token_selection_strategy.py#L22), you see that we receive our logits as a `device_array`, then call:

```text
(top_k, top_p) is (None, None) => self.sampler.select_greedy
(top_k, top_p) is (int, None) => self._sample_logits_top_k
(top_k, top_p) is (None, float) => self.sampler.select_top_k, self._to_softmax, self._sample_logits_top_p
```
`BeamSearchBeam` calls similar, but slightly different functions.

Basically, if we want a quick patch to `sampler.select_top_k` then this works.
If we want a patch to `top_p`, then we have a couple more spots to update.
But, I think the cleanest thing to do would be to replace `xtensor` with numpy throughout the post-processing.

stbaione (approved):
This is fine as an incremental change. See comment above as to the broader changes that should be made to optimize `top_p` and other `sampling/selection` algorithms further.

PR #1421, opened 2025-05-09
Restore separate threads for prefill and decode

rsuderman:
Threading pool actually slowed existing throughput if enabling multiple threads on decode. Simplified to have separate threads between prefill and decode. This should allow pre-work to decouple streams between the two workloads.


stbaione (approved):
LGTM, as soon as CI passes

PR #1422, opened 2025-05-09
Add `topk` to `export_paged_llm_v1`

stbaione:
Stacked on: #1454 

This PR adds a `topk` export to `export_paged_llm_v1`. You can export varying sizes of topk, through the cli:

```bash
--top-k 1 # argmax
--top-k 4 # topk_k4
```

This will cause the `prefill` and `decode` functions to return the top `k` logit indices, along with a subset of the value logits.

We also add a `_split_topk` implementation to make execution faster. The results still leave a lot to be desired:

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>

<div style='direction:ltr'>

Function | Input | Time
-- | -- | --
Topk_k2 | <1x1x131072,   f16> | 179ms
Topk_k4 | <1x1x131072,   f16> | 194ms
Topk_k8 | <1x1x131072,   f16> | 260ms
Topk_k16 | <1x1x131072,   f16> | 533ms
Topk_k32 | <1x1x131072,   f16> | 1538ms

</div>

</body>

</html>

But, it's better than just exporting `topk` by itself. If we did the normal `ops.topk`, the execution time for the `mistral` logits, with `k==4` is in the millions of miliseconds.

Issue for this perf opened [here](https://github.com/iree-org/iree/issues/20772)


sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -465,16 +465,50 @@ def _(
         ):
             return ops.argmax(logits, axis, chunk_size=hp.context_length // 1024)
 
+    def generate_top_k(k: int):
+        dtype = llama_config.activation_dtype
+        if "float8" in str(dtype) or dtype == torch.bfloat16:
+            dtype = torch.float16
```
IanNod, line 471: Why is this being set to fp16? Fp32 would be the safest lossless dtype to use

sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -465,16 +465,50 @@ def _(
...
+
+        logits: torch.Tensor = torch.empty(
+            1,
+            1,
+            131072,
```
IanNod, line 476: Is there a reason this is hardcoded? Should just be hp.context_length, no?

sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -465,16 +465,50 @@ def _(
...
+        )
+        def _(
+            _,
+            logits=logits,
+            k=k,
+            dim=-1,
+            largest=True,
+            _sorted=True,
+        ):
+            return ops.topk(logits, k, dim, largest, _sorted, chunk_size=1024)
```
IanNod, line 495: There is no real way to modify chunk_size through the generate_top_k function. Can that be an input arg?

sharktank/sharktank/examples/export_paged_llm_v1.py
stbaione: I think we originally went with `fp16` to export the `prefill/decode` logits, because it's just what we default to. In this PR, it's just to match that expectation

sharktank/sharktank/examples/export_paged_llm_v1.py
stbaione: Yep, thanks for catching this. Was hardcoding to generate some mlir test cases and forgot to remove it... updated

sharktank/sharktank/examples/export_paged_llm_v1.py
stbaione: Updated

shortfin/python/shortfin_apps/llm/components/config_struct.py
```diff
@@ -150,8 +150,9 @@ class ModelParams:
     # TODO(stbaione): Update comment when export with `top_k > 1` is enabled.
```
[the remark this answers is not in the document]
stbaione: Removed TODO

sharktank/tests/ops/ops_test.py
```diff
@@ -417,6 +418,87 @@ def test(self):
         assert ops.equal(actual, expected)
 
 
+class TestTopK(unittest.TestCase):
```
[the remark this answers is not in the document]
stbaione: It seems like the standard is to keep the sharded tests separate from the unsharded ones

sharktank/sharktank/ops/signatures.py
```diff
@@ -1519,7 +1519,14 @@ def _sum_trampoline(
...
 
 @overridable
-def topk(tensor, k: int, dim: int, largest: bool, sorted: bool) -> AnyTensor:
+def topk(
+    tensor,
+    k: int,
+    dim: int,
+    largest: bool,
+    sorted: bool,
+    chunk_size: Optional[int] = None,
```
Alex-Vasile, line 1528: The sharded versions haven't been updated with this arg.

stbaione: I added a line in signature to call `top_k` without `chunk_size`, if the tensor is of type `SplitPrimitiveTensor`

sharktank/sharktank/ops/default_impls.py
```diff
@@ -649,12 +649,56 @@ def squeeze_default(tensor, dim: Optional[int] = None) -> AnyTensor:
...
+    if chunk_size is None:
+        result = torch.topk(
+            unbox_tensor(tensor), k=k, dim=dim, largest=largest, sorted=sorted
+        )
+        return result.values, result.indices
+
+    return _split_topk(tensor, k, dim, largest, sorted, chunk_size)
+
+
+def _split_topk(
```
IanNod, line 668: There are many topk algorithms that are likely more performant than this. If we are going to address split-k performance at sharktank level, could you add a todo to investigate others like bucket selection?

sharktank/sharktank/ops/signatures.py
IanNod: Wouldn't we still want this functionality for SplitPrimitiveTensor's? Could we at least add a todo to reconcile that later?

sharktank/sharktank/examples/export_paged_llm_v1.py
IanNod: I think the expectation would be to return the same type as the activations we specify. I'm guessing you are doing this to work around some non-supported types in numpy when we transfer data back to host? If so there are workarounds to view the data as a different type

sharktank/sharktank/ops/default_impls.py
stbaione: Updated

sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -308,14 +308,20 @@ def _(model, tokens, seq_lens, seq_block_ids, cs):
...
                 logits = ops.elementwise(torch.log, ops.softmax(logits, dim=-1))
 
-            if args.top_k is None:
+            top_k = args.top_k
+            if top_k is None:
                 return logits
 
-            if args.top_k == 1:
+            if top_k == 1:
                 max_logits, indices = model.argmax(
```
rsuderman, line 316: Just make each of these returns. No point to assign just to return afterwards.

sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -444,45 +450,21 @@ def _(
...
```
rsuderman, line 458: See comment about returning rather than falling through for a return.

sharktank/sharktank/models/llm/llm.py
```diff
@@ -153,6 +153,16 @@ def argmax(
...
         return max_logits, indices
 
+    def topk(
+        self,
+        logits: torch.Tensor,
+        k: int,
+        chunk_size: int,
+    ):
+        return ops.topk(
+            logits, k=k, dim=-1, largest=True, sorted=True, chunk_size=chunk_size
```
rsuderman, line 163: I wouldn't bother setting sorted. We should just do that sorting on the CPU. Small sorts are basically free there. Also not all sampling methods will care about sorting so no need to do extra work.

shortfin/python/shortfin_apps/llm/components/config_struct.py
```diff
@@ -147,20 +147,20 @@ class ModelParams:
     # The element type of the attention caches.
     attn_dtype: sfnp.DType = sfnp.float16
 
-    # TODO(stbaione): Update comment when export with `top_k > 1` is enabled.
     # Define the `top_k` kernel that model was exported with.
```
rsuderman, line 150: Rename from exporting `top_k` kernels to `taking top_k of logits on return`. This is the behavior we are actually performing.

rsuderman (changes requested):

sharktank/sharktank/examples/export_paged_llm_v1.py
stbaione: Updated

sharktank/sharktank/examples/export_paged_llm_v1.py
stbaione: Updated

sharktank/sharktank/models/llm/llm.py
stbaione: Updated

shortfin/python/shortfin_apps/llm/components/config_struct.py
stbaione: Updated

stbaione:
@rsuderman had to set `sorted` back to `True` when running `topk`, it caused a compiler error:

```bash
ERROR    app_tests.integration_tests.llm.model_management:model_management.py:539 Compilation failed with code 1
ERROR    app_tests.integration_tests.llm.model_management:model_management.py:540 STDOUT: 
ERROR    app_tests.integration_tests.llm.model_management:model_management.py:541 STDERR: <unknown>:0: error: failed to legalize operation 'torch.constant.int'
<unknown>:0: note: see current operation: %25 = "torch.constant.int"() <{value = 4 : i64}> : () -> !torch.int
<unknown>:0: error: failed to legalize operation 'torch.constant.int'
<unknown>:0: note: see current operation: %22 = "torch.constant.int"() <{value = 3 : i64}> : () -> !torch.int
```

Verified that the same inputs for prefill/decode work in unit tests with `sorted == False`, so will file an issue in IREE

PR #1430, opened 2025-05-12
Shortfin llm numpy postprocessing

stbaione:
Move the post-processing sampling/selection logic entirely from `xtensor` to `numpy`.

### DIFF
--- a/shortfin/python/shortfin_apps/llm/components/token_selection_strategy/beam_group.py
+++ b/shortfin/python/shortfin_apps/llm/components/token_selection_strategy/beam_group.py
@@ -5,6 +5,7 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 import logging
+import numpy as np
 
 from abc import ABC, abstractmethod
 from asyncio import gather
@@ -12,19 +13,10 @@ from dataclasses import dataclass, field
 from typing import Callable, Dict, List, Set
 from uuid import uuid4
 
-import shortfin as sf
-import shortfin.array as sfnp
-
 from .base_token_selection_strategy import DecodeConfig
 from .config import LogitsNormalization
 from .sampler import Sampler
 from ..messages import LlmInferenceExecRequest
-
-from shortfin_apps.utils import (
-    convert_int_to_float,
-    convert_float_to_int,
-    convert_list_to_device_array,
-)
 
 logger = logging.getLogger(__name__)
 
@@ -43,7 +35,7 @@ class Beam(ABC):
     accumulated_normalization: float = 0.0
     last_token: int | None = None
 
-    def apply_temperature(self, logits: sfnp.device_array):
+    def apply_temperature(self, logits: np.array) -> np.array:
         """Apply temperature to the logits of a decode invocation.
 
         Args:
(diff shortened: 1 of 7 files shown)


shortfin/python/shortfin_apps/llm/components/token_selection_strategy/beam_group.py
```diff
@@ -43,35 +35,46 @@ class Beam(ABC):
...
 
         Args:
             temperature (float): Value to use for `temperature`.
         """
         if self.decode_config.temperature == 1.0:
             return logits
-        return sfnp.divide(logits, self.decode_config.temperature)
+        return np.divide(logits, self.decode_config.temperature)
+
+    def _softmax(self, logits: np.array) -> np.array:
```
rsuderman, line 48: Can we change this to take `np.array` or `sfnp.device_array` then perform a conversion from `sfnp` to `np` if necessary? This should prevent us from needing to change the test implementations.

Similarly we can do this for the results (e.g. keep returning `list[int]`.

rsuderman (changes requested):
Can we get a small evaluation verifying that `numpy` releases the GIL during invocations of `topk`. This could be useful for validating that we see a boost when multithreading is enabled.

stbaione:
> Can we get a small evaluation verifying that `numpy` releases the GIL during invocations of `topk`. This could be useful for validating that we see a boost when multithreading is enabled.

Setup a quick [benchmarking script](https://gist.github.com/stbaione/9780996d52695a041dc6a0e713bb012a) to compare numpy/shortfin performance in multi-threaded context, testing `argpartition`. Benchmark results seem to indicate that `numpy` releases the GIL, while `shortfin` doesn't. 

It can be observed that `shortfin's argpartition` benchmark scales linearly with the number of threads.

However, `numpy's argpartition` maintains close to the same performance as threads increases, jumps at the transition from 16 to 32, then stabilizes again.

Results:

Shortfin Results:
|   Threads |   Total Time (s) |   Time per Thread (s) |   Repeitions |
|----------:|-----------------:|----------------------:|-------------:|
|         1 |        0.0813365 |             0.0813365 |           50 |
|         2 |        0.165406  |             0.0827032 |           50 |
|         4 |        0.528919  |             0.13223   |           50 |
|         8 |        1.26031   |             0.157539  |           50 |
|        16 |        2.3192    |             0.14495   |           50 |
|        32 |        5.14513   |             0.160785  |           50 |
|        64 |       11.1559    |             0.174312  |           50 |

Numpy Results:
|   Threads |   Total Time (s) |   Time per Thread (s) |   Repeitions |
|----------:|-----------------:|----------------------:|-------------:|
|         1 |        0.0701847 |            0.0701847  |           50 |
|         2 |        0.0715857 |            0.0357928  |           50 |
|         4 |        0.0720103 |            0.0180026  |           50 |
|         8 |        0.0749917 |            0.00937396 |           50 |
|        16 |        0.0738442 |            0.00461526 |           50 |
|        32 |        0.115939  |            0.0036231  |           50 |
|        64 |        0.117396  |            0.00183431 |           50 |

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/beam_group.py
stbaione: For the results part, do you mean return the same type received? So, if I receive a numpy array, return a numpy array.

If I receive a device_array, do the softmax in numpy, then return a device_array?

stbaione: Added the conversion from `sfnp.device_array` to `np.array`

rsuderman (approved):

PR #1431, opened 2025-05-12
Changes workers flag to shortfin_workers

IanNod:
Flag change for shortfin cli workers flag due to conflicting fastapi's workers flag

### DIFF
--- a/shortfin/python/shortfin_apps/llm/cli.py
+++ b/shortfin/python/shortfin_apps/llm/cli.py
@@ -120,7 +120,7 @@ def add_service_args(parser: argparse.ArgumentParser):
         help="Number of workers to use when running in `offline` mode.",
     )
     parser.add_argument(
-        "--workers",
+        "--shortfin_workers",
         type=int,
         default=1,
         help="Number of workers to use when running in `server` mode.",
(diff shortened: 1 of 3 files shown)


shortfin/python/shortfin_apps/llm/cli.py
```diff
@@ -120,7 +120,7 @@ def add_service_args(parser: argparse.ArgumentParser):
         help="Number of workers to use when running in `offline` mode.",
     )
     parser.add_argument(
-        "--workers",
+        "--shortfin_workers",
```
stbaione, line 123: Can just remove this CLI/config arg, along with `fibers_per_worker`. The only spot that they're referenced now is for a logging statement. They're not doing anything functionally anymore

stbaione (changes requested):

shortfin/python/shortfin_apps/llm/cli.py
IanNod: My understanding is we were planning to still use these flags once we have multiple hip streams implemented. I could remove them now but figured that is less work to plumb through once those are added

stbaione: Gotcha, can leave them in place.

stbaione (approved):
`LGTM`, fyi, you'll need to pass through a cli arg to `server.py` and rewrite the `uvicorn.run(...)` step to use the `import string` method to enable multiple fastapi workers, when you get to that part

PR #1432, opened 2025-05-13
Make improved reservation scheduler with strobing

rsuderman:
Reservations should be made unique to original requests. Prioritizes throughput by ensuring related requests are executed together. Strobing is still used to dispatch work known to the reservation system.

It will attempt to minimize the number of dispatches by maximizing to the ideal size. Any remaining work post-reservation filled will be filled with non-reservation work.

### DIFF
--- a/shortfin/python/shortfin_apps/llm/components/messages.py
+++ b/shortfin/python/shortfin_apps/llm/components/messages.py
@@ -22,7 +22,13 @@ class InferencePhase(Enum):
 class LlmInferenceExecRequest(InferenceExecRequest):
     """Performs a prefill operation."""
 
-    def __init__(self, phase: InferencePhase, input_token_ids: list[int], rid=None, orig_instance_id=None):
+    def __init__(
+        self,
+        phase: InferencePhase,
+        input_token_ids: list[int],
+        rid=None,
+        orig_instance_id=None,
+    ):
         super().__init__()
         self.phase = phase
         self.start_position: int = 0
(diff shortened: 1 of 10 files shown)


shortfin/python/shortfin_apps/llm/components/token_selection_strategy/beam_search_token_selection_strategy.py
```diff
@@ -282,18 +282,21 @@ async def decode(
         )
 
         reservations = beam_group.active_beam_count
-        config.decode_begin_callback(reservations)
+        config.decode_begin_callback(rid=exec_req.orig_instance_id, count=reservations)
```
stbaione, line 285: Will need to update `Greedy`/`MultiGreedy`. Multigreedy is most relevant to our current effort

shortfin/python/shortfin_apps/llm/components/messages.py
```diff
@@ -33,6 +33,7 @@ def __init__(self, phase: InferencePhase, input_token_ids: list[int], rid=None):
         # Unique `instance_id` for token selection strategies that may need
         # to differentiate between an original req and a copy of a req.
         self.instance_id = str(uuid4())
+        self.orig_instance_id = self.instance_id if orig_instance_id is None else orig_instance_id
```
stbaione, line 36: Can you add a comment to explain difference between `orig_instance_id` and `instance_id`?

stbaione (changes requested):

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/beam_search_token_selection_strategy.py
stbaione: (Multigreedy w/ top_p)

rsuderman: Fixed.

shortfin/python/shortfin_apps/llm/components/messages.py
rsuderman: Added a comment.

stbaione (approved):

PR #1424, opened 2025-05-10
[sharktank] Refactor llm sharding 

archana-ramalingam:
Refactor llm sharding 

These changes are part of Deepseek model bringup.

PR #1433, opened 2025-05-13
[sharktank] Refactor ffn_norm and add_residual from ffn blocks

archana-ramalingam:
Refactor `ffn_norm` and `add_residual` out of ffn blocks.

These changes are part of Deepseek model bringup.

Context:
`ffn_norm` and `add_residual` should be applied only once at the end of an MoE/FFN block.
Although `ffn_norm` belongs to the ffn layer, in llms like Deepseek, where multiple ffn blocks are present (shared & routed expert blocks), `ffn_norm` tends to be applied twice, once to each blocks leading to numeric issues. Hence, the refactor resolves this issue.


sharktank/sharktank/layers/ffn_block.py
```diff
@@ -40,42 +36,34 @@ def __init__(
...
-        self.ffn_norm = torch.nn.Identity()
-        self.add_residual = add_residual
+
+        ffn_suffix = ""
+        if theta.optional_tensor("ffn_gate_shexp") is not None:
+            ffn_suffix = "_shexp"
+
+        ffn_gate = "ffn_gate" + ffn_suffix
+        ffn_up = "ffn_up" + ffn_suffix
+        ffn_down = "ffn_down" + ffn_suffix
```
Alex-Vasile, line 46: Same comment as: https://github.com/nod-ai/shark-ai/pull/1424#discussion_r2085777282

sharktank/sharktank/layers/ffn_moe_block.py
```diff
@@ -150,7 +145,7 @@ def forward(
 
         # (self.num_experts, num_tokens)
         router_indices = (
-            ops.reshard_like(torch.arange(num_tokens), router_scores)
+            ops.reshard_like(torch.arange(num_tokens, device=h.device), router_scores)
```
Alex-Vasile, line 148:
```suggestion
            ops.reshard_like(torch.arange(num_tokens, device=router_scores.device), router_scores)
```
?

sharktank/sharktank/layers/ffn_block.py
archana-ramalingam: Fixed [here](https://github.com/nod-ai/shark-ai/blob/deepseek-refactor-3/sharktank/sharktank/models/llm/llm.py#L84).

sharktank/sharktank/layers/ffn_moe_block.py
archana-ramalingam: Reverted back to `h.device`, because if `router_scores` is sharded, the test fails as `router_scores.device` is not valid.

archana-ramalingam: Nvm, revert fails for same reason too.

sogartar: `h` can also be sharded. in the feature branch `deepseek-v3`. I added `device` to the `InferenceTensor`.

sogartar: It is a bit confusing as `InferenceTensor.device` mimics `torch.Tensor.device`, but we also have `ShardedTensor.devices`, which returns device ordinals for our export.

Alex-Vasile: I think for now it would make more sense to leave it on the default device for eager mode and add in the device flag when we have proper support for multi-device.

archana-ramalingam: This test passes on deepseek branch, so it's some sort of sharding going wrong here, as it's missing the sharding changes.

Alex-Vasile: I'd say remove the changes relative to the deepseek branch, mark the test as skipped, and unskip it when the rest of the deepseek work comes in.

archana-ramalingam: Yes, I am going to skip it here and leave it unskipped on deepseek branch.

sharktank/sharktank/layers/mixture_of_experts_block.py
```diff
@@ -28,18 +28,19 @@ class MoeBlock(ThetaLayer):
...
         theta: Theta,
-        expert_used_count: int,
         rms_epsilon: float,
         moe_activation=torch.nn.functional.silu,
         *,
         experts_ffn_moe_block: PreGatherFFNMOE | DenseFFNMOE | str = "DenseFFNMOE",
         score_experts=softmax,
         normalize_experts=True,
-        add_residual=True,
+        shard_count: int = 1,
```
sogartar, line 37: I think this not used.
I recall that I removed it on the deepseek-v3 feature branch.

sharktank/sharktank/types/theta.py
```diff
@@ -219,6 +221,35 @@ def rename_tensors_to_paths(self):
         for path, tensor in self.flatten().items():
             tensor.name = path
 
+    def rename_tensors(self, name_map: dict):
```
sogartar, line 224: Could add a return type hint and a better type hint for the arg.

sharktank/sharktank/layers/mixture_of_experts_block.py
```diff
@@ -60,41 +61,53 @@ def __init__(
...
                     f'Unknown experts_ffn_moe_block "{experts_ffn_moe_block}"'
                 )
         else:
-            self.experts = experts_ffn_moe_block
+            self.routed_experts = experts_ffn_moe_block
 
-        if "shared_experts" in theta:
-            self.shared_experts = FFN(theta("shared_experts"))
+        if self.expert_shared_count is not None:
+            ffn_theta = theta.pop("shexp", inplace=True)
```
sogartar, line 108: It would be quite surprising for a caller to realize that the theta was modified.

sharktank/sharktank/layers/mixture_of_experts_block.py
```diff
@@ -60,41 +61,53 @@ def __init__(
...
         # Add expert_count x FFN
         if isinstance(experts_ffn_moe_block, str):
             if experts_ffn_moe_block == "PreGatherFFNMOE":
-                self.experts = PreGatherFFNMOE(theta, activation=moe_activation)
+                self.routed_experts = PreGatherFFNMOE(
+                    theta, activation_fn=moe_activation
+                )
             elif experts_ffn_moe_block == "DenseFFNMOE":
-                self.experts = DenseFFNMOE(theta, activation_fn=moe_activation)
+                ffn_theta = theta.pop("exps", inplace=True)
```
sogartar, line 95: It would be quite surprising for a caller to realize that the theta was modified.

archana-ramalingam: Yes, that's the reason `inplace` was added, meaning although it's a pop function, the original theta is unaltered if `inplace=True`

sharktank/sharktank/types/theta.py
archana-ramalingam: `name_map` is expected to be a dict. Can you elaborate on what is a better type hint here?

sharktank/sharktank/types/theta.py
```diff
@@ -114,16 +114,18 @@ def transform(self, *transforms: InferenceTensorTransform) -> "Theta":
...
         name_path = ".".join(_norm_name_path(name_path))
         flat = self.flatten()
         accum = {}
         key_list = list(flat.keys())
         for key in key_list:
-            if key.startswith(name_path):
+            if name_path in key:
                 accum[key] = flat.pop(key)
-        self._tree = flat_to_nested_dict(flat)
+        if not inplace:
```
sogartar, line 126: Shouldn't the meaning of this argument be reversed. If `inplace=True` I would assume that the theta should be changed.

sharktank/sharktank/layers/mixture_of_experts_block.py
archana-ramalingam: Removed

sharktank/sharktank/types/theta.py
archana-ramalingam: Valid point, will reverse it now.

archana-ramalingam: Reversed

sharktank/sharktank/types/theta.py
```diff
@@ -114,16 +114,18 @@ def transform(self, *transforms: InferenceTensorTransform) -> "Theta":
     def to(self, *, device: Optional[Union[str, torch.device]] = None) -> "Theta":
         return self.transform(InferenceTensorTransforms.to_device(device))
 
-    def pop(self, *name_path: str | int) -> "Theta":
+    def pop(self, *name_path: str | int, inplace: bool = False) -> "Theta":
```
sogartar: I think the `inplace` argument needs some explanation.
It is more natural to have a function with another name as `pop` should mean that you modify the object.
Here we are doing a filtering by subpath, so something like
```
def match(self, *name_path: str | int) -> "Theta":
```

sharktank/sharktank/types/theta.py
sogartar: It is not clear if we are doing a complete name renaming or a subname renaming, which seems what is going on.

sharktank/sharktank/types/theta.py
```diff
@@ -219,6 +221,35 @@ def rename_tensors_to_paths(self):
...
+        theta_dict = {}
+        for k, v in self.flatten().items():
+            layer_parts = k.split(".")
+            new_layer_name = k
+            layer_name = layer_parts[0]
+            v_globals = v.globals
+            # v_layer_name can be differant than k in sharded cases
+            v_layer_name = list(v_globals.keys())[0]
+
+            if "blk" in k:
```
sogartar, line 234: There is no place for knowledge about specific names in such a general class as Theta.

sharktank/sharktank/layers/mixture_of_experts_block.py
```diff
@@ -60,43 +60,53 @@ def __init__(
...
         # Add expert_count x FFN
         if isinstance(experts_ffn_moe_block, str):
             if experts_ffn_moe_block == "PreGatherFFNMOE":
-                self.experts = PreGatherFFNMOE(theta, activation=moe_activation)
+                self.routed_experts = PreGatherFFNMOE(
+                    theta, activation_fn=moe_activation
+                )
             elif experts_ffn_moe_block == "DenseFFNMOE":
-                self.experts = DenseFFNMOE(theta, activation_fn=moe_activation)
+                ffn_theta = theta.pop("exps", inplace=False)
```
sogartar, line 93: It would be quite surprising for a caller to realize that the argument theta was modified.
The same problem is for the shared experts.

sharktank/sharktank/types/theta.py
```diff
@@ -219,6 +221,35 @@ def rename_tensors_to_paths(self):
...
+
+            if isinstance(v, DefaultPrimitiveTensor):
+                v = v_globals[v_layer_name]
+                theta_dict[new_layer_name] = DefaultPrimitiveTensor(
+                    name=new_layer_name, data=v
+                )
+            else:
+                theta_dict[new_layer_name] = v.clone(
+                    name=v_layer_name.replace(layer_name, new_layer_name)
+                )
```
sogartar, line 249: Is this function renaming the path within the Theta or is it renaming the global tensor names? Isn't it doing too much? This will cause export problems renaming the globals in the MLIR. It is used in the MoE block's theta, which is a subtheta of the top theta used for exporting the model.

sharktank/sharktank/types/theta.py
archana-ramalingam: True, I was following along the lines of irpa file format in [dump_gguf](https://github.com/nod-ai/shark-ai/blob/main/sharktank/sharktank/tools/dump_gguf.py#L72)

sharktank/sharktank/types/theta.py
archana-ramalingam: Global and layer path names are changed. But this is only applied on a popped temp theta. The original theta is unchanged. Not sure if this causes export problems. I can remove it, if you are sure it will.

sharktank/sharktank/layers/mixture_of_experts_block.py
archana-ramalingam: Yes, that's the reason inplace was added, meaning although it's a pop function, the original theta is unaltered if inplace=True

sharktank/sharktank/types/theta.py
sogartar: I assume it is not any `dict` but a `dict[str, str]`.

sharktank/sharktank/layers/mixture_of_experts_block.py
Alex-Vasile: I agree here. Rather than change the semantics of `pop`, I think this would be more intuitive as a `deepcopy` followed by a `pop` on the copy.

sharktank/sharktank/types/theta.py
sogartar: The GGUF dump also seems weird. I would not put it there as well. I can give it the benefit of the doubt. Maybe `blk` is special for the GGUF format, but I doubt it.

sharktank/sharktank/types/theta.py
```diff
@@ -114,16 +114,18 @@ def transform(self, *transforms: InferenceTensorTransform) -> "Theta":
     def to(self, *, device: Optional[Union[str, torch.device]] = None) -> "Theta":
         return self.transform(InferenceTensorTransforms.to_device(device))
 
-    def pop(self, *name_path: str | int) -> "Theta":
+    def pop(self, *name_path: str | int, inplace: bool = True) -> "Theta":
```
sogartar, line 117: I think I left a comment on my previous review, but could not find it. I am posting it again.
The `pop` function seems weird that it would not change the contents.
What we are doing is filtering on a subpath.
I think It would be more clear if there as a function `match` or `match_subpath`.
```
def match(self, *name_path: str | int) -> "Theta"
```

sharktank/sharktank/types/theta.py
sogartar: Most likely it will not, depending on what the rest of the code does. But it may causes problems as the same underlying torch tensor would be shared by InferenceTensors with different names.

sharktank/sharktank/layers/mixture_of_experts_block.py
archana-ramalingam: A deepcopy of a theta for llms like deepseek might add unnecessary constraint on memory?

sharktank/sharktank/types/theta.py
archana-ramalingam: Reverting the expert renaming change to the default method until we find a reliable, elegant method to resolve this.

sogartar (approved):

PR #1435, opened 2025-05-13
Fix test_grok and remove xfail on it

### LINKED ISSUE #1270: test_grok doesn't work on runners
[test_grok](https://github.com/nod-ai/shark-ai/blob/fd7d7e4e61cec2738ca91064b7d3f5885ff5ba48/sharktank/tests/models/grok/test_grok.py#L13) is run but the cross-entropy is not being checked because it apparently doesn't pass on the runners ("Unknown why but this does not reproduce on the buildbots" per @rsuderman ).

Alex-Vasile:
The expected cross_entropy in the test does could not be replicated even with the commit that initially implemented it. The new value is the value optained by running the test on the original commit. Same result whether using cpu or rocm version of torch.

Closes #1270.

### DIFF
--- a/sharktank/tests/models/grok/test_grok.py
+++ b/sharktank/tests/models/grok/test_grok.py
@@ -11,11 +11,6 @@ import torch
 import pytest
 
 
-@pytest.mark.xfail(
-    raises=AssertionError,
-    strict=False,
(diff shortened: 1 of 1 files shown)


Alex-Vasile:
The current value used to pass. Further investigation needed to see why it's changed.

PR #1437, opened 2025-05-13
Decrease numy of `await device` in Batcher

rsuderman:
`await device0` is extremely expensive so all transfers should be enqueued then awaited on. Doing one big wait will substantially prevent stalling.


shortfin/python/shortfin_apps/llm/components/batcher.py
```diff
@@ -432,6 +432,7 @@ async def get_args(self, bs, device0):
 
     async def get_results(self, logits, req_count, device0):
         # Return results.
+        to_wait = False
```
stbaione, line 435: Nit: Should name this var `await device` as well

rsuderman: Fixed.

stbaione (approved):

PR #1438, opened 2025-05-13
Add support for rejecting requests and making sure that the server is robust.

zphoenixrises:
- We now abandon generation of tokens at the server if the client request is disconnected
- Max queue size is now set to the largest decode length
- Propagated a handle to check the current status of the client request disconnection state to LlmInferenceExecRequest so that prefill/decode loops can exit early if client disconnects
- Tested with 200 batch size and the server is robust


shortfin/python/shortfin_apps/llm/components/batcher.py
```diff
@@ -114,10 +114,10 @@ async def board_flights(self):
             return
         target_size = min(self._current_workitems, self.ideal_batch_size)
         if waiting_count < target_size:
-            logger.info("Pending workitems to be enqueued")
+            logger.debug("Pending workitems to be enqueued")
```
IanNod, line 117: Is there a reason we are changing some of the logging in this PR?

shortfin/python/shortfin/interop/fastapi/__init__.py
```diff
@@ -53,7 +53,19 @@ def __init__(self, request: Request):
         self.response = asyncio.Future(loop=self._loop)
         self.responded = False
         self._streaming_queue: asyncio.Queue | None = None
-        self.is_disconnected = False
+        self._is_disconnected = False
+
+        self._loop.create_task(self._monitor_disconnection())
```
IanNod, line 58: This thread will be spun up for all shortfin instances, but is only actually used for llms. Overhead should be minimal so probably fine, but could you at least add a todo comment to add this functionality for our text_to_image pipelines like flux and sdxl? Should also verify that this overhead doensn't affect mlperf type throughput for those pipelines

IanNod (commented):
couple minor comments but otherwise looks good to me

shortfin/python/shortfin_apps/llm/components/batcher.py
zphoenixrises: These logs clutter the output and make it hard to spot anything during debugging the code. I often have to resort to using printfs instead of logging to spot any meaningful logs, which defeats the whole purpose of these logs. Info logging level is usually enabled during normal operation of the software. I do not think all these logs are needed during normal operation and are only needed while debugging.

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/multi_greedy_token_selection_strategy.py
```diff
@@ -84,9 +84,10 @@ async def decode(
         reservations = beam_group.active_beam_count
         config.decode_begin_callback(reservations)
         for _ in range(config.decode_config.max_completion_tokens):
+            if exec_req.is_disconnected():
```
stbaione, line 87: Should add this line to `beam_search_token_selection_strategy` as well

shortfin/python/shortfin/interop/fastapi/__init__.py
zphoenixrises: Added an issue https://github.com/nod-ai/shark-ai/issues/1439

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/multi_greedy_token_selection_strategy.py
zphoenixrises: Agreed

shortfin/python/shortfin_apps/llm/components/messages.py
```diff
@@ -22,7 +23,13 @@ class InferencePhase(Enum):
 class LlmInferenceExecRequest(InferenceExecRequest):
     """Performs a prefill operation."""
 
-    def __init__(self, phase: InferencePhase, input_token_ids: list[int], rid=None):
+    def __init__(
+        self,
+        phase: InferencePhase,
+        input_token_ids: list[int],
+        rid=None,
+        is_disconnected: Callable[[], bool] = lambda: False,
```
rsuderman, line 31: Rather than passing in just a function pointer, lets create a `AbstractStatusTracker` that can include an `is_disconnected` function. This would make it easier to plumb future states into the service.

shortfin/python/shortfin/interop/fastapi/__init__.py
```diff
@@ -53,7 +53,20 @@ def __init__(self, request: Request):
         self.response = asyncio.Future(loop=self._loop)
         self.responded = False
         self._streaming_queue: asyncio.Queue | None = None
-        self.is_disconnected = False
+        self._is_disconnected = False
+        # TODO: Enable the disconnection monitor to interrupt running inferences in SDXL and Flux when the client disconnects.
+        # See: https://github.com/nod-ai/shark-ai/issues/1439
+        self._loop.create_task(self._monitor_disconnection())
```
rsuderman, line 59: Do you ever kill this task after completion? If not the task may just live forever.

zphoenixrises: It should stop when the client disconnects, right?
`while not self._is_disconnected`

shortfin/python/shortfin_apps/llm/components/messages.py
zphoenixrises: Done

zphoenixrises:
> We should also try to have a basic test of disconnecting a request.

Done

shortfin/examples/python/fastapi/server.py
```diff
@@ -56,11 +56,22 @@ async def run(self):
                     responder.send_response(
                         JSONResponse({"answer": request.request_value})
                     )
+                elif request.request_value == 2:
+                    count = 0
+                    while not responder.is_disconnected() and count < 5:
```
rsuderman, line 61: Could we move this logic into `is_disconnected`? I.e. include a time disconnected before termination. We can default set to `1` but it feels reasonable to say `is_disconnected(timeout=1)`.

rsuderman (approved):
Just one nit

PR #1440, opened 2025-05-14
Create Program for each logical device

dezhiAmd:
- Why
Without changing compiler, the VMFB assumes only one logical device is being used. To use multiple logical devices, each logical device need create a program for it

- How
In class LlmGenerateService(GenerateService), use **inference_programs** which can store multiple programs to replace a single inference_program before this change. So that each logical device can use one inference_program

- Todo
Change devices from a 1D array to 2D array to support multiple devices and each device can have same number of logical devices


shortfin/python/shortfin_apps/llm/components/service.py
```diff
@@ -137,9 +137,17 @@ def _initialize_page_cache(self):
...
-        self.inference_program = self.create_program(
-            modules=component_modules, devices=self.sysman.ls.devices
-        )
+        logical_device_num = self.server_params.workers
+        self.inference_programs = {}
+
+        for index in range(logical_device_num):
+            logical_device = []
+            logical_device.append(self.sysman.ls.devices[index])
+            program = self.create_program(
```
dezhiAmd, line 146: Load the program twice, bound to two difference vectors of devices

PR #1444, opened 2025-05-14
Fix return of unshard_replicated

Alex-Vasile:
Current implementation returns a PrimitiveTensor instead of torch.Tensor.

### DIFF
--- a/sharktank/sharktank/ops/sharded_impls.py
+++ b/sharktank/sharktank/ops/sharded_impls.py
@@ -1615,7 +1615,7 @@ def unflatten_split(
 
 @unshard.override(ReplicatedTensor)
 def unshard_replicated(input: ReplicatedTensor) -> Tensor:
-    return input.shards[0]
+    return unbox_tensor(input.shards[0])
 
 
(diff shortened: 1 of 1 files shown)


sogartar (commented):
As is the current state of our codebase, both should be valid. If downstream ops can't ingest `InferenceTensor` its their problem.

sogartar:
For example if we had that then it would have been much easier to add the device ordinal to the actual primitive tensor.

Alex-Vasile:
> The argument for this is that you know that throughout the program you can rely that we have a common tensor base that can have customization points.

Do real model weights, when unsharded, come in as PrimitiveTensor types? I have seen torch.Tensors be propagated through the model when working with tp=1 and pp=1. I beleive that most of our ops only box the tensor, they don't also re-box them.

rsuderman:
Nope this is incorrect. Unboxing is meant to grab whatever underlying which is not true for ReplicatedTensors. If this was a replicated quantized tensor we would expect this to return the quantized tensor.

rsuderman (changes requested):
Incorrect behavior

PR #1454, opened 2025-05-15
Shortfin llm gpu argmax indices

stbaione:
- Enable using the gpu to perform argmax in the Shortfin server
- Changes the model on the `sharktank` side so that:
    - If `top-k is None`: Return the full logits
    - If `top-k == 1`: Return a subset of the logits, with the corresponding argmax indices
- Fixes a bug in the `split_argmax` implementation, adds a couple more tests for robustness/verification

Seeing marginal perf improvement when using `argmax` with `num_beams == 8`.
While not quite the jump I'd expect, it still shows promising results, as `argmax` was still pretty slow last I checked, even with the `split` implementation.

So, we're getting better results, even though the actual computation time for `argmax` is larger:

**Numpy Argmax Runs:**
```text
### Batch size: 4 ###

total requests 20
duration (s): 40.41542649269104
errors 0
mean length: 1.0

throughput (request / second): 0.49486054548049657
throughput (character / second): 0.49486054548049657
average request duration (s) 8.077639055252074
50%ile request duration (s) 8.309841752052307
75%ile request duration (s) 8.460618495941162
90%ile request duration (s) 8.737904477119447
95%ile request duration (s) 8.741231989860534
```

**Model-side Argmax Runs:**
```text
### Batch size: 4 ###

total requests 20
duration (s): 38.34156012535095
errors 0
mean length: 1.0

throughput (request / second): 0.5216271829996885
throughput (character / second): 0.5216271829996885
average request duration (s) 7.65543520450592
50%ile request duration (s) 7.543861269950867
75%ile request duration (s) 7.783713757991791
90%ile request duration (s) 8.81928973197937
95%ile request duration (s) 8.868924391269685

break: latency exceeds maximum of 5.000000s
```


sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -305,6 +305,14 @@ def _(model, tokens, seq_lens, seq_block_ids, cs):
             if args.logits_normalization == "log_softmax":
                 logits = ops.elementwise(torch.log, ops.softmax(logits, dim=-1))
 
+            if args.top_k is not None:
```
rsuderman, line 308: You can also do following to be more succinct
```
if args.top_k is None:
  return logits
```

sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -433,6 +441,14 @@ def _(
...
```
rsuderman, line 444: See comment above.

sharktank/sharktank/ops/default_impls.py
```diff
@@ -49,7 +49,8 @@ def _split_argmax(input_tensor, dim, keepdim: bool = False, chunk_size: int = 12
     input_tensor = unbox_tensor(input_tensor)
     dim = dim if dim >= 0 else input_tensor.dim() + dim
 
-    tensor_split = unflatten(input_tensor, dim, (chunk_size, -1))
+    n_chunks = input_tensor.shape[dim] // chunk_size
```
rsuderman, line 52: Validate that the chunk size is divideable into the chunking dimension

shortfin/python/shortfin_apps/llm/components/batcher.py
```diff
@@ -316,7 +322,13 @@ async def run(self):
                 )
 
             # Invoke VMFB. Logits are of shape [bs, bsl, d].
-            (logits,) = await fn(*args, fiber=self.fiber)
+            result = await fn(*args, fiber=self.fiber)
+
+            logits, indices = None, None
```
rsuderman, line 327:
```
logits = result[0]
if len(result) > 1:
  indices = result[1]
```

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/greedy_token_selection_strategy.py
```diff
@@ -33,9 +33,13 @@ def sample_logits(self) -> int:
         top_p = decode_config.top_p
 
         logits = np.array(exec_req.result_logits)
+        indices = exec_req.result_indices
 
         # Normal greedy selection based on max value
         if (top_k, top_p) == (None, None):
+            if indices is not None:
+                return indices.items[0]
```
rsuderman, line 41: You won't know this is true if `top_k` is set to `> 1` as topk returns the values unsorted. Indices should be proxied into `select_greedy` and used as n indirection. It should also be plumbed into the other token selection strategies.

sharktank/sharktank/examples/export_paged_llm_v1.py
rsuderman, line 1: You should also set the `top_k` value in the export json and check it `shortfin` side. This can be used to throw an error if the exported `top` selection falls outside of what we want to support service side.

rsuderman (changes requested):

sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -305,6 +305,14 @@ def _(model, tokens, seq_lens, seq_block_ids, cs):
...
+                if args.top_k == 1:
+                    max_logits, indices = model.argmax(
+                        logits, chunk_size=hp.context_length // 128
+                    )
+
```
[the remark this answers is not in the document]
stbaione: Updated

sharktank/sharktank/examples/export_paged_llm_v1.py
stbaione: Updated

sharktank/sharktank/examples/export_paged_llm_v1.py
stbaione: Updated

sharktank/sharktank/ops/default_impls.py
stbaione: Updated

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/greedy_token_selection_strategy.py
stbaione: For exporting the `topk` function, my thinking was I'll be setting the export to give us the sorted logits. Figured it'd be more efficient to get that over with on the GPU.

That would make it so that this line would always hold true

shortfin/python/shortfin_apps/llm/components/batcher.py
stbaione: Updated

sharktank/sharktank/examples/export_paged_llm_v1.py
stbaione: Added error check for `top_k != 1`

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/greedy_token_selection_strategy.py
rsuderman: Lets not worry too much about making sure logits are sorted. CPU wise sorting ~100 objects is basically instant so its unlikely we will even both on the GPU.

PR #1445, opened 2025-05-14
llama4 - patching

oyazdanb:
adding patching.py changes from Boian's llama4 branch 
https://github.com/nod-ai/shark-ai/commit/0b8b714a01e67b9979b9c2037f9058a583b2a1b9#diff-0b44c2142e71c1fd626319f8a7e0032dea83540af9c077482c83c88d4425c539

It adds tracing of arguments and nested arguments/results in a tree-like structure.

PR #1460, opened 2025-05-16
Fix relative import

Alex-Vasile:
Trying to run test locally raises an error:
```
    def test_export_compile(dummy_model_path: Path):
        with chdir(dummy_model_path):
            check_call(["shark", "model", "export", "dummy-model-local-llvm-cpu"])
            check_call(["shark", "model", "compile", "dummy-model-local-llvm-cpu"])
>           from .. import models
E           ImportError: attempted relative import with no known parent package

tests/tools/sharktank_test.py:32: ImportError
```

That line also does not appear to be needed as `register_all_models()` right below it does the same thing.


sogartar (approved):
My bad. I should have removed that.
Thank you for the fix.

Alex-Vasile:
Data-dependent test failure is casued by https://github.com/nod-ai/shark-ai/pull/1445#discussion_r2093125174

Rerunning unit tests to see if they'll pass this time. Issue appears to be huggingface/stabilityai file related.

PR #1464, opened 2025-05-16
Change setup_cache to use existing placement info

Alex-Vasile:
Instead of reading the placement info from a secondary place, use the existing tags inside cache_state.


sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -153,33 +153,33 @@ def setup_cache(model, shard_count):
             )
             page_dim = torch.export.Dim("page")
 
-            dynamic_shapes = [
-                {0: page_dim} for _ in range(llama_config.pipeline_parallelism_size)
-            ]
+            pipeline_parallelism_size = len(cache_state)
+            tensor_parallelism_size = (
+                cache_state[0].shard_count
```
rsuderman, line 158: Once it starts spreading across multiple lines like this the benefits of not having a regular if really disappear. I would just do

```
pipeline_parallelism_size = 1
if isinstance(cache_state[0], ShardedTensor):
  pipeline_parallelism_size = cache_state[0].shard_count
```

sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -153,33 +153,33 @@ def setup_cache(model, shard_count):
...
-                llama_config.tensor_parallelism_size > 1
-                or llama_config.pipeline_parallelism_size > 1
-            ):
+            if tensor_parallelism_size > 1 or pipeline_parallelism_size > 1:
                 shard_dim = cache_state[0].shard_dim
 
                 unpacked = [[shard._data for shard in cs.shards] for cs in cache_state]
                 dynamic_shapes = [
-                    [ds] * llama_config.tensor_parallelism_size for ds in dynamic_shapes
+                    [ds] * tensor_parallelism_size for ds in dynamic_shapes
```
rsuderman, line 174: Move this to where `dynamic_shapes` is constructed. Doing it after the fact feels wrong.

rsuderman (changes requested):

rsuderman (approved):

PR #1465, opened 2025-05-16
Reorder KV Cache contraction dimensions

rsuderman:
The KV Cache should have contracting dimensions placed together. Specifically the block sequence stride should be after the KV heads but before the attn_dim.


Groverkss (approved):
LGTM, but i don't see the page layout documentation being updated. I think it's in the PagedAttention class. Can you update that before landing?

rsuderman:
> LGTM, but i don't see the page layout documentation being updated. I think it's in the PagedAttention class. Can you update that before landing?

Updated

PR #1467, opened 2025-05-16
Deepseek tests

Alex-Vasile:
Tests are skipped for now, will be enabled by subsequent PR which adds the implementation.

### DIFF
--- a/sharktank/tests/layers/sharded_paged_latent_attention_block.py
+++ b/sharktank/tests/layers/sharded_paged_latent_attention_block.py
@@ -1,0 +1,140 @@
+# Copyright 2025 Advanced Micro Devices, Inc
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import unittest
+import pytest
+from copy import deepcopy
+
+import torch
+
+from sharktank.layers.paged_llama_attention_block import PagedLlamaAttentionBlock
+from sharktank.layers.rotary_embedding import RotaryEmbeddingLayer
+from sharktank.layers.testing import make_rand_torch
+
+from sharktank.models.deepseek.toy_deepseek import generate
+from sharktank.types.theta import Theta, flat_to_nested_dict
+from sharktank.types import unbox_tensor
+from sharktank.types.sharding import shard_theta, LatentAttentionBlockSharding
+from sharktank import ops
+from sharktank.utils.create_cache import *
+
+
+@pytest.mark.skip(
+    reason="Support will be added soon",
+)
+class ShardedPagedLatentAttentionBlockTest(unittest.TestCase):
+    """Verify that the sharded latent paged attention block behaves in PyTorch as the
+    unsharded variant."""
+
+    def testShardedLatentLayer(self):
+
+        bs = 1
+        start_index = 0
+        block_seqlen = 7
+        tensor_parallelism_size = 2
+        page_count = 64
+        dtype = torch.float16
+
+        theta, config = generate(12345)
+        theta = theta("blk", 0)
+        hp = config.hp
+        config.block_seq_stride = 16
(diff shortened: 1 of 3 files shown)


sharktank/tests/layers/sharded_paged_latent_attention_block.py
```diff
@@ -0,0 +1,140 @@
...
+
+        embedding = RotaryEmbeddingLayer(
+            rope_dimension_count=hp.rope_dimension_count,
+            rope_freq_base=hp.rope_freq_base,
+            max_seqlen=hp.context_length,
+        )
+
+        reference_model = PagedLlamaAttentionBlock(
+            theta=theta,
+            block_index=0,
```
sogartar, line 85: It is sad that the attention block needs to know its own index. Nothing can done in this PR.

sharktank/tests/models/deepseek/test_deepseek.py
```diff
@@ -4,28 +4,45 @@
...
-    ids = torch.asarray(
-        [[3, 22, 13, 114, 90, 232, 61, 13, 244, 13, 212]], dtype=torch.int64
-    )
-    logits = model.prefill(tokens=ids)
-    ids = ids[0, :-1]
-    logits = logits[0, 1:]
-    cross_entropy = torch.nn.functional.cross_entropy(logits, ids)
-
-    assert pytest.approx(5.3062, 1e-4) == cross_entropy
+class DeepseekShardedTest(unittest.TestCase):
```
sogartar, line 21: Why is this called `Sharded`?

sharktank/tests/models/deepseek/test_deepseek.py
```diff
@@ -4,28 +4,45 @@
...
+    def test_deepseek(self):
+        theta, config = generate(12345)
+        model = PagedLlmModelV1(theta=theta, config=config)
+
+        ids = [[3, 22, 13, 114, 90, 232, 61, 13, 244, 13, 212]]
```
sogartar, line 26: It is suspicious that these values have not change, since on the feature branch the model config has changed so I would assume that the tokens designed to return the highest cross-entropy would have to also change.

Alex-Vasile: @archana-ramalingam

sharktank/tests/models/deepseek/test_deepseek.py
Alex-Vasile, line 21:
```suggestion
class DeepseekTest(unittest.TestCase):
```

Alex-Vasile: Renamed

sharktank/tests/layers/sharded_paged_latent_attention_block.py
archana-ramalingam: Addressed here https://github.com/nod-ai/shark-ai/pull/1470#issuecomment-2887793785

sharktank/tests/models/deepseek/test_deepseek.py
sogartar: Since the test is skipped we can move things forward and address it when we land the other changes.

sogartar (approved):

sharktank/tests/layers/sharded_paged_latent_attention_block.py
```diff
@@ -0,0 +1,140 @@
...
+
+        spec = LatentAttentionBlockSharding(shard_count=tensor_parallelism_size)
+        sharding_keys = {k for k in spec.theta_sharding().keys()}
+        flattened = theta.flatten()
+        t = {}
+        for k in flattened:
+            if k.split(".")[0] in sharding_keys:
+                t[k] = flattened[k]
+
+        theta = Theta(flat_to_nested_dict(t))
```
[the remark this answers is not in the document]
archana-ramalingam: I rewrote this test, but the numerics didn't match with the original version. As we are already dealing with cache refactor issues, let's just add a TODO in deepseek branch and tackle this later.

PR #1470, opened 2025-05-16
Add missing names to toy weights

Alex-Vasile:
Part of deepseek model bringup.

### DIFF
--- a/sharktank/sharktank/layers/testing.py
+++ b/sharktank/sharktank/layers/testing.py
@@ -232,6 +232,7 @@ def make_mmdit_single_block_random_theta(
 
 
 def make_random_ffn_theta(
+    block_idx: int,
     in_dim: int,
     hidden_dim: int,
     dtype: torch.dtype,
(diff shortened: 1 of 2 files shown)


sogartar (commented):
I don't like this style of making the theta block having to know what block it is a part of or that the name must be prefixed by `blk`.
The correct approach is when you have generated your top level theta is for export to rename the tensors to match the theta structure.

You would have to call [rename_tensors_to_paths](https://github.com/nod-ai/shark-ai/blob/27998f83eebfa1290770138ee2ca4eb88dfdd866/sharktank/sharktank/types/theta.py#L234) and [mark_export_external_theta](https://github.com/nod-ai/shark-ai/blob/27998f83eebfa1290770138ee2ca4eb88dfdd866/sharktank/sharktank/types/theta.py#L253) before exporting to do that. We need to move towards a common path where these things happen automatically during export. This detail has been causing grief for all of us. @KyleHerndon is that last who complained about it.
For example [here](https://github.com/nod-ai/shark-ai/blob/6ef7ea633472d49cd2b85790e4f64888a8b6fdea/sharktank/sharktank/utils/export.py#L206) it is done like that but this is not what LLM models use for export. Maybe it should do the auto-rename there as well.
The point is that if we have a convergent export function that accepts a Theta as well then we would not be burned by this again.

archana-ramalingam (approved):

archana-ramalingam:
> I don't like this style of making the attention having to know what block it is a part of or that the name must be prefixed by `blk`. The correct approach is when you have generated your top level theta is for export to rename the tensors to match the theta structure.
> 
> You would have to call [rename_tensors_to_paths](https://github.com/nod-ai/shark-ai/blob/27998f83eebfa1290770138ee2ca4eb88dfdd866/sharktank/sharktank/types/theta.py#L234) and [mark_export_external_theta](https://github.com/nod-ai/shark-ai/blob/27998f83eebfa1290770138ee2ca4eb88dfdd866/sharktank/sharktank/types/theta.py#L253) before exporting to do that. We need to move towards a common path where these things happen automatically during export. This detail has been causing grief for all of us. @KyleHerndon is that last who complained about it. For example [here](https://github.com/nod-ai/shark-ai/blob/6ef7ea633472d49cd2b85790e4f64888a8b6fdea/sharktank/sharktank/utils/export.py#L206) it is done like that but this is not what LLM models use for export. Maybe it should do the auto-rename there as well. The point is that if we have a convergent export function that accepts a Theta as well then we would not be burned by this again.

Can we create an issue elaborating the pain points of our current methods, so that we can track and decide on a sensible solution?

PR #1474, opened 2025-05-19
Shortfin gpu topk

stbaione:
## Description

Stacked on #1422 

This PR wires up the `topk` kernel from `sharktank` through the `shortfin LLM server`.

Similarly to `argmax`, if the model is exported with `topk > 1`, it will return the indices and subset of values that correspond to the `k` highest scoring logits, in sorted order.

If an invalid combination of `SeverParams or GenReqInput` with `ModelParams` is used, the server returns a `400` error, with this body:

```json
{
    "error": "Requested top-k larger than exported top-k",
    "code": "INVALID_REQUEST_ARGS",
    "exported_topk": 4,
    "requested_topk": 8
}
```

We won't see any performance improvements from using the gpu based `topk` route, until the [codegen performance is fixed](https://github.com/iree-org/iree/issues/20772)


rsuderman (approved):

sharktank/sharktank/models/llm/llm.py
```diff
@@ -153,6 +153,16 @@ def argmax(
...
+    def topk(
+        self,
+        logits: torch.Tensor,
+        k: int,
+        chunk_size: int,
+    ):
+        return ops.topk(
+            logits, k=k, dim=-1, largest=True, sorted=True, chunk_size=chunk_size
+        )
+
```
Groverkss, line 165: Why is a llm model exposing a topk kernel?

Groverkss: I know topk/argmax should be done on the device side instead of the host side, but why is it part of the model? You should directly use ops.argmax/ops.topk in the export script. If you want to export topk/argmax as functions for shortfin to use then use fxb.export_program and export the op directly.

PR #1484, opened 2025-05-20
Fix KVCache for PP>1 TP=1

Alex-Vasile:
Missed in #1404.


sharktank/sharktank/layers/paged_attention.py
```diff
@@ -590,10 +592,17 @@ def read(
...
         page_ids = self.unwrap_pipelining(page_ids)
 
-        return self.caches[pipeline].read(
+        k, v = self.caches[pipeline].read(
             state=pipeline_state, transformer_block_index=block, page_ids=page_ids
         )
 
+        # If running purely pipelined, we need to rewrap the primitive tensors as replicated tensors.
+        if isinstance(state[pipeline], ReplicatedTensor):
+            k = ReplicatedTensor(ts=[k], devices=state[pipeline].devices)
```
rsuderman, line 601: This should be handled by `self.caches[pipeline]`. If it isn't setting the devices correctly thats an issue when passing to the subcaches..

rsuderman (changes requested):

rsuderman (approved):

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -144,6 +144,11 @@ def read(
         key = selected[:, :, 0, :].flatten(1, 2)
         value = selected[:, :, 1, :].flatten(1, 2)
 
+        if self.devices:
+            # Explicitly passing a list of one value to avoid redundant transfer inside ReplicateTensor.__init__.
+            key = ReplicatedTensor(ts=[key], devices=self.devices)
+            value = ReplicatedTensor(ts=[value], devices=self.devices)
+
```
Groverkss, line 151: Why are sharding details exposed in KVCache when there is a ShardedKVCache class?

Alex-Vasile: ShardedKVCache is used for tensor parallelism. For pipeline parallelim we have the PipelinedCache which uses both KVCache (PP only) and ShardedKVCache (TP and PP)

PR #1486, opened 2025-05-20
[sharktank] Add Latent attention changes and Deepseek config

archana-ramalingam:
- Add Latent attention changes 
- Add Deepseek config changes


sharktank/sharktank/layers/configs/llm_configs.py
```diff
@@ -45,36 +46,119 @@ class LlamaHParams:
...
             p, f"{name_prefix}.rope.dimension_count", default_rope_dimension_count
         )
+        expert_count = _optional_int_prop(
+            p, f"{name_prefix}.expert_count", default_expert_count
+        )
 
         attention_softcap = 30.0 if name_prefix == "grok" else None
 
+        if name_prefix == "deepseek2":
+            default_n_expert_groups = 8
```
rsuderman, line 107: Move this to a helper to avoid the single large block for deepseek. Just makes it easier for skimming. And have the helper return a `dict`. Comment below will help that make sense.

sharktank/sharktank/layers/configs/llm_configs.py
```diff
@@ -88,18 +172,33 @@ def from_gguf_props(p: dict[str, Any]):
             attention_head_count_kv=_optional_int_prop(
                 p, f"{name_prefix}.attention.head_count_kv", attention_head_count
             ),
-            attn_head_dim=rope_dimension_count,
+            attn_head_dim=attn_head_dim,
```
rsuderman, line 175: For any of these new parameters that are deepseek specific we can do the following.

If we have a dict with `{"q_lora_rank":q_lora_rank}` we can just pass this to the constructor `**architecture_specialization`. This expansion just lets architecture specific customization to be passed around as dictionaries instead.

sharktank/sharktank/layers/paged_llama_attention_block.py
```diff
@@ -108,9 +136,55 @@ def __init__(
...
+        embedding: RotaryEmbeddingLayer,
+        embedding_batch_mask: torch.Tensor,
+    ):
+        """
+        x:
+            input token embeddings.
+            shape is (batch_size, sequence_length, embedding_length)
+        """
+        if self.attn_type == "gqa":
+            bs, batch_seq_len, _ = x.shape
```
rsuderman, line 152: Move this to a `self.gqa_attention` helper that we invoke. Lets people dive into specifics a little more easily.

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -945,6 +958,8 @@ def repeat_kv(x: torch.Tensor) -> torch.Tensor:
                 scale=None,  # defaults to 1/sqrt(dim)
             )
 
+        return attn_output
```
rsuderman, line 961: Is there a reason we changed to this formatting? Generally returning from each `if ` makes readability a little easier as you can identify early termination (though if we do this the `else` above should just be remove and `scaled_dot_product_attention` should just happen as a follow up to each other else.

rsuderman (changes requested):

sharktank/sharktank/layers/paged_llama_attention_block.py
archana-ramalingam: Done

sharktank/sharktank/layers/configs/llm_configs.py
```diff
@@ -173,32 +103,19 @@ def from_gguf_props(p: dict[str, Any]):
...
-            n_limited_groups=n_limited_groups,
             rope_dimension_count=rope_dimension_count,
             rope_freq_base=_optional_float_prop(
                 p, f"{name_prefix}.rope.freq_base", default_rope_freq_base
             ),
-            rope_scaling_type=rope_scaling_type,
-            rope_scaling_factor=rope_scaling_factor,
-            rope_scaling_original_context_length=rope_scaling_original_context_length,
-            rope_scaling_yarn_log_multiplier=rope_scaling_yarn_log_multiplier,
+            custom=custom_configs,
```
rsuderman, line 118: So here I would do `**custom_config` and still have the named values in your `LLaMaHParams` this way you can still have validationg that the names match to params. Its just a nice trick that allows you to optionally populate args.

sharktank/sharktank/layers/configs/llm_configs.py
```diff
@@ -47,41 +48,19 @@ class LlamaHParams:
     attention_layer_norm_rms_epsilon: float
     attention_head_count_kv: int
 
-    # Deepseek Multi-Latent Attention config
-    q_lora_rank: Optional[int] = None
```
rsuderman: So I would keep these configs around, just use the `dict` trick to avoid specifying all of them when some are optional

rsuderman (approved):

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -219,6 +218,10 @@ def write_timestep(
             index = index * self.block_seq_stride + page_offset
 
             values = ops.to(cache_partition, dtype=page_table.dtype)
+
+            if isinstance(values, ShardedTensor) and type(values) != type(page_table):
+                values = ops.reshard_like(values, like=page_table)
+
```
Groverkss, line 224: Do we need to expose these sharding details here? Also there should be documentation why we are doing this, this is not obvious.

Groverkss (commented):
Can you add me as a reviewer next time changes are made to paged_attention.py?

sharktank/sharktank/layers/paged_attention.py
Groverkss: There is a ShardedCache class which should be exposing these sharding details. KVCache by itself isn't sharded right?

archana-ramalingam: Yes, KVCache isn't sharded by itself and ShardedCache should be able to handle this implicitly. 
This change was added before refactor, we will verify if it's plumbed through for tp and @sogartar will take care of this in a future PR that fixes current TP issues we are facing.

sogartar: We should not need this.

PR #1489, opened 2025-05-21
Fix fp8 attention for brevitas quantization

KyleHerndon:


KyleHerndon:
Accidentally closed with a bad merge or smthn, working on it

PR #1490, opened 2025-05-21
[sharktank] Add missing argument device affinity for tensor-parallel Resnet block test

sogartar:


sogartar:
On Ubuntu the test passes with IREE version `3.5.0rc20250602`.

sogartar:
My assumption is that torch is leaving some buffers alive even though all variables that reference them are out-of-scope and GC is called. These buffers are backed by the memory-mapped weights files. On NTFS you are not allowed to delete a file that is in use.

sogartar:
There is more work on this PR, I need to xfail for different reasons on Windows and on Linux, but the xfail that I recently wrote to support message matching requires a significant rework as it exposed problems with the approach of wrapping the `pytest.mark.xfail`. I am in the process of redoing it.

Alex-Vasile (approved):

PR #1491, opened 2025-05-21
[sharktank] Add mixture of experts(moe) support for Llama 4 model

vivekkhandelwal1:
This commit adds support for mixture of experts for the Llama 4 model.

The code is adopted from https://github.com/nod-ai/shark-ai/tree/users/sogartar/llama4.

Authored-by: Boian Petkantchin <Boian.Petkantchin@amd.com>
Co-authore-by: Vivek Khandelwal <vivekkhandelwal1424@gmail.com>

### DIFF
--- a/sharktank/sharktank/layers/configs/llm_configs.py
+++ b/sharktank/sharktank/layers/configs/llm_configs.py
@@ -75,6 +75,15 @@ class LlamaHParams:
     # MoE config
     expert_count: Optional[int] = None
     expert_used_count: Optional[int] = None
+    expert_feed_forward_length: Optional[int] = None
+    expert_shared_feed_forward_length: Optional[int] = None
+
+    # Specifies the interval at which Mixture of Experts (MoE) layers are inserted among the model's layers.
+    # For example:
+    #     1 - every layer is an MoE layer.
+    #     3 - every third layer (layers 0, 3, 6, ...) is an MoE layer; others are dense layers.
+    # If None, no specific interleaving pattern is applied.
+    interleave_moe_layer_step: Optional[int] = None
 
     # Deepseek MoE config
     expert_shared_count: Optional[int] = None
@@ -85,17 +94,14 @@ class LlamaHParams:
     n_dense_layers: Optional[int] = None
     route_scale: Optional[float] = None
 
-    # Deepseek MoE config
-    expert_shared_count: Optional[int] = None
-
     @staticmethod
     def from_gguf_props(p: dict[str, Any]):
         name_prefix = p.get("general.architecture", "llama")
-
-        default_rope_freq_base = 500000.0
-        default_rope_dimension_count = 128
         default_expert_count = 0
         default_expert_used_count = 0
+        default_interleave_moe_layer_step = None
+        default_rope_freq_base = 500000.0
+        default_rope_dimension_count = 128
 
         attention_head_count = _int_prop(p, f"{name_prefix}.attention.head_count")
         rope_dimension_count = _optional_int_prop(
@@ -128,7 +134,6 @@ class LlamaHParams:
             attention_head_count_kv=_optional_int_prop(
                 p, f"{name_prefix}.attention.head_count_kv", attention_head_count
             ),
-            expert_count=expert_count,
             expert_used_count=_optional_int_prop(
                 p, f"{name_prefix}.expert_used_count", default_expert_used_count
             ),
@@ -139,6 +144,17 @@ class LlamaHParams:
             rope_dimension_count=rope_dimension_count,
             rope_freq_base=_optional_float_prop(
                 p, f"{name_prefix}.rope.freq_base", default_rope_freq_base
+            ),
+            expert_feed_forward_length=_optional_int_prop(
+                p, f"{name_prefix}.expert_feed_forward_length", None
+            ),
+            expert_shared_feed_forward_length=_optional_int_prop(
+                p, f"{name_prefix}.expert_shared_feed_forward_length", None
+            ),
+            interleave_moe_layer_step=_optional_int_prop(
+                p,
+                f"{name_prefix}.interleave_moe_layer_step",
+                default_interleave_moe_layer_step,
             ),
             **custom_config,
         )
@@ -198,6 +214,18 @@ class LlamaHParams:
             res[
                 f"{self.model_arch}.rope.scaling.yarn_log_multiplier"
             ] = self.rope_scaling_yarn_log_multiplier
+        if self.expert_feed_forward_length is not None:
+            res[
+                f"{self.model_arch}.expert_feed_forward_length"
+            ] = self.expert_feed_forward_length
+        if self.expert_shared_feed_forward_length is not None:
+            res[
+                f"{self.model_arch}.expert_shared_feed_forward_length"
+            ] = self.expert_shared_feed_forward_length
+        if self.interleave_moe_layer_step is not None:
+            res[
+                f"{self.model_arch}.interleave_moe_layer_step"
+            ] = self.interleave_moe_layer_step
         return res
 
 
@@ -239,6 +267,11 @@ def get_custom_configs(p: dict[str, Any], name_prefix: str):
             p, f"{name_prefix}.rope.scaling.yarn_log_multiplier"
         )
         res["attn_head_dim"] = res["qk_nope_head_dim"] + res["qk_rope_head_dim"]
+
+    if name_prefix == "llama4":
+        res["expert_shared_count"] = _optional_int_prop(
+            p, f"{name_prefix}.expert_shared_count", None
+        )
 
     return res
 
@@ -314,6 +347,9 @@ class LlamaModelConfig:
     # Dtype to use for general FP activations not otherwise configured.
     activation_dtype: torch.dtype = torch.float16
 
+    # Activation function to use for the model. If None, will use the activation function specified in the model configuration.
+    activation_fn: Optional[str] = None
+
     # Dtype to use for attention.
     attention_dtype: torch.dtype = torch.float16
 
@@ -338,6 +374,12 @@ class LlamaModelConfig:
     # Which attention kernel to use.
     attention_kernel: str = "torch"
 
+    # Implementation of rotary embeddings.
+    # E.g. "llama3", "llama4", "default".
+    # "llama3" indicates running with HuggingFace implementation and ensures
+    # numerical equivalency to HuggingFace's LLaMa.
+    rope_type: str | None = None
+
     # Indicates if running with HuggingFace implementation and ensures
     # numerical equivalency to HuggingFace's LLaMa if true (by modifying
     # rotary embedding).
@@ -356,6 +398,55 @@ class LlamaModelConfig:
 
     # A list of layer indices where chunked attention is applied instead of full attention.
     chunked_attention_layers: Optional[set[int]] = None
+
+    # Indices of layers that are MoE.
+    moe_layers: Optional[list[int]] = None
+
+    # Indices of layers that use RoPE after the attention.
+    rope_layers: Optional[list[int]] = None
+
+    use_qk_norm: bool = False
+    # If True, applies normalization to the query and key vectors in attention.
+
+    # In HuggingFace transformers, this field is represented as an int, but it is only ever used as a boolean.
+    # For clarity and correctness, it should be a bool: if True, enables attention temperature tuning.
+    attn_temperature_tuning: Optional[bool] = None
+
+    # Scaling factor applied to attention scores.
+    attn_scale: Optional[float] = None
+
+    # Scaling factor applied as a floor value in attention computations.
+    floor_scale: Optional[int] = None
+
+    # If True, the Feed-Forward Network (FFN) block adds the residual connection at the end.
+    ffn_add_residual: bool = True
+
+    # The size of the model's vocabulary.
+    vocabulary_size: Optional[int] = None
+
+    # The default data type to use for model parameters and computations.
+    dtype: Optional[torch.dtype] = None
+
+    def __post_init__(self):
+        if self.moe_layers is None:
+            if self.hp.interleave_moe_layer_step is None:
+                self.moe_layers = []
+            else:
+                self.moe_layers = list(
+                    range(
+                        self.hp.interleave_moe_layer_step - 1,
+                        self.hp.block_count,
+                        self.hp.interleave_moe_layer_step,
+                    )
+                )
+        else:
+            if self.hp.interleave_moe_layer_step is not None:
+                raise ValueError(
+                    "moe_layers and hp.interleave_moe_layer_step are mutually exclusive."
+                )
+
+        if isinstance(self.dtype, str):
+            self.dtype = serialized_name_to_dtype(self.dtype)
 
     def to_properties(self) -> "PropertyValueType":
         res = self.hp.to_gguf_props()
--- a/sharktank/sharktank/layers/configs/llm_configs.py
+++ b/sharktank/sharktank/layers/configs/llm_configs.py
@@ -348,9 +348,6 @@ class LlamaModelConfig:
     # Dtype to use for general FP activations not otherwise configured.
     activation_dtype: torch.dtype = torch.float16
 
-    # Activation function to use for the model. If None, will use the activation function specified in the model configuration.
-    activation_fn: Optional[str] = None
-
     # Dtype to use for attention.
     attention_dtype: torch.dtype = torch.float16
 
@@ -418,9 +415,6 @@ class LlamaModelConfig:
 
     # Scaling factor applied as a floor value in attention computations.
     floor_scale: Optional[int] = None
-
-    # If True, the Feed-Forward Network (FFN) block adds the residual connection at the end.
-    ffn_add_residual: bool = False
 
     # The default data type to use for model parameters and computations.
     dtype: Optional[torch.dtype] = None
--- a/sharktank/sharktank/layers/configs/llm_configs.py
+++ b/sharktank/sharktank/layers/configs/llm_configs.py
@@ -224,6 +224,7 @@ class LlamaHParams:
 
 
 def get_custom_configs(p: dict[str, Any], name_prefix: str):
+    res = defaultdict(lambda: None)
 
     if name_prefix == "grok":
         res["attention_softcap"] = 30.0
--- a/sharktank/sharktank/layers/configs/llm_configs.py
+++ b/sharktank/sharktank/layers/configs/llm_configs.py
@@ -94,6 +94,9 @@ class LlamaHParams:
     n_dense_layers: Optional[int] = None
     route_scale: Optional[float] = None
 
+    # The size of the model's vocabulary.
+    vocab_size: Optional[int] = None
+
     @staticmethod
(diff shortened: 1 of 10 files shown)


sharktank/sharktank/layers/ffn_moe_block.py
```diff
@@ -32,21 +32,47 @@ def __init__(
...
+        expert_gate: torch.Tensor | None = None,
+        einstring="mk,menk->men",
+    ):
         inputs = inputs[:, :]
         weights = weights[experts, :, :]
-        matmul = einsum_2args(inputs, weights, "mek,menk->men")
+        if expert_gate is not None:
+            weights = weights * expert_gate.view(
+                expert_gate.shape[0], expert_gate.shape[1], 1, 1
+            )
```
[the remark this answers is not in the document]
sogartar: Actually, this is concerning as I think we would not be hitting this path as I look the rest of the code and also on the [users/sogartar/llama4](https://github.com/nod-ai/shark-ai/tree/users/sogartar/llama4) branch.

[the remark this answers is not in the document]
sogartar: Disregard what I wrote earlier about this block. We are actually multiplying by the `expert_gate` but after this operation. It is OK to do that as these operations commute.
This is all throw-away code. The only thin that is really different in this class is that we don't post multiply with the `expert_gate`/`router_scores` as mentioned in the other comment.

sharktank/sharktank/layers/configs/llm_configs.py
```diff
@@ -221,6 +272,55 @@ class LlamaModelConfig:
...
+    attn_scale: Optional[float] = None
+
+    # Scaling factor applied as a floor value in attention computations.
+    floor_scale: Optional[int] = None
+
+    # If True, the Feed-Forward Network (FFN) block adds the residual connection at the end.
+    ffn_add_residual: bool = True
+
+    # The size of the model's vocabulary.
+    vocabulary_size: Optional[int] = None
```
sogartar, line 298: I was wrong putting this here. It should be in `LlamaHParams` as it is available in GGUF. there also would need to be addition in the `from_gguf_prop` and `to_gguf_props` methods.
The difference is that there it is named `vocab_size`.

sharktank/sharktank/layers/ffn_moe_block.py
```diff
@@ -32,21 +32,47 @@ def __init__(
...
 
-    def pre_matmul_gather(self, inputs, weights, experts, einstring="mk,menk->men"):
+    def pre_matmul_gather_orig(
+        self,
+        inputs: torch.Tensor,
+        weights: torch.Tensor,
+        # (batch_size * sequence_length, num_top_experts)
+        experts: torch.Tensor,
+        einstring="mk,menk->men",
+    ):
```
sogartar, line 53: This can be removed.

sharktank/sharktank/layers/ffn_moe_block.py
```diff
@@ -32,21 +32,47 @@ def __init__(
...
 
-    def bigger_mmg(self, inputs, weights, experts):
+    def pre_matmul_gather(
+        self,
+        inputs: torch.Tensor,
+        weights: torch.Tensor,
+        # (batch_size * sequence_length, num_top_experts)
+        experts: torch.Tensor,
+        # (bs * sl, num_top_experts)
+        expert_gate: torch.Tensor | None = None,
```
sogartar, line 66: This argument should be removed.

sharktank/sharktank/models/llama/testing.py
```diff
@@ -101,54 +104,77 @@ def make_attention_block_ffn_theta_v2(
     return Theta(res_dict)
 
 
-def make_moe_block_theta(feature_dim=1024, ffn_dim=6144, num_experts=8) -> Theta:
-    return Theta(
+def make_attention_moe_block_random_theta(
```
sogartar, line 107: We must make sure to reconcile with the work coming from the Deepseek side. There are probably discrepancies.

sharktank/sharktank/layers/configs/llm_configs.py
```diff
@@ -150,8 +188,12 @@ def _optional_float_prop(p: dict[str, Any], name: str, default_value: float) ->
         raise ValueError(f"Property '{name}' expected to be a float and was not") from e
 
 
-def _optional_int_prop(p: dict[str, Any], name: str, default_value: int) -> int:
+def _optional_int_prop(
+    p: dict[str, Any], name: str, default_value: Optional[int] = None
+) -> int:
```
archana-ramalingam, line 193: If we are returning int and None, shouldn't this be `-> Union[int, None]:` ?

sharktank/sharktank/layers/ffn_block.py
```diff
@@ -60,4 +62,7 @@ def forward(
             ffn_activation = ops.elementwise(self.activation_fn, ffn_up)
             ffn_down = self.ffn_down(ffn_activation)
 
+        if self.add_residual:
```
archana-ramalingam, line 65: Determine the add_residual version for llama4 and move accordingly.

sharktank/sharktank/layers/mixture_of_experts_block.py
```diff
@@ -110,7 +110,7 @@ def __init__(
                 }
             )
             self.shared_experts = FFN(
-                theta=shared_ffn_theta, activation_fn=moe_activation
+                theta=shared_ffn_theta, activation_fn=moe_activation, add_residual=False
```
archana-ramalingam, line 113: Same comment as above for `add_residual`

sharktank/sharktank/models/llama/testing.py
archana-ramalingam: Deepseek, grok have their own `make_moe_theta()`. This change shouldn't affect them. We need to refactor all these functions into a single `testing.py`.

sharktank/sharktank/models/llm/llm.py
```diff
@@ -377,14 +383,25 @@ def __init__(
...
+        is_moe_block = block_index in config.moe_layers
+        # Always define add_residual
+        self.add_residual = False
+        if is_moe_block:
+            (
+                score_experts,
+                moe_activation,
+                self.add_residual,
+                normalize_experts,
+            ) = moe_func_map[config.hp.model_arch]
```
archana-ramalingam, line 404: Can move this outside the `if is_moe_block:`, so that `self.add_residual` is visible later on.

sharktank/sharktank/models/llm/llm.py
```diff
@@ -407,6 +424,8 @@ def __init__(
                 FFN(
                     theta=theta,
                     fake_quant=fake_quant,
+                    activation_fn=activation_fn,
+                    add_residual=config.ffn_add_residual,
```
archana-ramalingam, line 428: Same comment for `add_residual`. Ideally we want all models to fetch `add_residual` from config, till then self.add_residual will be `config.ffn_add_residual` or `self.add_residual` in this script.

sharktank/sharktank/layers/ffn_moe_block.py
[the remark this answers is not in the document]
vivekkhandelwal1: Hi Boian,sorry but I am confused by your comment. Can you please clarify whether we need this or not?

sharktank/sharktank/layers/configs/llm_configs.py
vivekkhandelwal1: It seems you have added this change.

sharktank/sharktank/layers/ffn_block.py
vivekkhandelwal1: Hi Archana, I took a look at the Llama4 code and the residual is added after the sum up of shared and routed experts in MOE. You can see [here](https://github.com/huggingface/transformers/blob/b369a65480cf1df22b3c853f086b832bc5785a19/src/transformers/models/llama4/modeling_llama4.py#L448) that the addition of residual happens after the execution of MOE block [here](https://github.com/huggingface/transformers/blob/b369a65480cf1df22b3c853f086b832bc5785a19/src/transformers/models/llama4/modeling_llama4.py#L443). Therefore, we have to keep this change.

sharktank/sharktank/models/llama/testing.py
```diff
@@ -101,54 +104,77 @@ def make_attention_block_ffn_theta_v2(
...
-            ),
-            "blk.0.ffn_up_exps.weight": DefaultPrimitiveTensor(
-                name="blk.0.ffn_up_exps.weight",
-                data=make_rand_torch((num_experts, feature_dim * num_experts, ffn_dim)),
-            ),
-            "blk.0.ffn_down_exps.weight": DefaultPrimitiveTensor(
-                name="blk.0.ffn_down_exps.weight",
-                data=make_rand_torch((num_experts, ffn_dim, feature_dim * num_experts)),
-            ),
+            "ffn_norm.weight": DefaultPrimitiveTensor(
```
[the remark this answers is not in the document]
vivekkhandelwal1: Fixed now.

sharktank/sharktank/models/llm/llm.py
vivekkhandelwal1: I have made the change. Please let me know if it's correct or not.

sharktank/sharktank/layers/ffn_moe_block.py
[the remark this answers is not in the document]
sogartar: I meant that you can remove this bit + the `expert_gate` argument as it will be unused.

sharktank/sharktank/layers/configs/llm_configs.py
```diff
@@ -81,17 +90,17 @@ class LlamaHParams:
     n_dense_layers: Optional[int] = None
     route_scale: Optional[float] = None
 
-    # Deepseek MoE config
-    expert_shared_count: Optional[int] = None
```
archana-ramalingam, line 85: I see this change was missed while rebasing.

sharktank/sharktank/layers/configs/llm_configs.py
```diff
@@ -123,7 +132,6 @@ def from_gguf_props(p: dict[str, Any]):
             attention_head_count_kv=_optional_int_prop(
                 p, f"{name_prefix}.attention.head_count_kv", attention_head_count
             ),
-            expert_count=expert_count,
```
archana-ramalingam, line 126: This is necessary for all MOE models. Please add this back.

sharktank/sharktank/layers/configs/llm_configs.py
```diff
@@ -135,6 +143,22 @@ def from_gguf_props(p: dict[str, Any]):
             rope_freq_base=_optional_float_prop(
                 p, f"{name_prefix}.rope.freq_base", default_rope_freq_base
             ),
+            expert_feed_forward_length=_optional_int_prop(
```
archana-ramalingam, line 146: A better way of adding architecture specific hyperparameters is to use this [function](https://github.com/nod-ai/shark-ai/blob/main/sharktank/sharktank/layers/configs/llm_configs.py#L197). 
Can be addressed in a later PR.

sharktank/sharktank/layers/configs/llm_configs.py
```diff
@@ -233,6 +271,11 @@ def get_custom_configs(p: dict[str, Any], name_prefix: str):
         )
         res["attn_head_dim"] = res["qk_nope_head_dim"] + res["qk_rope_head_dim"]
 
+    if name_prefix == "llama4":
+        res["expert_shared_count"] = _optional_int_prop(
```
archana-ramalingam, line 275: This doesn't have to be `_optional_int_prop`. Because the  LlamaHParams has already set `expert_shared_count` to None by default. This holds for all params added in this `get_custom_configs()` fn.

sharktank/sharktank/layers/ffn_block.py
```diff
@@ -27,6 +27,7 @@ def __init__(
         is_gated: bool = True,
         activation_fn: Callable[[torch.Tensor], torch.Tensor] = F.silu,
         fake_quant: bool = False,
+        add_residual: bool = False,
     ):
         """
         add_residual:
```
archana-ramalingam, line 33: Can remove this comment as well.

sharktank/sharktank/layers/ffn_block.py
archana-ramalingam: So we will have to remove all `add_residual` in `ffn_block.py` because this op has to happen at the end of ffn block in llm.py.

sharktank/sharktank/models/llm/llm.py
archana-ramalingam: Shouldn't have added my comment here. add_residual has to go anyway. Refer to comment below.

sharktank/sharktank/models/llm/llm.py
```diff
@@ -337,6 +338,11 @@ def __init__(
         config: LlamaModelConfig,
         fake_quant: bool = True,
     ):
+        activation_fn = (
+            ACT2FN[config.activation_fn]
```
archana-ramalingam, line 342: config.activation_fn and config.ffn_add_residual are not being set in llm_configs.py for llama4 specifically. Are we addressing this in another PR or the default values align with llama4?

sharktank/sharktank/models/llm/llm.py
```diff
@@ -390,6 +396,12 @@ def __init__(
                 False,
                 True,
             ),
+            "llama4": (
+                torch.nn.functional.sigmoid,
+                activation_fn,
+                True,
```
archana-ramalingam, line 402: This is the local equivalent of `config.ffn_add_residual`. Make sure this is the value llama4 requires. Since `config.ffn_add_residual` is not plumbed through/set in llm_configs.py, we can drop using it for now and stick to `self.add_residual` to avoid any confusion.

sharktank/sharktank/layers/configs/llm_configs.py
sogartar: But if we read with `_int_prop(...)` and the value is missing it will raise.

sharktank/sharktank/models/llm/llm.py
sogartar: It is probably better to remove the `ffn_add_residual` option from the config. Hardcode it here as is done for the other models. We can add it in another PR for all model variants.

sharktank/sharktank/layers/configs/llm_configs.py
archana-ramalingam: Since we have a condition that checks for llama4 architecture, it should work. 
If the model follows llama4 architecture, the model is expected to have this param.

sharktank/sharktank/layers/configs/llm_configs.py
vivekkhandelwal1: Done.

sharktank/sharktank/models/llm/llm.py
vivekkhandelwal1: `config.activation_fn` is not required since we support only the language version of the Llama 4 model which has a pre-determined activation function.

vivekkhandelwal1:
> > @sogartar @archana-ramalingam I have addressed the majority of the comments. The CI is passing since I have removed 2 lines of code changes from [here](https://github.com/nod-ai/shark-ai/compare/main...users/sogartar/llama4?diff=split&w#diff-003e19a39f8d1ebd5c9f5eade128e8d3652d2526f06145730ef1c555fa9a5b17R84-R93), which was the primary cause for CI failure. I am not sure atm whether we need those changes or not for Llama 4.
> > I will test the MOE implementation with the Llama4's implementation, till then we will not merge this PR.
> 
> So, I have verified the numerics and for Llama4 Moe it works with those changes in but only for the `PreGatherFFNMOE`, not for the Dense one. Now, I need to add those changes, conditional on the Llama 4 model. But I don't see a way to identify whether the call is made for which model. Shall I add a new arg for that?
> 
> @sogartar @archana-ramalingam

I had made a change for this by adding an argument. You may review the patch now, and I can make the changes accordingly.

sharktank/sharktank/models/llama/testing.py
```diff
@@ -84,54 +85,66 @@ def make_attention_block_ffn_theta_v2(
...
+        dtype=dtype,
+    )
+    res_dict.update(attention_theta.tree)
+    moe_theta = make_random_moe_block_theta(
+        in_dim=config.hp.embedding_length,
+        expert_hidden_dim=config.hp.expert_feed_forward_length,
+        num_experts=config.hp.expert_count,
+        with_ffn_norm=True,
+        num_shared_experts=config.hp.expert_shared_count,
+        shared_expert_hidden_dim=config.hp.expert_shared_feed_forward_length,
```
oyazdanb, line 107: shared_expert_hidden_dim is not not defined in make_random_moe_block_theta here https://github.com/vivekkhandelwal1/shark-ai/blob/llama4/sharktank/sharktank/layers/testing.py#L314

archana-ramalingam: I removed `shared_expert_hidden_dim` as it was similar to `expert_hidden_dim`. But if it's defined in the llama4 config and are diff values, we can add it back. 
PR that removed it: https://github.com/nod-ai/shark-ai/pull/1424/files#diff-01ecddc9cbe8d5f0006740065c61b914537e685b172f96b390bc529c26440d85L259

archana-ramalingam:
> > > > > @sogartar @archana-ramalingam I have addressed the majority of the comments. The CI is passing since I have removed 2 lines of code changes from [here](https://github.com/nod-ai/shark-ai/compare/main...users/sogartar/llama4?diff=split&w#diff-003e19a39f8d1ebd5c9f5eade128e8d3652d2526f06145730ef1c555fa9a5b17R84-R93), which was the primary cause for CI failure. I am not sure atm whether we need those changes or not for Llama 4.
> > > > > I will test the MOE implementation with the Llama4's implementation, till then we will not merge this PR.
> > > > 
> > > > 
> > > > So, I have verified the numerics and for Llama4 Moe it works with those changes in but only for the `PreGatherFFNMOE`, not for the Dense one. Now, I need to add those changes, conditional on the Llama 4 model. But I don't see a way to identify whether the call is made for which model. Shall I add a new arg for that?
> > > > @sogartar @archana-ramalingam
> > > 
> > > 
> > > I had made a change for this by adding an argument. You may review the patch now, and I can make the changes accordingly.
> > 
> > 
> > We should pass `config.hp.model_arch` like we do for [here](https://github.com/nod-ai/shark-ai/blob/main/sharktank/sharktank/models/llm/llm.py#L362) for `PagedLlamaAttentionBlock`. This is assuming the `config.hp.model_arch==llama4`. It's sensible to add a new arg, if it's a specific op like `add_residual`, `normalize_experts` because we might expect this in future llms as well.
> 
> Hi @archana-ramalingam, I have made the changes as requested. Can you please review it now?

Great job Vivek, LGTM after resolving conflicts!

archana-ramalingam (approved):

sharktank/sharktank/models/llama/testing.py
vivekkhandelwal1: I think we don't need that. If  required, I will add it in the follow-up patches.

PR #1516, opened 2025-05-27
Cache Storage to avoid allocate pinned memory for every prefill step and every decode step

dezhiAmd:
- Why
Before this change, storage in pinned memory returned by **for_transfer** are allocated every prefill step and every decode step. Cache storage can significantly reduce the allocation and release of pinned memory thus reduce the latency and improve the through-put

- How
Since logits are discarded every time after prefill step and decode step (and all other storage returned by for_transfer), cache them. The cache including prefill cache and decode cache is created when LlmGenerateService is created.

- Result (Single Server, SPX mode MI300X) 
 
Using script from https://github.com/nod-ai/benchmark_scripts

```
Batch size 8:
Before: "throughput": 0.9365385081863682, "latency_mean": 8.524904807806015
After:   "throughput": 0.953148448882024, "latency_mean": 8.375769250392914
 
Batch size 16:
Before:  "throughput": 1.0481085278749398, "latency_mean": 14.982843790054321
After:    "throughput": 1.1322175557670333, "latency_mean": 13.83236265540123
```

Using internal script:
```
Before:
Throughput Results:
Number of concurrent clients: 6
Total requests completed: 30
Test duration: 22.25 seconds
Average latency: 4.4514s
Min latency: 4.3440s
Max latency: 4.6043s
Throughput: 1.35 requests/second

After:
Throughput Results:
Number of concurrent clients: 6
Total requests completed: 30
Test duration: 20.91 seconds
Average latency: 4.1655s
Min latency: 4.0274s
Max latency: 4.2539s
Throughput: 1.43 requests/second
```

### DIFF
--- a/shortfin/python/shortfin_apps/llm/components/batcher.py
+++ b/shortfin/python/shortfin_apps/llm/components/batcher.py
@@ -404,7 +404,9 @@ class PrefillExecutorProcess(LlmExecutorProcess):
         # TODO: Better support in shortfin for h2d. The best way to do it is
         # device dependent.
         int_dtype = sfnp.int64
-        tokens = sfnp.device_array.for_device(self.device0, [self.bs, self.bsl], int_dtype)
+        tokens = sfnp.device_array.for_device(
+            self.device0, [self.bs, self.bsl], int_dtype
+        )
         seq_lens = sfnp.device_array.for_device(self.device0, [self.bs], int_dtype)
         seq_block_ids = sfnp.device_array.for_device(
             self.device0, [self.bs, block_count], int_dtype
@@ -555,7 +557,9 @@ class DecodeExecutorProcess(LlmExecutorProcess):
         # device dependent.
         int_dtype = sfnp.int64
         tokens = sfnp.device_array.for_device(self.device0, [self.bs, 1], int_dtype)
-        start_positions = sfnp.device_array.for_device(self.device0, [self.bs], int_dtype)
+        start_positions = sfnp.device_array.for_device(
+            self.device0, [self.bs], int_dtype
+        )
         seq_lens = sfnp.device_array.for_device(self.device0, [self.bs], int_dtype)
         seq_block_ids = sfnp.device_array.for_device(
             self.device0, [self.bs, block_count], int_dtype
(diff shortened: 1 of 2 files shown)


stbaione:
What were the parameters of the benchmark that yielded 30%?

dezhiAmd:
> What were the parameters of the benchmark that yielded 30%?

I followed the playbook

shortfin/python/shortfin_apps/llm/components/batcher.py
```diff
@@ -591,3 +649,30 @@ async def get_results(self, logits, indices, req_count):
...
+        )
+
+        if not transfer:
+            return buffers
+
+        logits, indices = buffers
+
+        # Lazily initialize host buffers only once
+        if DecodeExecutorProcess._host_logits_init == 0:
+            DecodeExecutorProcess._host_logits = logits.for_transfer()
```
rsuderman, line 665: This is not safe. Transfer buffers between invocations are not guaranteed to be the same size.

shortfin/python/shortfin_apps/llm/components/batcher.py
```diff
@@ -375,6 +357,14 @@ async def run(self):
 class PrefillExecutorProcess(LlmExecutorProcess):
     """Executes a prefill batch."""
 
+    tokens_host_cache: dict[tuple[int, int], sfnp.device_array] = {}
```
rsuderman, line 360: These would be globally shared between all executor processes with no cleanup. If we stopped the service then restarted it, then it would have invalid contents. We should not have static global state carried between processes as this would cause very bad outcomes in multi-instance.

rsuderman (changes requested):

stbaione:
> > What were the parameters of the benchmark that yielded 30%?
> 
> I followed the playbook

Was it with the internal script or the chai script? If chai script, which batch size?

shortfin/python/shortfin_apps/llm/components/batcher.py
dezhiAmd: You are absolutely right. Logits size is determined by [bs, bsl], I have replaced a single logits_cache with a dictionary of logits which use the shape [bs, bsl] as the key to retrieve from the cache. The performance improvement can remain and now it is safe

PR #1524, opened 2025-05-28
[sharktank] move and rename get_iree_compiler_flags

sogartar:
Rename to get_iree_compiler_flags_from_object for a more descriptive name.

### DIFF
--- a/sharktank/sharktank/utils/iree.py
+++ b/sharktank/sharktank/utils/iree.py
@@ -40,6 +40,47 @@ halelementtype_map = {
 }
 
 
+def get_iree_compiler_flags(
+    iree_hal_target_device: str,
+    iree_hal_local_target_device_backends: list[str] | None = None,
+    iree_hip_target: str | None = None,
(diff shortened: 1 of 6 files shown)


sharktank/sharktank/utils/iree.py
```diff
@@ -40,6 +40,47 @@
...
+            for i in range(device_count)
+        ]
+    if iree_hal_target_device.startswith("local"):
+        res += [
+            f"--iree-hal-local-target-device-backends={v}"
+            for v in iree_hal_local_target_device_backends
+        ]
+        res += ["--iree-llvmcpu-target-cpu=host"]
+    elif iree_hal_target_device.startswith("hip"):
+        res += [f"--iree-hip-target={iree_hip_target}"]
```
Alex-Vasile, line 65: Should raise an error on an else if they pass a different target deivce.

sogartar: Done.

PR #1529, opened 2025-05-28
[LLM Server] Merge `Greedy` strategies into one class

stbaione:
# Description

This PR merged the `Greedy` and `MultiGreedy` token_selection_strategy classes into a single `IndependentTokenSelectionStrategy` class.

This accomplishes a few things:

- Decreases duplicated code. Will just make it easier on us not having to update as many classes, especially since they do basically the same thing.
- Step towards combining `BeamSearch` with this class into a single class, further reducing spots to have to update.
- Clarifies class name. When we wrote those classes they really were for `greedy` token selection. However, we quickly found that they fit the use case well for things like `top_p` and `top_k` without `beam_search` in multi-hypothesis generation. From then, name `Greedy` in the names just causes confusion.
- By naming it `IndependentTokenSelectionStrategy`, I'm highlighting the difference between this and beam_search. Each beam independently selects its next token to form our next set of beams. In beam_search, each beam returns a candidate of selections, where we dependently select the next set of beams based on the global score.

I also add a smoke test for greedy multi-hypothesis and a smoke test for beam_search. `top_k` and `top_p` will require some more effort, since I'll need to do a probabilistic comparison with the logits and some expected response.

## Follow Up
- PR will be pushed shortly which removes `--token_selection_strategy` from the CLI. This is easier now since they're one class. Instead, there will be a `--use_beam_search` arg that will determine if we should use `IndependentBeam` or `BeamSearchBeam`.


shortfin/python/shortfin_apps/llm/components/token_selection_strategy/independent_token_selection_strategy.py
```diff
@@ -0,0 +1,211 @@
...
+        raise NotImplementedError("IndependentBeam does not track a score")
+
+    def normalize_score(self, value):
+        raise NotImplementedError("IndependentBeam does not track a score")
+
+    def update_final_score(self):
+        raise NotImplementedError("IndependentBeam does not track a score")
+
+
+class IndependentTokenSelectionStrategy(BaseTokenSelectionStrategy):
```
rsuderman, line 93: Is there a reason we need a single hypothesis strategy? Presumably we could just abandon single hypothesis and always do beam search with a width of one? Most of it just means the loop turns to a single interation and likely wouldn't affect performance.

rsuderman (commented):
Main question is whether its worth maintaining a separate single vs multi beam selection. Really we are just deciding how to take N hypothesis measures and make M new requests. The greedy vs multi-greedy is just a specialization of that.

If we wipe out the independent one and instead just rework how we process beams / hypothesis we can simplify to an interface of just specializing how we generate the new requests.

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/independent_token_selection_strategy.py
stbaione: I think that we're thinking of the same end goal. `MultiGreedy` was a really bad name for that class, because it's used for more than `greedy` token selection, despite what the name implies.

Really it's a `MultiHypothesis` class, that just scores locally within a beam, instead of globally across all beams.

So, `top_p` for example, goes through our `MultiGreedy` class. `top_p + beam_search` would be something different, but ultimately hurts diversity, and goes through the `BeamSearch` class.

So, it's the same idea of always doing `beam_search` with a width of 1, but with this class instead, since most sampling methods don't do beam_search by default.

My end goal is to merge `BeamSearchTokenSelectionStrategy` class into this class, and rename it something like `TokenSelector`. I was just thinking of an optional function to call is `--use_beam_search` was set, but I could create an abstraction, probably a `BeamSearchBeamGroup`.

rsuderman (approved):

PR #1530, opened 2025-05-29
[mlir_kernel] Use mlir_kernel for injecting Wave kernels

aviator19941:
This PR adds a Wave sdpa kernel invocation and uses the existing sharktank `mlir_kernel` decorator to inject the Wave asm that is generated from the `wave_compile` API. The tests passing depend on https://github.com/iree-org/iree-turbine/pull/787 being merged first.

It is a refactor of https://github.com/nod-ai/shark-ai/pull/1378.

### DIFF
--- a/sharktank/sharktank/kernels/mlir_kernel.py
+++ b/sharktank/sharktank/kernels/mlir_kernel.py
@@ -142,7 +142,7 @@ class MLIRTensor:
 
 class MLIRSpec:
     """
-    A class representing a MLIR Jinja2 template.
+    A class representing a MLIR Jinja2 template or a MLIR asm.
 
     The `subs` dictionary contains additional substitutions passed to the jinja
     template generator.
@@ -153,11 +153,15 @@ class MLIRSpec:
     """
 
     mlir: str
+    target_kernel_name: str
     subs: dict
+    use_jinja: bool
 
-    def __init__(self, mlir: str, subs: dict = {}):
+    def __init__(self, mlir: str, target_kernel_name: str = None, subs: dict = {}, use_jinja: bool = True):
         self.mlir = mlir
+        self.target_kernel_name = target_kernel_name
         self.subs = subs
+        self.use_jinja = use_jinja
 
 
 def mlir_kernel(
@@ -353,17 +357,21 @@ def mlir_kernel(
                 # the mlir spec.
                 if symbol_name is None:
                     # Generate the MLIR spec using jinja.
-                    asm = (
-                        _get_jinja2_env()
-                        .from_string(mlir)
-                        .render(
-                            {
-                                "kernel_name": kernel_name,
-                                **self._get_additional_aliases(dims),
-                                **mlir_spec.subs,
-                            }
+                    if mlir_spec.use_jinja:
+                        asm = (
+                            _get_jinja2_env()
+                            .from_string(mlir)
+                            .render(
+                                {
+                                    "kernel_name": kernel_name,
+                                    **self._get_additional_aliases(dims),
+                                    **mlir_spec.subs,
+                                }
+                            )
                         )
-                    )
+                    else:
+                        asm = mlir_spec.mlir
+                        kernel_name = mlir_spec.target_kernel_name
                     try:
                         module_op = Operation.parse(asm, context=kb.context)
                     except MLIRError as e:
--- a/sharktank/sharktank/kernels/wave/attention.py
+++ b/sharktank/sharktank/kernels/wave/attention.py
@@ -1,0 +1,137 @@
+# Copyright 2025 Advanced Micro Devices, Inc.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+from sharktank.kernels.base import *
+from iree.turbine.kernel.wave.templates.vanilla_attention import (
+    get_bhsd_attention_kernel,
+)
+from iree.turbine.kernel.wave.scheduling.schedule import SchedulingType
+from iree.turbine.kernel.wave.compile import wave_compile, WaveCompileOptions
+from iree.turbine.kernel.wave.templates.attention_common import AttentionShape
+from iree.turbine.kernel.wave.constraints import MMAType
+from iree.turbine.kernel.wave.utils.general_utils import (
+    get_default_scheduling_params,
+)
+from iree.turbine.kernel.wave.utils.run_utils import (
+    set_default_run_config,
+)
+from sharktank.kernels.mlir_kernel import *
+
+__all__ = [
+    "get_wave_flash_attention_asm"
+    "wave_bhsd_flash_attention",
+]
+
+def get_wave_flash_attention_asm(target_function_name: str, shape: AttentionShape, mfma_variant: list[MMAType], dynamic_dims: bool, is_causal: bool = False, is_custom_mask: bool = False,) -> str:
+    (
+        base_attention_func,
+        hyperparams,
+        dynamic_symbols,
+        dynamic_symbols_map,
+    ) = get_bhsd_attention_kernel(
+        shape,
+        mfma_variant,
+        dynamic_dims,
+        is_causal=is_causal,
+        is_custom_mask=is_custom_mask,
+    )
+    hyperparams.update(get_default_scheduling_params())
+    options = WaveCompileOptions(
+        subs=hyperparams,
+        schedule=SchedulingType.NONE,
+        dynamic_symbols=dynamic_symbols,
+        dynamic_symbols_map=dynamic_symbols_map,
+        waves_per_eu=2,
+        denorm_fp_math_f32="preserve-sign",
+        func_name=target_function_name,
+        compile_to_mlir=True,
+    )
+    options = set_default_run_config(options)
+    base_attention = wave_compile(options, base_attention_func)
+
+    asm = base_attention.asm
+    return asm
+
+# Wave Attention Kernels
+# Each kernel is put into its own class to create a namespace for it
+B = StaticDim.B # batch_size
+H = StaticDim.H # num_query_heads
+M = StaticDim.M # query_seq_len
+N = StaticDim.N # head_size_kv
+K1 = StaticDim.K1 # head_size
+K2 = StaticDim.K2 # kv_seq_len
+
+F16 = Dtype.F16
+F32 = Dtype.F32
+
+@mlir_kernel(
+    inputs=(
+        MLIRTensor[
+            B,
+            H,
+            M,
+            K1,
+            F16
+        ],
+        MLIRTensor[
+            B,
+            H,
+            K2,
+            K1,
+            F16
+        ],
+        MLIRTensor[
+            B,
+            H,
+            K2,
+            N,
+            F16
+        ],
+        MLIRTensor[
+            B,
+            H,
+            M,
+            N,
+            F32
+        ],
+    ),
+    results=(
+        MLIRTensor[B, H, M, N, F32],
+    ),
+)
+def wave_bhsd_flash_attention(
+    q, k, v, c, result=None
+):
+    batch_size, num_heads, q_s, q_d = q.type.shape
+    v_batch_size, num_heads_kv, v_s, v_d = v.type.shape
+    shape = AttentionShape(
+        batch_size=batch_size,
+        num_query_heads=num_heads,
+        num_kv_heads=num_heads_kv,
+        query_seq_len=q_s,
+        head_size_kv=v_d,
+        head_size=q_d,
+        kv_seq_len=v_s,
+    )
+    mfma_variant = (MMAType.F32_32x32x8_F16, MMAType.F32_32x32x8_F16)
+    dynamic_dims = False
+    is_causal = True
+    is_custom_mask = False
+    i_type_str = "f16"
+    o_type_str = "f32"
+
+    target_function_name = f"wave_flash_attention_{batch_size}_{num_heads}_{q_s}_{v_d}_{i_type_str}_{o_type_str}"
+
+    asm = get_wave_flash_attention_asm(
+        target_function_name,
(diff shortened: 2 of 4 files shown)


sharktank/sharktank/kernels/mlir_kernel.py
```diff
@@ -153,11 +153,15 @@ class MLIRSpec:
     """
 
     mlir: str
+    target_kernel_name: str
```
aviator19941, line 156: I'm not sure if there is a better way to deal with setting `func_name` in the `wave_compile` call and making sure it uses the specified `func_name` instead of the generated kernel_name by `get_kernel_name`. By default, the func_name is set to `isolated_benchmark`, but can be specified by the user.

However, the dims and dtypes that are used in `self._get_kernel_name(func.__name__, dims, dtypes)` are defined after the Wave asm is created, so it would be repetitive to update the func_name again in the generate function. This was the simplest way I could think of to solve that problem, but maybe @Groverkss would know a better way?

Groverkss: We don't want to do any changes to mlir_kernel and kernels/base.py

What you want to do instead is create a util.func with the kernel name that the mlir_kernel wants and then func.call the wave kernel. This is really important, because the kernel name is also used to check if we have already generated a kernel for this specific mlir_kernel. We cannot hook into the wave caching mechanism, so we cannot use the wave kernel name.

You want to generate the MLIRSpec as:

```
mlir = "module {" + asm + """
util.func private @{{kernel_name}}(%q : !q, %k : !k, %v : !v, %c : !c) -> !result {
  %result = func.call {{wave_kernel_name}} (%q, %k, %v, %c) : (!q, !k, !v, !c) -> !result
  util.return %result : !result
}
"""
```

You want to do something like this.

This way, the kernel_name is unique and is used for caching, but the wave kernel name can be anything, as during module merging, the collision renaming will catch it.

I don't think we can land it without the kernel name being `kernel_name`, because otherwise caching will not work.

Groverkss (changes requested):

Groverkss (approved):
LGTM, nice!

.github/workflows/ci-sharktank.yml
```diff
@@ -197,6 +197,59 @@ jobs:
             --durations=0 \
             --timeout=800
 
+  test_wave_integration:
```
Groverkss, line 200: Tagging @ScottTodd for review on CI, I'm not very familiar with what should be put in CI

sharktank/sharktank/kernels/wave/attention.py
```diff
@@ -0,0 +1,150 @@
...
+from iree.compiler.ir import (
+    Module,
+    Context,
+)
+
+__all__ = [
+    "get_module_body",
+    "get_wave_flash_attention_asm",
+    "wave_bhsd_flash_attention",
+]
```
Groverkss, line 31: Do we need to expose all of these?

Groverkss (commented):
LGTM, I'm not familiar with CI testing though, is there no CI which already has pytorch rocm installed?

.github/workflows/ci-sharktank.yml
aviator19941: Yeah I couldn't find a CI which already has pytorch rocm installed

ScottTodd: Why does this require pytorch rocm?

It's okay in shark-ai to pull in torch with rocm support for comparisons against a reference, as in https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/ci_eval_short.yaml, but I don't want to artificially limit ourselves to only being able to import models on AMD GPUs, or only run on AMD GPUs. AMD's tech stack should also support CPU and Vulkan, ideally also Metal, CUDA, etc. Note that IREE and iree-turbine are the most general, supporting all backends, while shark-ai is AMD focused.

This may also involve https://github.com/iree-org/iree-turbine/pull/587#discussion_r2013067870, so let's make sure we know specifically what is needed technically. One of the main reasons we built all this compiler tech was to be able to introduce more deliberate boundaries between ML frameworks and deployment environments.

.github/workflows/ci-sharktank.yml
```diff
@@ -197,6 +197,59 @@ jobs:
...
+    name: "Wave Integration Tests"
+    strategy:
+      matrix:
+        python-version: [3.11]
+        runs-on: [linux-mi300-1gpu-ossci-nod-ai]
```
ScottTodd, line 205: I was hoping that ci-sharktank.yml would only include unit tests and could run quickly on github-hosted runners without needing any special machine setup (e.g. populated caches). Keeping unit test workflows deliberately simple helps ensure that development on the project does not require special access, GPU machines, or other complicated requirements. This continues the trend that I want to break.

My preference would be to build off of https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/pkgci.yml for any integration tests workflows that need more exotic resources (GPUs, large files). The tracking issue with more information is here: https://github.com/nod-ai/shark-ai/issues/584

.github/workflows/ci-sharktank.yml
```diff
@@ -197,6 +197,59 @@ jobs:
...
+      fail-fast: false
+    runs-on: ${{matrix.runs-on}}
+    defaults:
+      run:
+        shell: bash
```
ScottTodd, line 210: This bash as the default shell is only needed if a workflow runs on Windows too.

.github/workflows/ci-sharktank.yml
aviator19941: @ScottTodd Wave currently only compiles for AMD GPUs and not CPU, which is why it requires pytorch rocm. The Wave API used in this PR will not work unless it is using pytorch rocm, which is expected. However, there is no hard design that will stop Wave from supporting other GPU's in the future.

I'll try to make another job in https://github.com/nod-ai/shark-ai/blob/main/.github/workflows/pkgci.yml and make it specific for Wave tests.

sharktank/sharktank/kernels/wave/attention.py
aviator19941: Nope, good catch. Just wave_bhsd_flash_attention.

.github/workflows/ci-sharktank.yml
aviator19941: Done

.github/workflows/ci-sharktank.yml
ScottTodd: > We need it to run e2e for Wave

Be **specific**. What is needed, where, and why?

> also to detect hw/target architecture

IREE and shark both use ahead-of-time compilation, and you can detect that information using the ROCm SDK instead of PyTorch.

raikonenfnu: >Be specific. What is needed, where, and why?

Just pointing out high level when we need it, remember I didn't write the code/PR here just giving extra context from Wave perspective.

> IREE and shark both use ahead-of-time compilation, and you can detect that information using the ROCm SDK instead of PyTorch.
Agreed, that's why I said we may get away without installing pytorch-rocm in my previous comment

stellaraccident (changes requested):
Sounds like people are guessing about things that have real answers. Blocking merge until resolved. I'll release my lock when I feel this is done correctly.

sharktank/tests/kernels/attention_wave_test.py
```diff
@@ -0,0 +1,47 @@
...
+
+import torch
+
+from sharktank.kernels.wave.attention import wave_bhsd_flash_attention
+from iree.turbine.kernel.wave.utils.reference_kernel_utils import (
+    scaled_dot_product_attention_bhsd,
+)
+
+
+@unittest.skipIf(not torch.cuda.is_available(), "CUDA is required for Wave tests")
```
raikonenfnu, line 22: @ScottTodd if you are still curious, to run e2e on this test, we would need pytorch rocm. As of now, Wave only works on ROCm and on top of that Wave expects that tensors fed into Wave already lives in GPU => we'd need pytorch-rocm to generate and move tensors to GPU. 

However, as I previously mentioned, the main use case for Wave integration here is for injecting IR. Hence, I reckon a more apt test is to have a LIT test where we mix wave kernel and another torch Op e.g torch.add or torch.ReLU. If the test is updated to this LIT test, we would not need pytorch-rocm dependency anymore. 

CC: @aviator19941

sharktank/tests/kernels/attention_wave_test.py
```diff
@@ -0,0 +1,47 @@
...
+        accum_dtype = torch.float32
+        q = torch.randn([4, 32, 128, 128], device="cuda", dtype=dtype)
+        k = torch.randn([4, 32, 128, 128], device="cuda", dtype=dtype)
+        v = torch.randn([4, 32, 128, 128], device="cuda", dtype=dtype)
+        output = torch.zeros([4, 32, 128, 128], device="cuda", dtype=accum_dtype)
+        result = wave_bhsd_flash_attention(q, k, v, output)
+
+        # Tolerances are empirical and results are not expected to match exactly.
+        ref = scaled_dot_product_attention_bhsd(q, k, v, is_causal=True)
+        torch.testing.assert_close(result, ref, atol=atol, rtol=rtol)
```
zjgarvey, line 43: Testing eager-mode custom op execution requires pytorch with rocm/cuda, since the backend used by iree-compile/iree-run-module is determined by the `torch.device` of input pytorch tensors.

If we don't want to add gpu ci specifically for this one case, we could:

1. upstream this custom op to iree-turbine, which already does e2e eager gpu testing against pytorch
2. only test aot (e.g. write an nn module with this op, export it to mlir, and just verify the func name and whatever else you want matches expectations). 

cc: @ScottTodd

zjgarvey: Ah, just refreshed the page and saw https://github.com/nod-ai/shark-ai/pull/1530#issuecomment-2960657677 

please disregard.

aviator19941: No worries, I agree with you. I think testing aot is the way to go here, thanks!

sharktank/tests/kernels/attention_wave_test.py
ScottTodd: Merging thread with @zjgarvey 's comment:

> Testing eager-mode custom op execution requires pytorch with rocm/cuda, since the backend used by iree-compile/iree-run-module is determined by the `torch.device` of input pytorch tensors.
> 
> If we don't want to add gpu ci specifically for this one case, we could:
> 
>   1. upstream this custom op to iree-turbine, which already does e2e eager gpu testing against pytorch
> 
>   2. only test aot (e.g. write an nn module with this op, export it to mlir, and just verify the func name and whatever else you want matches expectations).

Thanks, this is what I was reaching for, and what I would typically hope to see in a tracking issue, RFC, or some other design discussion prior to code being written and pull requests getting sent - "weeks of coding can save you hours of planning". The specific architecture across these projects and the PyTorch integration matters significantly, as the design space is large and full of pitfalls. I also want us to avoid painting ourselves into a corner and limiting our deployment reach artificially, as "we" (our team at AMD, AMD as a whole, IREE contributors within the Linux foundation, etc.) benefit from having the flexibility to target a variety of targets in a variety of ways.

We are also restructuring both ROCm and framework integrations of it (e.g. PyTorch) and are expanding support to new devices and platforms (e.g. Windows) so I want to know very specifically what these projects need from that foundation they are building on. If there are workarounds like https://github.com/iree-org/iree/pull/20991#discussion_r2132413709, I want us to flag them early and discuss across the team, so we can plan long term solutions. The IREE and SHARK projects historically have had no dependency on kernel libraries and should continue to support efficient execution in their absence (e.g. with just a kernel mode driver for your GPU and no user-mode ROCm install, or only a limited install without the BLAS/math libraries you should be able to run an IREE/SHARK program).

If the integration here is limited to injecting IR (no GPU requirement) and interop with torch device tensors in tests, that is acceptable. I want to keep the line between AOT and JIT as clear as we can make it, ideally building off of existing PyTorch concepts like torch.compile and torch.export. We should still be able to export and compile a program on a CPU-only machine and then run that program with good performance on a GPU machine - that is critical for deployment at scale and keeping the development model democratized so that anyone (even without datacenter GPU access) can contribute. For research, experimentation, development, tuning, and other activities, the line _can_ blur between JIT/AOT and pull in more complicated dependencies, but it is important to keep that part of the project strongly isolated from the AOT path.

sharktank/sharktank/kernels/wave/utils.py
```diff
@@ -0,0 +1,27 @@
...
+    to private, so that it gets removed when inlined.
+    """
+    block = module.operation.regions[0].blocks[0]
+    ops_asm = []
+    for op in block.operations:
+        if op.operation.name == "func.func":
+            op.attributes["sym_visibility"] = StringAttr.get("private")
+        ops_asm.append(op.get_asm())
+
+    return "\n".join(ops_asm)
```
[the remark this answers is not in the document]
aviator19941: The `wave_compile()` call is located here: https://github.com/nod-ai/shark-ai/pull/1530/files#diff-0dfb9bd5ceab215e003ac11cdf3a07d1ea3ced4d8cc025cc669a7c01e26a8d7dR66.

Yeah I can add a unit test for it. It's for formatting the Wave asm string so it can be concatenated properly to the template mlir_kernel string.

sharktank/sharktank/kernels/wave/attention.py
```diff
@@ -0,0 +1,138 @@
...
+# Each kernel is put into its own class to create a namespace for it
+B = StaticDim.B  # batch_size
+H = StaticDim.H  # num_query_heads
+M = StaticDim.M  # query_seq_len
+N = StaticDim.N  # head_size_kv
+K1 = StaticDim.K1  # head_size
+K2 = StaticDim.K2  # kv_seq_len
+
+F16 = Dtype.F16
+F32 = Dtype.F32
```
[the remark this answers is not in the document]
aviator19941: Done

stellaraccident (commented):
Removing my block and deferring to other reviews.

PR #1531, opened 2025-05-29
[sharktank] Add toy deepseek eager perplexity test

archana-ramalingam:
- Add toy deepseek eager perplexity test


sharktank/tests/evaluate/baseline_perplexity_scores.json
```diff
@@ -635,5 +635,14 @@
...
     "mean_perplexity": 14.87456
+  },
+  "deepseek_v3_torch": {
+    "perplexities": [
+      512.066162,
+      8907.019531,
+      4864.950684,
+      508.603119
+    ],
+    "mean_perplexity": 3698.159874
```
Alex-Vasile, line 646: Test is currently failing:
`FAILED sharktank/tests/evaluate/perplexity_torch_test.py::PerplexityTest::test_deepseek_v3 - AssertionError: np.float64(3698.159874) != np.float64(1830.414139) within 0.5 delta (np.float64(1867.745735) difference) : Current perplexity deviates baseline by -1867.745735`

sharktank/sharktank/evaluate/perplexity_torch.py
```diff
@@ -171,32 +232,41 @@ def get_logits(self, skip_decode: bool) -> torch.tensor:
...
+            logger.debug(f" Prompts for Evaluation:")
+            for idx, prompt in enumerate(test_prompts):
+                logger.debug(
+                    f" Prompt {idx}: \nTokens: {prompt.encode()}\nToken ids: {self.token_ids[idx]}\n"
+                )
 
-        self.max_prompt_length = max(seq_lens)
+            self.page_cache_size = (
+                len(self.token_ids[0]) // self.generator.model.config.block_seq_stride
+            ) * len(test_prompts) + 1
```
Alex-Vasile, line 265: `self.start` and `self.page_cache_size` are being set in two different places. The default for `self.start` (set in init) is for a full mode while the default for `self.page_cache_size` is for the toy model. This and having these value be set in different places is confusing, should be consolidated more.

.github/workflows/ci-llama-large-tests.yaml
```diff
@@ -70,7 +70,7 @@ jobs:
           source ${VENV_DIR}/bin/activate
           pytest sharktank/tests/models/llama/benchmark_amdgpu_test.py \
             -v -s \
-            --run-nightly-llama-tests \
+            --run-nightly-tests \
```
Alex-Vasile, line 73: Should we switch the pytorch version on this and the ci-llama-quick-tests.yaml to also use the rocm version?

sharktank/tests/evaluate/perplexity_torch_test.py
```diff
@@ -63,6 +66,10 @@ def run_and_check_perplexity(self):
         )
         gc.collect()
 
+    @pytest.mark.xfail(
```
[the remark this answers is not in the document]
archana-ramalingam: Done

.github/workflows/ci-llama-large-tests.yaml
archana-ramalingam: We need torch+rom only for eager, most other tests don't need it. Reason to fall back on cpu version is that, torch+rocm is a large package and takes time and space to install. 
Ideally we want to setup a single env and reuse it across sharktank CIs, I beleive shortfin uses this approach to some extent.

sharktank/sharktank/evaluate/perplexity_torch.py
archana-ramalingam: Makes sense. Moved setting them under respective model versions.

Issue #1509, opened 2025-05-22
Llama8B Perplexity Issues on TOM


Alex-Vasile:
Mean perplexity has gone from `17.9` to `772,120.1`. Run command is:
```
python -m sharktank.evaluate.perplexity_torch \
  --irpa-file=/shark-dev/8b/instruct/weights/llama3.1_8b_instruct_fp16.irpa \
  --tokenizer-config-json=/shark-dev/8b/instruct/tokenizer_config.json \
  --num-prompts=32 \
  --device='cuda:2'
```

Bisected to https://github.com/nod-ai/shark-ai/pull/1481.

- [x] There are no eager perplexity tests that run on PRs. We should add one to run Llama 8B on the 1GPU runner, similar to how we have it for IREE perplexity.

Groverkss:
Fixed by https://github.com/iree-org/iree-turbine/pull/887

Issue #1495, opened 2025-05-21
--use-attention-mask flag negatively impacts perplexity results


Alex-Vasile:
Compiling an unsharded model with and without the `--use-attention-mask` flag and then running perplexity on it results in different values. Below are results for LLama 8B.
```
Llama3 8b, bs=4, IREE ppl

unsharded 
with attn mask - 13.356212
{
  "perplexities": [
    7.410479,
    22.796532,
    10.047695,
    13.17014
  ],
  "mean_perplexity": 13.356212
}
w/o attn mask - 13.016442
{
  "perplexities": [
    7.182783,
    21.818655,
    9.878222,
    13.18611
  ],
  "mean_perplexity": 13.016442
}
```
@archana-ramalingam 

A difference was also found with a toy deepseek model 2263 vs 2266 (model weights are random so actual values are meaningless). For deepseek only the prefill logits were affected, the decode logits were identical.

Alex-Vasile:
Update: Does not appear cause a difference in eager mode. This is a tracing/IREE issue. Working on getting a smaller repro.

Alex-Vasile:
Actual issue was that the perplexity test was grabbing the wrong prefill logits, specifically it was looking at the padded logits, whose values are undefined and so would likely change since the two approach (with and without the flag) use a slightly different mask.

archana-ramalingam:
We are seeing this on eager perplexity test too. 
Llama8b fp16, decode logits only match: 0.093%, while prefill matches 100% and perplexity score deviates largely. (Will reproduce eager numbers once regression is fixed ToM.)

PR #1535, opened 2025-05-29
[sharktank] Fix which prefill logits IREE perplexity calculations are using

Alex-Vasile:
The current implementation had two issues:
1. IREE and eager perplexity calculations were using different prefill logits
2. Neither were taking into account the sequence length of a given prompt within the batch when grabbing a slice of the logits.

This PR changes the IREE perplexity calculation to use the correct prefill logits; each entry in the batch the slice corresponding to the final real token is being used to calculate perplexity.

Only fixing the IREE side, the eager side is untouched (and left incorrect) because of #1509 prevent rerunning to get new values. Leaving eager calc as is also allows for #1509 to be more easily fixed.

TODO:
- [x] Update perplexity scores in `baseline_perplexity_scores.json`

For: #1495


sharktank/tests/evaluate/baseline_perplexity_scores.json
```diff
@@ -372,268 +372,268 @@
...
+      5.914172,
+      16.327044,
+      14.681435,
+      6.076033,
+      11.515375,
+      9.710052,
+      10.319663
     ],
-    "mean_perplexity": 14.629444
+    "mean_perplexity": 14.009884
```
Alex-Vasile, line 504: Iree perplexity going down is great to see!

archana-ramalingam: Yes, this is great!

PR #1545, opened 2025-05-30
[sharktank] Fix eager pipeline parallelism regression for toy deepseek

archana-ramalingam:
- Fix eager pipeline parallelism regression for toy deepseek

### DIFF
--- a/sharktank/tests/evaluate/perplexity_torch_test.py
+++ b/sharktank/tests/evaluate/perplexity_torch_test.py
@@ -14,6 +14,8 @@ import gc
 from sharktank.evaluate import perplexity_torch
 from sharktank.utils.testing import (
     is_nightly,
+    is_llama_8b,
+    is_deepseek,
 )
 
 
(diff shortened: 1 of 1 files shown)


sharktank/tests/evaluate/perplexity_iree_test.py
```diff
@@ -160,8 +163,12 @@ def test_deepseek_v3(self):
         self.prepare_argv(extra_args=(f"--use-toy-model",))
         self.run_and_check_perplexity()
 
-    @pytest.mark.skip(reason="https://github.com/iree-org/iree/issues/20914")
-    @is_nightly
+    @pytest.mark.xfail(
```
Alex-Vasile, line 166: Are you able to use the xfail from #1521 so we can make sure it xfails for the same reason?

archana-ramalingam: Will have to try 405b and gen numbers in a diff PR.

Alex-Vasile (approved):

PR #1547, opened 2025-05-30
[sharktank] Fix deepseek sharding

archana-ramalingam:
Add leftover changes to fix deepseek sharding

### DIFF
--- a/sharktank/sharktank/types/sharding.py
+++ b/sharktank/sharktank/types/sharding.py
@@ -166,20 +166,25 @@ class Conv2DSplitOutputChannelSharding(ThetaLayerSharding):
 
 
 class FFNSharding(ThetaLayerSharding):
-    def __init__(self, shard_count: int):
+    def __init__(self, shard_count: int, suffix: str = ""):
         super().__init__()
         self.shard_count = shard_count
+        self.suffix = suffix
 
     def theta_sharding(self) -> ThetaSharding:
+        ffn_gate = "ffn_gate" + self.suffix
+        ffn_up = "ffn_up" + self.suffix
+        ffn_down = "ffn_down" + self.suffix
+
         return ThetaSharding(
             {
-                "ffn_gate": LinearSplitParallelWeightAndBiasSharding(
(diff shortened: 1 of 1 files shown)


sogartar (commented):
The reason that
[sharktank/tests/layers/mixture_of_experts_block_test.py::MoeBlockTest::testTensorParalle](https://github.com/nod-ai/shark-ai/actions/runs/15355768194/job/43214371219?pr=1547#step:6:3471) fails and the other test does not is that in the other test we are adding sharding specs for both FFN and MoE.
https://github.com/nod-ai/shark-ai/blob/f4c620465153a449f1b4530e4353f60c8cfb9412/sharktank/sharktank/types/sharding.py#L124
https://github.com/nod-ai/shark-ai/blob/f4c620465153a449f1b4530e4353f60c8cfb9412/sharktank/sharktank/types/sharding.py#L134

The failing test creates directly `MoeBlockSharding`. There is no FFN block sharding, so it fails.

We don't have a mechanism that would fail if there is a sharding spec without a tensor. If we had this it would have caught this problem. But this would require somewhat of a large refactor as all sharding specs must be precise on what tensors are expected based on more precise configuration.

sogartar:
For the other direction if there is a tensor without a sharding spec it will fail. That is why I added the [Ignore](https://github.com/nod-ai/shark-ai/blob/f4c620465153a449f1b4530e4353f60c8cfb9412/sharktank/sharktank/types/sharding.py#L48). If you really want to ignore something you would have to be specific.

We need a strict 1-to-1 mapping between Theta tensor and tensor sharding specs.

sogartar (approved):

sogartar:
> Added a [patch ](https://github.com/nod-ai/shark-ai/pull/1547/commits/5cfd116747664f9e2ae5751c2d6f80a41f8aa26b) that fixes the moe test, just wondering if there are alternate ways to fix this, like by implicitly inferring if the ffn has the suffix in it's name.

It OK for the model config to control completely the sharding spec. It should contain all required information. The alternative is to complicate theta resharding with polymorphic behavior where the sharding spec receives the theta and do decisions based on what tensors are available in the theta.

PR #1551, opened 2025-06-01
[shortfin][WIP] Implement run_in_executor for PyWorkerEventLoop

vinayakdsci:
All of the CPU work done in the LLM app is *non-async*, which means that it will not allow anything else to be scheduled on the event loop until it completes. While not perfect, a workaround is to offload the blocking functions to a `ThreadPoolExecutor`, which integrated well with other async code.

This patch implements `run_in_executor` for `PyWorkerEventLoop`, which will allow using `asyncio.to_thread` to offload blocking calls in async contexts to a `ThreadPoolExecutor`.

Also uses `asyncio.to_thread` to call `beam_group.process_beams` which includes expensive blocking functions in its call stack.

### DIFF
--- a/shortfin/python/_shortfin/asyncio_bridge.py
+++ b/shortfin/python/_shortfin/asyncio_bridge.py
@@ -8,7 +8,8 @@ import asyncio
 import inspect
 
 from . import lib as sfl
-
+import asyncio.futures as futures
+import concurrent.futures
 
 # Feature detect some versions where signatures changes.
 if "context" in inspect.signature(asyncio.Task).parameters:
@@ -21,6 +22,7 @@ else:
 class PyWorkerEventLoop(asyncio.AbstractEventLoop):
     def __init__(self, worker: sfl.local.Worker):
         self._worker = worker
+        self._closed = False
 
     def get_debug(self):
         # Requirement of asyncio.
@@ -82,6 +84,16 @@ class PyWorkerEventLoop(asyncio.AbstractEventLoop):
             ).with_traceback(source_exception.__traceback__)
         else:
             raise RuntimeError(f"Async exception on {self._worker}: {context}")
+
+    def run_in_executor(self, executor, func, *args):
+        executor = concurrent.futures.ThreadPoolExecutor(thread_name_prefix="asyncio")
+        return futures.wrap_future(executor.submit(func, *args), loop=self)
+
+    def is_closed(self):
+        return self._closed
+
+    def close(self):
+        self._closed = True
 
     def _timer_handle_cancelled(self, handle):
         # We don't do anything special: just skip it if it comes up.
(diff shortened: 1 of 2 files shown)


shortfin/python/_shortfin/asyncio_bridge.py
```diff
@@ -83,6 +85,16 @@ def call_exception_handler(self, context) -> None:
         else:
             raise RuntimeError(f"Async exception on {self._worker}: {context}")
 
+    def run_in_executor(self, executor, func, *args):
+        executor = concurrent.futures.ThreadPoolExecutor(thread_name_prefix="asyncio")
```
stbaione, line 89: Do we need to instantiate a `ThreadPoolExecutor` each time, or can this be a static or instance variable that's initialized when the PyWorkerEventLoop is created?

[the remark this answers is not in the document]
stbaione: This is a good point ^, it's CPU bound, not I/O bound, so theoretically multi-processing should work better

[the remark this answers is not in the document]
vinayakdsci: `ProcessPoolExecutor()` cannot be used in the scope of a function (due to design idiosyncrasies), and must be created in the `if __name__ == "__main__"` block. `ThreadPoolExecutor` is fine here, the only purpose is to not block the event loop when a non-async function executes, which this achieves.

shortfin/python/_shortfin/asyncio_bridge.py
```diff
@@ -83,10 +87,49 @@ def call_exception_handler(self, context) -> None:
         else:
             raise RuntimeError(f"Async exception on {self._worker}: {context}")
 
+    def run_in_executor(self, executor, func, *args):
+        self._check_closed()
+        if executor is None:
+            executor = self._default_executor
+            self._check_default_executor()
+            if executor is None:
+                executor = concurrent.futures.ThreadPoolExecutor(
```
stbaione, line 96: I'm not sure if it would be a problem for us right now, but should be noted that we don't have any control on the `max_workers` these `ThreadPoolExecutors` can create

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/independent_token_selection_strategy.py
```diff
@@ -185,7 +186,7 @@ async def decode(
                 config.decode_callback(req)
 
             await beam_group.wait()
-            beam_group.process_beams()
+            await asyncio.to_thread(beam_group.process_beams)
```
stbaione, line 189: nit: Could just import `to_thread` instead of the entire package:

```python
from asyncio import to_thread as asyncio_to_thread
```

StephenBaione:
> Did you observe a speedup from this change? I tried [something similar](https://github.com/nod-ai/shark-ai/pull/1515), although I submitted each task to the `ThreadPoolExecutor` in the `selection_callback`. It seems like this PR parallelizes at a better point. Also allows the beam post-processing to be parallelized. I think that thread overhead was cancelling out the potential gains in my PR.
> 
> Also be sure to test against varying batch sizes. When I was doing my parallelization work, the % performance diff would vary a lot depending on the load the server was under

Do you have some metrics showing a speedup for different batch size settings in the "customer" script?

PR #1552, opened 2025-06-01
remove redundant object

dan-garvey:
somewhere between the refactor and some of Archana's changes, we ended up with two `self.paged_attention=` statements in our init method. I'm not sure if this is the correct solution.

### DIFF
--- a/sharktank/sharktank/layers/paged_llama_attention_block.py
+++ b/sharktank/sharktank/layers/paged_llama_attention_block.py
@@ -47,19 +47,6 @@ class PagedLlamaAttentionBlock(ThetaLayer):
         pipeline_to_device_map: tuple[tuple[int, ...], ...] | None = None,
     ):
         super().__init__(theta)
-
-        self.paged_attention = PagedAttention(
-            transformer_block_count=cache.transformer_block_count,
-            block_to_pipeline_map=block_to_pipeline_map,
-            pipeline_to_device_map=pipeline_to_device_map,
-            attn_head_count=head_count_kv,
-            attn_head_dim=head_dim,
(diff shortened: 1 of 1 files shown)


archana-ramalingam (approved):
Well, that's embarrassing. I see 2 pipeline args are being loaded from the attn block rather than from cache. They originate from config and shouldn't matter. 
So, let's also get Alex's feedback.

Alex-Vasile:
Creating the paged attention was added in [1098](https://github.com/nod-ai/shark-ai/pull/1098/files#diff-4fb0b22b51621e1860e33258ac9cd1fa3597ce0c7c1598e263761e9a5d39446f) and the `self.cache = cache` call was removed.  [1486](https://github.com/nod-ai/shark-ai/pull/1486/files#diff-4fb0b22b51621e1860e33258ac9cd1fa3597ce0c7c1598e263761e9a5d39446f) added it back in, but calling it `self.paged_attention` instead.

The version provided in the argument (produced by [create_paged_kv_cache](https://github.com/nod-ai/shark-ai/blob/b95ab3b578c0016f4f86121fc9e41dbc5fabe5f5/sharktank/sharktank/utils/create_cache.py#L10)) and the one created in in this init appear to be the same.

Looking into it though `self.paged_attention` is never used. All of `PageAttention`'s methods that operate on a cache take it as an argument. So, unless we plan on refactoring the functions calls to use `self.paged_attention` I think both instances of `self.paged_attention` should be removed.

Also yes. The `block_to_pipeline_map` and `pipeline_to_device_map` argument can be removed, they're also used to create `self.paged_attention`.

Issue #1549, opened 2025-05-30
Torch is letting us read beyond the bounds of arrays


Alex-Vasile:
Most of our tests and eager runs are with torch 2.5.1+rocm6.2 on the GPU. On that version (and possibly later, haven't tested) we are able to access memory beyond the end of an array if the array is on a GPU, and we index with a list (or tensor) of indices.

One place where this manifests is in perplexity tests using toy models. Tokens are ran one at a time:
https://github.com/nod-ai/shark-ai/blob/806762fe7bc41c47f697dc00613b443f1a986940/sharktank/sharktank/evaluate/perplexity_iree.py#L340-L341

If `self.max_prompt_length - 1` is greater than the model's context_length, then the decode step **should** error out with an index out of bounds error here:
https://github.com/nod-ai/shark-ai/blob/806762fe7bc41c47f697dc00613b443f1a986940/sharktank/sharktank/layers/rotary_embedding.py#L221

When running on CPU it does, when running on GPU **it does not**, it just returns memory from somewhere, presumably whatever is just after the end of the original array. The end result is that the model **may** return `NaN` for some/all/none of the perplexity scores. This is not consistent as it depends on the content of the memory being read in (e.g. for bs=4 and sequence lengths of [458, 213, 482, 336], only 1 of the 4 return NaN.). In my limited testing, the pattern of which returns NaN appears to be deterministic.

Easy repro using torch:
```Python
import torch
a = torch.tensor([1,2,3,4])
b = torch.tensor([1,2,3,4], device='cuda:0')

a[4] # Errors out
b[4] # Errors out

a[[4, 4]] # Errors out
b[[4, 4]] # Runs >:(
```

I will investigate what happens when running through IREE if we provide long enough prompts.

I have a feeling this may be the cause of our many NaN results: https://github.com/nod-ai/shark-ai/issues/1410

Alex-Vasile:
Tracked in the torch repo at https://github.com/pytorch/pytorch/issues/155045.

PR #1557, opened 2025-06-02
Check input lengths for RotaryEmbeddingLayer.compute_batch_mask

Alex-Vasile:
This will prevent us from accidentally indexing beyond the end of the RotaryEmbedding table, at least in Eager mode. (See #1549)


sharktank/sharktank/layers/rotary_embedding.py
```diff
@@ -208,6 +208,9 @@ def compute_batch_mask(
         positions_seq = torch.arange(0, batch_seq_len, device=self.device).unsqueeze(
             0
         ) + start_positions.unsqueeze(1)
+        assert not any(
```
archana-ramalingam, line 211: This can be moved before the previous step using the `start_positions`.

PR #1560, opened 2025-06-02
Limit get_prompts and perplexity tests to prompts shorter than the context length

Alex-Vasile:


sharktank/sharktank/evaluate/perplexity_iree.py
```diff
@@ -390,9 +390,18 @@ def run_iree_module(devices: list[iree.runtime.HalDevice]):
                 modules=(hal_module, parameters_module, self.vm_module),
             )
 
+            self.last_token = self.max_prompt_length
```
archana-ramalingam, line 393: Probably for a future PR, `last_token` can be replaced by `last_token_index`.

Alex-Vasile: renamed to last_token_index. I'm leaving max_prompt_length in there for this PR. There are a bunch of refactoring and renaming to do, and I'll leave it for that. It is used in other places and need to see if it can be renamed.

Issue #1512, opened 2025-05-27
[shortfin_apps.sd] Fix gray image responses from "/generate" endpoint


bjacobgordon:
Works as expected through [this commit](https://github.com/nod-ai/shark-ai/commit/2d07eb5dfdc416aad2b5bb0d1d931af2bc0f9d7c), but breaks in [the very next commit](https://github.com/nod-ai/shark-ai/commit/87548933f89e9f4695ec5fb439d1e82f750a3550).

monorimet:
Working on a bisect through the IREE commits now. Since the IREE bump from [iree-3.5.0rc20250513](https://github.com/iree-org/iree/commits/iree-3.5.0rc20250513/) -> [iree-3.5.0rc20250520](https://github.com/iree-org/iree/commits/iree-3.5.0rc20250520/) is what's introducing the regression, we bisect the commit diffs until we find the problem change.

This should have been caught in the sdxl mlperf CI that we don't have -- that CI should be checking accuracy in terms of CLIP and FID scores, which is the best way to validate the output.

ScottTodd:
FYI you can use https://github.com/iree-org/iree/tree/main/build_tools/pkgci/bisect to run an automated bisect e.g. with `bisect_packages.py --good-ref=iree-3.5.0rc20250513 --bad-ref=iree-3.5.0rc20250520 --test-command=...`

bjacobgordon:
Sure, @monorimet
```shell
python -m shortfin_apps.sd.server --device=amdgpu --device_ids=0 --build_preference=precompiled
```

monorimet:
Thanks. Do you get the black image from the simple client as well or just through shark UI? Can you send me a pip freeze of the env encountering the issue on https://github.com/nod-ai/shark-ai/commit/87548933f89e9f4695ec5fb439d1e82f750a3550 ?

bjacobgordon:
## Using `simple_client.py`
### [The commit just before 8754893](https://github.com/nod-ai/shark-ai/commit/2d07eb5dfdc416aad2b5bb0d1d931af2bc0f9d7c)
![Image](https://github.com/user-attachments/assets/290d3e8c-c268-4160-a62a-629d9e2c21f6)

###  [8754893](https://github.com/nod-ai/shark-ai/commit/87548933f89e9f4695ec5fb439d1e82f750a3550)
![Image](https://github.com/user-attachments/assets/9bad665d-e14b-431f-973a-74b230470433)

[...]

PR #1563, opened 2025-06-03
shortfin_apps.sd: Updates artifacts version for sdxl vmfbs, adds gfx1201 flagfile

monorimet:


stbaione (approved):
Verified that image generation is fixed with this

bjacobgordon:
> Verified that #1512 is fixed with this

Thanks for grabbing this, @stbaione!

PR #1564, opened 2025-06-03
[sharktank] IREE vs Eager Test Framework

Alex-Vasile:
Adds functionality to easily perform eager vs iree tests for LLMs. Using the Deepseek test as the first example.

TP and PP have not been tested and are not supported. Will be added in subsequent PR.


sharktank/sharktank/utils/testing.py
```diff
@@ -108,6 +110,234 @@ def assertFileWritten(self, p: Path):
...
+        if isinstance(node, AnyTensor):
+            return node.to(dev)
+
+        if isinstance(node, dict):
+            for key in node.keys():
+                node[key] = self.inplace_transfer_theta(node[key], dev)
+            return node
+
+        for key in node.keys:
+            self.inplace_transfer_theta(node.tensor(key), dev)
```
sogartar, line 160: It seems that the current theta would not change as when a new tensor gets returned our theta would not get updated.

sharktank/sharktank/utils/testing.py
```diff
@@ -108,6 +110,234 @@ def assertFileWritten(self, p: Path):
...
+                    actual=iree_i, expected=eager_i, rtol=rtol, atol=atol
+                )
+            except AssertionError as error:
+                raise AssertionError(
+                    f"Outputs do not match for {stage_name} batch index {i}:\n"
+                    f"Eager: {eager_i}\n"
+                    f"IREE: {iree_i}\n"
+                ) from error
+
+    def inplace_transfer_theta(self, node: Theta | dict | AnyTensor, dev: torch.device):
```
sogartar, line 147: It seems that this function would be better as a part of the `Theta`.
Maybe `Theta.to(...) -> Theta`.
Is it important that this be in-place?

sharktank/sharktank/utils/testing.py
```diff
@@ -108,6 +110,234 @@ def assertFileWritten(self, p: Path):
         self.assertGreater(p.stat().st_size, 0, msg=f"Expected file {p} had zero size")
 
 
+class IreeVsEagerLLMTest(TempDirTestBase):
```
sogartar, line 113: It would be better if this was not derived from `TempDirTestBase`. I would be more flexible.
You could have supply an optional temp dir in the init.

sharktank/sharktank/utils/testing.py
```diff
@@ -108,6 +110,234 @@ def assertFileWritten(self, p: Path):
...
+        self.exporter = ExportArtifacts.from_config(
+            self.config,
+            irpa_path=self.dataset_path,
+            iree_hip_target=self.iree_hip_target,
+            iree_hal_target_device=self.iree_hal_target_device,
+            hip_device_id=self.iree_device,
+            output_name=work_dir / "model",
+        )
+
+        self.config.device = torch.device("cuda:0")
```
sogartar, line 257: Did you intend for this to be hardcoded?
This would fail if you don't have this device?

sharktank/sharktank/utils/testing.py
```diff
@@ -108,6 +110,234 @@ def assertFileWritten(self, p: Path):
...
+        """
+        eager_decode_tokens = self.eager_batch.prefill()
+        self.eager_batch.dump_args(
+            phase="prefill",
+            arg_name="eager_logits",
+            arg=self.eager_batch.prefill_logits,
+        )
+
+        self.eager_prefill_logits = torch.tensor(
+            np.load(self.prefill_eager_logits_path)
```
sogartar, line 288: bfloat16 and the various f8 types would be a problem form numpy.
You could use [safetensors](https://github.com/huggingface/safetensors?tab=readme-ov-file#yet-another-format-).

sogartar: I am not sure if you could pass safetensors as iree-run-module function args. I recall you can't. Maybe it got support.
If not, you could use [iree_tool_prepare_input_args](https://github.com/iree-org/iree-turbine/blob/9cc8c9a53d910504581e26170565a63489ff12df/iree/turbine/support/tools.py#L36C5-L36C33), which would support any torch dtype.

sharktank/sharktank/utils/testing.py
Alex-Vasile: `setup_variables()` accepts an optional `work_dir` argument which it uses instead of the temp_dir.

I wanted temp folder as default behaviour with the option to keep the outputs for local testing.

sharktank/sharktank/utils/testing.py
Alex-Vasile: node.tensor(key) gets updated in place

sharktank/sharktank/utils/testing.py
sogartar: But how would I use this in the context of lets say a test written for pytest
```
def test_my_thing():
```

We already made some of our fixtures unusable with non `unittest` tests, which needs to be addressed at some point.
I would like to have more flexible components.
Also with not much change this class could be made usable outside of tests.
Even if you could instantiate this inside a function I am not sure if we should consider it an abuse of `unittest.TestCase`.

sharktank/sharktank/utils/testing.py
Alex-Vasile: I am relying on `load_llm.py:Batch` to run the torch model and dump these values. If fp8 support requires special work then I think that should be in a subsequent PR as it will require redoing portions of the `Batch` class (at a minimum).

sharktank/sharktank/utils/testing.py
```diff
@@ -108,6 +110,227 @@ def assertFileWritten(self, p: Path):
...
+        work_dir: Path | None = None,
+        skip_decode: bool = False,
+    ):
+        """
+        Setup the variables and objectes needed for the IREE vs eager test.
+
+        Args:
+            theta: The Theta object containing the model parameters.
+            config: The configuration for the model.
+            raw_token_ids: The raw token ids to use for the prefill stage.
```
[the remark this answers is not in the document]
Alex-Vasile: Done

Alex-Vasile:
Looks like bumping iree broke the Deepseek compile.

sogartar (approved):

Alex-Vasile:
> Looks like merging in main broke the Deepseek compile. Between [b7a3517](https://github.com/nod-ai/shark-ai/commit/b7a35179469f2e0f9a3a5d08374668c3357c65dc) and [e4934bf](https://github.com/nod-ai/shark-ai/commit/e4934bff35515b3b49c38888d66911f6e184a7d8) (inclusive of both).

Issue seems to be related to using float32 for attention and activation dtypes. Will file a separate issue, but not relevant here since deepseek should have been using float16 (for now) and not float32.

archana-ramalingam (approved):

PR #1565, opened 2025-06-03
[sharktank] Make attention kernel functions private

IanWood1:
There is no reason to publicly export these functions (`util.func` defaults to public) and it is easier to read the IREE IR dumps if they get removed after inlining.


Groverkss (approved):
oof this is my fault

PR #1577, opened 2025-06-05
Rework topk kernel and hip code

rsuderman:
Reworked to be dynamic batch dimension and support for arbitrary batch dims to better support invocations.

Improved the build system to use absolute paths for the hsaco files and target specification.


sharktank/sharktank/kernels/topk.py
```diff
@@ -43,17 +43,16 @@ def generate(self, ksel: KernelSelection, kb: KernelBuilder):
 
         # Generate specialization signature and types.
         bs = input.type.shape[0]
-        sl = input.type.shape[1]
-        sl = "?" if sl < 0 else sl  # Use ? for dynamic dimensions in MLIR
+        bs = "?" if bs < 0 else bs  # Use ? for dynamic dimensions in MLIR
```
zphoenixrises, line 46: I don't think we need this

sharktank/sharktank/ops/default_impls.py
```diff
@@ -733,8 +733,14 @@ def topk_default(
     if chunk_size is not None:
         return _split_topk(tensor, k, dim, largest, sorted, chunk_size)
 
-    if largest and not sorted and len(tensor.shape) == 3 and dim == 2:
+    if largest and not sorted and dim == len(tensor.shape) - 1:
```
zphoenixrises, line 736: This feels hacky and could give rise to this code path being invoked unintentionally . I personally feel that having the explicit flag would be appropriate here.

zphoenixrises (approved):
Awesome, Let's get it in!

sharktank/sharktank/ops/default_impls.py
rsuderman: Added a flag and check. Ultimately we should support `linalg_ext.topk` for everything but it has some flaky compilation which we need to sort out before we can permanently turn on.

sharktank/sharktank/kernels/topk.py
rsuderman: Removed.

PR #1580, opened 2025-06-05
[sharktank] Add Llama4 test for toy prefill

vivekkhandelwal1:


sharktank/requirements.txt
```diff
@@ -7,7 +7,7 @@ numpy
 
 # Model deps.
 huggingface-hub
-transformers==4.50.0
+transformers==4.52.4
```
Alex-Vasile, line 10: We should bump this separately

sharktank/tests/models/llama4/moe_test.py
```diff
@@ -0,0 +1,220 @@
...
+        )
+
+        input = (
+            torch.rand([batch_size, sequence_length, feature_dim], dtype=dtype) - 0.5
+        )
+        moe_output = moe_block(input)
+
+        from sharktank.types import unbox_tensor, Theta
+
+        if num_shared_experts is not None:
```
Alex-Vasile, line 56: Are you planning on paramaterizing this test to run with and without shared experts?

sharktank/tests/models/llama4/moe_test.py
```diff
@@ -0,0 +1,220 @@
...
+
+        router_top_value, router_indices = torch.topk(
+            router_logits.transpose(0, 1), self.top_k, dim=1
+        )
+        router_scores = (
+            torch.full_like(router_logits.transpose(0, 1), float("-inf"))
+            .scatter_(1, router_indices, router_top_value)
+            .transpose(0, 1)
+        )
+        # We do this to make sure we have -inf for non topK tokens before going through the !
```
Alex-Vasile, line 141: "before going through the !" ?

sharktank/tests/models/llama4/moe_test.py
```diff
@@ -0,0 +1,220 @@
...
+        super().__init__()
+        self.top_k = num_experts_per_tok
+        self.hidden_dim = hidden_size
+        self.num_experts = num_local_experts
+        self.experts = Llama4TextExperts(
+            num_local_experts, intermediate_size, hidden_size
+        )
+        self.router = torch.nn.Linear(hidden_size, num_local_experts, bias=False)
+        self.shared_expert = Llama4TextMLP(hidden_size, intermediate_size)
+        # self.shared_expert = shared_expert
```
Alex-Vasile, line 125:
```suggestion
```

sharktank/tests/models/llama4/moe_test.py
```diff
@@ -0,0 +1,220 @@
...
+        # now that we finished expert computation -> we scatter add because we gathered previously
+        # we have to do this because we used all experts on all tokens. This is faster than the for loop, tho you are compute bound
+        # this scales a lot better if you do EP!
+        out.scatter_add_(
+            dim=0, index=router_indices, src=routed_out.view(-1, hidden_dim)
+        )
+        return out, router_scores
+
+
+class Llama4TextExperts(torch.nn.Module):
```
archana-ramalingam, line 169: For 'Llama4TextExperts' and `Llama4TextMLP`, is there a reason we are not importing these HF functions and initializing with local config but instead have implemented them here?

sharktank/sharktank/layers/paged_llama_attention_block.py
```diff
@@ -214,6 +229,30 @@ def forward(
...
 
+        if self.use_qk_norm:
+            xq = self.qk_norm(xq)
+            xk = self.qk_norm(xk)
+
+        # Use temperature tuning from https://arxiv.org/abs/2501.19399
+        # Ken M. Nakanishi - Scalable-Softmax Is Superior for Attention (2025)
+        if self.attn_temperature_tuning and not self.use_rope:
+            if start_positions is None:
+                cache_position = torch.arange(0, h.shape[1], dtype=torch.long)
```
sogartar, line 240: This is a part that would need to be refactored in a subsequent PR. There are questions like, are we going to use padding at the start and then the decode step would need to be supported.
Nevertheless, `cache_position` is not the right name for this. I think it will get confused with the KV cache. We are here essentially taking the batch sequence length.

sharktank/sharktank/layers/paged_llama_attention_block.py
```diff
@@ -214,6 +229,30 @@ def forward(
...
+            else:
+                assert False, "TODO: decode step"
+            attn_scales = (
+                torch.log(
+                    torch.floor((cache_position.float() + 1.0) / self.floor_scale) + 1.0
+                )
+                * self.attn_scale
+                + 1.0
+            )
+            input_shape = h.shape[:-1]
```
sogartar, line 250: Maybe call this `input_tokens_shape`.

sharktank/sharktank/layers/rotary_embedding.py
```diff
@@ -174,6 +174,22 @@ def forward_unsharded(
...
             return xt_out
 
+        if self.rope_scaling_type == "llama4":
+            freqs_cis_real = rotary_embed_table[0][
+                :, : rotary_embed_table[0].shape[1] // 2
+            ]
+            freqs_cis_imag = rotary_embed_table[1][
+                :, : rotary_embed_table[0].shape[1] // 2
+            ]
+            # TODO: don't use complex numbers as the compiler does better without them.
```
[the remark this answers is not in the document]
sogartar: Let's do this later. It is hard enough to get this functional. This path will not be taken by other models.

sharktank/sharktank/models/llama4/testing.py
```diff
@@ -0,0 +1,172 @@
...
+    moe_layers = None
+    if config.hp.interleave_moe_layer_step is None:
+        moe_layers = config.moe_layers
+    rope_scaling = {
+        "factor": 8.0,
+        "high_freq_factor": 4.0,
+        "low_freq_factor": 1.0,
+        "original_max_position_embeddings": 8192,
+        "rope_type": "llama3",
+    }
```
sogartar, line 24: It is strange that this says `rope_type` is `llama3` but our config has `rope_scaling_type` `llama4`.
I forgot why Llama4 needed a separate RoPE path.

sharktank/sharktank/models/llama4/testing.py
```diff
@@ -0,0 +1,172 @@
...
+        moe_layers=moe_layers,
+        interleave_moe_layer_step=config.hp.interleave_moe_layer_step,
+        rope_scaling=rope_scaling,
+        attention_chunk_size=config.attention_chunk_size,
+        torch_dtype=config.dtype,
+        attn_temperature_tuning=config.attn_temperature_tuning,
+        floor_scale=config.floor_scale,
+        attn_scale=config.attn_scale,
+        # attn_implementation="eager",
+        # attn_implementation="flex_attention",
```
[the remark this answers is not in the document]
sogartar: This was for testing. I think `sdpa` is the default, which is what we want.

sharktank/sharktank/models/llm/llm.py
```diff
@@ -196,16 +196,16 @@ def prefill(
                     count=len(self.cache.pipeline_to_device_map[pipeline]),
                     devices=self.cache.pipeline_to_device_map[pipeline],
                 )
-                for pipeline in range(self.cache.pipeline_count)
+                for pipeline in range(len(self.cache.pipeline_to_device_map))
```
sogartar, line 199: Why do we need this change?
What would happen if we are not using `PipelinedCache`? Do we know that we always have `pipeline_to_device_map` in the polymorphic KV cache? The other KV cache classes don't have this.
Maybe we should be getting this from the config.

sharktank/sharktank/models/llm/llm.py
```diff
@@ -362,6 +369,13 @@ def __init__(
                 fake_quant=fake_quant,
                 softcap=config.hp.attention_softcap,
                 model_arch=config.hp.model_arch,
+                use_rope=use_rope,
+                use_qk_norm=block_index in config.rope_layers and config.use_qk_norm
+                if config.rope_layers
+                else False,
```
sogartar, line 375: Could you prepare this arguments before the call? Lets not put if-else at the call site. People like clean calls.

sharktank/tests/models/llama4/llama4_test.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+class Llama4Test(TempDirTestBase):
+    def setUp(self):
+        super().setUp()
+        torch.random.manual_seed(12345)
+
+    @parameterized.expand(
+        [
+            (torch.float32, 1e-5),
```
sogartar, line 29: This atol is not being used.

sharktank/tests/models/llama4/moe_test.py
sogartar: I am not sure about adding the `moe_test`. The problem with it is that it relies on a copy of some internal details of the HF transformers implementation for reference. We cant depend on the HF transformers package as these classes are internal details of the implementation and we should not expose them. The only way to implement this is to copy as we get a frozen variant.

sogartar: I needed this test to play with the MoE to get the port right. Maybe the test would have merit in the future.

sharktank/sharktank/layers/paged_llama_attention_block.py
sogartar: Long-term it would probably have to be an arange over the current accumulated batch sequence length across all prefill/decode invocations.

sharktank/tests/models/llama4/moe_test.py
vivekkhandelwal1: For now, I have removed it.

sharktank/tests/models/llama4/moe_test.py
vivekkhandelwal1: The latest Transformers version does not contain these comments. I have updated the code file.

sharktank/sharktank/models/llama4/testing.py
vivekkhandelwal1: I have changed it to `llama3` in our config to keep it consistent with the Transformers.

sharktank/sharktank/models/llm/llm.py
vivekkhandelwal1: Also, I have to make this change because of the changes made here: https://github.com/nod-ai/shark-ai/pull/1404

sharktank/sharktank/layers/paged_llama_attention_block.py
vivekkhandelwal1: But the name was adopted from Transformers only.

sharktank/tests/models/llama4/llama4_test.py
```diff
@@ -0,0 +1,96 @@
...
+class Llama4Test(TempDirTestBase):
+    def setUp(self):
+        super().setUp()
+        torch.random.manual_seed(12345)
+
+    @pytest.mark.xfail(
+        is_mi300x,
+        raises=TypeError,
+        strict=False,
+        reason="argument of type 'NoneType' is not iterable",
```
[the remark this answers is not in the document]
vivekkhandelwal1: I have created a tracking issue here: https://github.com/nod-ai/shark-ai/issues/1613/

vivekkhandelwal1:
@archana-ramalingam @Alex-Vasile If the PR looks good to you, can I merge this?

archana-ramalingam (approved):

PR #1584, opened 2025-06-05
 Rework topk kernel and hip code #1577 

rsuderman:
Reworked to be dynamic batch dimension and support for arbitrary batch dims to better support invocations.

Improved the build system to use absolute paths for the hsaco files and target specification.


zphoenixrises (approved):
LGTM!

PR #1589, opened 2025-06-06
adds qkv quantization to testing of llama

dan-garvey:


sharktank/tests/models/llama/test_llama.py
```diff
@@ -7,14 +7,16 @@
...
 from sharktank.models.llm import *
 from sharktank.models.llama.toy_llama import generate
+from parameterized import parameterized
 
 import pytest
 import torch
 
 
-def test_llama():
+@parameterized.expand([(True, 2.378), (False, 0.577)])  # quantize qkv
```
archana-ramalingam, line 16: Looks like the `cross_entropy_golden_value` needs to be updated for the first case to get the tests passing.

PR #1590, opened 2025-06-06
Rework kernels to include per indices mapping

rsuderman:
`linalg_ext.scatter` provides its input indices. Having the inputs invoke in a chained fashion optimizes back-to-back top-ks for split-k.


zphoenixrises (approved):
Awesome works! I think this is good to go! We can fine tune the compilation of the custom kernels in a follow up.

PR #1592, opened 2025-06-07
[sharktank] Nightly CI - Flux 

PhaneeshB:


.github/workflows/ci-sharktank-nightly.yml
```diff
@@ -7,10 +7,10 @@
 name: CI - sharktank nightly
 
 on:
-  workflow_dispatch:
-  schedule:
-    # Weekdays at 11:00 AM UTC = 03:00 AM PST / 04:00 AM PDT
-    - cron: "0 11 * * 1-5"
+  push:
+    branches:
+      - 'fix_flux_ci'
```
sogartar, line 12: This would need to be reverted.
You can trigger manually a workflow from the [actions panel](https://github.com/nod-ai/shark-ai/actions). There you can select a branch that is in this repo.

sharktank/tests/models/vae/vae_test.py
```diff
@@ -149,7 +149,7 @@ def testVaeIreeVsHuggingFace(self):
             "--iree-opt-outer-dim-concat=true",
             "--iree-llvmgpu-enable-prefetch=true",
             "--iree-hip-waves-per-eu=2",
-            "--iree-dispatch-creation-enable-aggressive-fusion=true",
+            # "--iree-dispatch-creation-enable-aggressive-fusion=true", #TODO: https://github.com/iree-org/iree/issues/20875
```
sogartar, line 152: I copied these options from SDXL's config several months ago. It is curious if we really want this one and the rest.

PhaneeshB: Most of these flags are relevant for perf, some maybe be the same as default behavior. 
IMO good have explicit flag config that we want working to catch any regressions.

PR #1600, opened 2025-06-09
[RoPE] Add RoPEv2

Groverkss:


KyleHerndon (commented):
LGTM. Do you think that this should eventually completely replace the existing rotary_embedding usages?

sharktank/tests/kernels/mlir_kernel_test.py
```diff
@@ -58,3 +59,8 @@ def test_mlir_kernel(self):
         torch.testing.assert_close(source[3], out[0])
         torch.testing.assert_close(source[7], out[1])
         torch.testing.assert_close(source[54], out[2])
+
+
+source = torch.randn((64, 1024), dtype=torch.float16, device="cuda:0")
+indices = torch.randint(0, 1024, (4,), device="cuda:0")
+print(sharktank_gather(source, indices))
```
sogartar, line 66: Is this some debugging remanent?
Removing it should fix the CI error
```
ERROR sharktank/tests/kernels/mlir_kernel_test.py - AssertionError: Torch not compiled with CUDA enabled
```

sharktank/tests/layers/rotary_embedding_v2_test.py
```diff
@@ -0,0 +1,147 @@
...
+
+from sharktank.layers.rotary_embedding_v2 import RotaryEmbeddingV2Layer
+from transformers.models.llama.modeling_llama import (
+    LlamaRotaryEmbedding,
+    apply_rotary_pos_emb,
+)
+from transformers import LlamaConfig
+import pytest
+
+torch.manual_seed(123456)
```
sogartar, line 17: Like that tests don't get a frozen seed. If the execution order of tests changes their seed will change.
```
@pytest.fixture
def seed_torch_rng():
    with torch.random.fork_rng():
        yield

def test_thing(seed_torch_rng):
    ...
```

sogartar (approved):

rsuderman:
I am going to take over this PR to see if we can get it in.

PR #1602, opened 2025-06-09
[tuner] build td spec using config list

bangtianliu:
This PR is the first step toward supporting attention in the tuner.

- Generalizes build_td_spec to accept a config_list argument, allowing multiple configuration attributes to be applied.
Example usage for attention tuning:
```
config_list = [
    ("compilation_info", compilation_info_attr),
    ("decomposition_config", decomposition_config_attr),
] 
```
- For existing contraction/conv ops, the function remains compatible:

```
config_list = [("compilation_info", compilation_info_attr)]
```

### DIFF
--- a/sharktuner/sharktuner/spec_builder.py
+++ b/sharktuner/sharktuner/spec_builder.py
@@ -35,7 +35,7 @@ def get_placeholder_spec(context: ir.Context) -> ir.Module:
 def build_td_spec(
     context: ir.Context,
     op: ir.Operation,
-    compilation_info: iree_codegen.CompilationInfoAttr,
+    config_list: list[tuple[str, ir.Attribute]],
     func_name: str,
 ) -> ir.Module:
     bbargs = []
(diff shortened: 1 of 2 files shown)


sharktuner/sharktuner/spec_builder.py
```diff
@@ -76,35 +76,57 @@ def build_td_spec(
...
+        transform.named_sequence
+        @__kernel_config(%variant_op: !transform.any_op {{transform.consumed}}) -> !transform.any_op
+            attributes {{ iree_codegen.tuning_spec_entrypoint }} {{
+            %res = transform.foreach_match in %variant_op
+                @{func_name} -> @apply_op_config
+            : (!transform.any_op) -> !transform.any_op
+            transform.yield %res : !transform.any_op
         }}
-        """
+    }}"""
```
kuhar, line 131: How could we test this? I think it should be possible to generate some TD IR with more than one config and then make sure that is parses.

kuhar (commented):
What's the overall plan?
> This PR is the first step toward supporting attention in the tuner as suggested by Max.

Do you have it written down somewhere?

sharktuner/sharktuner/spec_builder.py
bangtianliu: We don't currently have a test for this file. I’ll look into how I can add one.

Max191:
Yeah, I can write something up, but Bangtian basically summed it up. The assumption is that the tuner will not be doing anything other than setting attributes on the root operation, so we just need a way to apply a list of attributes instead of only compilation_info. I didn't see a good immediate reason to try to generalize any further than what we have here in this PR, and I think this should put us in a position where we can start doing attention tuning properly.

Max191 (commented):
This looks like what I had in mind! The next step is to connect it with the ConstraintGenerator class. We also need a test for the new string formatting stuff.

kuhar (approved):
LGTM, thanks for adding tests

PR #1603, opened 2025-06-10
[sharktank] Expand perplexity test coverage for 70B/405B models

archana-ramalingam:
Expand perplexity test coverage for 70B/405B models


Alex-Vasile:
@archana-ramalingam Benchmark tests seem to be working. Only current failure is due to new flags accidentally activating `@is_llama_8b`, getting a commit to fix that, and then undoing the changes to get nightlies to run. The rest is good to go.

PR #1605, opened 2025-06-10
[sharktank] Add module patcher that uses ops.trace_tensor

sogartar:
This can be used to trace tensors both in eager and during execution with IREE. Usually it allows to get adequate tracing density when models are decomposed into multiple nested torch modules.

### DIFF
--- a/sharktank/sharktank/utils/debugging.py
+++ b/sharktank/sharktank/utils/debugging.py
@@ -144,7 +144,7 @@ set_trace_tensor_callback(trace_tensor_to_safetensors_callback)
 
 def trace_tensors_to_safetensors(key: str, tensors: Dict[str, torch.Tensor]):
     # Sanitize as path.
-    key = re.sub("[" + re.escape(r"""#~!@$%^&*()[]{}:;"'""") + "]", "", key)
+    key = re.sub("[" + re.escape(r"""*?"<>|""") + "]", "", key)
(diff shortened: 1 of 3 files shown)


sharktank/sharktank/utils/debugging.py
```diff
@@ -144,7 +144,7 @@ def trace_tensor_to_safetensors_callback(key: str, *tensors: Tuple[torch.Tensor]
 
 def trace_tensors_to_safetensors(key: str, tensors: Dict[str, torch.Tensor]):
     # Sanitize as path.
-    key = re.sub("[" + re.escape(r"""#~!@$%^&*()[]{}:;"'""") + "]", "", key)
```
PhaneeshB, line 147: seems like we want to escape some more characters, would it be better to just add new chars to the existing sequence?
something like this `"[" + re.escape(r"""#~!@$%^&*()[]{}:;"'"*?"<>|""") + "]"`

assuming the existing pattern is useful for some cases atleast.

PhaneeshB (approved):

sharktank/sharktank/utils/debugging.py
sogartar: I don't want to escape all these characters. They are legal file name characters on all platforms so they should be OK. In particular I don't want `%` to be escaped.

PR #1606, opened 2025-06-10
[shortfin][LLM] Optimize d2h transfers and tune strobe delays

vinayakdsci:
This patch moves `await device0` in `_transfer_buffer()` to before the d2h transfer starts. This ensures that the transfer does not happen simultaneously with kernel invocations, which can stall both the GPU and the CPU as the work is happening, reducing overall throughput.

Also tunes the values of `STROBE_SHORT_DELAY` and `STROBE_LONG_DELAY` for the batchers, as it does not make sense to wait for  the same amount of time when requests are available as when they are not.

### DIFF
--- a/shortfin/python/shortfin_apps/llm/components/batcher.py
+++ b/shortfin/python/shortfin_apps/llm/components/batcher.py
@@ -31,6 +31,7 @@ from .service_debug_dumper import SERVICE_DEBUG_DUMPER
 
 logger = logging.getLogger(__name__)
 
+STROBE_DELAY_COEFF = 4
 
 ########################################################################################
 # Batcher
@@ -42,8 +43,8 @@ import math
 class LlmBatcherProcess(BatcherProcess):
     """This batcher provides a high-level mechanism for dispatching LLM tasks."""
 
-    STROBE_SHORT_DELAY = 0.065
-    STROBE_LONG_DELAY = 0.065
+    STROBE_SHORT_DELAY = 0.065 // STROBE_DELAY_COEFF
+    STROBE_LONG_DELAY = 0.065 * STROBE_DELAY_COEFF
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


shortfin/python/shortfin_apps/llm/components/batcher.py
```diff
@@ -311,7 +313,6 @@ async def _transfer_buffer(self, req_count, device0, buffers):
             host_buffer.copy_from(buffer)
             new_buffers.append(host_buffer)
 
-        await device0
```
rsuderman, line 314: You cannot do this - you need to know that the data has finished transferring before you can return / process the buffer.

rsuderman (changes requested):

shortfin/python/shortfin_apps/llm/components/batcher.py
stbaione: For additional info, you can see the issue with this in the smoke test failures thrown in this run.

For example:

```text
Expected to be one of: {', there was a little girl named Lily. She loved to play with her'}
Actual response: , there was a little girl Onceing the world. She was so excited
```

Happens for [CPU](https://github.com/nod-ai/shark-ai/actions/runs/15557263351/job/43801016920?pr=1606) and [GPU](https://github.com/nod-ai/shark-ai/actions/runs/15557263351/job/43801016928?pr=1606).

shortfin/python/shortfin_apps/llm/components/batcher.py
```diff
@@ -42,8 +43,8 @@
 class LlmBatcherProcess(BatcherProcess):
     """This batcher provides a high-level mechanism for dispatching LLM tasks."""
 
-    STROBE_SHORT_DELAY = 0.065
-    STROBE_LONG_DELAY = 0.065
+    STROBE_SHORT_DELAY = 0.065 // STROBE_DELAY_COEFF
```
stbaione, line 46: Was there a measured throughput gain from this change? Seems logical, but would be good to comment on the PR the benefit from it to be sure it behaves the way we expect it to

PR #1617, opened 2025-06-11
[sharktank] In trace tensor module patching allow for exclusion by regex key search

sogartar:
In some strange cases torch does not want to trace tensors during export.
It complains about that we are trying to modify frozen tensors. We need our signature to be
```
trace_tensor(str trace_key, Tensor(a!) tensor) -> ()
```
This means it modifies the tensor in-place and returns nothing. Then the op
will not be removed.
Cloning the tensor before tracing does not help as torch export is smart and
removes the whole expression.
Ideally we would like to say that the op is side-effecting but that seems
unavailable in torch.

With this change you can exclude some trace key patterns to be excluded from the trace.
Thus circumventing the torch weirdness.

### DIFF
--- a/sharktank/sharktank/utils/patching.py
+++ b/sharktank/sharktank/utils/patching.py
@@ -9,6 +9,7 @@ from typing import Any, TYPE_CHECKING, Union
 from collections.abc import Mapping, Iterable
 from sharktank.types import InferenceTensor, unbox_tensor
 import logging
+import re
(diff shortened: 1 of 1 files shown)


Alex-Vasile (commented):
I don't understand how this change addresses the description of the problem you have listed, or how this change would be used. Could you please elaborate.

PR #1618, opened 2025-06-11
[sharktank] allow for different dtypes and sort by file modification time in assert_close_safetensors

sogartar:
Sometimes when comparing traces the reference may be of higher precision. Here a flag is added to allow for this auto promotion.

Sort by file modification time to try and get an order closer to the reference execution order. When we trace, tensors are written in order as the module is executed and we want to preserve this order when doing the comparison. It would be ideal if we could have a global trace tensor counter. Then we would be able to do a lexicographical sort, but this does not play nice during torch export.

### DIFF
--- a/sharktank/sharktank/utils/testing.py
+++ b/sharktank/sharktank/utils/testing.py
@@ -186,6 +186,7 @@ def assert_close_safetensors(
     rtol: Optional[float] = None,
     atol: Optional[float] = None,
     fail_fast: bool = True,
+    allow_different_dtype: bool = False,
 ):
     """Asserts that actual and reference safetensors files are within tolerances.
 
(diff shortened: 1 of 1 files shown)


oyazdanb (approved):

sharktank/sharktank/utils/testing.py
```diff
@@ -257,16 +264,10 @@ def print_stats(label, t):
         # Then assert.
         for name in ref_f.keys():
             actual = actual_f.get_tensor(name)
+            if allow_different_dtype:
+                actual = actual.to(dtype=ref.dtype)
```
Alex-Vasile, line 268: Why not use `check_dtype` arg for `assert_close`?

Alex-Vasile (changes requested):
For the

sharktank/sharktank/utils/testing.py
```diff
@@ -196,55 +197,61 @@ def assert_close_safetensors(
...
+        not_close_list: list[tuple[Path, Path]] = []
+        for ref_file_path, actual_file_path in ref_actual_file_path_map.items():
             try:
                 assert os.path.isfile(actual_file_path)
                 assert_close_safetensors(
                     actual_file_path,
                     ref_file_path,
                     rtol=rtol,
                     atol=atol,
-                    fail_fast=fail_fast,
```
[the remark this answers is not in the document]
sogartar: I made a mistake since there may be more then one tensor in the archive. I will fix it.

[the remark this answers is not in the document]
sogartar: Done.

sharktank/sharktank/utils/testing.py
sogartar: Done.

sharktank/sharktank/utils/testing.py
```diff
@@ -196,21 +197,25 @@ def assert_close_safetensors(
...
 
-    if os.path.isdir(ref_path):
-        assert os.path.isdir(actual_path)
-        actual_path = os.path.abspath(actual_path)
-        ref_path = os.path.abspath(ref_path)
-        ref_paths = [
-            Path(root) / file
-            for root, dirs, files in os.walk(ref_path)
-            for file in files
+    if not ref_path.is_file():
```
Alex-Vasile, line 203: `ref_path.is_file()` will return file if the path doesn't exist. It's better to check that it exists and that it's a dir.

sharktank/sharktank/utils/testing.py
```diff
@@ -196,21 +197,25 @@ def assert_close_safetensors(
...
+        # Get all files in ref_path recursively.
+        ref_file_paths: list[Path] = [
+            file_path for file_path in Path(ref_path).rglob("*") if file_path.is_file()
```
Alex-Vasile, line 206: Should this not filter to just safetensors?

Alex-Vasile (changes requested):

sharktank/sharktank/utils/testing.py
sogartar: I added a check if the path exists.

PR #1624, opened 2025-06-11
[Llama] Convert output normalization and linear layer to activation data type

jtuyls:
Currently, the output normalization and linear layer of Llama3_8b_fp8 are performed in f32, even when the activation data type is set to bf16 and the weights are provided in bf16 as well:
```
...
%770 = torch_c.from_builtin_tensor %__auto.output.weight : tensor<128256x4096xbf16> -> !torch.vtensor<[128256,4096],bf16>
...
%11955 = torch.prims.convert_element_type %770, %int6_12214 : !torch.vtensor<[128256,4096],bf16>, !torch.int -> !torch.vtensor<[128256,4096],f32>
%int-2_12215 = torch.constant.int -2
%int-1_12216 = torch.constant.int -1
%11956 = torch.aten.transpose.int %11955, %int-2_12215, %int-1_12216 : !torch.vtensor<[128256,4096],f32>, !torch.int, !torch.int -> !torch.vtensor<[4096,128256],f32>
%int6_12217 = torch.constant.int 6
%11957 = torch.prims.convert_element_type %11956, %int6_12217 : !torch.vtensor<[4096,128256],f32>, !torch.int -> !torch.vtensor<[4096,128256],f32>
%int4_12218 = torch.constant.int 4
%11958 = torch.aten.mul.int %int4_12218, %777 : !torch.int, !torch.int -> !torch.int
%int4096_12219 = torch.constant.int 4096
%11959 = torch.prim.ListConstruct %11958, %int4096_12219 : (!torch.int, !torch.int) -> !torch.list<int>
%11960 = torch.aten.view %11954, %11959 : !torch.vtensor<[4,?,4096],f32>, !torch.list<int> -> !torch.vtensor<[?,4096],f32>
torch.bind_symbolic_shape %11960, [%773], affine_map<()[s0] -> (s0 * 128, 4096)> : !torch.vtensor<[?,4096],f32>
%11961 = torch.aten.mm %11960, %11957 : !torch.vtensor<[?,4096],f32>, !torch.vtensor<[4096,128256],f32> -> !torch.vtensor<[?,128256],f32>
```
This can be reproduced using the following export command:
```
python3 -m sharktank.examples.export_paged_llm_v1 --irpa-file=8b_fp8.irpa \
--output-mlir=8b_fp8.mlir \
--output-config=8b_fp8_config.json \
--bs-prefill=4 --bs-decode=4 --attention-kernel sharktank \
--attention-dtype=float8_e4m3fnuz --activation-dtype=bfloat16  --use-hf --use-attention-mask --kv-cache-dtype=float8_e4m3fnuz
```
I think we should always convert the activation back to the activation_dtype to make sure that those layers are computed in the desired data type. In this case, this will result in the normalization and linear layer being computed in bf16, which results in significant **speedup of 16.9% on the entire model latency** compared with f32 as shown in the following table containing latencies for Llama3_8b_fp8 on sequence lengths of 2048:

| Configuration | Latency (ms) |
| -------- | ------- |
| Base_f32  | 372   |
| Base_f32_tuned | **337** |
| Bf16_norm_linear | 326 |
| Bf16_norm_linear_tuned | **280**  |

Notes:
- Base_f32: the model as exported before this PR, with the output normalization and linear layer being computed in f32.
- Base_f32_tuned: same as Base_f32, but with `--iree-codegen-enable-default-tuning-specs=true`.
- Bf16_norm_linear: the model as exported with this PR, with the output normalization and linear layer being computed in bf16.
- Bf16_norm_linear_tuned: same as Bf16_norm_linear, but with `--iree-codegen-enable-default-tuning-specs=true`.
- The numbers are averages, with `--benchmark_repetitions=3` set in `iree-benchmark-module`.

I have also checked the perplexity to make sure that the results are still accurate using 'perplexity_iree_test.py':

| Configuration | Perplexity |
| -------- | ------- |
| **Base_expected** | **12.54**  |
| Base_f32 | 12.40  |
| Bf16_norm_linear | 12.42  |
| Bf16_linear | 12.41  |

From this, I don't see much difference in perplexity, which is what I would have expected as the activations are already bf16 at other places and the linear layer's weights are provided in bf16 as well.

Notes:
- Base_f32: the model as exported before this PR, with the output normalization and linear layer being computed in f32.
- Bf16_norm_linear: the model as exported with this PR, with the output normalization and linear layer being computed in bf16.
- Bf16_linear : added for completeness to check what happens if we keep the normalization layer in f32, but compute the linear layer in bf16.

Let me know if there are other perplexity tests I should run to make sure that this doesn't regress anything else.

cc @AmosLewis @MaheshRavishankar


AmosLewis (approved):
I just test the llm.py change with what I run daily for llama fp8 performance [llama8b_fp8_export_run.sh](https://gist.github.com/AmosLewis/61d729e0cb3e3ecbd92a9b9edbf31867#file-llama8b_fp8_export_run-sh). Prefill improve, decode stay the same. 128: **30.4ms -> 26ms**. 2048: **340ms -> 279 ms**
Detailed results:
[llama3_fp8_128&2048_benchmark_jorn.txt](https://gist.github.com/AmosLewis/594b86503ca1e10187f7efe0d2212467#file-llama3_fp8_128-2048_benchmark_jorn-txt)

PR #1625, opened 2025-06-11
Initial code coverage commit for sharktank

zphoenixrises:
This PR is to test out code coverage in the shartank package. Once we tweak this and are satisfied with the results, we can enable it for all the code in the repo. 

Example: https://app.codecov.io/github/nod-ai/shark-ai/commit/f24ce4c301b042c50c6fd2192579300fa0b66128
image

### DIFF
--- a/.github/workflows/ci-sharktank.yml
+++ b/.github/workflows/ci-sharktank.yml
@@ -84,7 +84,7 @@ jobs:
             --log-cli-level=info \
             -v
         env:
-          COVERAGE_FILE: .coverage.${{ matrix.python-version }}.${{ matrix.torch-version }}.${{ matrix.runs-on }}
+          COVERAGE_FILE: coverage.${{ matrix.python-version }}.${{ matrix.torch-version }}.${{ matrix.runs-on }}
 
       - name: Upload coverage data
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02
--- a/.github/workflows/ci-sharktank.yml
+++ b/.github/workflows/ci-sharktank.yml
@@ -80,6 +80,7 @@ jobs:
         run: |
           pytest -n 20 sharktank/ \
             --cov=sharktank \
+            --cov-report xml:cov.xml \
             --durations=10 \
             --log-cli-level=info \
             -v
@@ -96,7 +97,7 @@ jobs:
         if: ${{ github.event_name == 'pull_request' || github.event_name == 'push' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' }}
         uses: codecov/codecov-action@18283e04ce6e62d37312384ff67231eb8fd56d24
         with:
-          files: coverage.${{ matrix.python-version }}.${{ matrix.torch-version }}.${{ matrix.runs-on }}
+          files: cov.xml
 
   coverage:
     name: Coverage
@@ -112,7 +113,7 @@ jobs:
       - uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093
         id: download
         with:
-          pattern: coverage-*
+          pattern: coverage*
           merge-multiple: true
 
       - name: Coverage comment
--- a/.github/workflows/ci-sharktank.yml
+++ b/.github/workflows/ci-sharktank.yml
@@ -88,9 +88,6 @@ jobs:
         env:
           COVERAGE_FILE: .coverage.${{ matrix.python-version }}.${{ matrix.torch-version }}.${{ matrix.runs-on }}
 
-      - name: Display structure of files
-        run: ls -la -R .
-
       - name: Upload coverage data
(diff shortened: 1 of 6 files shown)


.github/workflows/ci-sharktank.yml
```diff
@@ -71,14 +76,41 @@ jobs:
...
+            --cov-report=term \
             --durations=10 \
             --log-cli-level=info \
             -v
 
+      - name: Annotate coverage in PR
+        if: ${{ github.event_name == 'pull_request' }}
+        uses: py-cov-action/python-coverage-comment-action@v3
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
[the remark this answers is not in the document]
zphoenixrises: I agree. It is definitely worth trying out

stbaione (approved):
Looks good to me, but would wait on @ScottTodd's approval. Do you have plans to add coverage for the `shortfin` unit tests as well?

zphoenixrises:
I'll add it to the rest of the repo after we are try it out on sharktank

.github/workflows/ci-sharktank.yml
```diff
@@ -71,14 +75,35 @@ jobs:
...
+            --cov-report=term \
             --durations=10 \
             --log-cli-level=info \
             -v
 
+      - name: Annotate coverage in PR
+        if: ${{ github.event_name == 'pull_request' && runner.os != 'Windows' }}
+        uses: py-cov-action/python-coverage-comment-action@970a227e0c16ef4589a99a9970ab0ceb8c53059a
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
ScottTodd, line 93: Does this need to be passed explicitly? Does this action work from forks? It looks like it did, based on this PR itself coming from a fork.

.github/workflows/ci-sharktank.yml
```diff
@@ -71,14 +75,35 @@ jobs:
...
+
+      - name: Upload coverage to Codecov
+        if: ${{ github.event_name == 'pull_request' || github.event_name == 'push' }}
+        uses: codecov/codecov-action@18283e04ce6e62d37312384ff67231eb8fd56d24
+        with:
+          files: coverage.xml
```
ScottTodd, line 99: These events don't include `workflow_dispatch` or `schedule`. Is that intentional? (This workflow in particular is set up to run on `workflow_dispatch` for testing, but other nightly jobs may use `schedule` and coverage reports could be useful there too)

ScottTodd (approved):
Thanks! Eager to see what data and development practices this will gain us :D

.github/workflows/ci-sharktank.yml
zphoenixrises: Added both events

.github/workflows/ci-sharktank.yml
zphoenixrises: The token does need to be passed explicitly it seems. Most likely to be able to post comments on PRs and to be able to update the PR that is managed by this action.  

Reference: https://github.com/py-cov-action/python-coverage-comment-action/tree/970a227e0c16ef4589a99a9970ab0ceb8c53059a/

.github/workflows/ci-sharktank.yml
```diff
@@ -71,13 +75,60 @@ jobs:
...
+
+      - name: Upload coverage to Codecov
+        if: ${{ github.event_name == 'pull_request' || github.event_name == 'push' || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' }}
+        uses: codecov/codecov-action@18283e04ce6e62d37312384ff67231eb8fd56d24
+        with:
+          files: coverage.xml
+
+  coverage:
+    name: Coverage
+    runs-on: ubuntu-latest
```
ScottTodd, line 104: We usually pin to a specific version here also (e.g. `ubuntu-24.04`)

.github/workflows/ci-sharktank.yml
```diff
@@ -71,13 +75,60 @@ jobs:
...
+    needs: test
```
ScottTodd, line 105: If you want this coverage job to block PR merging, you can add it to `needs:` down in `ci_sharktank_summary` at the bottom of the file.

sharktank/sharktank/kernels/topk.py
ScottTodd, line 1: Want to revert the changes to this file before merge? Looks like they were just for testing.

ScottTodd (approved):

.github/workflows/ci-sharktank.yml
zphoenixrises: I think we can enforce that after we have a few runs of this.

sharktank/sharktank/kernels/topk.py
zphoenixrises: Good point!

PR #1627, opened 2025-06-12
[sharktank] Load parameter index using a file descriptor instead of host pointers

archana-ramalingam:
Load parameter index using a file descriptor instead of host pointers as requested in this [issue](https://github.com/iree-org/iree/issues/20900#issuecomment-2931891750)

### DIFF
--- a/sharktank/sharktank/evaluate/perplexity_iree.py
+++ b/sharktank/sharktank/evaluate/perplexity_iree.py
@@ -217,29 +217,17 @@ class PerplexityIree:
 
         shard_count = self.tensor_parallelism_size
 
-        vm_instance = ireert.VmInstance()
         self.devices: list[iree.runtime.HalDevice] = get_iree_devices(
             device=self.iree_devices,
             device_count=self.pipeline_parallelism_size * shard_count,
             allow_repeating=True,
         )
-        hal_module = iree.runtime.create_hal_module(
-            instance=vm_instance, devices=self.devices
-        )
-        self.weight_path = Path(self.weight_path_str)
-        parameter_index = iree.runtime.ParameterIndex()
-        handles = get_file_handle(shard_count=shard_count, weight_path=self.weight_path)
-        for handle in handles:
-            parameter_index.load_from_file_handle(handle, "irpa")
 
-        parameter_provider = parameter_index.create_provider(scope="model")
-        parameters_module = iree.runtime.create_io_parameters_module(
-            vm_instance, parameter_provider
-        )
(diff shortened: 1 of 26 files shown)


sharktank/sharktank/evaluate/perplexity_iree.py
```diff
@@ -215,6 +215,33 @@ def load_model(
...
+
+        parameter_provider = parameter_index.create_provider(scope="model")
+        parameters_module = iree.runtime.create_io_parameters_module(
+            vm_instance, parameter_provider
+        )
+        self.vm_module = iree.runtime.VmModule.mmap(vm_instance, str(self.output_vmfb))
+        self.vm_context = iree.runtime.VmContext(
+            instance=vm_instance,
+            modules=(hal_module, parameters_module, self.vm_module),
+        )
```
sogartar, line 243: Is there a reason to not use `load_iree_module` here? If `load_iree_module` does not support something that is required we should add it there.

sharktank/sharktank/utils/iree.py
```diff
@@ -112,6 +112,22 @@ def f(
...
+) -> List[FileHandle]:
+    handles = []
+    modified_weight_path = weight_path
+    for i in range(shard_count):
+        if shard_count > 1:
+            modified_weight_path = str(
+                weight_path.with_suffix(f".rank{i}{weight_path.suffix}")
+            )
+        with open(str(modified_weight_path), "rb") as f:
+            handles.append(FileHandle.wrap_fd(f.fileno(), True, False))
```
sogartar, line 127: You could skip the `FileHandle` default args or use named parameters. Just `True, False` is not that informative.

sharktank/tests/layers/sharded_conv2d_with_iree_test.py
```diff
@@ -75,12 +75,9 @@ def run_iree_module(devices: list[iree.runtime.HalDevice]):
...
         parameter_index = iree.runtime.ParameterIndex()
-        for i in range(shard_count):
-            parameter_index.load(
-                file_path=str(
-                    Path(params_path).with_suffix(f".rank{i}{params_path.suffix}")
-                )
-            )
+        handles = get_file_handle(shard_count=shard_count, weight_path=params_path)
+        for handle in handles:
+            parameter_index.load_from_file_handle(handle, "irpa")
```
sogartar, line 80: This test was written before `load_iree_module`, but it would be nice to use it.

sogartar (approved):
This can be done in another PR, but as you can see refactoring when we have code duplication of IREE module loading has a cost.

sogartar (approved):

PR #1629, opened 2025-06-12
Makes topk kernel much faster

rsuderman:


zeeshanhaque21 (approved):
LGTM!

ScottTodd:
Drive by: the PR description is empty here. Can you quantify or qualify the "much faster"? Would be good to get in the habit of including benchmark results, traces, etc. on PRs until we have automation doing that for us.

PR #1635, opened 2025-06-12
Add sdxl serving test for release

pravg-amd:


shortfin/tests/apps/sd/serving_test.py
```diff
@@ -0,0 +1,186 @@
...
+    "seed": [0],
+    "output_type": ["base64"],
+    "rid": ["string"],
+}
+
+
+def start_server(fibers_per_device=1, isolation="per_fiber"):
+    # Start the server
+    srv_args = [
+        "python",
```
ScottTodd, line 47: Prefer to use `sys.executable` instead of `python` to launch a python subprocess from another python script (that's more portable across operating systems and inside/outside of a venv)

shortfin/tests/apps/sd/serving_test.py
```diff
@@ -0,0 +1,186 @@
...
+        "-m",
+        "shortfin_apps.sd.server",
+    ]
+    with open("sdxl_config_i8.json", "wb") as f:
+        r = requests.get(
+            "https://sharkpublic.blob.core.windows.net/sharkpublic/sdxl/04082025/configs/sdxl_config_i8.json",
+            allow_redirects=True,
+        )
+        f.write(r.content)
```
ScottTodd, line 56: Why is production code depending on a json file from a remote server? This should be bundled as part of shortfin itself.

shortfin/tests/apps/sd/serving_test.py
```diff
@@ -0,0 +1,186 @@
...
+        [
+            f"--model_config=sdxl_config_i8.json",
+            f"--fibers_per_device={fibers_per_device}",
+            f"--isolation={isolation}"
+        ]
+    )
+    runner = ServerRunner(srv_args)
+    # Wait for server to start
+    time.sleep(3)
+    return runner
```
ScottTodd, line 67: Is there a more reliable way to wait here? This could be a source of flaky tests.

shortfin/tests/apps/sd/serving_test.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+@pytest.fixture(scope="module")
+def sd_server_fpd1():
+    runner = start_server(fibers_per_device=1)
+
+    yield runner
+
+    if os.path.exists("sdxl_config_i8.json"):
+        os.remove("sdxl_config_i8.json")
```
ScottTodd, line 77: Oh, this is writing into the source directory? If keeping the file download, prefer to use a temp directory (pytest has utilities for this), so there is no risk of leaving a loose file in the source directory that could accidentally get committed to git.

(also prefer to use pathlib over `os.path`)

shortfin/tests/apps/sd/serving_test.py
```diff
@@ -0,0 +1,186 @@
...
+
+    # Teardown: kill the server
+    del runner
+
+@pytest.mark.system("amdgpu")
+def test_sd_server(sd_server_fpd1):
+    imgs, status_code = send_json_file(sd_server_fpd1.url)
+    assert len(imgs) == 1
+    ref_path = os.path.dirname(__file__)
+    ref_path = os.path.join(ref_path, "refs/snow_cat.png")
```
ScottTodd, line 87: prefer pathlib over `os.path`

shortfin/tests/apps/sd/serving_test.py
```diff
@@ -0,0 +1,186 @@
...
+    if not os.path.exists(ref_path):
+        assert 0, f"Reference image does not exist at {ref_path}"
+    ref_image = Image.open(ref_path)
+    ref_image_numpy = numpy.array(ref_image)
+    gen_image_numpy = numpy.array(imgs[0])
+    assert (ref_image_numpy == gen_image_numpy).all()
```
ScottTodd, line 93: Exact image comparison against a compressed image? How reliable is that? Shouldn't this use a perplexity check or some other ML-friendly metric?

shortfin/tests/apps/sd/serving_test.py
```diff
@@ -0,0 +1,186 @@
...
+            each_image = image_from(Base64CharacterEncodedByteSequence(each_png))
+            imgs.append(each_image)
+
+    except requests.exceptions.RequestException as e:
+        print(f"Error sending the request: {e}")
+
+    return imgs, response.status_code
+
+
+def find_free_port():
```
ScottTodd, line 170: I've seen similar code in other parts of shortfin. Can we use a common helper function instead of reimplementing? (We should have multiple tests using servers like this, so such tests should have a robust library of utils to call on)

shortfin/tests/apps/sd/serving_test.py
```diff
@@ -0,0 +1,186 @@
...
+    """
+    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
+        s.bind(("localhost", 0))
+        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        return s.getsockname()[1]
+
+
+def test_placeholder():
+    # Here in case this pytest is invoked via CPU CI and no tests are run.
+    pass
```
ScottTodd, line 186: What is the actual failure in that case? This shouldn't be required I think...

PR #1640, opened 2025-06-12
Move attn_type_map to a single location

Alex-Vasile:
One copy is easier to maintain and means we can't forget about updating both.


archana-ramalingam (approved):
LGTM, ideally this and few other parameters need to be setup in the `llm_configs.py`.

Alex-Vasile:
> LGTM, ideally this and few other parameters need to be setup in the `llm_configs.py`.

We would have to pass it in directly to the llama configs when creating it. We had previously discussed something similar for pipeline parallelism and we didn't want the config object modifying itself after initialization, which is what would likely need to happen for it do the lookup.

PR #1642, opened 2025-06-13
First Mooncake KV cache implementation for serving llm models

lisaliu1:
This PR adds the connection between shortfin server and Mooncake kv store. It includes the following work:

1. A Mooncake kv store client wrapper namely the MooncakeStore class (defined in mooncake.py).   This class is used to set up connection between Mooncake kv store server, and send PUT, GET requests to store the KV pairs
2. A KV cache implementation namely the MooncakePagedAttentionCache class (defined in mooncake_attention_cache.py). This class inherits TriePagedAttentionCache class and serves as a second level cache of the published KV pairs. Each published page is stored in mooncake with a key that corresponds to its token_ids.
3. A cache refill implementation namely the MooncakePagedAttentionCacheAllocation class (defined in mooncake_attention_cache.py). This class populates the trie structure and the pages in page_pool with mooncake cache when a page miss happens.

### DIFF
--- a/shortfin/python/shortfin_apps/llm/components/kvcache/mooncake.py
+++ b/shortfin/python/shortfin_apps/llm/components/kvcache/mooncake.py
@@ -17,7 +17,6 @@ from safetensors.numpy import load as safetensors_load
 from safetensors.numpy import save as safetensors_save
 
 from typing import Optional
-from mooncake.store import MooncakeDistributedStore
 
 logger = logging.getLogger(__name__)
 
@@ -62,6 +61,17 @@ class MooncakeStore:
     """
 
     def __init__(self, config: MooncakeConfig):
+
+        try:
+            from mooncake.store import MooncakeDistributedStore
+        except ImportError as e:
+            raise ImportError(
+                "Mooncake is not installed. "
+                "Please run "
+                "pip3 install mooncake-transfer-engine "
+                "to install the Mooncake package."
+            ) from e
+
         try:
             self.config = config
             self.store = MooncakeDistributedStore()
--- a/shortfin/docker/Dockerfile.shortfin-rocm-mooncake
+++ b/shortfin/docker/Dockerfile.shortfin-rocm-mooncake
@@ -43,6 +43,5 @@ ENV OUTPUT_CONFIG_PATH=$EXPORT_DIR/config.json
 ENV VMFB_PATH=$EXPORT_DIR/model.vmfb
 ENV BS=4
 ENV ROCR_VISIBLE_DEVICES=0
-#ENV MOONCAKE_CONFIG_PATH=/app/models/Mooncake/mooncake.json
 
 CMD ["/bin/bash"]


shortfin/python/shortfin_apps/llm/components/kvcache/mooncake_attention_cache.py
```diff
@@ -0,0 +1,312 @@
+from typing import Dict, Set, List, Tuple, Optional
+from dataclasses import dataclass
+import time
+import math
+import heapq
```
stbaione, line 5: In this file in general, I'm not sure if inheritance is the best move here. Maybe opt for composition instead.

My reasoning is:

1. If I understand correctly, mooncake should work with many prefix sharing algorithms. I.e. when we add a new hashing algorithm class, or whatever new technique they come up with in the future, it should also benefit from mooncake. Due to this, I'm not sure if it makes sense to implement mooncake, such that it requires inheritance from one of the kvcache classes.
2. Looking through the way the inheritance is written, it seems more like these should be standalone classes that can be accessed from a cache or an allocation, if mooncake is enabled. There are a couple spots, where we override, repeat logic from the base class, then just add an extra line to make a call to the mooncake store.
3. We may find ourselves in a spot in the future, where we need to duplicate changes between these classes and the `Trie` classes. We also would eventually find ourselves in a position where we'll want to add another algorithm, but will be blocked by first needing to de-couple this from the `Trie` class

shortfin/python/shortfin_apps/llm/components/kvcache/mooncake_attention_cache.py
```diff
@@ -0,0 +1,312 @@
...
+        Args:
+            token_ids: List of token ids as the key for the page
+            page: PageInfo object containing page details
+        """
+        if not self.mooncake_store:
+            raise ValueError("Mooncake store is not initialized")
+
+        key = self.token_ids_to_key(token_ids)
+        logger.debug(f"Sending page with key {key} to Mooncake store.")
+        value = self.page_pool.get_page_data(page)
```
stbaione, line 212: This will run into concurrency issues.

The `page_table` is stored on device. The `device_array` that gets created in this function is on host. You need to `await` the device that the `page_table` is stored on to ensure that the data is finished copying over.

See [LlmExecutorProcess._transfer_buffer](https://github.com/nod-ai/shark-ai/blob/main/shortfin/python/shortfin_apps/llm/components/batcher.py#L296).

If you don't await the device, then you'll end up sending only partially copied data over to mooncake.

The reason the other `PagePool` functions didn't have this await close to the function call, is because it's defaulting to the `await` that they know is coming eventually in `_transfer_buffers`. Since you send to mooncake before then, it won't be fully copied.

stbaione (changes requested):

shortfin/python/shortfin_apps/utils.py
```diff
@@ -391,6 +391,11 @@ def __init__(
         self.fibers_per_device = fibers_per_device
         self.validate_fiber_configuration()
 
+        # Mooncake configuration path
+        self.mooncake_config_path = None
+        if "MOONCAKE_CONFIG_PATH" in os.environ:
```
stbaione, line 396: It would be better to plumb through the `mooncake_config_path` as a server argument as opposed to an env var.

You can add it as a cli arg [here](https://github.com/nod-ai/shark-ai/blob/main/shortfin/python/shortfin_apps/llm/server.py#L54).

Then you can create a field for it in [ServerParams](https://github.com/nod-ai/shark-ai/blob/main/shortfin/python/shortfin_apps/llm/components/config_struct.py#L218), which should automatically make it accessible to [LlmGenerateService](https://github.com/nod-ai/shark-ai/blob/main/shortfin/python/shortfin_apps/llm/components/service.py#L33).

Also, this is a base class used by other shortfin servers (i.e. SDXL), so we don't want to put anything in here that's Llm specific.

shortfin/python/shortfin_apps/llm/components/kvcache/mooncake_attention_cache.py
lisaliu1: that's a very good point.  I'll change it to composition.

shortfin/python/shortfin_apps/llm/components/service.py
```diff
@@ -122,10 +122,17 @@ def _initialize_page_cache(self):
         page_pool = PagePool(devices=self.devices, config=page_pool_config)
 
         if self.server_params.prefix_sharing_algorithm == "trie":
-            self.page_cache = TriePagedAttentionCache(
-                page_pool=page_pool,
-                tokens_per_page=self.model_params.paged_kv_cache.block_seq_stride,
-            )
+            if self.mooncake_config_path != None:
```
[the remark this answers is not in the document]
lisaliu1: it's corrected now.

shortfin/python/shortfin_apps/llm/components/kvcache/mooncake.py
```diff
@@ -0,0 +1,193 @@
...
+        self.engine.initialize_ext(
+            config.local_hostname,
+            config.metadata_server,
+            config.protocal,
+            config.device_name,
+            config.master_server_address,
+        )
+
+
+class MooncakeStore:
```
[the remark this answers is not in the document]
lisaliu1: I agree. I can add the data zip function in the next PR.

shortfin/python/shortfin_apps/llm/components/kvcache/mooncake_attention_cache.py
lisaliu1: Currently, MooncakeAttentionCache is using other caching algorithms. It's certainly more generic this way.

shortfin/python/shortfin_apps/llm/components/kvcache/mooncake.py
```diff
@@ -0,0 +1,193 @@
...
+    """
+    MooncakeStore is a wrapper around the Mooncake KVCache.
+    It provides methods to connect to the Mooncake master and perform operations on the KVCache.
+    """
+
+    def __init__(self, config: MooncakeConfig):
+        try:
+            from mooncake.store import MooncakeDistributedStore
```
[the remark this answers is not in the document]
stbaione: Is there a reason for the conditional import, instead of just adding it as a dependency?

shortfin/python/shortfin_apps/llm/components/kvcache/mooncake_cache.py
```diff
@@ -0,0 +1,265 @@
+from typing import Dict, Set, List, Tuple, Optional
```
stbaione, line 1: I'd like to get @rsuderman's opinion on this design choice. 

I'm not sure that we NEED a new type of allocation and new type of kvcache for what we have here.

The actual functional diff between this class and the classes we already have are the addition of `write_back_pages` and `update_pages`. Otherwise, it directly pipes to the base class. 

If this `is-a` Allocation and `is-a` Cache, why do we need to store instances of other Allocation/Cache types for the class to work properly? Or is it that a kvcache may `have-a` mooncake storage layer?

I'd argue that introducing a new cache type and a new allocation type is a bigger change than adding a `CacheManager` to coordinate between your `MooncakeStore` and `KVCache` classes, or just storing this as a client on the existing caches and optionally calling it.

I'm not sure if it fits conceptually either. `Base` and `Trie` are different `kvcache algorithms`, where the functionality of the cache itself varies quite a bit.

Mooncake isn't really a `kvcache algorithm` or `kvcache type`. It looks like it's distributed blob storage, same concept as redis, but vRAM instead of RAM. So, it's more like a different `storage` type for `kvcache`, but it's not a kvcache in the sense that it has it's own prefix matching, sharding behavior, is read/written to *by the model* during invocation, etc.

I know there were disagreements with my concerns, so just wanna get another look at this specific choice. Committing to it could cause a lot of down the road head-aches, if it is in fact the wrong abstraction.

shortfin/python/shortfin_apps/llm/components/kvcache/mooncake_cache.py
```diff
@@ -0,0 +1,265 @@
...
+        await device
+
+        mooncake_store = self.cache.mooncake_store
+        for key, value in zip(keys, values):
+            logger.debug(f"Writing back page to Mooncake store with key: {key}")
+            mooncake_store.put_int_list(key, value)
+        logger.info(f"Successfully wrote back {len(keys)} pages to Mooncake store.")
+
+    async def update_pages(self, device: sf.ScopedDevice, token_ids: List[int]) -> bool:
+        """Update pages in the device."""
```
stbaione, line 125: nit: this docstring could use more detail on what the function does

shortfin/python/shortfin_apps/llm/components/kvcache/mooncake.py
```diff
@@ -0,0 +1,154 @@
...
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class MooncakeConfig:
+    """
+    Configuration for connecting to mooncake master.
+    """
+
+    local_hostname: str = "localhost"
```
stbaione, line 27: Should some of these fields be required?

stbaione (changes requested):
A couple nits/small things. I would like to get more feedback on the design choice of creating a new `KVCache/KVCacheAllocation` type.

Needs tests for either of us to be able to speak on the accuracy of the component

amd-vivekag (commented):
Added some review comments.

stbaione:
@amd-vivekag thanks for detailed response! Yeah, I don't have a problem so much with the concept of decorator pattern. We have decorators in the code base already, although without the inheritance: [WrappedAllocation](https://github.com/nod-ai/shark-ai/blob/main/shortfin/python/shortfin_apps/llm/components/device_array_cache.py#L49), [PyDisableBarrier](https://github.com/nod-ai/shark-ai/blob/main/shortfin/python/array_binding.cc#L392).

Concern is more that the inheritance might not conceptually fit, which could make code messier over time as we try to keep fitting it. 

I also think that this might be achievable with a smaller change. I can see the idea of minimizing changes in `batcher`, but I also see adding a new `KVCache` and `KVCacheAllocation`, along with the client classes themselves as a `big` change.

Like you said, there are multiple ways to implement. I could definitely be wrong on this. Just wanna make sure, because KVCache is a pretty important area for the service

shortfin/python/shortfin_apps/llm/components/batcher.py
```diff
@@ -466,6 +464,15 @@ async def get_args(self, bs):
         for page_table in self.page_tables:
             args.append(WrappedAllocation(sfnp.disable_barrier(page_table)))
 
+        updated = True
+        for req in self.exec_requests:
+            if req.allocation is not None:
```
rsuderman, line 469: We should just sub this out so that we can always call `req.allocation.update_pages` or `req.update_pages`. It is likely we will want to do similar things to handle page copies for deferred page forking.

shortfin/python/shortfin_apps/llm/components/batcher.py
```diff
@@ -466,6 +464,15 @@ async def get_args(self, bs):
...
+                updated = await req.allocation.update_pages(
+                    device=self.device0, token_ids=req.input_token_ids
+                )
+            if not updated:
```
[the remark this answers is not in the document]
rsuderman: If we step out for safety as mentioned above we can log and return a failure.

shortfin/python/shortfin_apps/llm/components/batcher.py
```diff
@@ -492,6 +499,18 @@ async def get_results(self, logits, indices, req_count):
 
         for req in self.exec_requests:
             req.done.set_success()
+            # write back pages to mooncake
+            logger.debug("Writing back pages for request %s", req)
+            if req.allocation is not None:
```
rsuderman, line 504: Same thing here about stubbing out the behavior.

shortfin/python/shortfin_apps/llm/components/kvcache/mooncake.py
```diff
@@ -0,0 +1,154 @@
...
+            raise TypeError("Mooncake Store Get Type Error.") from e
+
+        if not value_bytes:
+            return None
+
+        if value_bytes:
+            data = safetensors_load(value_bytes)
+            device_id = int(data["device_id"].item())
+            device = (
+                torch.device("cuda", device_id)
```
rsuderman, line 137: We should avoid using the accelerators. Honestly we should avoid using pytorch at all. If we are running within the service we need to avoid invoking `torch` as it can seize GPU resources and fail to release them. Can we load safetensors into something other than torch?

shortfin/python/shortfin_apps/llm/components/kvcache/mooncake_cache.py
rsuderman: This appears fine to me. We will likely end up re-arrange how mooncake interacts with existing caches over time but for a first pass implementation we shouldn't have any issues here. Its better to get this in and see where we can continue from here.

rsuderman (approved):
I have some minor changes but for the most part I am comfortable with the PR where its at. The only other major improvement is that it would be a good idea to include some basic tests for the mooncake_cache. While we aren't doing cache refactorings yet, I expect they will come up in the next couple of weeks.

rsuderman (approved):
I have some minor changes but for the most part I am comfortable with the PR where its at. The only other major improvement is that it would be a good idea to include some basic tests for the mooncake_cache. While we aren't doing cache refactorings yet, I expect they will come up in the next couple of weeks.

shortfin/python/shortfin_apps/llm/components/kvcache/mooncake.py
lisaliu1: The global_segment_size and local_buffersize are optional. But it's better to keep these fields to make them configurable.

shortfin/python/shortfin_apps/llm/components/kvcache/mooncake.py
lisaliu1: use numpy array instead.

shortfin/python/shortfin_apps/llm/components/kvcache/mooncake.py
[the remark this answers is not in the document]
lisaliu1: The reason is to make Mooncake cache optional, that is, users can still use shortfin if they only want to deploy shortfin on a single node and do not want to have Mooncake package installed.
It also allows our unit test to run smoothly without having to install Mooncake in our unit test environment.

PR #1647, opened 2025-06-16
Precheck for testcase

amd-vivekag:
Add "name-tests-test" as a pre-commit hook for testcase filename convention.

I've verified that the pre-commit is working as expected. I ran following commands to verify the same:
```
pip install pre-commit
pre-commit install

```

Tried committing a wrong filename i.e. renaming back `real_pool_test.py` to `real_pool_tests.py`, `git  commit` failed with following error:
```
python tests naming......................................................Failed
- hook id: name-tests-test
- exit code: 1

shortfin/tests/apps/llm/components/kvcache/trie_attention_cache/real_pool_tests.py does not match pattern ".*_test\.py"

```

### DIFF
--- a/sharktank/tests/models/deepseek/sharded_test.py
+++ b/sharktank/tests/models/deepseek/sharded_test.py
@@ -175,7 +175,6 @@ class DeepseekShardedTest(TempDirTestBase):
         )
 
         def run_iree_module(iree_devices: list[iree.runtime.HalDevice]):
-
             iree_module, vm_context, vm_instance = load_iree_module(
                 module_path=iree_module_path,
                 devices=iree_devices,


amd-vivekag:
@ScottTodd  There are tests which are starting with "test_", which should be valid from pytest point of view, but pre-commit hook is complaining about it. It seems that we can have only one type. What would you suggest? Should I rename those files as well?

ScottTodd:
> @ScottTodd There are tests which are starting with "test_", which should be valid from pytest point of view, but pre-commit hook is complaining about it. It seems that we can have only one type. What would you suggest? Should I rename those files as well?

https://github.com/pre-commit/pre-commit-hooks?tab=readme-ov-file#name-tests-test 

> 
> #### `name-tests-test`
> verifies that test files are named correctly.
> - `--pytest` (the default): ensure tests match `.*_test\.py`
> - `--pytest-test-first`: ensure tests match `test_.*\.py`
> - `--django` / `--unittest`: ensure tests match `test.*\.py`

Yeah looks like either but not both. I don't mind the enforced consistency (all `.*_test.py`).

If we did want to allow both, we could fork https://github.com/pre-commit/pre-commit-hooks/blob/main/pre_commit_hooks/tests_should_end_in_test.py, or add a pre-commit hook using regex and the `fail` language: https://pre-commit.com/#fail.

ScottTodd (approved):
LGTM. Why did workflows fail on https://github.com/nod-ai/shark-ai/actions/runs/15676648349?pr=1647 ? Looks like they were cancelled?

amd-vivekag:
> LGTM. Why did workflows fail on https://github.com/nod-ai/shark-ai/actions/runs/15676648349?pr=1647 ? Looks like they were cancelled?

I'm also not sure why the run got cancelled. Is it because of llama tests taking too much time and all tests got cancelled due to this? I'm just wondering. It could be something else as well.

ScottTodd:
> I liked your suggestion of forking the repo and add an option there. But it will require regular bump (adding maintenance). If this new option gets merged in the repo, then that would be the best (I think).

If going that route, I would fork just that python file and use it as a local hook: https://pre-commit.com/#repository-local-hooks . All the file does is check file paths against a regex, and that's easy enough to reimplement without copyright or maintenance concerns.

It also needs this, so the regex only runs on python files in `tests/` directories:

```yml
-   id: name-tests-test
    name: python tests naming
    description: verifies that test files are named correctly.
    entry: name-tests-test
    language: python
    files: (^|/)tests/.+\.py$
```

> I also checked about `fail` language, that will require listing of all testcase directories in the yaml file. Will that be possible from maintenance point of view?

It might be possible with just regex include/exclude, but a script is definitely easier to work with and test. We have this hook in IREE, for example: https://github.com/iree-org/iree/blob/c514bb7435bb1a9305a4a7cb552a64eb6d9d0941/.pre-commit-config.yaml#L103-L110

> IMHO, having a common convention seems a better option considering maintenance issues in future. Let me know your thoughts.

Consistency SGTM!

PR #1665, opened 2025-06-17
Llama-4 toy prefill Iree vs eager

oyazdanb:
The iree vs eager test for prefill toy llama4.

### DIFF
--- a/sharktank/sharktank/examples/export_paged_llm_v1.py
+++ b/sharktank/sharktank/examples/export_paged_llm_v1.py
@@ -53,7 +53,7 @@ def main():
     dataset_type = cli.get_input_data_files(args)
     dataset_type = "irpa" if "irpa" in dataset_type else "gguf"
     dataset = cli.get_input_dataset(args)
-    hp = configs.LlamaHParams.from_gguf_props(dataset.properties)
+    hp = configs.LlamaHParams(**dataset.properties["hparams"])
     if "tensor_parallelism_size" in dataset.properties:
         dataset_tensor_parallelism_size = dataset.properties["tensor_parallelism_size"]
         if dataset_tensor_parallelism_size != args.tensor_parallelism_size:
@@ -69,6 +69,8 @@ def main():
     block_to_pipeline, pipeline_to_devices = pipeline_parallelize_theta(
         dataset.root_theta, args.pipeline_parallelism_size
     )
+    if hp.model_arch == "llama4":
+        rope_layers = [i for i in range(hp.block_count) if int((i + 1) % 4 != 0)]
 
     llama_config = LlamaModelConfig(
         hp,
@@ -83,6 +85,14 @@ def main():
         activation_dtype=args.activation_dtype,
         attention_dtype=args.attention_dtype,
         kv_cache_dtype=args.kv_cache_dtype,
+        use_qk_norm=True if hp.model_arch == "llama4" else args.use_qk_norm,
+        rope_layers=rope_layers if hp.model_arch == "llama4" else None,
+        attention_chunk_size=args.attention_chunk_size,
+        attn_temperature_tuning=True
+        if hp.model_arch == "llama4"
+        else args.attn_temperature_tuning,
+        floor_scale=args.floor_scale,
+        attn_scale=args.attn_scale,
     )
     llama_config.fake_quant = args.fake_quant
 
--- a/sharktank/sharktank/examples/export_paged_llm_v1.py
+++ b/sharktank/sharktank/examples/export_paged_llm_v1.py
@@ -53,7 +53,7 @@ def main():
     dataset_type = cli.get_input_data_files(args)
     dataset_type = "irpa" if "irpa" in dataset_type else "gguf"
     dataset = cli.get_input_dataset(args)
-    hp = configs.LlamaHParams(**dataset.properties["hparams"])
+    hp = configs.LlamaHParams.from_gguf_props(dataset.properties)
     if "tensor_parallelism_size" in dataset.properties:
         dataset_tensor_parallelism_size = dataset.properties["tensor_parallelism_size"]
         if dataset_tensor_parallelism_size != args.tensor_parallelism_size:
--- a/sharktank/sharktank/layers/rotary_embedding.py
+++ b/sharktank/sharktank/layers/rotary_embedding.py
@@ -166,12 +166,18 @@ class RotaryEmbeddingLayer(BaseLayer):
         _, sl, _, _ = xt_.shape
 
         if self.model_arch == "llama4":
+            if isinstance(rotary_embed_table, ReplicatedTensor):
+                rotary_embed_table = unbox_tensor(rotary_embed_table)
+            bs, sl, _, dim = xt.shape
+            dim_half = rotary_embed_table.size(-1) // 2
+
             freqs_cis_real = rotary_embed_table[0][
-                :, : rotary_embed_table[0].shape[1] // 2
+                start_index : start_index + sl, :dim_half
             ]
             freqs_cis_imag = rotary_embed_table[1][
-                :, : rotary_embed_table[0].shape[1] // 2
+                start_index : start_index + sl, :dim_half
             ]
+
             # TODO: don't use complex numbers as the compiler does better without them.
             freqs_cis = torch.view_as_complex(
                 torch.stack([freqs_cis_real, freqs_cis_imag], dim=-1)
@@ -235,6 +241,26 @@ class RotaryEmbeddingLayer(BaseLayer):
             freqs_cis = (cos[:, None, None, :], sin[:, None, None, :])
             return freqs_cis
 
+        if self.model_arch == "llama4":
+            shape = positions_seq.shape
+
+            def _rows_for(t):
+
+                return (
+                    self._compute_rotary_embed_table(t.flatten())
+                    .unflatten(1, shape)
+                    .permute(1, 2, 0, 3)
+                )
+
+            if isinstance(positions_seq, ReplicatedTensor):
+                freqs_cis = ReplicatedTensor(
+                    ts=[_rows_for(s) for s in positions_seq.shards],
+                    devices=positions_seq.devices,
+                )
+            else:
+                freqs_cis = _rows_for(positions_seq)
+
+            return freqs_cis
         if self.use_table:
             freqs_cis = self.rotary_embed_table[positions_seq.flatten()]
         else:
@@ -244,6 +270,7 @@ class RotaryEmbeddingLayer(BaseLayer):
                     self._compute_rotary_embed_table(s.flatten()).unflatten(0, shape)
                     for s in positions_seq.shards
                 ]
+
                 freqs_cis = ReplicatedTensor(ts=ts)
             else:
                 freqs_cis = self._compute_rotary_embed_table(positions_seq.flatten())
@@ -332,6 +359,29 @@ class RotaryEmbeddingLayer(BaseLayer):
             sin = torch.sin(emb).to(self.dtype)
             return (cos, sin)
 
+        if self.model_arch == "llama4":
+            # half-dimension (cos/sin share it)
+            dim_half = dim // 2
+
+            # inverse frequencies  (same formula as classic RoPE)
+            inv_freq = 1.0 / (
+                self.rope_freq_base
+                ** (torch.arange(0, dim_half, device=self.device).float() / dim_half)
+            )  # [dim/2]
+
+            # outer-product  → [max_sl, dim/2]
+            emb = t.unsqueeze(1).float() * inv_freq.unsqueeze(0)
+
+            cos = torch.cos(emb).to(self.dtype)  # [max_sl, dim/2]
+            sin = torch.sin(emb).to(self.dtype)  # [max_sl, dim/2]
+
+            # shape [2, max_sl, dim/2]  : 0 = cos, 1 = sin
+            table = torch.stack((cos, sin), dim=0)
+
+            # duplicate to full `dim` so later code can slice [:, :, :dim//2]
+            table = torch.cat((table, table), dim=-1)  # [2, max_sl, dim]
+
+            return table
         freqs = 1.0 / (
             self.rope_freq_base ** ((torch.arange(0, dim) // 2).float() / dim * 2.0)
         ).to(device=self.device)
--- a/sharktank/sharktank/utils/cli.py
+++ b/sharktank/sharktank/utils/cli.py
@@ -165,6 +165,30 @@ def add_model_options(parser: argparse.ArgumentParser):
         help="Return the log softmax of the logits",
         choices=["none", "softmax", "log_softmax"],
     )
+    parser.add_argument(
+        "--attention-chunk-size",
+        help="chunk sizes in chunk attention used for llama4",
+        type=int,
+        default=37,
+    )
+    parser.add_argument(
+        "--attn-temperature-tuning",
+        help="tunning temperature for llama4",
+        action="store_true",
+    )
+    parser.add_argument(
+        "--attn-scale",
+        help="attention scale used for llama4",
+        type=float,
+        default=0.2,
+    )
+    parser.add_argument(
+        "--floor-scale",
+        help="attention scale used for llama4",
+        type=int,
+        default=31,
+    )
+    parser.add_argument("--use-qk-norm", help="qk_norm for llama4", action="store_true")
 
 
 def add_model_input_options(parser: argparse.ArgumentParser):
--- a/sharktank/tests/models/llama4/llama4_test_iree_eager.py
+++ b/sharktank/tests/models/llama4/llama4_test_iree_eager.py
@@ -1,0 +1,442 @@
+from sharktank.utils.iree import (
+    with_iree_device_context,
+    load_iree_module,
+    prepare_iree_module_function_args,
+    flatten_for_iree_signature,
+)
+
+from sharktank.layers.configs import LlamaModelConfig
+from sharktank.models.llm import PagedLlmModelV1
+from sharktank.models.llama4.toy_llama4 import generate
+from sharktank.types import *
+from sharktank.utils.evaluate import *
+from sharktank.utils.iree import (
+    get_iree_devices,
+    make_hal_buffer_view_trace_default_callback,
+    TorchLikeIreeModule,
+)
+from sharktank.utils.export_artifacts import ExportArtifacts
+from sharktank.examples import export_paged_llm_v1
+from sharktank.layers.mixture_of_experts_block import MoeBlock
+from sharktank.layers.norm import L2Norm
+from sharktank.ops import topk, zeros_like, reshard_like
+
+import torch
+import logging
+from os import PathLike
+from dataclasses import asdict
+import iree.runtime
+from collections import OrderedDict
+import argparse
+from pathlib import Path
+import sys
+
+logging.basicConfig(level=logging.DEBUG)
+logger = logging.getLogger(__name__)
+
+iree_compile_flags = []  # TODO; fill this if we need any flag
+
+
+def _sample_inputs(self, batch_size=1, dtype=torch.int64):
+    return llama4_toy_sample_inputs(
+        self,
+        config=self.config,
+        batch_size=batch_size,
+        dtype=dtype,
+    )
+
+
+def deterministic_attn_mask(inverted_mask):
+    L = inverted_mask.shape[-1]
+    causal = torch.tril(
+        torch.ones(L, L, dtype=inverted_mask.dtype, device=inverted_mask.device)
+    )
+    return causal.unsqueeze(0)  # (1, L, L)
+
+
+def deterministic_mask(inverted_mask):
+    seq_len = inverted_mask.shape[-1]
+    causal = torch.tril(
+        torch.ones(
+            seq_len, seq_len, dtype=inverted_mask.dtype, device=inverted_mask.device
+        )
+    )
+    return causal.unsqueeze(0)
+
+
+def deterministic_input_mask(seq_lens, sl, *, device="cpu"):
+    return torch.zeros(seq_lens.size(0), sl, dtype=torch.bool, device=device)
+
+
+def deterministic_page_table(page_count, config, dtype=torch.float16):
+
+    sub_page_dims = (
+        config.hp.attention_head_count_kv,  # 4
+        config.hp.expert_used_count,  # 2
+        config.hp.block_count,  # 4
+        config.block_seq_stride,  # 13
+        config.hp.attn_head_dim,  # 8
+    )
+    flat_dim = math.prod(sub_page_dims)
+    table = torch.arange(page_count * flat_dim, dtype=dtype).view(page_count, flat_dim)
+
+    return [table]
+
+
+def llama4_toy_sample_inputs(
+    model: PagedLlmModelV1,
+    config: LlamaModelConfig,
+    batch_size: int = 1,
+    dtype=torch.int64,
+) -> tuple[tuple[AnyTensor], OrderedDict[str, AnyTensor]]:
+
+    hp = config.hp
+    context_len = hp.context_length
+    base = torch.arange(context_len) % hp.vocab_size
+    tokens = base.repeat(batch_size, 1).long()
+
+    page_count = (len(tokens[0]) // config.block_seq_stride) * batch_size
+    kv_cache_state = deterministic_page_table(
+        page_count, config
+    )  # model.cache.allocate(page_count)
+
+    seq_block_ids = torch.arange(
+        start=0, end=tokens.numel() // config.block_seq_stride, dtype=dtype
+    ).view(batch_size, context_len // config.block_seq_stride)
+
+    hf_2d_attention_mask = torch.ones_like(tokens, dtype=dtype)
+    inverted_mask = hf_2d_attention_mask == 0
+    attention_mask = deterministic_mask(
+        inverted_mask
+    )  # model.attention_mask(inverted_mask)
+    seq_lens = torch.full((batch_size,), context_len, dtype=dtype)
+
+    args = tuple()
+    kwargs = OrderedDict(
+        (
+            ("tokens", tokens),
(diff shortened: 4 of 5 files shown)


sharktank/tests/models/llama4/llama4_test_iree_eager.py
```diff
@@ -0,0 +1,411 @@
...
+    ffn_input = h.view(-1, feature_dim)
+
+    # For each token, the router calculates the router weights for all experts
+    # shape: (batch_size * sequence_length, expert_count)
+    router_logits = self.ffn_gate_inp(ffn_input)
+    router_weights = self.score_experts(router_logits.to(torch.float))
+
+    router_weights = reshard_like(router_weights, like=ffn_input)
+
+    def pick_first_expert(scores: torch.Tensor, k: int = 1):
```
Alex-Vasile, line 140: We shouldn't be using patched layers since we will no longer be testing the actual layers. Any change to the MoE block that breaks it won't be caught by this test.

sharktank/tests/models/llama4/llama4_test_iree_eager.py
```diff
@@ -0,0 +1,411 @@
...
+        moe_output = moe_output + self.shared_experts(ffn_input)
+
+    moe_output = moe_output.reshape(batch_size, sequence_length, feature_dim)
+
+    moe_output = self.layer_output_norm(moe_output)
+
+    return moe_output
+
+
+def export_llama4_model_mlir(
```
Alex-Vasile, line 201: Much of this can be re-written to take advantage of ExportArtifact. If there is stuff missing from it it needs to be added. I recently added the iree_run_vmfb function, and am refactoring things futher.

sharktank/tests/models/llama4/llama4_test_iree_eager.py
```diff
@@ -0,0 +1,411 @@
...
+    torch.manual_seed(seed)
+
+    work_dir = Path(args.work_dir)
+
+    MoeBlock.forward = patch_forward_moe
+
+    # add sample_inputs
+    PagedLlmModelV1.sample_inputs = _sample_inputs
+    PagedLlmModelV1.input_mask = staticmethod(deterministic_input_mask)
+    PagedLlmModelV1.attention_mask = staticmethod(deterministic_attn_mask)
```
Alex-Vasile, line 302: Why do the input and attention masks need to be overrided?

sharktank/tests/models/llama4/llama4_test_iree_eager.py
```diff
@@ -0,0 +1,411 @@
...
+        target_torch_model,
+        output_path=work_dir,
+        mlir_output_path=mlir_path,
+        parameters_output_path=parameters_path,
+        batch_size=args.batch_size,
+        config=target_config,
+    )
+
+    iree_devices = get_iree_devices(
+        device="hip://5",
```
Alex-Vasile, line 334: You need to use hip://0 since this will likely run on a single GPU runner.

sharktank/tests/models/llama4/llama4_test_iree_eager.py
```diff
@@ -0,0 +1,411 @@
...
+    batch_size: int,
+    config: LlamaModelConfig,  # TODO; it can be read from HF when comparing the actual model
+):
+    export_llama4_iree_parameters(model, parameters_output_path, config=config)
+    export_llama4_model_mlir(
+        model, output_path=output_path, config=config, batch_size=batch_size
+    )
+
+
+def runCompareIreeAgainstTorchEager(args, atol):
```
Alex-Vasile, line 290: Need to rewrite this so that it gets picked up by pytest and automatically run.

Also likely needs a `@is_mi300x` decorator.

Alex-Vasile (changes requested):

sharktank/tests/models/llama4/llama4_test_iree_eager.py
```diff
@@ -0,0 +1,411 @@
...
+        config.block_seq_stride,  # 13
+        config.hp.attn_head_dim,  # 8
+    )
+    flat_dim = math.prod(sub_page_dims)
+    table = torch.arange(page_count * flat_dim, dtype=dtype).view(page_count, flat_dim)
+
+    return [table]
+
+
+def llama4_toy_sample_inputs(
```
sogartar, line 85: Aren't these prefill sample inputs?

sharktank/tests/models/llama4/llama4_test_iree_eager.py
```diff
@@ -0,0 +1,411 @@
...
+
+    hp = config.hp
+    context_len = hp.context_length
+    base = torch.arange(context_len) % hp.vocab_size
+    tokens = base.repeat(batch_size, 1).long()
+
+    page_count = (len(tokens[0]) // config.block_seq_stride) * batch_size
+    kv_cache_state = deterministic_page_table(
+        page_count, config
+    )  # model.cache.allocate(page_count)
```
sogartar, line 100: Why not use `model.cache.allocate`?

sharktank/tests/models/llama4/llama4_test_iree_eager.py
sogartar: Yes, if there is something wrong or not implemented in the MoE, we should do that instead.

sharktank/tests/models/llama4/llama4_test_iree_eager.py
```diff
@@ -0,0 +1,411 @@
...
+    )
+
+
+def export_llama4_iree_parameters(
+    model: PagedLlmModelV1,
+    parameters_output_path: PathLike,
+    config: LlamaModelConfig,  # TODO; it can be read from HF when comparing the actual model
+):
+    config_dict = {
+        "hparams": asdict(config.hp),
```
sogartar, line 269: The whole config should go into the metadata.
```
config_dict = config.to_properties()
```

sharktank/tests/models/llama4/llama4_test_iree_eager.py
```diff
@@ -0,0 +1,411 @@
...
+
+    moe_output = moe_output.reshape(batch_size, sequence_length, feature_dim)
+
+    moe_output = self.layer_output_norm(moe_output)
+
+    return moe_output
+
+
+def export_llama4_model_mlir(
+    model_or_parameters_path: PagedLlmModelV1 | PathLike,
```
sogartar, line 202: This parameters looks unused.

sharktank/tests/models/llama4/llama4_test_iree_eager.py
```diff
@@ -0,0 +1,411 @@
...
+    )
+
+    iree_devices = get_iree_devices(
+        device="hip://5",
+        device_count=1,
+    )
+    iree_module_path = work_dir / "model.vmfb"
+
+    def run_iree_module(iree_devices: list[iree.runtime.HalDevice]):
+        cpu_device = get_iree_devices(driver="local-task", device_count=1)
```
sogartar, line 340: You should be using only the devices provided by the function args.
They need to outlive this function. There is something weird going on during torch tensors destruction and the device may get destroyed before all tensors backed by IREE buffers gets destroyed. Which would lead to a crash.

sharktank/tests/models/llama4/llama4_test_iree_eager.py
```diff
@@ -0,0 +1,411 @@
...
+        iree_buffere_view_trace_callback = make_hal_buffer_view_trace_default_callback(
+            cpu_device[0]
+        )
+        debug_sink = iree.runtime.HalModuleDebugSink(iree_buffere_view_trace_callback)
```
sogartar, line 344: Tensor tracing should be remove as it is debugging remnants.

sharktank/tests/models/llama4/llama4_test_iree_eager.py
sogartar: This test should use the `get_iree_flags` pytest fixture to acquire the target device and related compilation flags.
Then the test becomes retargetable and can run on CPU, HIP, etc.

sharktank/tests/models/llama4/llama4_test_iree_eager.py
```diff
@@ -0,0 +1,411 @@
...
+            debug_sink=debug_sink,
+        )
+        iree_kw = {
+            "tokens": kw["tokens"],
+            "seq_lens": kw["seq_lens"],
+            "seq_block_ids": kw["seq_block_ids"][0],
+            "cache_state": kw["cache_state"],
+        }
+
+        flat_args = flatten_for_iree_signature(iree_kw)
```
sogartar, line 359: This is unused.

sharktank/tests/models/llama4/llama4_test_iree_eager.py
```diff
@@ -0,0 +1,411 @@
...
+        msg=f"Actual vs expected results diff > {atol}",
+    )
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--work_dir", type=str)
+    parser.add_argument("--batch_size", type=int)
+    args = parser.parse_args()
+    runCompareIreeAgainstTorchEager(args, atol=1e-1)
```
sogartar, line 411: Instead of this we should have function with a name prefixed by `test` so it gets picked up by `pytest`.
Then you run the tests with
```
pytest sharktank/tests/models/llama4/llama4_test_iree_eager.py
```

sogartar (commented):
My major objection is the conditional logic based on the `model_arch` config.
We have opened the flood gates with the expectation that we would fix this, but with every addition the problem gets worse.

The same hold for the slew of flags for the LLM export script.

sharktank/sharktank/layers/rotary_embedding.py
```diff
@@ -235,6 +241,26 @@ def compute_batch_mask(
...
+
+            if isinstance(positions_seq, ReplicatedTensor):
+                freqs_cis = ReplicatedTensor(
+                    ts=[_rows_for(s) for s in positions_seq.shards],
+                    devices=positions_seq.devices,
+                )
+            else:
+                freqs_cis = _rows_for(positions_seq)
+
+            return freqs_cis
```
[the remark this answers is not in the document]
oyazdanb: When I run the code as it is, I see an error[ here](https://github.com/nod-ai/shark-ai/blob/fa71e409419189c8ef1eae9c06245ed5b878d270/sharktank/sharktank/layers/rotary_embedding.py#L187) 

`xt_out = torch.view_as_real(xt_ * freqs_cis[:, :, None, :]).flatten(3)`

and the error is 
`RuntimeError: The size of tensor a (s0) must match the size of tensor b (143) at non-singleton dimension 1)`

it turned out that can be iree export issue (I might need to open an iree ticket). when we use this 
`freqs_cis_real = rotary_embed_table[0][: , :dim_half]`

`xt_` is trying to be broadcasted to [bs,s0,n_head,dim/2] where s0 is a symbolic shape and s0 !=143. But when we slice it as start_index : start_index + sl, s0=sl=143 and it runs.

sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -69,6 +70,8 @@ def main():
     block_to_pipeline, pipeline_to_devices = pipeline_parallelize_theta(
         dataset.root_theta, args.pipeline_parallelism_size
     )
+    if hp.model_arch == "llama4":
+        rope_layers = [i for i in range(hp.block_count) if int((i + 1) % 4 != 0)]
```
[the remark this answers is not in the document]
vivekkhandelwal1: Also, I am not sure if this is the correct way for how the rope layers are determined.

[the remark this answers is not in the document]
oyazdanb: i fixed this part and removed the conditions based on llama4 in the file.

sharktank/sharktank/utils/cli.py
```diff
@@ -165,6 +165,30 @@ def add_model_options(parser: argparse.ArgumentParser):
...
+        type=float,
+        default=0.2,
+    )
+    parser.add_argument(
+        "--floor-scale",
+        help="attention scale used for llama4",
+        type=int,
+        default=31,
+    )
+    parser.add_argument("--use-qk-norm", help="qk_norm for llama4", action="store_true")
```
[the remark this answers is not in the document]
oyazdanb: fixed.

sharktank/tests/models/llama4/llama4_test_iree_eager.py
```diff
@@ -0,0 +1,411 @@
...
+        )
+    )
+    return causal.unsqueeze(0)
+
+
+def deterministic_input_mask(seq_lens, sl, *, device="cpu"):
+    return torch.zeros(seq_lens.size(0), sl, dtype=torch.bool, device=device)
+
+
+def deterministic_page_table(page_count, config, dtype=torch.float16):
```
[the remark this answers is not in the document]
oyazdanb: fixed.

sharktank/tests/models/llama4/llama4_test_iree_eager.py
```diff
@@ -0,0 +1,411 @@
...
+    seq_len = inverted_mask.shape[-1]
+    causal = torch.tril(
+        torch.ones(
+            seq_len, seq_len, dtype=inverted_mask.dtype, device=inverted_mask.device
+        )
+    )
+    return causal.unsqueeze(0)
+
+
+def deterministic_input_mask(seq_lens, sl, *, device="cpu"):
```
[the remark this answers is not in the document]
oyazdanb: i removed this function as we did not need it.

sharktank/tests/models/llama4/llama4_test_iree_eager.py
oyazdanb: fixed.

sharktank/tests/models/llama4/llama4_test_iree_eager.py
oyazdanb: if i flatten all the configs here, it gives me some error in from_gguf_props; as this is for the toy model, i added a patch for from_gguf_props for the toy model.

sharktank/tests/models/llama4/llama4_test_iree_eager.py
oyazdanb: removed.

sharktank/tests/models/llama4/llama4_test_iree_eager.py
oyazdanb: removed cpu_device.

sharktank/tests/models/llama4/llama4_test_iree_eager.py
oyazdanb: removed.

sharktank/tests/models/llama4/llama4_test_iree_eager.py
oyazdanb: fixed.

sharktank/sharktank/utils/cli.py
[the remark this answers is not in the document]
sogartar: The arguments are still here.
The export script should either read the args form the meta parameters of the IRPA/GGUF using `LlamaModelConfig.from_properties` or have 1 argument to supply a serialized file.

Long-term we need to decouple parameters/weights and meta parameters as weights can be shared between multiple model configurations.

sharktank/tests/models/llama4/llama4_test_iree_eager.py
sogartar: This seems unchanged.

sogartar (commented):
I think you may have forgotten to push some changes.

sharktank/sharktank/utils/cli.py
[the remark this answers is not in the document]
oyazdanb: fixed.

sharktank/tests/models/llama4/llama4_test_iree_eager.py
```diff
@@ -0,0 +1,39 @@
...
+import pytest
+
+logging.basicConfig(level=logging.DEBUG)
+logger = logging.getLogger(__name__)
+
+iree_compile_flags = []  # TODO; fill this if we need any flag
+
+
+@pytest.mark.usefixtures("get_iree_flags", "device")
+@is_mi300x
```
sogartar, line 20: You don't need `@is_mi300x` if the test is retargetable using the `get_iree_flags` fixture. Then we would be able to run it for any backend by passing the appropriate flags.

I actually see that the `IreeVsEagerLLMTester` class does not handle CPU flags appropriately and does not use [get_iree_compiler_flags](https://github.com/nod-ai/shark-ai/blob/3f196d60c33dec590dca29aa2ce120b629a5b3b5/sharktank/sharktank/utils/iree.py#L131C5-L131C28).

sharktank/tests/models/llama4/llama4_test_iree_eager.py
```diff
@@ -0,0 +1,39 @@
...
+class TestLlama4IreeEager(TempDirTestBase):
+    def testUnshardedToySizedModelIREEVsEager(self):
+        seed = 1234
+        random.seed(seed)
+        torch.manual_seed(seed)
+        config = testing.make_toy_model_config(dtype=torch.float16)
```
sogartar, line 26: It would be nice to have a f32 test as well with more stringent tolerances.
You could parametrize the test on the dtype, rtol and atol.

sharktank/sharktank/utils/testing.py
```diff
@@ -36,8 +36,12 @@
     reason="Run quick tests if --run-quick-test is passed",
 )
 is_nightly = pytest.mark.skipif(
-    'not config.getoption("run-nightly-tests")',
-    reason="Run large tests if --run-nightly-tests is passed",
+    'not config.getoption("run-nightly-test")',
+    reason="Run large tests if --run-nightly-test is passed",
```
sogartar, line 40: Actually all nightly tests run whatever is marked [expensive](https://github.com/nod-ai/shark-ai/blob/3f196d60c33dec590dca29aa2ce120b629a5b3b5/sharktank/pyproject.toml#L53). At some point we need to reconcile the various parallel approaches for marking tests for nightly. We should use markers as then you could construct logical expressions to make your test selection.

sogartar (approved):
Thank you. Just a few minor comments.

sharktank/sharktank/layers/rotary_embedding.py
```diff
@@ -166,12 +166,18 @@ def forward_unsharded(
...
 
         if self.model_arch == "llama4":
+            if isinstance(rotary_embed_table, ReplicatedTensor):
+                rotary_embed_table = unbox_tensor(rotary_embed_table)
+            bs, sl, _, dim = xt.shape
+            dim_half = rotary_embed_table[0].shape[1] // 2
+
             freqs_cis_real = rotary_embed_table[0][
-                :, : rotary_embed_table[0].shape[1] // 2
+                start_index : start_index + sl, :dim_half
```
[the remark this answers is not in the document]
sogartar: When looking at the code bellow this change
```
            # TODO: don't use complex numbers as the compiler does better without them.
            freqs_cis = torch.view_as_complex(
                torch.stack([freqs_cis_real, freqs_cis_imag], dim=-1)
            )
            freqs_cis = freqs_cis.unsqueeze(0)
            xt_ = torch.view_as_complex(xt.float().reshape(*xt.shape[:-1], -1, 2))
            xt_out = torch.view_as_real(xt_ * freqs_cis[:, :, None, :]).flatten(3)
            return xt_out.type_as(xt)
```
It really does make sense as if the seq len is not matching
```
xt_ * freqs_cis[:, :, None, :]
```
will fail.

sharktank/sharktank/utils/testing.py
```diff
@@ -188,7 +192,7 @@ def __init__(
             use_attention_mask=True,
         )
 
-        self.config.device = torch.device(torch_device)  # Switch to gpu for eager mode
+        # self.config.device = torch.device(torch_device)  # Switch to gpu for eager mode
```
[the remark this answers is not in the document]
oyazdanb: returned it back; it was done by mistake.

sharktank/sharktank/utils/export_artifacts.py
```diff
@@ -360,13 +359,8 @@ def export_llm_to_mlir(
...
             export_args.append("--use-hf")
-
-        self._run_cmd(
-            cmd=subprocess.list2cmdline(export_args),
-            run_msg="Exporting MLIR",
-            success_msg="Exported to MLIR successfully",
-            exception=ExportMlirException,
-        )
+        sys.argv = export_args
+        export_paged_llm_v1.main()
```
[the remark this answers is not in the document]
archana-ramalingam: We might want to retain the `ExportMlirException` handling.

sharktank/sharktank/utils/export_artifacts.py
```diff
@@ -356,13 +355,8 @@ def export_llm_to_mlir(
...
```
Alex-Vasile, line 359: We shouldn't do this. If there is an infinite loop or the program hands, we should find out why. This works for both llama and deepseek

Alex-Vasile (changes requested):

sharktank/sharktank/layers/configs/llm_configs.py
```diff
@@ -112,9 +112,10 @@ def from_gguf_props(p: dict[str, Any]):
         expert_count = _optional_int_prop(
             p, f"{name_prefix}.expert_count", default_expert_count
         )
-        defaut_n_dense_layers = 0 if expert_count and expert_count > 0 else None
+        # Changing default_n_dense_layers to 0, gives error in Moe block condition for llama4 in llm.py
+        default_n_dense_layers = None
         n_dense_layers = _optional_int_prop(
```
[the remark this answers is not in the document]
oyazdanb: i've moved res["n_dense_layers"] to deepseek custom configs.

sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -69,21 +69,19 @@ def main():
...
+    llama_config.pipeline_parallelism_size = args.pipeline_parallelism_size
+    llama_config.block_to_pipeline_map = block_to_pipeline
+    llama_config.pipeline_to_device_map = pipeline_to_devices
+    llama_config.use_hf = args.use_hf
+    llama_config.static_tables = (False,)  # Rely on the compiler for hoisting tables.
+    llama_config.attention_kernel = args.attention_kernel
+    llama_config.block_seq_stride = args.block_seq_stride
+    llama_config.activation_dtype = args.activation_dtype
+    llama_config.attention_dtype = args.attention_dtype
+    llama_config.kv_cache_dtype = args.kv_cache_dtype
```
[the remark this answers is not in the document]
oyazdanb: I have to get the config from `from_properties` as it flattens some configs we have for llama4 which are here https://github.com/nod-ai/shark-ai/blob/users/oyazdanb/llama4/sharktank/sharktank/models/llama4/testing.py#L47-L51, otherwise I have to pass them as args.

sharktank/tests/models/llama4/llama4_test_iree_eager.py
archana-ramalingam: CPU is not our goal, but still @Alex-Vasile can take a look.

sharktank/sharktank/utils/export_artifacts.py
Alex-Vasile: `capture_output=True` is needed in order to use the xfail that matches on a string.

sharktank/sharktank/utils/export_artifacts.py
```diff
@@ -242,8 +242,14 @@ def _run_cmd(
...
         logger.info(f"{run_msg}:\n" f"cd {self.cwd} && {cmd}")
+
         proc = subprocess.run(
-            cmd, shell=True, capture_output=True, text=True, cwd=self.cwd
+            cmd,
+            shell=True,
+            cwd=self.cwd,
+            stdout=sys.stdout,
+            stderr=sys.stderr,
+            text=True,
```
archana-ramalingam, line 252: Can we add `capture_output=True` back here? This is required to match the error with the export/compile exception.

sharktank/tests/models/llama4/llama4_test_iree_eager.py
Alex-Vasile: Fixing `IreeVsEagerLLMTester` to be retargetable and use the `get_iree_flags` fixture is a good extension, but that's beyond this PR. Until then the `@is_mi300x` is needed.

sharktank/tests/models/llama4/llama4_test.py
```diff
@@ -94,3 +94,32 @@ def run_hf_model():
...
+            iree_hal_target_device=self.iree_hal_target_device,
+            skip_decode=True,
+            use_qk_norm=True,
+            attention_chunk_size=37,
+        )
+        tester.run_and_compare_iree_vs_eager(atol=atol, rtol=rtol)
+
+    def testUnshardedToySizedModelIREEVsEager(self):
+        for dtype, atol, rtol in [(torch.float16, 1e-1, 1e-1)]:
+            self.helper_run(dtype=dtype, atol=atol, rtol=rtol)
```
Alex-Vasile, line 125: This should be implemented with a `@paramaterize`.

sharktank/tests/models/llama4/llama4_test.py
```diff
@@ -28,7 +28,7 @@ def setUp(self):
 
     @pytest.mark.xfail(
         is_mi300x,
-        raises=TypeError,
+        raises=(TypeError, AssertionError),
```
Alex-Vasile, line 31: Why two different errors?

Alex-Vasile (commented):
Most of it looks good, two last comments left.

sharktank/tests/models/llama4/llama4_test.py
Alex-Vasile: If it's an assertion error, leave only that error there to ensure that the reason for the failure doesn't switch for an unrelated reason.

PR #1668, opened 2025-06-18
[sharktank] in perplexity eval add an optional arg to specify prefill length

sogartar:
We start calculating perplexity with the last prefill token.
This allows for more flexibility how perplexity is calculated.


Alex-Vasile:
The rotary_test is being flaky, Not sure what's wrong with it as I haven't been able to reproduce locally.

archana-ramalingam:
> The rotary_test is being flaky, Not sure what's wrong with it as I haven't been able to reproduce locally.

A re-run fixed it. Couple of these test failures pass on a re-run.

archana-ramalingam (approved):
Thanks for adding this flexibility to ppl tests.

PR #1672, opened 2025-06-18
Further ExportArtifacts and benchmark_amd_gpu refactors

Alex-Vasile:
This should be the end of the major refactors required for ExportArtifacts.

Had to refactor a few other files to get their usage of ExportArtifacts inline with the new interfaces.

### DIFF
--- a/sharktank/sharktank/utils/export_artifacts.py
+++ b/sharktank/sharktank/utils/export_artifacts.py
@@ -13,6 +13,9 @@ import time
 from pathlib import Path
 from datetime import timedelta
 from typing import Any, List, Optional, TYPE_CHECKING
+
+import numpy as np
+import torch
 from sharktank.utils.iree import get_iree_compiler_flags_from_object
 
 if TYPE_CHECKING:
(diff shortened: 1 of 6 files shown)


sharktank/tests/models/llama/benchmark_amdgpu_test.py
```diff
@@ -89,136 +91,73 @@ def save_benchmarks(
...
-                tensor_parallelism_size=self.tensor_parallelism_size,
+                input_path=self.artifact_dir / "prefill_args_bs4_128_stride_32_tp1",
+                tensor_parallelism_size=1,
             ),
             2048: self.save_benchmarks(
                 benchmark_fn="prefill_bs4",
-                input_path=artifact_dir / "prefill_args_bs4_2048_stride_32",
-                tensor_parallelism_size=self.tensor_parallelism_size,
+                input_path=self.artifact_dir / "prefill_args_bs4_2048_stride_32",
+                tensor_parallelism_size=1,
```
Alex-Vasile, line 125: Simplifying this further is possible, but will ideally involved renaming files so they have a consistent pattern that we can create using f-strings

archana-ramalingam (approved):

sharktank/sharktank/utils/export_artifacts.py
```diff
@@ -313,8 +408,8 @@ def compile_to_vmfb(
                 f"--iree-hal-dump-executable-files-to={hal_dump_path}/files"
             ]
         # Append optional arguments if provided
-        if args:
-            compile_args += args
+        if extra_args:
+            compile_args += extra_args
         else:
             compile_args += [
                 "--iree-opt-level=O3",
```
sogartar, line 415: This should not be called `extra_args` if it causes to override the default args.
Ideally, we should be able to toggle all options multiple times with the last flag superseding previous flags. I hope IREE does this right like other compilers.
I think we should leave it being `extra_args` and then not override.

sogartar: I will check if we can provide a flag multiple times and change this.

PR #1673, opened 2025-06-18
Add new model test CI script

AmosLewis:
- Add llama-8b-fp8 tp1 ci script
- Add llama-70b-fp16-tp8 ci script


AmosLewis:
Check if it is Monday(1), if so, run the steps related to the serving. We can change the number from 1-7 for Monday to Sunday. 
```
if [ "$(date -u +%u)" = "1" ]; then
  echo "IS_SERVING_DAY=true" >> $GITHUB_ENV
else
  echo "IS_SERVING_DAY=false" >> $GITHUB_ENV
fi
```

PR #1676, opened 2025-06-19
[sharktank] in perplexity allow for passing as CLI of extra compiler flags

sogartar:
Now the flags are really extra, meaning they do not override existing flags.

Also refactor to always skip file generation if they exist.

### DIFF
--- a/sharktank/sharktank/utils/cli.py
+++ b/sharktank/sharktank/utils/cli.py
@@ -240,6 +240,21 @@ def add_export_artifacts(parser: argparse.ArgumentParser):
         help="Output file path for compiled vmfb file",
         type=str,
     )
+    parser.add_argument(
+        "--extra-compile-arg",
+        help=(
+            "Additional flag(s) to provide to the IREE compiler. "
+            "E.g. `--extra-compile-arg=--compile-mode=vm --extra-compile-arg=--iree-vm-target-extension-f32`"
+        ),
+        action="append",
+        type=str,
+        default=[],
+    )
+    parser.add_argument(
+        "--skip-if-file-exists",
+        help="Skip generation of files if they exist. E.g. MLIR is already exported.",
+        action=argparse.BooleanOptionalAction,
+    )
 
 
 def add_save_tensor_options(parser: argparse.ArgumentParser):
--- a/sharktank/sharktank/utils/cli.py
+++ b/sharktank/sharktank/utils/cli.py
@@ -250,11 +250,6 @@ def add_export_artifacts(parser: argparse.ArgumentParser):
         type=str,
         default=[],
     )
-    parser.add_argument(
-        "--skip-if-file-exists",
-        help="Skip generation of files if they exist. E.g. MLIR is already exported.",
-        action=argparse.BooleanOptionalAction,
-    )
 
 
 def add_save_tensor_options(parser: argparse.ArgumentParser):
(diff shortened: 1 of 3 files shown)


sharktank/sharktank/utils/export_artifacts.py
```diff
@@ -529,4 +530,4 @@ def export_and_compile_llm(
         """
         self.export_llm_to_mlir(skip_decode=skip_decode)
         self.compile_to_vmfb(extra_args=extra_compile_args, hal_dump_path=hal_dump_path)
-        return str(self.output_vmfb.resolve())
+        return str(Path(self.output_vmfb).resolve())
```
Alex-Vasile, line 533: All of these paths should have been Paths rather than strings. I should have forced them to be paths in the initializer. Then this and the other defensive conversions to Path are not needed

sharktank/sharktank/utils/export_artifacts.py
```diff
@@ -407,16 +407,17 @@ def compile_to_vmfb(
             compile_args += [
                 f"--iree-hal-dump-executable-files-to={hal_dump_path}/files"
             ]
+
+        compile_args += [
```
Alex-Vasile, line 411: This will force these values of the flags. I don't think it will be possible to export **without**, e.g., `--iree-opt-level=O3`.

sharktank/sharktank/utils/cli.py
```diff
@@ -240,6 +240,21 @@ def add_export_artifacts(parser: argparse.ArgumentParser):
...
+        "--extra-compile-arg",
+        help=(
+            "Additional flag(s) to provide to the IREE compiler. "
+            "E.g. `--extra-compile-arg=--compile-mode=vm --extra-compile-arg=--iree-vm-target-extension-f32`"
+        ),
+        action="append",
+        type=str,
+        default=[],
+    )
+    parser.add_argument(
```
Alex-Vasile, line 253: Should have a default of False in order to keep current behaviour the same.

sogartar: It has `False` as default, through [argparse.BooleanOptionalAction](https://docs.python.org/3/library/argparse.html#argparse.BooleanOptionalAction).

sharktank/sharktank/utils/export_artifacts.py
sogartar: You can specify in your extra compile args
```
--iree-opt-level=O2
```

and it would take that as the optimization level.

Alex-Vasile: It will take the last value rather than error out about duplicate flags?

sharktank/sharktank/utils/cli.py
```diff
@@ -240,6 +240,21 @@ def add_export_artifacts(parser: argparse.ArgumentParser):
...
+        "--skip-if-file-exists",
```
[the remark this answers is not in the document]
Alex-Vasile: We only check (and use pre-existing results) if this `skip_if_file_exists=True` is passed to the ExportArtifacts initializer

https://github.com/nod-ai/shark-ai/blob/2fd48b19dfc3159de7f95109b15cea574b4c9a0e/sharktank/sharktank/utils/export_artifacts.py#L334-L341


https://github.com/nod-ai/shark-ai/blob/2fd48b19dfc3159de7f95109b15cea574b4c9a0e/sharktank/sharktank/utils/export_artifacts.py#L393-L395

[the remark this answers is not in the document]
archana-ramalingam: Right, 2 things about this flag:
- `skip_if_file_exists` is just an additional check and I see the paths are not None at this point, so not having this flag doesn't make a diff.
- User now has to pass mlir, config and `skip_if_file_exists` to let us know to use pre-exported paths. Just makes it more complicated than it has to be.
 
If we really want to keep `skip_if_file_exists`, it should be initialized based on the passed args not a burden to the user.

Alex-Vasile (approved):

PR #1678, opened 2025-06-19
Add CI tests against torch version 2.6.0 and remove for 2.4.1

PhaneeshB:
- bump torch to 2.6.0 and then to 2.7.1 in upcoming weeks
- needed for failing T5 encoder tests against Hugging Face due model loading vulnerability fix.

### DIFF
--- a/build_tools/torch_rocm_version_map.py
+++ b/build_tools/torch_rocm_version_map.py
@@ -1,0 +1,21 @@
+# Copyright 2025 Advanced Micro Devices, Inc.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import argparse
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Get the ROCm version corresponding to the torch version of PyTorch ROCm."
+    )
+    parser.add_argument("torch_version", type=str)
+    args = parser.parse_args()
+    map = {"2.5.1": "6.2", "2.6.0": "6.2.4"}
+    print(map[args.torch_version])
+
+
+if __name__ == "__main__":
+    main()
--- a/sharktank/build_tools/install_test_dependencies.sh
+++ b/sharktank/build_tools/install_test_dependencies.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+set -eu
+set -o pipefail
+
+SCRIPT_DIR=$(dirname $0)
+
+SRC_DIR="$SCRIPT_DIR/../.."
+PYTORCH_ROCM=0
+IREE_UNPINNED=0
+
+show_help() {
+    echo "Usage:"
+    echo "$(basename "$0") [-h] [--torch-version <value>] [--pytorch-rocm] [--iree-unpinned] [--help]"
+    echo "Args:"
+    echo "--torch-version: Version of PyTorch. If omitted will install the default from pytorch-*-requirements.txt."
+    echo "--pytorch-rocm: Install PyTorch for ROCm instead of for CPU."
+    echo "--iree-unpinned: Install IREE unpinned(latest)/pinned version."
+    exit 0
+}
+
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        -h|--help) show_help ;;
+        --torch-version) TORCH_VERSION="$2"; shift ;;
+        --pytorch-rocm) PYTORCH_ROCM=1 ;;
+        --iree-unpinned) IREE_UNPINNED=1 ;;
+        *) echo "Unknown parameter passed: $1"; exit 1 ;;
+    esac
+    shift
+done
+
+python -m pip install --no-compile --upgrade pip
+
+if [[ -v TORCH_VERSION ]]; then
+    if (($PYTORCH_ROCM)); then
+        ROCM_VERSION=$(python "$SRC_DIR/build_tools/torch_rocm_version_map.py" $TORCH_VERSION)
+        pip install --no-compile \
+            --index-url https://download.pytorch.org/whl/rocm$ROCM_VERSION \
+            torch==$TORCH_VERSION+rocm$ROCM_VERSION
+    else
+        pip install --no-compile \
+            --index-url https://download.pytorch.org/whl/cpu \
+            torch==$TORCH_VERSION+cpu
+    fi
+else
+    if (($PYTORCH_ROCM)); then
+        pip install --no-compile -r "$SRC_DIR/pytorch-rocm-requirements.txt"
+    else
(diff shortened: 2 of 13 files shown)


pytorch-rocm-requirements.txt
```diff
@@ -1,2 +1,2 @@
---index-url https://download.pytorch.org/whl/rocm6.2
-torch>=2.5.1
+--index-url https://download.pytorch.org/whl/rocm6.3
+torch>=2.6.0
```
sogartar, line 2: Did you check that romc6.3 has torch 2.6.0?

sogartar (commented):
Aside from the failing `T5EncoderIreeTest::testCompareToyIreeVsEager_1` looks good.
It seems that the absolute error is not much higher than the tolerance.
It may be that the tolerance can be relaxed. f32 has 23 fraction bits and bf16 has 7.
f32 has 6.5e4 more precision. Meaning that the fraction can represent that many more numbers. If 1e-5 is the atol for f32. It is not unreasonable that 1e-1 be the atol for bf16.

PhaneeshB:
> Aside from the failing `T5EncoderIreeTest::testCompareToyIreeVsEager_1` looks good. It seems that the absolute error is not much higher than the tolerance. It may be that the tolerance can be relaxed. f32 has 23 fraction bits and bf16 has 7. f32 has 6.5e4 more precision. Meaning that the fraction can represent that many more numbers. If 1e-5 is the atol for f32. It is not unreasonable that 1e-1 be the atol for bf16.

Agree that it makes sense to have lower tolerances for f32 vs bf16 comparison.
atol=1e-1 might be too lenient for encoding since small changes in encoding could potentially lead to a different output path in the sampler phase.
this would need to be tested out e2e to for better clarity, 
for this test we can perhaps start with `atol=1e-3` and `rtol=1e-2`

PhaneeshB:
Sharktank Nightly Run 
https://github.com/nod-ai/shark-ai/actions/runs/15827991778

sharktank/tests/models/t5/t5_test.py
```diff
@@ -480,19 +480,16 @@ def runTestV1_1CompareIreeAgainstTorchEager(
...
         [
             (torch.float32, torch.float64, 1e-5, 1e-5),
-            (torch.bfloat16, torch.float64, 2e-2, 1e-2),
+            (
+                torch.bfloat16,
+                torch.float64,
+                2e-2,
+                1e-2,
+                "ISSUE: https://github.com/nod-ai/shark-ai/issues/1689",
+            ),  # mark.xfail
```
sogartar, line 492: You should use this pattern for xfailing for a specific parameter set
https://stackoverflow.com/a/30575822/324204

sogartar: Also use our xfail with a regex match.
https://github.com/nod-ai/shark-ai/blob/26e83175729864d159cbfcd56f17e322ead28d51/sharktank/sharktank/utils/testing.py#L656

sharktank/tests/models/t5/t5_test.py
```diff
@@ -501,7 +498,11 @@ def testCompareToyIreeVsEager(
         reference_dtype: torch.dtype,
         atol: float,
         rtol: float,
+        xfail_reason: str | None = None,
     ):
+        if xfail_reason is not None:
+            pytest.xfail(reason=xfail_reason)
+
```
sogartar, line 505: No need for this if using the aforementioned pattern.
https://stackoverflow.com/a/30575822/324204

sharktank/tests/models/t5/t5_test.py
PhaneeshB: > missed that there is a param wrapper in parametrized lib too.
https://github.com/wolever/parameterized/blob/b9f6a640452bcfdea08efc4badfe5bfad043f099/parameterized/parameterized.py#L161

seems like this is also just a wrapper for args kwargs unlike pytest.param which can mark tests too.

sogartar: The match makes sure that the exception message does not change unexpectedly. If the test starts failing due to another reason. This is needed because some tests were failing during compilation, but then before they got fixed they started failing with another compiler bug.
Like that we even don't check if the exception type is as expected.

sogartar: I changed the atol as it is too stringent.

sogartar: Regarding the problems with xfailing a parametrized test here is an issue https://github.com/nod-ai/shark-ai/issues/1708.

sharktank/tests/models/llama4/llama4_test.py
```diff
@@ -1,4 +1,6 @@
+import re
 import transformers.models
+from sharktank.utils.export_artifacts import IreeCompileException
```
archana-ramalingam, line 3: Double import

archana-ramalingam:
Requested Rob to review to make sure this doesn't interfere with our rework project.

rsuderman (changes requested):

rsuderman:
This bump includes substantial hackery to xfail tests and change installs. This needs to be better cleaned up and simplified.

build_tools/torch_rocm_version_map.py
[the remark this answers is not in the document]
sogartar: The pip package version and index is dependent on the torch version. I wanted to have to say what only the torch version is.
```
pip install \
  --index-url https://download.pytorch.org/whl/rocm6.2 \
  torch==2.5.1+rocm6.2
```

sharktank/tests/models/vae/vae_test.py
```diff
@@ -308,40 +309,108 @@ def testCompareToyEagerVsHuggingFace(
...
-                atol=atol,
-                rtol=rtol,
+        self.runTestCompareIreeVsEager(
+            target_dtype=target_dtype,
+            reference_model=reference_model,
+            atol=atol,
+            rtol=rtol,
+        )
+
+    @pytest.mark.xfail(
```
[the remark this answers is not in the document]
sogartar: I reverted to non-strict and xfail without regex matching.

sharktank/build_tools/install_test_dependencies.sh
```diff
@@ -0,0 +1,61 @@
...
+fi
+
+if (($IREE_UNPINNED)); then
+    pip install --no-compile --upgrade -r "$SRC_DIR/requirements-iree-unpinned.txt"
+else
+    pip install --no-compile -r "$SRC_DIR/requirements-iree-pinned.txt"
+fi
+
+pip install --no-compile -r "$SRC_DIR/sharktank/requirements-tests.txt"
+pip install --no-compile -e "$SRC_DIR/sharktank"
```
archana-ramalingam, line 61: Looks good, in a later PR we want to merge redundant scripts like [setenv.sh](https://github.com/nod-ai/shark-ai/blob/main/scripts/setenv.sh) and ultimately move to share envs across pre-submits and nightly CIs, like shortfin.

PR #1683, opened 2025-06-20
Fix trie code base for multi beam

amd-vivekag:
Earlier, server code was crashing in case of multiple workers due to race condition. This PR fixes that.

### DIFF
--- a/shortfin/python/shortfin_apps/llm/components/kvcache/base_attention_cache.py
+++ b/shortfin/python/shortfin_apps/llm/components/kvcache/base_attention_cache.py
@@ -170,6 +170,8 @@ class BasePagedAttentionCache:
         return BasePagedAttentionCacheAllocation(pages, cache=self)
 
     def increment_pages(self, pages: List[PageInfo]):
+        if not self.use_ref_counts:
+            return
         with self._ref_count_lock:
             for page in pages:
                 self.ref_counts[page.index] += 1
@@ -177,6 +179,9 @@ class BasePagedAttentionCache:
     def decrement_pages(
         self, pages: List[PageInfo], return_empty_pages: bool = False
     ) -> None | List[PageInfo]:
+        if not self.use_ref_counts:
+            return [] if return_empty_pages else None
+
         with self._ref_count_lock:
             if return_empty_pages:
                 empty_pages = []
--- a/shortfin/python/shortfin_apps/llm/components/kvcache/trie_attention_cache.py
+++ b/shortfin/python/shortfin_apps/llm/components/kvcache/trie_attention_cache.py
@@ -9,26 +9,9 @@ from .base_attention_cache import (
     CacheAllocationFailure,
     PageAllocation,
 )
-
-
-@dataclass
-class RefCount:
-    """
-    A reference counter to replace simple int.
-    """
-
-    count: int = 0
-
-    def increment(self) -> int:
-        self.count += 1
-        return self.count
-
-    def decrement(self) -> int:
-        self.count -= 1
-        return self.count
-
-    def is_empty(self) -> bool:
-        return self.count <= 0
+from .kvcache_utils import RefCount
+import sys
+from threading import Lock
 
 
 @dataclass
@@ -54,6 +37,7 @@ class TrieNode:
     parent: Optional["TrieNode"] = None
     ref_count: RefCount = None
     access_time: float = 0.0
+    _lock: Lock = Lock()
 
     def __post_init__(self) -> None:
         """Initialize children dict and access time if not provided."""
@@ -72,15 +56,18 @@ class TrieNode:
         Returns:
             The newly created child node
         """
-        new_node = TrieNode(tokens=tokens, page=page, parent=self)
-        self.children[tokens] = new_node
-        return new_node
+        with self._lock:
+            new_node = TrieNode(tokens=tokens, page=page, parent=self)
+            self.children[tokens] = new_node
+            return new_node
 
     def unlink(self) -> None:
         """Remove this node from its parent's children."""
-        if self.parent is not None:
-            del self.parent.children[self.tokens]
-            self.parent = None
+        with self._lock:
+            if self.parent is not None:
+                if self.tokens in self.parent.children:
+                    del self.parent.children[self.tokens]
+                self.parent = None
 
     def __hash__(self) -> int:
         """Nodes are uniquely identified by their memory address."""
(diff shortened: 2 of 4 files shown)


shortfin/python/shortfin_apps/llm/components/kvcache/base_attention_cache.py
```diff
@@ -170,13 +170,18 @@ def acquire_pages_for_tokens(
         return BasePagedAttentionCacheAllocation(pages, cache=self)
 
     def increment_pages(self, pages: List[PageInfo]):
+        if not self.use_ref_counts:
```
stbaione, line 173: Are these needed? `TrieAttentionCache` calls `super().__init__(...)`, with constant `use_ref_counts=False` [here](https://github.com/nod-ai/shark-ai/blob/main/shortfin/python/shortfin_apps/llm/components/kvcache/trie_attention_cache.py#L294), because `Trie` increments at the `node` level.

I like this as a `runtime error` check though. If they call this function and `use_ref_counts==False`, you throw a RuntimeError that "BaseAttentionCache must have `use_ref_counts==True` to increment/decrement."

Little nervous this could lead us to a future bug, where we call ref count related functions with no issue, and don't realize that it's actually been off the whole time. Or visa/versa, it's on when it shouldn't be. Seems clearer to explicitly disallow that functionality if it isn't enabled.

I could be missing something though

shortfin/python/shortfin_apps/llm/components/kvcache/trie_attention_cache.py
```diff
@@ -9,26 +9,9 @@
...
-        return self.count
-
-    def decrement(self) -> int:
-        self.count -= 1
-        return self.count
-
-    def is_empty(self) -> bool:
-        return self.count <= 0
+from .kvcache_utils import RefCount
+import sys
```
stbaione, line 13: nit: unused import

shortfin/python/shortfin_apps/llm/components/kvcache/trie_attention_cache.py
```diff
@@ -9,26 +9,9 @@
...
+from threading import Lock
```
stbaione, line 14: nit: move this up to the other `from [standard_package] import ...` imports (line 3)

shortfin/python/shortfin_apps/llm/components/kvcache/trie_attention_cache.py
```diff
@@ -79,7 +62,8 @@ def create_child(self, tokens: Tuple[int, ...], page: PageInfo) -> "TrieNode":
     def unlink(self) -> None:
         """Remove this node from its parent's children."""
         if self.parent is not None:
-            del self.parent.children[self.tokens]
+            if self.tokens in self.parent.children:
```
stbaione, line 65: Was this a patch you had to add from a case you came across or a theoretical check in case key doesn't exist, but not one you came across while developing? i.e. does it work the same if this line is removed? 

Reason being:

We're calling `unlink`, so the assumption is that this node is currently `linked`. I would expect that a `linked` node would HAVE to have it's tokens stored as a child of its own parent, since we're unlinking this node to no longer be a child of that parent.

Feeling some caution that this could actually be a kind of "dangerous" catch where it masks a bug elsewhere that caused the child to be improperly linked to its parent, now or in the future.

Was this maybe a concurrency fix, where unlink was getting called multiple times for the same node?

shortfin/python/shortfin_apps/llm/components/kvcache/trie_attention_cache.py
```diff
@@ -141,61 +125,69 @@ def publish_pages_for_tokens(
...
-
-        def has_common_prefix(tokens1, tokens2):
-            for t1, t2 in zip(tokens1, tokens2):
-                if t1 != t2:
-                    return False
-            return True
-
-        if not has_common_prefix(self.tokens, tokens):
+        min_len = min(len(self.tokens), len(tokens))
+        if list(self.tokens)[:min_len] != list(tokens)[:min_len]:
```
stbaione, line 129: nit: Aren't `self.tokens` and `tokens` already lists?

shortfin/python/shortfin_apps/llm/components/kvcache/trie_attention_cache.py
```diff
@@ -141,61 +125,69 @@ def publish_pages_for_tokens(
...
+            ]
+
+            # Add unpublished pages to trie
+            if publish_incomplete_page:
+                raise NotImplementedError(
+                    "Additional work needed here to support publishing incomplete pages to ensure that we finish up a page before attaching child nodes to it."
+                )
 
-        tokens_per_page = self.cache.tokens_per_page
+            # cur_node = self.last_cached_node
```
stbaione, line 167: nit: remove comment

shortfin/python/shortfin_apps/llm/components/kvcache/trie_attention_cache.py
```diff
@@ -223,9 +215,8 @@ def extend_allocation(self, tokens: List[int], *, extra_token_slots=0) -> None:
             raise ValueError("New tokens must be longer than current tokens")
 
         # Check that current tokens are a prefix of new tokens
-        for old_token, new_token in zip(self.tokens, tokens):
-            if old_token != new_token:
-                raise ValueError("New tokens must extend current token sequence")
+        if tokens[: len(self.tokens)] != list(self.tokens):
```
stbaione, line 218: Same about `self.tokens` already being a list

stbaione (changes requested):
Mostly looks good. A few nits, and a couple spots that may be problematic

shortfin/python/shortfin_apps/llm/components/kvcache/trie_attention_cache.py
amd-vivekag: done

shortfin/python/shortfin_apps/llm/components/kvcache/base_attention_cache.py
```diff
@@ -170,13 +170,23 @@ def acquire_pages_for_tokens(
...
+            raise RuntimeError(
+                "BaseAttentionCache should not be initialized with use_ref_counts as False"
```
stbaione, line 175: This phrasing might mislead a developer a little bit, since `BasePagedAttentionCache` does function with `use_ref_counts == False` and `TriePagedAttentionCache` explicity sets `use_ref_counts == False`.

Could say something like:

"BaseAttentionCache must have use_ref_counts enabled to increment/decrement"

shortfin/python/shortfin_apps/llm/components/kvcache/base_attention_cache.py
```diff
@@ -170,13 +170,23 @@ def acquire_pages_for_tokens(
...
         with self._ref_count_lock:
             for page in pages:
                 self.ref_counts[page.index] += 1
 
     def decrement_pages(
         self, pages: List[PageInfo], return_empty_pages: bool = False
     ) -> None | List[PageInfo]:
+        if not self.use_ref_counts:
+            raise RuntimeError(
+                "BaseAttentionCache should not be initialized with use_ref_counts as False"
```
stbaione, line 187: same here

shortfin/python/shortfin_apps/llm/components/kvcache/trie_attention_cache.py
stbaione: Gotcha. Yeah, all of the annotations point to `List[int]`, but I see the conversions to `Tuple` in the `TrigePagedAttentionCache` class. That seems like a type bug to me. We should probably make sure that at the least `self.tokens` is a consistent type,

stbaione (approved):
LGTM

PR #1684, opened 2025-06-20
[sharktank] move functions out of utils.testing to avoid unwanted dependency on pytest

sogartar:
The recent fix https://github.com/nod-ai/shark-ai/pull/1662 was not enough to mitigate the problem.
This change removes all imports of `sharktank.utils.testing` form non-testing code by moving the functions `make_rand_torch` and `make_random_mask` to another location.


stbaione (approved):
Seems to do the trick. Double-verified locally by uninstalling `pytest` and running through the `llama_serving` flow

PR #1694, opened 2025-06-23
Create RequestQueueManager

dezhiAmd:
Why:
Prepare for rate limiting

How:
Refactor function add_to_queue and remove_from_queue into a separate class

### DIFF
--- a/shortfin/tests/apps/llm/components/request_queue_manager_test.py
+++ b/shortfin/tests/apps/llm/components/request_queue_manager_test.py
@@ -1,0 +1,45 @@
+# Copyright 2025 Advanced Micro Devices, Inc.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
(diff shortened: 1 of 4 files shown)


shortfin/tests/apps/llm/components/request_queue_manager_test.py
```diff
@@ -0,0 +1,29 @@
...
+def test_request_queue_manager():
+    queue_manager = RequestQueueManager(6)
+
+    assert queue_manager.current_queue_size == 0
+
+    # Add to queue
+    assert queue_manager.add_to_queue(4) == True
+    assert queue_manager.current_queue_size == 4
+
+    # Add more to queue
```
stbaione, line 19: Let's add one more case above this one for the bug that you just fixed, and PR is good-to-go!

Assuming `current_queue_size == 4` from case above:

```python
# Try to add beyond max, when `current_queue_size` < `max_queue_size`
assert not queue_manager.add_to_queue(3)
assert queue_manager.current_queue_size == 4
```

Unit tests look like they'll fully exhaustive after that addition

dezhiAmd: Done

stbaione (approved):

PR #1711, opened 2025-06-25
Fix Issue 1700

dezhiAmd:
Why:
The finally block is always executed if the try block is entered. As a result, even the **add_to_queue** failes, **remove_from_queue** is called.

How:
1) Move the **add_to_queue** out of try code block
2) Only call **remove_from_queue** if **add_to_queue** succeeded

### DIFF
--- a/shortfin/tests/apps/llm/components/request_queue_manager_test.py
+++ b/shortfin/tests/apps/llm/components/request_queue_manager_test.py
@@ -4,26 +4,7 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-import pytest
 from shortfin_apps.llm.components.request_queue_manager import RequestQueueManager
-
-
-def test_remove_from_queue_raises_exception():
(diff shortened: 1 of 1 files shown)


stbaione:
Code looks good, but suddenly seeing an ASAN leak. This seems like a flake. I'm not sure how this code could have caused a leak. Rerun should fix it. If not, we need to figure out what's going on

stbaione (approved):

marbre:
@dezhiAmd the CI at HEAD is failing for this commit. In particular [`Unit tests :: Ubuntu (Clang)(full) :: 3.12`](https://github.com/nod-ai/shark-ai/actions/runs/16003219788/job/45143351688#logs) is failing:

```
FAILED tests/apps/llm/components/kvcache/base_attention_cache_test.py::test_fork_pages[tokens5-3-three_exact_pages] - AssertionError: Fork Error in three_exact_pages: New last page should be filled with 1s.
assert False
 +  where False = all(<generator object test_fork_pages.<locals>._test_fork_pages.<locals>.<genexpr> at 0x7f3f37212400>)
 ```

PR #1715, opened 2025-06-25
Separated rotary sharding behavior from rotary layer

rsuderman:
Rotary layer should have sharding behavior split away from the main implementation. This avoids handling the desharding - resharding of the rotary embedding table.

### DIFF
--- a/sharktank/sharktank/layers/rotary_embedding.py
+++ b/sharktank/sharktank/layers/rotary_embedding.py
@@ -18,6 +18,112 @@ from sharktank.types import (
 )
 
 
+def build_rotary_layer(
+    tensor_parallelism_size: int = 1,
+    pipeline_parallelism: bool = False,
+    devices=None,
+    **kwargs,
+):
+    rotary_layer = RotaryEmbeddingLayer(**kwargs)
+    return ShardedRotaryLayer(
+        tensor_parallelism_size=tensor_parallelism_size,
+        pipeline_parallelism=pipeline_parallelism,
+        rotary_layer=rotary_layer,
+        devices=devices,
+    )
+
+
+# We wrap a shardless rotary layer so the sharding behavior can be handled independently of the numerics.
+class ShardedRotaryLayer(BaseLayer):
+    def __init__(
+        self,
+        *,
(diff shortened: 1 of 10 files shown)


sharktank/sharktank/layers/rotary_embedding.py
```diff
@@ -18,133 +18,137 @@
...
 
-            xt_shards = [
-                self.forward_unsharded(
-                    xt=unbox_tensor(xt_shard),
-                    start_index=start_index,
-                    rotary_embed_table=rotary_shard,
+    def compute_batch_mask(
+        self, start_positions: Union[torch.Tensor, ReplicatedTensor], batch_seq_len: int
+    ) -> torch.Tensor:
+        if isinstance(start_positions, ShardedTensor):
```
stbaione, line 86: nit: Include `ShardedTensor` in arg type annotation

sharktank/sharktank/layers/rotary_embedding.py
```diff
@@ -18,133 +18,137 @@
...
+            shards = []
+            table = self.rotary_embed_table()
+            for s, ts in zip(start_positions.shards, table.shards):
```
stbaione, line 89: Not needed for this PR, but wanted to note something. 

I see this `zip -> callback -> append to list -> clone` iteration pretty much everywhere when it comes to sharded. Simple utility function could be added that accepts a callback and reduces duplicating this loop

stbaione (approved):
Looks good, just one nit for type annotation. 

Existing tests validate functionality of new wrapping class. 

Something to note is this still has the `seq_len > 1` bug for `compute_batch_mask`. But, makes sense to keep existing behavior here for the refactor

sharktank/sharktank/layers/rotary_embedding.py
rsuderman: Seems reasonable for a follow up.

sharktank/sharktank/layers/rotary_embedding.py
rsuderman: Fixed.

PR #1725, opened 2025-06-26
Lisal.mooncake update write back

lisaliu1:


shortfin/tests/apps/llm/components/kvcache/mooncake_cache_test.py
```diff
@@ -0,0 +1,144 @@
...
+        allocation.release_pages()
+
+    lsys.run(_test_write_back_pages())
+
+
+# fmt: off
+@pytest.mark.parametrize(
+   "tokens,expected_keys,case_name", test_data
+)
+def test_update_pages(
```
stbaione, line 131: This test doesn't currently execute on test runs `(lsys.run(...))`

stbaione (approved):
LGTM

Documenting the decision to ignore the ASAN issue for gardener:

The ASAN issue is unrelated to this PR. From syncs, our suspicion is that it's a leak from a dependency (likely IREE), with the expectation that our next bump will resolve it.

We've seen this exact leak occur across multiple PRs, including #1768, which triggered the exact same leak by just adding a simple pytest func that adds two integers together. 

From syncs yesterday, we discussed that we will temporarily let this fail, and revisit if not cleared up by bumping IREE.

PR #1730, opened 2025-06-26
[LLM Server] Fix issue with batch `input_id` requests

stbaione:
## Description

We were hitting an issue when attempting to send batch requests through the `mlperf harness`, where we specify `input_ids` instead of text.

Simple fix here, we weren't checking if the request was pre-tokenized or not, so we were attempting to index into a `None` value, which causes the error.

Seeing some gaps in the CI regarding requests that supply `input_token_ids` instead of `text`. Will follow up with coverage PR + any fixes that unearths.


jinchen62 (approved):
LGTM, thanks!

PR #1733, opened 2025-06-27
[tuner] add parser for attention op

bangtianliu:
After https://github.com/iree-org/iree/pull/21170 and https://github.com/iree-org/iree/pull/21216 exposing python bindings from iree side, This PR adds parser for attention op.

### DIFF
--- a/sharktuner/sharktuner/dispatch_parser.py
+++ b/sharktuner/sharktuner/dispatch_parser.py
@@ -11,6 +11,7 @@ from abc import ABCMeta, abstractmethod
 
 from iree.compiler import ir  # type: ignore
 from iree.compiler.dialects import linalg, func  # type: ignore
+from iree.compiler.dialects import iree_codegen  # type: ignore
 
 from . import common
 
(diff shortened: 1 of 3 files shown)


sharktuner/sharktuner/dispatch_parser.py
```diff
@@ -81,3 +81,13 @@ def has_valid_root_op(self) -> bool:
...
         return True
+
+
+class AttentionOpInterfaceParser(DispatchParser):
+    def __init__(self, root_op: ir.Operation):
+        super().__init__(root_op)
+
+    def has_valid_root_op(self) -> bool:
+        root_op = self.get_root_op()
+        # TODO(Bangtian): will see whether to use `iree_codegen.isa_attention_op`
```
Max191, line 92: nit: I think we do want to use `iree_codegen.isa_attention_op`, so maybe update the TODO to say that we should use it once it is available? (is it already available?)

Max191 (approved):
LGTM

sharktuner/sharktuner/dispatch_parser.py
bangtianliu: yeah, I will update this when https://github.com/iree-org/iree/pull/21216 is landed.

kuhar, line 92:
```suggestion
        # TODO(Bangtian): Switch to `iree_codegen.isa_attention_op` once available
```

kuhar (approved):

PR #1734, opened 2025-06-27
[Shortfin] Correct KVCache size calculation for pipeline parallelism

Alex-Vasile:
Current implementation assumes a uniform split of the kvcache size between all devices. This change has `export_paged_llm_v1` specify the size of the kvcache size needed for each device index allowing for unsharded, tensor parallelism, and pipeline parallelism to be handled seamlessly. It also moves the responsibility of this calculation to sharktank rather than having both sharktank and shortfin needing to do the same calculation.

### DIFF
--- a/shortfin/python/shortfin_apps/llm/components/service.py
+++ b/shortfin/python/shortfin_apps/llm/components/service.py
@@ -47,14 +47,12 @@ class LlmGenerateService(GenerateService):
         model_params: ModelParams,
         server_params: "ServerParams",
         program_isolation: str = "per_call",
-        max_queue_size: int = 3,  # Maximum number of requests in queue
     ):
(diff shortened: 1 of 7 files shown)


shortfin/python/shortfin_apps/llm/components/service.py
```diff
@@ -70,6 +68,7 @@ def _initialize_max_queue_size(self):
         """Initialize request and response queues"""
         if self.model_params.decode_batch_sizes:
             self.max_queue_size = max(self.model_params.decode_batch_sizes)
+            # self.max_queue_size *= self.model_params.pipeline_parallelism_size
```
rsuderman, line 71: We should delete this.

rsuderman (approved):

PR #1735, opened 2025-06-27
Switch to the new rotary embedding layer

rsuderman:
This performs the appropriate test updates and replacements for the rotary embedding layer. This also refactors some test code to avoid cross implementation dependencies.


sharktank/sharktank/layers/latent_attention_block.py
```diff
@@ -55,7 +54,7 @@ def forward(
         self,
         h: torch.Tensor | ShardedTensor,
         start_index: int,
-        embedding: RotaryEmbeddingLayer,
+        embedding,
```
Groverkss, line 57: please don't remove type hints

sharktank/sharktank/layers/paged_llama_attention_block.py
```diff
@@ -209,7 +208,7 @@ def forward(
         self,
         h: torch.Tensor | ShardedTensor,
         *,
-        embedding: RotaryEmbeddingLayer,
+        embedding,
```
Groverkss, line 211: type hinting needed

sharktank/sharktank/layers/rotary_embedding.py
```diff
@@ -9,273 +9,119 @@
...
-        self.dtype = dtype
-        self.rope_freq_base = rope_freq_base if rope_freq_base is not None else 10000.0
+        self.head_dim = head_dim
+        self.rope_theta = rope_theta
+        self.interleaved = interleaved
+
         self.yarn_beta_slow = yarn_beta_slow
         self.yarn_beta_fast = yarn_beta_fast
         self.yarn_factor = yarn_factor
         self.yarn_original_context_len = yarn_original_context_len
```
Groverkss, line 110: i dont know why these parameters are here when they are not used.

sharktank/sharktank/layers/rotary_embedding_sharded.py
```diff
@@ -0,0 +1,156 @@
...
+                sincos_cache=(
+                    unbox_tensor(mask_sin_shard),
+                    unbox_tensor(mask_cos_shard),
+                ),
+            )
+            for xt_shard, mask_sin_shard, mask_cos_shard in zip(
+                xt.shards, mask[0].shards, mask[1].shards
+            )
+        ]
+        return xt.clone(ts=xt_shards)
```
Groverkss, line 156: Let's not change the interface to rotary please. This interface is why i rewrote rotary. It's really bad. compute_batch_mask/apply_batch_mask/forward is bad. The only thing that needs to be there is compute_cache, and apply_rotary. nothing else please.

Groverkss (changes requested):

sharktank/sharktank/layers/rotary_embedding.py
```diff
@@ -314,29 +160,125 @@ def _apply_yarn(self, freqs):
             freqs = torch.where(is_medium_freq, smoothed_inv_freq, inv_freq_llama)
```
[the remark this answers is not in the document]
rsuderman: Your rope implementation triggered numerical failures. We can land the rope improvements separately. We need to avoid monolithic patches that potentially break numeriocs.

archana-ramalingam (approved):

Alex-Vasile:
The torch perplexity for the toy deepseek model changed:
```FAILED sharktank/tests/evaluate/perplexity_torch_test.py::PerplexityTest::test_deepseek_v3 - AssertionError: np.float64(2634.61493) != np.float64(2636.684089) within 0.5 delta (np.float64(2.0691589999996722) difference) : Current perplexity deviates baseline by 2.0691589999996722```

The toy deepseek is failing torch perplexity, but it looks like the real llama 8b tests pass. These numbers may just need to be updated.

PR #1736, opened 2025-06-29
Rate Limiting based on available page

dezhiAmd:
Why:
When new request coming in, the number of pages needed to be checked to ensure there is enough number of pages.

How:
Use three parameters to predict number of pages needed

1. Length of input tokens
2. Block sequence stride
3. Number of beams

### DIFF
--- a/shortfin/python/shortfin_apps/llm/components/generate.py
+++ b/shortfin/python/shortfin_apps/llm/components/generate.py
@@ -9,6 +9,7 @@ import dataclasses
 import io
 import json
 import logging
+import math
 
 from copy import deepcopy
 from typing import List, Tuple
@@ -283,9 +284,9 @@ class ClientGenerateBatchProcess(sf.Process):
                 )
 
                 # return error reponse if no pages available
-                needed_pages_num = math.ceil(len(request.input_token_ids) / self.service.model_params.paged_kv_cache.block_seq_stride) + total_requested_beams - 1
+                needed_pages_num = math.ceil(len(input_token_ids) / self.service.model_params.paged_kv_cache.block_seq_stride) + total_requested_beams - 1
                 available_pages_num = len(self.service.page_pool.available_pages)
-                if needed_pages > available_pages_num:
+                if needed_pages_num > available_pages_num:
                     self._return_error_response(
                         status.HTTP_400_BAD_REQUEST,
                         error_message="Not enough pages for the new request",
@@ -305,7 +306,7 @@ class ClientGenerateBatchProcess(sf.Process):
                     self.gen_req,
                     index,
                     input_text=input_text,
-                    input_token_ids,
+                    input_token_ids=input_token_ids,
                     eos_token_id=self.tokenizer.eos_token_id,
                     decode_config=decode_config,
                     status_tracker=self.responder.get_status_tracker(),
--- a/shortfin/python/shortfin_apps/llm/components/generate.py
+++ b/shortfin/python/shortfin_apps/llm/components/generate.py
@@ -236,8 +236,6 @@ class ClientGenerateBatchProcess(sf.Process):
             if streaming:
                 self.responder.stream_start()
 
-            # Launch all individual generate processes and wait for them to finish.
-            gen_processes = []
             input_ids = self.gen_req.input_ids
             is_pretokenized = input_ids is not None
             # TODO: We should send this to an executor and await the results.
@@ -246,6 +244,8 @@ class ClientGenerateBatchProcess(sf.Process):
             else:
                 input_batch = self.tokenize()
 
+            # Launch all individual generate processes and wait for them to finish.
+            gen_processes = []
             for index, input_tokens in enumerate(input_batch):
                 decode_config = decode_configs[index]
 
@@ -270,23 +270,42 @@ class ClientGenerateBatchProcess(sf.Process):
                     )
                     return
 
-                idx, fiber = await self.service.main_fiber_pool.get()
-                indices.append(idx)
-
                 input_text = (
                     self.gen_req.text[index]
                     if not is_pretokenized and not self.gen_req.is_single
                     else self.gen_req.text
                 )
 
+                input_token_ids= (
+                    input_tokens
+                    if is_pretokenized
+                    else input_tokens.ids
+                )
+
+                # return error reponse if no pages available
+                needed_pages_num = math.ceil(len(request.input_token_ids) / self.service.model_params.paged_kv_cache.block_seq_stride) + total_requested_beams - 1
+                available_pages_num = len(self.service.page_pool.available_pages)
+                if needed_pages > available_pages_num:
+                    self._return_error_response(
+                        status.HTTP_400_BAD_REQUEST,
+                        error_message="Not enough pages for the new request",
+                        code=ResponseErrorCodes.PAGE_FULL,
+                        extra_fields={
+                            "available_page": available_pages_num,
+                            "requested_page": needed_pages_num,
+                        },
+                    )
+                    return
+
+                idx, fiber = await self.service.main_fiber_pool.get()
+                indices.append(idx)
+
                 gen_process = GenerateItemProcess(
                     self,
                     self.gen_req,
                     index,
                     input_text=input_text,
-                    input_token_ids=input_tokens
-                    if is_pretokenized
-                    else input_tokens.ids,
+                    input_token_ids,
                     eos_token_id=self.tokenizer.eos_token_id,
                     decode_config=decode_config,
                     status_tracker=self.responder.get_status_tracker(),
--- a/shortfin/python/shortfin_apps/llm/components/generate.py
+++ b/shortfin/python/shortfin_apps/llm/components/generate.py
@@ -277,14 +277,17 @@ class ClientGenerateBatchProcess(sf.Process):
                     else self.gen_req.text
                 )
 
-                input_token_ids= (
-                    input_tokens
-                    if is_pretokenized
-                    else input_tokens.ids
-                )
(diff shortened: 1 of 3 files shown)


shortfin/python/shortfin_apps/llm/components/generate.py
```diff
@@ -270,23 +271,45 @@ async def run(self):
...
                     self.gen_req.text[index]
                     if not is_pretokenized and not self.gen_req.is_single
                     else self.gen_req.text
                 )
 
+                input_token_ids = input_tokens if is_pretokenized else input_tokens.ids
+
+                # return error reponse if no pages available
+                needed_pages_num = (
+                    math.ceil(
```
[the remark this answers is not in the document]
dezhiAmd: I can create a separate class called RateLimtier and include tests.
But it is not clear to me the error in this equation. Here is what I see:
1) Frist part is based on input tokens and block sequence stride, refer [here](https://github.com/nod-ai/shark-ai/blob/a505b24906bb5bde39d41e8bf29db75c9784a388/shortfin/python/shortfin_apps/llm/components/batcher.py#L195)

2) Second part is based on number of beams, the last page is copied (num_beams -1 ), refer [here](https://github.com/nod-ai/shark-ai/blob/a505b24906bb5bde39d41e8bf29db75c9784a388/shortfin/python/shortfin_apps/llm/components/token_selection_strategy/beam_group.py#L445) which eventually calls **n** times acquire_free_pages(1) where **n**=num_beams -1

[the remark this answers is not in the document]
dezhiAmd: When you say "the remaining number of tokens left over from the input tokens", are referring to in partial prefill, only part of the input is new, while the rest has already been processed?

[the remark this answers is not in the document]
dezhiAmd: Added output_pages needed. Thanks

shortfin/python/shortfin_apps/llm/components/request_queue_manager.py
```diff
@@ -6,13 +6,50 @@
...
+import math
+from .config_struct import ModelParams
 from typing import Optional
 
 logger = logging.getLogger(__name__)
 
 from .token_selection_strategy.config import DecodeConfig
 
 
+class RateLimiter:
```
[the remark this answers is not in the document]
dezhiAmd: Fixed

shortfin/python/shortfin_apps/llm/components/request_queue_manager.py
```diff
@@ -6,45 +6,58 @@
...
-            True if the request was added successfully, False if the queue is full.
+            ID if the request was added successfully, None if the queue is full.
         """
         request_size = sum(config.num_beams for config in decode_configs)
 
         with self._lock:
             if self._current_queue_size + request_size > self._max_queue_size:
                 logger.debug(
-                    f"Add failed: queue size {self._current_queue_size}, request size {request_size}"
+                    f"Request rejected: {self._current_queue_size} (current) + {request_size} (new) > {self._max_queue_size} (max)."
```
dezhiAmd, line 60: Add max queue size to clarify why request is rejected

shortfin/python/shortfin_apps/llm/components/generate.py
```diff
@@ -203,7 +204,7 @@ async def run(self):
                 error_message="Server queue is full. Please try again later.",
                 code=ResponderErrorCodes.QUEUE_FULL,
                 extra_fields={
-                    "current_size": self.service.queue_manager.current_queue_size,
+                    "current_size": self.service.queue_manager._current_queue_size,
```
rsuderman, line 207: We generally don't want to access privately owned member variables (e..g. prefixed with a `_`).

shortfin/python/shortfin_apps/llm/components/request_queue_manager.py
```diff
@@ -80,3 +91,37 @@ def remove_from_queue(self, id: Optional[int]) -> None:
             logger.debug(
                 f"Removed from queue: new queue size {self._current_queue_size}"
             )
+
+    def check_memory_availability(
```
rsuderman, line 95: We should be adding this behavior to `add_to_queue`. Any of the rate limiter work should check all resources rather than individually checking each. The idea is we can ask the manager whether we have all resources available to run. Having separate checks and returns just clutters up the python code.

So you will need to augment `add_to_queue` to take any other available work.

rsuderman (changes requested):

shortfin/python/shortfin/interop/fastapi/__init__.py
```diff
@@ -29,6 +29,7 @@
 _fastapi_response_map = {
     ResponderErrorCodes.INVALID_REQUEST_ARGS: status.HTTP_400_BAD_REQUEST,
     ResponderErrorCodes.QUEUE_FULL: status.HTTP_503_SERVICE_UNAVAILABLE,
```
stbaione, line 31: You're replacing `QUEUE_FULL` with `PAGE_FULL` here right? It so, you can remove `QUEUE_FULL`.

Also, something like `KVCACHE_FULL`, `KVCACHE_PAGES_FULL` or `PAGE_POOL_FULL` might be a little more expressive of what the error code means, and easier for a new developer to understand. Singular `PAGE_FULL` is a little confusing

shortfin/python/shortfin_apps/llm/components/request_queue_manager.py
```diff
@@ -6,65 +6,194 @@
...
+                )
+                return None
 
-        Returns:
-            True if the request was added successfully, False if the queue is full.
-        """
-        request_size = sum(config.num_beams for config in decode_configs)
+        # Step 2: Pre-calculate total needed pages
+        stride = self.model_params.paged_kv_cache.block_seq_stride
+        available_pages = self.page_pool.get_available_pages_num()
```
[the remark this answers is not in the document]
dezhiAmd: In Python, everything is passed by object reference. in `__init__` method:
self.page_pool = page_pool
It does not create a copy of `page_pool`. It stores a reference to the same object that was passed in. This means:
If another thread or component modifies `page_pool`, those changes will be visible to `RequestQueueManager`.
And available_pages = self.page_pool.get_available_pages_num()is thread-safe
all other accesses to `page_pool` are also synchronized using the same lock

**With that being said, I am not sure how to add an internally tracked value which can make a difference. Please provide more guidance here.**

shortfin/python/shortfin_apps/llm/components/request_queue_manager.py
```diff
@@ -6,65 +6,188 @@
...
+            else max_queue_size
+        )
         self._lock = threading.Lock()
         self._current_queue_size = 0
         self._current_id = 0
         self._current_tasks = {}
+        self._request_pages = {}
+
+        self.model_params = model_params
+        self.available_page_count = self.model_params.paged_kv_cache.device_block_count
```
stbaione, line 46: I'm not sure if we can rely on this as a stateful variable. I believe that it would be better to pass the `PagePool` to the `RequestQueueManager`, and accessing `page_pool.available_page_count`. You could make a property on `RequestQueueManager` called `available_page_count` that calls that function to save some keystrokes, and keep the code that you have `mostly` the same. That ensures that we're in sync with the actual number of available pages, and keeps us from having to track it twice. There's also a functional reason for this:

The reason being:

Our `total static page count` is determined correctly by `device_block_count`.

However, the pages that a given request might use, per decode-step, when `multi-hypothesis` is involved isn't so straightforward.

We can make a good estimate, like the one you have in `_calculate_needed_pages`, as to the max kvcache pages a request may use. But, non-obvious stuff can happen under the hood that causes our estimate to be off.

For example, let's say we have a `beam_search` request, that looks like this after our a few decode steps. Let's say that each token takes up a page (simplifying for explanation):

```text
beam_1: The quick brown fox jumped over the bush
beam_2: The quick brown dog dug under the fence
```

Then, we have the following scenario:

```text
shared_pages = 3 - The quick brown
beam_1_pages = 5 - fox jumped over the bush
beam_2_pages = 5 - dog dug under the fence
total_pages = 13
```

However, let's say we do our next decode step, and the next two tokens are both candidates of `beam_2`, making our new beams (for example):

```text
beam_1: The quick brown dog dug under the fence quickly
beam_2: The quick brown dog dug under the fence eagerly
```

That changes how many pages we are using for this request:

```text
shared_pages = 8
beam_1_pages = 1 - quickly
beam_2_pages = 1 - eagerly
total_pages = 10
```

In this (simplified) case, you can see that while we can calculate a maximum number of pages for a request, the number of pages actually in use will change and depend on the decode strategy.

shortfin/tests/apps/llm/components/request_queue_manager_test.py
```diff
@@ -4,36 +4,144 @@
 # See https://llvm.org/LICENSE.txt for license information.
```
stbaione, line 4: See message in slack on adding `batch` tests

shortfin/python/shortfin_apps/llm/components/request_queue_manager.py
dezhiAmd: Refer to Rob's comment [here](https://github.com/nod-ai/shark-ai/pull/1736#discussion_r2207913154)

shortfin/tests/apps/llm/components/request_queue_manager_test.py
dezhiAmd: [Parametrizing Tests](https://docs.pytest.org/en/stable/example/parametrize.html) added

PR #1737, opened 2025-06-30
[sharktank] Fix and add tests for running models in eager with  paged_llm_v1

sogartar:
We don't have tests for the script.

Expand functionality with
1. Use LlamaModelConfig.from_properties to parse config instead of just the GGUF properties.
2. Actually generate a random prompt with --prompt-seq-len.
3. Add max decode steps option.

### DIFF
--- a/sharktank/sharktank/utils/load_llm.py
+++ b/sharktank/sharktank/utils/load_llm.py
@@ -31,10 +31,15 @@ class TorchGenerator:
         tokenizer: Optional[InferenceTokenizer] = None,
         # Need to look at the model more for this.
         end_token: int = 2,
+        max_decode_steps: int | None = None,
     ):
+        """
+        max_decode_steps: maximum number of decode steps to perform.
+        """
         self.model = model
         self.tokenizer = tokenizer
         self.end_token = end_token
+        self.max_decode_steps = max_decode_steps
 
     @property
     def block_seq_stride(self) -> int:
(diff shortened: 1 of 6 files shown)


sogartar:
Yes, I accidently include the changes from this PR into the other.

sharktank/sharktank/utils/load_llm.py
```diff
@@ -345,7 +379,7 @@ def decode(self, token_batch=None):
                 _decode_attention_mask,
             )
 
-        if self.dump_path is not None:
+        if self.dump_path is not None and self.dump_decode_steps < self.decode_step:
```
archana-ramalingam, line 382: Is `self.dump_decode_steps` supposed to be `max_decode_steps`? What's the diff if retaining both?

sogartar: dump_decode_steps is the number of steps to dump args for. `max_decode_steps` is how many decode steps to run.

sharktank/tests/examples/paged_llm_v1_test.py
```diff
@@ -0,0 +1,42 @@
...
+    dataset_path = tmp_path / "model.irpa"
+    dataset.save(dataset_path)
+    paged_llm_v1.main(
+        [
+            f"--irpa-file={dataset_path}",
+            "--tokenizer-type=fake",
+            "--bs=4",
+            "--prompt-seq-len=2",
+            "--max-decode-steps=1",
+            f"--tensor-parallelism-size={tensor_parallelism_size}",
```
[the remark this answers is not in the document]
sogartar: I made it run only when `torch.cuda.is_available()`.

PR #1740, opened 2025-06-30
Add a flag for converting quark's non-matmul parameters to a provided flag

KyleHerndon:
This changes:
- Output scales (output_scale)
- KV cache scales (kv_cache_scale)
- Attention/probability output scales (prob_output_scale/attn_scale)
- All norm layer weights
- Embedding and output weight

### DIFF
--- a/sharktank/sharktank/models/llama/tools/import_quark_dataset.py
+++ b/sharktank/sharktank/models/llama/tools/import_quark_dataset.py
@@ -23,6 +23,7 @@ import sys
 import torch
 
 from sharktank.types import *
+from sharktank.types.tensors import serialized_name_to_dtype
 from sharktank.layers.configs.llm_configs import (
     _int_prop,
     _float_prop,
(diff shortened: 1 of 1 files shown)


sharktank/sharktank/models/llama/tools/import_quark_dataset.py
```diff
@@ -61,6 +61,109 @@ def as_torch_or_none(tensor: Optional[InferenceTensor]) -> Optional[torch.Tensor
     return tensor.as_torch()
 
 
+def get_torch_dtype(dtype_str: Optional[str]) -> Optional[torch.dtype]:
```
IanNod, line 64: This is effectively the same as serialized_name_to_dtype here https://github.com/nod-ai/shark-ai/blob/a505b24906bb5bde39d41e8bf29db75c9784a388/sharktank/sharktank/types/tensors.py#L1573C5-L1573C29 correct? Can we just use that instead?

IanNod (commented):
Could you provide a more detailed commit message?

sharktank/sharktank/models/llama/tools/import_quark_dataset.py
KyleHerndon: Nice catch, I also accidentally included changes from another patch I'm working on.

PR #1743, opened 2025-07-01
[sharktank] Generate sample tokens

archana-ramalingam:
- Update instructions to generate sample tokens
- Update `max_decode_steps` in perplexity scripts
- Cleanup


archana-ramalingam:
> This PR has overlap with #1737.

I will let you merge yours as it handles OOM better. I will leave my PR to update only the relevant instructions/cleanup.

sharktank/sharktank/utils/load_llm.py
```diff
@@ -100,6 +101,9 @@ def begin_batch(
         cache_state = self.model.cache.allocate(self.page_cache_size)
         self.free_pages = list(range(1, self.page_cache_size))
 
+        if max_decode_steps < dump_decode_steps:
+            max_decode_steps = dump_decode_steps
+
```
sogartar, line 106: If anything this should throw as this is an impossible request.

sogartar (approved):

sharktank/sharktank/utils/load_llm.py
sogartar: Max decode steps is useful. It was useful for me when I wanted to run just one step.

sogartar: I guess we could increase the default CLI value.

PR #1750, opened 2025-07-01
Remove streaming from shortfin-llm

rsuderman:
Streaming is basically non-functional for beam search. We should be ripping out support so that we can refactor processing layers.


stbaione (approved):
Tracking issue created for re-enablement: #1752

PR #1754, opened 2025-07-02
Cleanup the build for the ItemProcessor

rsuderman:
Constructor reached into sub objects (batchers, page cache), didn't cache useful information (rid), and did not leverage data grouping (DecodeConfig)

### DIFF
--- a/shortfin/python/shortfin_apps/llm/components/generate.py
+++ b/shortfin/python/shortfin_apps/llm/components/generate.py
@@ -53,6 +53,7 @@ class GenerateItemProcess(sf.Process):
         self,
         *,
         gen_req: GenerateReqInput,
+        client,
         prefill_batcher,
         decode_batcher,
         page_cache,
@@ -64,6 +65,7 @@ class GenerateItemProcess(sf.Process):
         fiber: sf.Fiber,
     ):
         super().__init__(fiber=fiber)
+        self.client = client
         self.gen_req = gen_req
         self.rid = rid
         self.input_text = input_text
@@ -263,6 +265,7 @@ class ClientGenerateBatchProcess(sf.Process):
                 )
 
                 gen_process = GenerateItemProcess(
+                    client=self,
                     prefill_batcher=self.service.prefill_batcher,
                     decode_batcher=self.service.decode_batcher,
                     page_cache=self.service.page_cache,
--- a/shortfin/python/shortfin_apps/llm/components/generate.py
+++ b/shortfin/python/shortfin_apps/llm/components/generate.py
@@ -196,8 +196,8 @@ class ClientGenerateBatchProcess(sf.Process):
 
         # Try to add request to queue
         # TODO(@zphoenixrises): Add load testing and integration tests for this.
-        if not self.service.queue_manager.add_to_queue(decode_configs):
-        if not added_to_queue:
+        run_request = self.service.queue_manager.add_to_queue(decode_configs)
+        if not run_request:
             self.responder.send_error(
                 error_message="Server queue is full. Please try again later.",
                 code=ResponderErrorCodes.QUEUE_FULL,
@@ -285,7 +285,8 @@ class ClientGenerateBatchProcess(sf.Process):
         finally:
             self.service.main_fiber_pool.return_fiber(indices)
             self.responder.ensure_response()
-            self.service.queue_manager.remove_from_queue(decode_configs)
+            if run_request:
+                self.service.queue_manager.remove_from_queue(decode_configs)
 
     def generate_response(
         self,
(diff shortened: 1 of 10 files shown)


shortfin/python/shortfin_apps/llm/components/generate.py
```diff
@@ -302,9 +303,7 @@ async def run(self):
         finally:
             self.service.main_fiber_pool.return_fiber(indices)
             self.responder.ensure_response()
-
-            if added_to_queue:
-                self.service.queue_manager.remove_from_queue(total_requested_beams)
+            self.service.queue_manager.remove_from_queue(decode_configs)
```
stbaione, line 306: This `finally` block always runs, even if the request was rejected: #1711. That was causing us to call:

`self.service.queue_manager.remove_from_queue(total_requested_beams)` even when the request was rejected.

That would cause weirdness in the queue_manager:

```text
Let max_queue_size = 4, total_requested_beams = 8
```

Previously, we just returned `False` in cases where it would cause the `current_queue_size` to drop below zero, so this bug was silently around for awhile. 

It was my advice to throw an error if `current_queue_size - requested_removal_size < 0`. Need to know if our `current_queue_size` is out of sync or not.

Anyways, long winded way of saying we should keep the boolean check or another way to ensure `remove_from_queue` doesn't get invoked when a request is rejected

shortfin/python/shortfin_apps/llm/components/request_queue_manager.py
```diff
@@ -40,7 +44,7 @@ def add_to_queue(self, request_size: int) -> bool:
             logger.debug(f"Added to queue: new queue size {self.current_queue_size}")
             return True
 
-    def remove_from_queue(self, request_size: int) -> None:
+    def remove_from_queue(self, decode_configs: list[DecodeConfig]) -> None:
```
dezhiAmd, line 47: There is Docstring Mismatch:
Args:
    decode_configs: A list of DecodeConfig objects representing the request.

shortfin/python/shortfin_apps/llm/components/request_queue_manager.py
```diff
@@ -30,6 +32,8 @@ def add_to_queue(self, request_size: int) -> bool:
         Returns:
             True if the request was added successfully, False if the queue is full.
         """
+        request_size = sum(config.num_beams for config in decode_configs)
```
dezhiAmd, line 35: Maybe we can add this line before calling function add_to_queue, so there is no need to change this class and the test for this class

rsuderman: We want to move this logic to the workload manager. This delegates how the workload requirements are computed to the component that actually tracks them.

shortfin/python/shortfin_apps/llm/components/generate.py
rsuderman: Also, we should make the workload manage return a unique id, then we can always remove that ID. If no ID is provided then it can't run.

shortfin/python/shortfin_apps/llm/components/request_queue_manager.py
dezhiAmd: A small caveat is that now if the loop, 
request_size = sum(config.num_beams for config in decode_configs)
are called twice.
One in add_to_queue, one in remove_from_queue

shortfin/python/shortfin_apps/llm/components/generate.py
```diff
@@ -172,37 +172,32 @@ def _check_topk_params(
         )
         return False
 
-    def _pre_processing_sampling_params(self) -> Tuple[List[DecodeConfig], int]:
+    def get_decode_configs(self) -> Tuple[List[DecodeConfig], int]:
```
stbaione, line 175: Should update type to just `List[DecodeConfig]`

shortfin/tests/apps/llm/components/request_queue_manager_test.py
```diff
@@ -4,30 +4,28 @@
...
 
     # Try to add beyond max
-    assert queue_manager.add_to_queue(2) == False
-    assert queue_manager.current_queue_size == 6
+    assert not queue_manager.add_to_queue(get_decode_configs(2))
 
     # Remove from queue
-    queue_manager.remove_from_queue(3)
-    assert queue_manager.current_queue_size == 3
+    queue_manager.remove_from_queue(get_decode_configs(3))
```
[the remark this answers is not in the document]
rsuderman: I ended up expanding the implementation to include the ID system. Overall we want to check that it can basically track whether things can be enqueued but we shouldn't look at internal data. E.g. not look at parallel beams, total pages, etc. The idea is we restrict to looking at the interface.

PR #1756, opened 2025-07-02
[shortfin] Add support for an awaitable Task

vinayakdsci:
Plumbs initial coroutine support for use in C++ by implementing an awaitable `Task` class, which has support for coroutine chaining and generic types. 

All the coroutine logic is placed in a new directory, which will be the home of all the subsequent coroutine related work. Non-trivial template code made it difficult to split the code in a header/source layout, so all the code is implemented as a `.hpp` file.


vinayakdsci:
I'll add tests in the next commit.

PR #1759, opened 2025-07-03
[shortfin] Fix license formatting in pyproject.toml

Alex-Vasile:
As currently implemented I get the following error when running `./dev_me.py`:
```
 ValueError: invalid pyproject.toml config: `project.license`.
  configuration error: `project.license` must be valid exactly by one definition (2 matches found):

      - keys:
          'file': {type: string}
        required: ['file']
      - keys:
          'text': {type: string}
        required: ['text']

  error: subprocess-exited-with-error
  ```

### DIFF
--- a/shortfin/pyproject.toml
+++ b/shortfin/pyproject.toml
@@ -15,7 +15,7 @@ authors = [
 ]
 description = "SHARK inference library and serving engine"
 readme = "README.md"
-license = "Apache-2.0"
+license = {"text" = "Apache-2.0"}
 classifiers = [
   "Development Status :: 3 - Alpha",
   "Programming Language :: Python :: 3",


marbre:
Following https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license, the format we have should be correct. Can you provide some further details?

marbre (commented):
According to https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license, the format we have should be correct. Can you provide some further details on the environment you're using?

Alex-Vasile:
Issue was the version of setuptools being used. I had `65.5`, upgrading to `80.9.0`. I think this is related to https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files and PEP 639.

The only line in our setup that sets a restriction on the version of setuptools required is:
https://github.com/nod-ai/shark-ai/blob/854f6498787dac9a532a401550da9bc172427f5c/shortfin/pyproject.toml#L4

Which isn't high enough to satisfy PEP 639, needs to be at least `77.0.3`

marbre:
> Issue was the version of setuptools being used. I had `65.5`, upgrading to `80.9.0`. I think this is related to https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files and PEP 639.
> 
> The only line in our setup that sets a restriction on the version of setuptools required is:
> 
> https://github.com/nod-ai/shark-ai/blob/854f6498787dac9a532a401550da9bc172427f5c/shortfin/pyproject.toml#L4
> 
> Which isn't high enough to satisfy PEP 639, needs to be at least `77.0.3`

Thanks for digging deeper! I've sent #1765.

marbre:
Closing as #1765 was landed.

PR #1767, opened 2025-07-07
[Sharktank] Do not bitcast f8->i8 for index_copy_ and index_put_ during export to help compiler fusion

sogartar:
PyTorch Eager CPU does not support index_copy_ and index_put_ for f8. Because of that we need to view as int8.
On the other had the IREE compiler can't fuse these bitcasts.

With this change we are taking different paths during export and eager, such that both limitations are circumvented.
The paged KV cache is adapted to just call `ops.index_copy/put_`.


dan-garvey:
I think in general keeping parity between eager and export is desirable, but this does seem like a good exception case. (things that should be a no-op but can't get fused)

IanNod (approved):

PR #1768, opened 2025-07-07
Test pytest

dezhiAmd:


shortfin/tests/apps/llm/components/request_queue_manager_test.py
```diff
@@ -31,3 +31,9 @@ def test_request_queue_manager():
     # Remove from queue
     queue_manager.remove_from_queue(3)
     assert queue_manager.current_queue_size == 3
+
+
+def test_py_test_leak():
```
dezhiAmd, line 36: Could this code change cause a memory leak?

PR #1777, opened 2025-07-08
Integrate Wave mxfp4 bmm kernel into Sharktank

aviator19941:
This PR integrates the Wave mxfp4 batched gemm kernel into Sharktank.


sharktank/sharktank/types/layout_utils.py
```diff
@@ -126,17 +126,15 @@ def pack_nibbles(low: torch.Tensor, high: torch.Tensor) -> torch.Tensor:
...
-        max_val = low.max().item()
-        raise ValueError(
-            f"Low nibble values must be in range [{_FP4_MIN_INDEX}, {_FP4_MAX_INDEX}], got maximum value {max_val}."
-        )
-    if torch.any(high > _FP4_MAX_INDEX):
-        max_val = high.max().item()
-        raise ValueError(
-            f"High nibble values must be in range [{_FP4_MIN_INDEX}, {_FP4_MAX_INDEX}], got maximum value {max_val}."
-        )
+    # Validate nibble value ranges - using torch._assert for export compatibility
```
KyleHerndon, line 129: We shouldn't leave this as a comment, unfortunately. Not sure how to fix this yet.

aviator19941: yeah that's the only thing I'm not sure about yet, what if we made it into a tensor?

sharktank/tests/kernels/wave/mxfp4_gemm_test.py
```diff
@@ -0,0 +1,64 @@
+# Copyright 2025 Advanced Micro Devices, Inc.
```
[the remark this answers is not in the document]
aviator19941: Sure

sharktank/sharktank/types/layout_utils.py
KyleHerndon: I think torch._check should work

sharktank/sharktank/types/layout_utils.py
```diff
@@ -126,17 +126,15 @@ def pack_nibbles(low: torch.Tensor, high: torch.Tensor) -> torch.Tensor:
...
-        raise ValueError(
-            f"Low nibble values must be in range [{_FP4_MIN_INDEX}, {_FP4_MAX_INDEX}], got maximum value {max_val}."
-        )
-    if torch.any(high > _FP4_MAX_INDEX):
-        max_val = high.max().item()
-        raise ValueError(
-            f"High nibble values must be in range [{_FP4_MIN_INDEX}, {_FP4_MAX_INDEX}], got maximum value {max_val}."
-        )
+    # Validate nibble value ranges - using torch._check for export compatibility
+    torch._check(
```
Alex-Vasile, line 130: There's one more such change needed in `fp4_e2m1_to_float32` inside ocp_floats.py. If not in this PR, than I can handle it in another.

aviator19941: Done

sharktank/sharktank/ops/custom_impls.py
```diff
@@ -103,6 +109,37 @@ def matmul_generic_tensor_block_scaled_i4(
...
+        return NotImplemented
+    layout = rhs.layout_type
+    if layout is not BlockScaledFp4Layout:
+        return NotImplemented
+    rhs_unpacked = rhs.unpack()
+    quantizer = DynamicFp4BlockQuantizer(
+        block_size=32, use_fe8m0_scale=True, name="matmul_input_quantizer"
+    )
+    lhs_quantized = quantizer.quantize(lhs)
+    lhs_unpacked = lhs_quantized.unpack()
```
sogartar, line 128: Why does `lhs` get unconditionally unboxed/dequantized and then quantized again?
Can't we skip requantization if it already has the expected qunatization?

aviator19941: the reason is that the lhs is expected to come in as a f16 or f32 tensor, so we need to quantize it. It doesn't have the expected quantization.

KyleHerndon: The input to this override is lhs is a normal tensor, rhs is a QuantizedTensor. The `unbox_tensor` is realistically just for extracting primitive tensors into torch tensors.

sogartar: My bad. I misread the override, thinking that `lhs` can be something else.

PR #1790, opened 2025-07-09
Correctly support --use-hf flag for tensor and pipeline parallelism

Alex-Vasile:


dan-garvey (approved):
good cleanup too

dan-garvey:
A lot of this code seems untested, if you don't have time to add them, it wouldn't hurt to file an issue

Alex-Vasile:
> A lot of this code seems untested, if you don't have time to add them, it wouldn't hurt to file an issue

Good point.


I'll add that to the llama iree vs eager test and parameterize that further. Should be a trivial change to add an export flag.

PR #1796, opened 2025-07-10
[sharktank] make view op support dtype overload

sogartar:
The troch.view op has dtype overload. This aligns our op API with torch.

### DIFF
--- a/sharktank/sharktank/ops/default_impls.py
+++ b/sharktank/sharktank/ops/default_impls.py
@@ -917,8 +917,18 @@ def _split_topk(
 
 
 @view.override(Tensor)
-def view_default(tensor: Union[Tensor, PrimitiveTensor], shape: List[int]) -> Tensor:
-    return unbox_tensor(tensor).view(*shape)
+def view_default(
+    tensor: Union[Tensor, PrimitiveTensor],
+    shape: List[int] | None,
+    dtype: torch.dtype | None,
+) -> Tensor:
+    assert (shape is None) ^ (
+        dtype is None
+    ), "Exactly one of shape or dtype must be provided"
+    if shape is not None:
+        return unbox_tensor(tensor).view(*shape)
+    else:
+        return unbox_tensor(tensor).view(dtype)
 
 
 @view.override(QuantizedTensor)
--- a/sharktank/sharktank/ops/signatures.py
+++ b/sharktank/sharktank/ops/signatures.py
@@ -1698,18 +1698,23 @@ def _topk_trampoline(
 
 
 @overridable
-def view(tensor: AnyTensor, shape: List[int]) -> AnyTensor:
+def view(
+    tensor: AnyTensor, shape: List[int] | None = None, dtype: torch.Tensor | None = None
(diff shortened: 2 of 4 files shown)


sharktank/sharktank/ops/default_impls.py
```diff
@@ -917,8 +917,18 @@ def _split_topk(
...
+    shape: List[int] | None,
+    dtype: torch.dtype | None,
+) -> Tensor:
+    assert (shape is None) ^ (
+        dtype is None
+    ), "Exactly one of shape or dtype must be provided"
+    if shape is not None:
+        return unbox_tensor(tensor).view(*shape)
+    else:
+        return unbox_tensor(tensor).view(dtype)
```
Alex-Vasile, line 931: This should return back a primitive tensor if it the input was a primitive tensor.

sharktank/sharktank/types/tensors.py
```diff
@@ -538,15 +538,38 @@ def unsqueeze(self, dim: int) -> "AnyTensor":
...
+        ...
+
+    @overload
     def view(self, *args: Union[List[List[int]], List[int]]) -> "AnyTensor":
+        ...
+
+    def view(
+        self,
+        *args: Union[List[List[int]], List[int], torch.dtype],
+        dtype: torch.dtype | None = None,
```
Alex-Vasile, line 552: We have Tensor.to that needs to handle a similar case, but that one is implemented in a different way, the main func takes (*args, **kwargs). I think it would be better to stay consistent with how we implement these overloaded funcs.

sogartar: I don't think 
```
def view(self, *args, **kwargs)
```
is more understandable compared to
```
def view(
    self,
    *args: Union[List[List[int]], List[int], torch.dtype],
    dtype: torch.dtype | None = None,
)
```

Why not make every function accept just `*args, **kwargs` then?

Alex-Vasile (approved):

sharktank/sharktank/ops/signatures.py
```diff
@@ -1698,18 +1698,23 @@ def _topk_trampoline(
...
     ...
 
 
 @view.trampoline
 def _view_trampoline(
-    d: SignatureDispatcher, tensor: AnyTensor, shape: List[int]
+    d: SignatureDispatcher,
+    tensor: AnyTensor,
+    shape: List[int] | None = None,
+    dtype: torch.Tensor | None = None,
```
archana-ramalingam, line 1713: Same dtype change needed here too.

sharktank/sharktank/ops/signatures.py
```diff
@@ -1698,18 +1698,23 @@ def _topk_trampoline(
 
 
 @overridable
-def view(tensor: AnyTensor, shape: List[int]) -> AnyTensor:
+def view(
+    tensor: AnyTensor, shape: List[int] | None = None, dtype: torch.Tensor | None = None
```
archana-ramalingam, line 1702: Same dtype change needed here too.

PR #1805, opened 2025-07-10
Initial support for QuantizedTensors as shards of ReplicatedTensors

Alex-Vasile:
Depends on #1803.

### DIFF
--- a/sharktank/sharktank/ops/__init__.py
+++ b/sharktank/sharktank/ops/__init__.py
@@ -28,6 +28,7 @@ from .utils import trivially_replicable
 from . import default_impls
 from . import custom_impls
 from . import sharded_impls
+from . import quantized_impls
 
 from . import attention_impls
 


sharktank/sharktank/types/pipelining.py
```diff
@@ -20,6 +20,8 @@
 
 from typing import Tuple
 
+from sharktank.types.tensors import QuantizedTensor
```
archana-ramalingam, line 23: Can we move this import as part of existing import above `from sharktank.types import (`?

sharktank/sharktank/types/tensors.py
```diff
@@ -1385,16 +1387,19 @@ def __init__(
         will be replicated that many times.
         """
         if devices is None:
-            num_shards = shard_count if isinstance(ts, torch.Tensor) else len(ts)
+            num_shards = len(ts) if isinstance(ts, list) else shard_count
             devices = tuple(range(num_shards))
 
-        if isinstance(ts, torch.Tensor):
+        if isinstance(ts, (torch.Tensor, QuantizedTensor)):
```
rsuderman, line 1393: We should make this a check for whether its any type of tensor. `torch.Tensor` or sharktank tensor type. We could also just check whether its a list.

rsuderman (changes requested):
Could we include a test for the tensor transfer.

archana-ramalingam (approved):

sharktank/sharktank/ops/quantized_impls.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+
+
+def transfer_or_barrier(
+    operation: Callable, tensor: QuantizedTensor | QuantizerTensor, ordinal: int
+):
+    def operation_transform(globals: dict[str, Tensor]) -> dict[str, Tensor]:
+        return {k: operation(f"{ordinal}", v) for k, v in globals.items()}
+
+    return tensor.transform_globals(operation_transform)
```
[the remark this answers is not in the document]
rsuderman: Nope, its meant for only globals, this has you transferring them on operation.

rsuderman (approved):

PR #1809, opened 2025-07-11
Sharktank extend rotary mask

stbaione:
This is a requirement for implementing `offset prefill`.

Previously, our `compute_batch_mask` function only supported `batch_seq_len == 1`.

However, for `offset prefill`, we need to support a varying `batch_seq_len`, which is often greater than one.

Also did some test cleanup, so that they're all self-contained in their own `unittest` class, and so that the `RotaryEmbeddingLayer` tests are centralized to one location.


dan-garvey:
#1792 #1800 
You may want to take a look at these as there is some overlap. 

I don't remember why I was under the impression that the function only supported bs_seq_len == 1, but as long as we're testing things that aren't that its fine to approve by me.

rsuderman (changes requested):

PR #1812, opened 2025-07-12
Add target gfx950 for flux model on MI355 amd device

AmosLewis:


jinchen62 (approved):

shortfin/python/shortfin_apps/flux/server.py
```diff
@@ -304,7 +304,7 @@ def main(argv, log_config=UVICORN_LOG_CONFIG):
         type=str,
         required=False,
         default="gfx942",
-        choices=["gfx942", "gfx1100", "gfx90a"],
+        choices=["gfx942", "gfx1100", "gfx90a", "gfx950"],
```
IanNod, line 307: We should have a precompiled vmfb to support that build_preference too

AmosLewis: here [sharkpublic/flux.1/20250711_mi355/](https://sharkpublic.blob.core.windows.net/sharkpublic/flux.1/20250711_mi355/)

PR #1817, opened 2025-07-14
[sharktank] Update perplexity logging

archana-ramalingam:
- Update perplexity logging to print prefill & decode times
- Skip mlir export step, if vmfb is passed
- Update `time()` to `time_ns()` for improved precision


sharktank/sharktank/evaluate/perplexity_iree.py
```diff
@@ -372,8 +347,11 @@ def decode_vmfb(
         self.print_token_comparison(i)
         return decode_logits
 
-    @timeit
     def get_logits(self, skip_decode: bool) -> torch.Tensor:
+        self.prefill_time = 0
+        self.decode_time = []
+        self.decode_steps = 0
```
Alex-Vasile, line 353: This var isn't needed, use len(self.decode_time)

sharktank/sharktank/evaluate/perplexity_iree.py
```diff
@@ -387,7 +365,9 @@ def run_iree_module(devices: list[iree.runtime.HalDevice]):
                 if skip_decode or len(out_logits) == 0:
                     token_batch = self.token_ids[:, : i + 1]
 
+                    start = time.time()
```
Alex-Vasile, line 368: Accuracy issues in time.time() will likely start being an issue here for our smaller models. I think it would be best to replace this with time.time_ns() instead and do the conversion from that.

sharktank/sharktank/evaluate/perplexity_torch.py
```diff
@@ -56,38 +56,14 @@ def __init__(
...
-
-            if total_seconds < 1:
-                time_taken = f" {round(total_seconds * 1000, 3)} ms"
-            elif total_seconds < 60:
-                time_taken = "{:.2f} secs".format(round(float(total_seconds), 2))
-            else:
-                time_taken = "{:02d} hrs : {:02d} mins : {:.2f} secs".format(
-                    int(hours), int(minutes), round(float(seconds), 2)
-                )
+            time_taken = calc_time(start, end)
```
Alex-Vasile, line 64: Same comment about the time.time() above

sharktank/sharktank/evaluate/perplexity_iree.py
archana-ramalingam: Removed

PR #1819, opened 2025-07-14
Check that pages are released on shutdown

rsuderman:
When the service is shutdown we should track and verify no pages were lost. This guarantees that if a decoding process crashes or fails to release pages an exception is thrown.

### DIFF
--- a/sharktank/sharktank/kernels/wave/mxfp4_gemm.py
+++ b/sharktank/sharktank/kernels/wave/mxfp4_gemm.py
@@ -1,0 +1,212 @@
+# Copyright 2025 Advanced Micro Devices, Inc.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+from sharktank.kernels.base import *
+from sharktank.kernels.mlir_kernel import *
+from sharktank.kernels.wave.utils import get_wave_module_body_asm
+import iree.turbine.kernel.lang as tkl
+import iree.turbine.kernel.wave as tkw
+from iree.turbine.kernel.lang.global_symbols import *
+from iree.turbine.kernel.wave.scheduling.schedule import SchedulingType
+from iree.turbine.kernel.wave.compile import wave_compile, WaveCompileOptions
+from iree.turbine.kernel.wave.templates.attention_common import AttentionShape
+from iree.turbine.kernel.wave.constraints import ScaledMMAType
+from iree.turbine.kernel.wave.utils.general_utils import (
+    get_default_scheduling_params,
+)
+from iree.turbine.kernel.wave.utils.run_utils import (
+    set_default_run_config,
+)
+from iree.compiler.ir import (
+    Module,
+    Context,
+)
+import torch
+
+
+__all__ = [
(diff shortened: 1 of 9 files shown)


shortfin/python/shortfin_apps/llm/components/batcher.py
```diff
@@ -111,23 +111,25 @@ async def board_flights(self):
 
         to_schedule = self.scheduler.should_execute(rid_map, self.strobes)
 
-        cache = self.page_cache
+        page_cache = self.page_cache
```
zeeshanhaque21, line 114: Do we really need this? it might be better to keep referencing it as self.page_cache in the following lines as that makes it explicit that this is what is being manipulated

shortfin/python/shortfin_apps/llm/components/batcher.py
```diff
@@ -111,23 +111,25 @@ async def board_flights(self):
...
         ...
 
-    def board_request(self, cache, request: LlmInferenceExecRequest):
+    def board_request(self, page_cache, request: LlmInferenceExecRequest):
         ...
 
-    def board(self, cache: BasePagedAttentionCache, fiber: Fiber, to_schedule: set):
+    def board(
+        self, page_cache: BasePagedAttentionCache, fiber: Fiber, to_schedule: set
+    ):
```
zeeshanhaque21, line 132: Do we need to be passing these variables in if they are already accessible through self? Ideally, I think these should be "private" variables as they are most likely not being used by objects/functions external to the BatcherProcess

shortfin/python/shortfin_apps/llm/components/batcher.py
```diff
@@ -590,13 +593,13 @@ async def get_args(
...
         int_dtype = sfnp.int64
 
-        tokens = cache.allocate([bs, 1], int_dtype)
-        start_positions = cache.allocate([bs], int_dtype)
-        seq_lens = cache.allocate([bs], int_dtype)
-        seq_block_ids = cache.allocate([bs, block_count], int_dtype)
+        tokens = array_cache.allocate([bs, 1], int_dtype)
+        start_positions = array_cache.allocate([bs], int_dtype)
+        seq_lens = array_cache.allocate([bs], int_dtype)
+        seq_block_ids = array_cache.allocate([bs, block_count], int_dtype)
```
zeeshanhaque21, line 602: Same as above

shortfin/python/shortfin_apps/llm/components/kvcache/base_attention_cache.py
```diff
@@ -145,6 +145,12 @@ def __init__(
             None if not use_ref_counts else threading.Lock()
         )
 
+    def shutdown(self):
+        available = self.page_pool.available_page_count()
+        total = self.page_pool.total_page_count()
+        if available != total:
+            raise ValueError(f"Pages lost: {total - available} of {total} unfreed")
+
```
zeeshanhaque21, line 153: Nice!!

zeeshanhaque21 (approved):
Looks great! Minor nits.

shortfin/python/shortfin_apps/llm/components/batcher.py
rsuderman: Same reason - accessing member variables multiple times for performance sensitive code is usually recommended against.

PR #1821, opened 2025-07-14
Cleanup scheduler to update rather than add / remove

rsuderman:
Makes more sense to notify the expected size instead of sending deltas. This is more realistic as we would not want a constant update on what future job sizes will look like.

### DIFF
--- a/shortfin/python/shortfin_apps/llm/components/scheduler.py
+++ b/shortfin/python/shortfin_apps/llm/components/scheduler.py
@@ -39,9 +39,6 @@ class Workgroup:
     def members(self):
         return set(self._members.keys())
 
-    def get_members(self, members):
-        return set(members) & set(self._members.keys())
-
     def has_member(self, rid):
         return rid in self._members
 
--- a/shortfin/python/shortfin_apps/llm/components/scheduler.py
+++ b/shortfin/python/shortfin_apps/llm/components/scheduler.py
@@ -12,14 +12,7 @@ import threading
 logger = logging.getLogger(__name__)
 
 
-class NewWorkItem(sf.Message):
-    def __init__(self, *, count: int, rid: int):
-        super().__init__()
-        self.count = count
-        self.rid = rid
-
-
-class DoneWorkItem(sf.Message):
+class UpdateWorkload(sf.Message):
     def __init__(self, *, count: int, rid: int):
         super().__init__()
         self.count = count
@@ -49,7 +42,7 @@ class Workgroup:
     def get_members(self, members):
         return set(members) & set(self._members.keys())
 
-    def is_member(self, rid):
+    def has_member(self, rid):
         return rid in self._members
 
     def member_count(self, rid):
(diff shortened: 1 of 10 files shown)


shortfin/python/shortfin_apps/llm/components/scheduler.py
```diff
@@ -49,7 +42,7 @@ def members(self):
     def get_members(self, members):
```
dezhiAmd, line 42: This function is not called anywhere. And it redundant with members property. Can it be deleted?

shortfin/python/shortfin_apps/llm/components/scheduler.py
```diff
@@ -241,42 +226,38 @@ def schedule_new():
...
         assert rid in self._workgroup_placement
 
         wid = self._workgroup_placement[rid]
         workgroup = self._workgroups[wid]
 
-        workgroup.remove(rid=rid, count=count)
-        if workgroup.is_empty():
-            self._workgroups.pop(wid)
+        workgroup.resize(rid=rid, count=0)
+        self._workgroups.pop(wid)
```
dezhiAmd, line 239: The code before this change appears to be safer since workgroup may not be empty

shortfin/python/shortfin_apps/llm/components/scheduler.py
rsuderman: Removed.

shortfin/python/shortfin_apps/llm/components/token_selection_strategy/token_selector.py
```diff
@@ -45,26 +45,18 @@ async def decode(
...
-        for _ in range(config.decode_config.max_completion_tokens):
+        active_beam_count = beam_group.active_beam_count
+        config.decode_reserve_callback(
+            rid=exec_req.orig_instance_id, count=active_beam_count
+        )
+        for _ in range(config.decode_config.max_completion_tokens - 1):
             if self.cancelled:
                 break
 
             active_beam_count = len(beam_group.active_beams)
```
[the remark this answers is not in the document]
rsuderman: Removed the add as its at the start of the loop anyway.

shortfin/python/shortfin_apps/llm/components/scheduler.py
```diff
@@ -202,13 +187,13 @@ def _schedule(self, *, rid, count):
...
-                return
             existing = workgroup.member_count(rid=rid)
-            workgroup.remove(rid=rid, count=existing)
-            count = count + existing
+            if workgroup.can_add(count - existing):
+                workgroup.resize(rid=rid, count=count)
+                return
 
+            # Remove the existing workload and readd
+            workgroup.resize(rid=rid, count=0)
```
zeeshanhaque21, line 196: Maybe move this logic to a separate remove_and_readd_workload. I think it is better to stay away from implicit behavior.

zeeshanhaque21 (approved):
LGTM

PR #1826, opened 2025-07-15
[sharktank] Add tensor slicing for f4 quantization

sogartar:


sogartar:
Depends on PR #1825.

sharktank/sharktank/types/misc.py
```diff
@@ -0,0 +1,113 @@
...
+    stop = size if s.stop is None else s.stop
+    stop = stop if stop >= 0 else size + stop
+
+    step = 1 if s.step is None else s.step
+    return slice(start, stop, step)
+
+
+def squeeze_slice(s: Slice) -> Slice:
+    """Remove Nones that represent insertion of a singleton dimensions."""
+    if not isinstance(s, tuple):
```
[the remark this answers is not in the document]
sogartar: When it comes to slicing tuple is special.

```python
import torch
t = torch.arange(0, 6, 1, dtype=int, device="cpu").view([2, 3])
print(f"t =\n{t}")
print(f"t[[0, 1]] =\n{t[[0, 1]]}")
print(f"t[(0, 1)] =\n{t[(0, 1)]}")
```

```
t =
tensor([[0, 1, 2],
        [3, 4, 5]])
t[[0, 1]] =
tensor([[0, 1, 2],
        [3, 4, 5]])
t[(0, 1)] =
1
```

A tuple means indexing into different dimension. A list means indexing into the same dimension.

sharktank/sharktank/ops/quantized_impls.py
```diff
@@ -52,3 +129,66 @@ def operation_transform(globals: dict[str, Tensor]) -> dict[str, Tensor]:
...
         operation_transform, copy_external_tensor_trait=False
     )
+
+
+@extract_slice.override(PlanarQuantizedTensor)
+@quantized_tensor_layout_of_type(tensor=BlockScaledFp4Layout)
+def extract_slice_BlockScaledFp4Layout(tensor: PlanarQuantizedTensor, key: Slice):
+    layout: BlockScaledFp4Layout = tensor.layout
+    slice_ = canonicalize_slice_descriptor(squeeze_slice(key), tensor.shape)
+    assert all(
```
KyleHerndon, line 139: Using `torch._check` might be necessary for compatibility with torch dynamo, but I haven't thought about how it interacts with the trampoline to sharding.

KyleHerndon (approved):

sharktank/sharktank/ops/quantized_impls.py
sogartar: What do you mean? The dispatch mechanism is like for any other tensor type.

PR #1835, opened 2025-07-15
Fix `view` override for `QuantizedTensor`

paulzzy:
`_view_trampoline` in sharktank/ops/signatures.py passes three parameters (with the third being `dtype`) to `view_QuantizedTensor` via `override`


dan-garvey (commented):
I can't imagine a case where we would want to do a `view(dtype)` on a quantized tensor, so this is probably fine? Why is removing the arg better than adding support though? 

nit: can you type annotate shape while you're at it? like the function above.

dan-garvey:
Looks fine to me. Your patch is unrelated to the failure in perplexity.

dan-garvey (approved):
Please add a test that does view on two quantized tensors before merging

PR #1841, opened 2025-07-15
Create a new Decoder framework

rsuderman:
A new framework for decoding should drastically improve decoding speed by limiting the time between steps. This avoids many of the heavy data structures / choices made during decoding.


zeeshanhaque21 (approved):
LGTM!!

Issue #1833, opened 2025-07-15
Fix discrepency in expected dimension between `BlockScaledLayout` and `BlockScaledFp4Layout`


Alex-Vasile:
There is a discrepancy of the expected dimension structure between `BlockScaledLayout` and `BlockScaledFp4Layout`.
The `BlockScaledLayout` expects a trailing dimension of size 1 in the block scales tensor.
`BlockScaledFp4Layout` does not have this. It is better to have it because then dimensions are aligned and broadcasting is natural during dequantization. It is also important to stay consistent.
```
layout.d.to(dtype) * layout.qs.to(dtype)
```

instead of having to do an unsqueeze also
```
layout.d.to(dtype).unsqueeze(-1) * layout.qs.to(dtype)
```

This would require generating new IRPA files for Llama 3.1 FP4.
Here is one example tensor from the quantized Llama 405B model.
```
BlockScaledFp4Layout(d([16384, 1664], dtype=torch.uint8), qs([16384, 1664, 16], dtype=torch.uint8), block_size=32, use_fe8m0_scale=True)
```

_Originally posted by @sogartar in https://github.com/nod-ai/shark-ai/pull/1810#discussion_r2203896133_

PR #1847, opened 2025-07-16
[sharktank] Fix erroneous last dim squeezing in compute_fp4_block_scales

sogartar:
This squeezing does not make sense. I suspect it was never caught because doing a squeeze on a dimension of size other than 1 does not raise an error.


KyleHerndon (approved):

KyleHerndon:
Related to #1833

PR #1848, opened 2025-07-16
Handle `StaticDim` scalars in `mlir_kernel` decorator

paulzzy:
Scalars cannot be specialized since they have no dimensions.

`StaticDim.EXAMPLE(0)` produces a scalar.


paulzzy:
I just realized scalars don't really make sense. E.g. it would allow something like `MLIRTensor[X, Y, 0, Z]`. Is there any neat way to handle scalars? Or should `MLIRTensor` only be allowed to support positive nonzero dimensions?

paulzzy:
Looks like [masked_flash_attention.py](https://github.com/nod-ai/shark-ai/blob/61ea387f57a830a67b8b7720e84732396d434a59/sharktank/sharktank/kernels/attention.py#L35) has an example for this.

PR #1849, opened 2025-07-16
Prepare FP8 quantization for sharded support

Alex-Vasile:
A collection of small changes needed to get FP8 quantized llama running with pipeline parallelism. 
- Comment out unused `prob_output_scale` in quark dataset conversion tool
- Change `unpack_to_raw_tensor` to work with ShardedTensor
- Change `masked_flash_attention` to work with ReplicatedTensors. Ideally we would like this to be handled by `is_trivially_replicable`, but it would be very messy to hook up with the current implementation of `is_trivially_replicable`.

Depends on #1820.


archana-ramalingam (approved):

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -131,9 +132,20 @@ def paged_attention_kv_cache_gather(
 kv_cache_gather = KVCacheGatherKernel()
 
 
-def unpack_raw_tensor(tensor):
+def unpack_to_raw_tensor(tensor: AnyTensor) -> AnyTensor:
```
dan-garvey, line 135: I think the initial commit also includes qualms about the name. I think raw might just be the wrong word and the whole name could just be "maybe_unpack_tensor"

sharktank/sharktank/models/llama/tools/import_quark_dataset.py
```diff
@@ -393,20 +393,22 @@ def update_norm_layer(
...
-                .to(torch.float32)
-                * 2.0
-            )
-            if weight_dtype_override is not None:
-                prob_output_scale = prob_output_scale.to(weight_dtype_override)
-            new_name = f"blk.{layer_idx}.attn_scale"
-            updated_tensors[new_name] = DefaultPrimitiveTensor(
-                name=new_name, data=prob_output_scale
-            )
+        # NOTE: Disabled since not needed.
```
dan-garvey, line 396: I think commenting out isn't necessary, the name should be different anyway so just re-enabling it would be bad. Git keeps history anyway so I'd prefer you just delete

dan-garvey (approved):
small change but otherwise looks good to me

sharktank/sharktank/layers/paged_attention.py
KyleHerndon: I'd assume that a maybe_unpack_tensor has a chance of failing and returning None, or something similar. I think the unconditional implication is fine. It's true of "unbox_tensor" already that we use everywhere.

sharktank/sharktank/ops/attention_impls.py
```diff
@@ -39,6 +39,20 @@ def _extract_linear_scale(t):
 
 
 def masked_flash_attention(q, k, v, a):
+    if (
```
dan-garvey, line 42: After chatting with Kyle about this, I think this is okay as a temporary thing because we're in a rush to support this, but it would be good to add a TODO suggesting we use the trampoline like every other op instead of calling this kernel directly

sogartar: I am not sure why we need a special handling for the case of all-replicated tensors. If this op were a proper registered op, you would get the all-replicated overload for free like it is for all other ops now.

sogartar: https://github.com/nod-ai/shark-ai/blob/c7c5f4d7973a521a2a48c2e9a518ce6cc1295aaa/sharktank/sharktank/ops/_registry.py#L350-L354

Alex-Vasile: Yes, the issue is that it's not being registered normally. It looks like it's being conditionally registered:
https://github.com/nod-ai/shark-ai/blob/14b4a97fc7871454495bb561b9ea45e421875989/sharktank/sharktank/ops/attention_impls.py#L102-L105

https://github.com/nod-ai/shark-ai/blob/14b4a97fc7871454495bb561b9ea45e421875989/sharktank/sharktank/examples/export_paged_llm_v1.py#L50-L53

I am not sure why we're conditionally overriding @dan-garvey

sogartar: I think the correct is to register `flash_attention` as a separate op. Also merge the implementation of `masked_flash_attention` into it. The presence of the mask argument should drive which is used.
Then the model config drives what is called at the call site. The argument `--attention-kernel` should be changing the model config. I suppose its purpose is to override the model config. We already have it.

PR #1856, opened 2025-07-17
Fix toy deepseek perplexity failures

Alex-Vasile:


sogartar (approved):
Does this get changed because of changes in IREE? Is it also flaky, I think I saw it being flaky but did not take a note, so I am not sure.
Isn't it better to make the tolerance really high as we know numerics are off?

PR #1870, opened 2025-07-17
[sharktank][mlir_kernel] Check for cache before generating kernel IR

Groverkss:
Before this PR, every time we used a wave kernel in a loop, the kernel would get generated every time by `wave_compile` but the actual kernel used would always be cached. After this patch, we first check for the cache, and then generate it.


aviator19941 (approved):
LGTM!

PR #1871, opened 2025-07-18
Refactor decoder with stateful tools

rsuderman:
Both decoding and stepping can be done without updating large sets of pages and ids every step. This implementation attempts to reduce the overhead by both tracking tokens / beams choices per step and only building the final token ids at the end invocation.


shortfin/python/shortfin_apps/llm/components/decoder/decoder.py
```diff
@@ -94,63 +92,171 @@ def select_topk(scores: np.ndarray, decode_config: DecodeConfig):
...
+        self._decode_config = decode_config
+        self._eos_token_id = self._decode_config.eos_token_id
+        self._hypothesis = self._decode_config.num_beams
+
+        self._select_function = None
+        self._select_function = (
+            select_topk if decode_config.use_beam_search else select_greedy
+        )
+
+        self._score_function = None
```
IanNod, line 184: why set _score_function and _select_function = None, then immediately getting a real value after?

shortfin/python/shortfin_apps/llm/components/decoder/decoder.py
```diff
@@ -94,63 +92,171 @@ def select_topk(scores: np.ndarray, decode_config: DecodeConfig):
...
+        beams, tokens = self._select(logits, indices)
+
+        return beams, tokens
+
+    def done(self):
+        return len(self._completed) >= self._hypothesis
+
+    def _build_response(self, beam, end_step):
+        tokens = []
+        for step in range(end_step - 1, -1, -1):
```
IanNod, line 236: Why are we stepping through the tokens in reverse order, then reversing back at the end?

IanNod (commented):
Are these changes already covered in our unit tests properly?

shortfin/python/shortfin_apps/llm/components/decoder/decoder.py
rsuderman: Beam search exploration means that each step needs to build off the prior selection. If you build forward you have to copy the entire history before appending the new token selectected, this means per hypothesis the build time is `O(n^2)` if you build forward (copying the history each step) whereas building in reverse you get `O(n)` as you do not need to copy the history. Its a little bit weird but you can think of traversing the path of a dijkstras once you hit the target location. You do not need to look at the unexplored leafs nodes.

IanNod (approved):

PR #1872, opened 2025-07-18
[Fusili] Test harness for MLIR asm emitter

AaronStGeorge:
This PR sets up `lit` + `FileCheck` tests for MLIR asm emitter. `lit` + `FileCheck`  are a well known and battle tested formula for testing MLIR asm, this approach should minimize fiddling with exact string matches.

### DIFF
--- a/.github/workflows/ci-sharkfuser.yml
+++ b/.github/workflows/ci-sharkfuser.yml
@@ -112,6 +112,10 @@ jobs:
       working-directory: ${{ env.SHARKFUSER_DIR }}
       run: |
         ctest --test-dir build -T test -T coverage --timeout 30
+        # LLVM includes some broken .gcda files to test llvm-cov. There doesn't
+        # seem to be a way to convince lcov to not scan the broken files, only
+        # to ignore them after the scan, hence the `--ignore-errors` flag as
+        # well as the `--exclude`.
         lcov --capture --directory build --ignore-errors gcov,version --exclude "*/_deps/*" --output-file build/coverage.info
         lcov --remove build/coverage.info 'build/*' '/usr/*' --ignore-errors unused --output-file build/coverage.info
         genhtml build/coverage.info --output-directory coverage_report
--- a/.github/workflows/ci-sharkfuser.yml
+++ b/.github/workflows/ci-sharkfuser.yml
@@ -35,7 +35,8 @@ jobs:
     runs-on: ${{ matrix.runs-on }}
     defaults:
       run:
-        shell: bash
+        # github actions swaps the {0} at the end for your script
+        shell: bash --noprofile --norc -exo pipefail {0}
     strategy:
       fail-fast: false
       matrix:
@@ -110,10 +111,9 @@ jobs:
       if: ${{ matrix.name == 'Ubuntu (GCC 13, Code Coverage)' }}
       working-directory: ${{ env.SHARKFUSER_DIR }}
       run: |
-        set -x
         ctest --test-dir build -T test -T coverage --timeout 30
-        lcov --capture --directory build --ignore-errors gcov,version --output-file build/coverage.info
-        lcov --remove build/coverage.info 'build/*' '/usr/*' '/tmp/*' --output-file build/coverage.info
+        lcov --capture --directory build --ignore-errors gcov,version --exclude "*/_deps/*" --output-file build/coverage.info
+        lcov --remove build/coverage.info 'build/*' '/usr/*' --output-file build/coverage.info
         genhtml build/coverage.info --output-directory coverage_report
 
     - name: Upload code coverage report
--- a/.github/workflows/ci-sharkfuser.yml
+++ b/.github/workflows/ci-sharkfuser.yml
@@ -113,7 +113,7 @@ jobs:
       run: |
         ctest --test-dir build -T test -T coverage --timeout 30
         lcov --capture --directory build --ignore-errors gcov,version --exclude "*/_deps/*" --output-file build/coverage.info
-        lcov --remove build/coverage.info 'build/*' '/usr/*' --output-file build/coverage.info
+        lcov --remove build/coverage.info 'build/*' '/usr/*' --ignore-errors unused --output-file build/coverage.info
         genhtml build/coverage.info --output-directory coverage_report
 
     - name: Upload code coverage report
--- a/.github/workflows/ci-sharkfuser.yml
+++ b/.github/workflows/ci-sharkfuser.yml
@@ -35,7 +35,7 @@ jobs:
     runs-on: ${{ matrix.runs-on }}
     defaults:
       run:
-        shell: bash -x
+        shell: bash
     strategy:
       fail-fast: false
       matrix:
@@ -110,6 +110,7 @@ jobs:
       if: ${{ matrix.name == 'Ubuntu (GCC 13, Code Coverage)' }}
       working-directory: ${{ env.SHARKFUSER_DIR }}
       run: |
+        set -x
         ctest --test-dir build -T test -T coverage --timeout 30
         lcov --capture --directory build --ignore-errors gcov,version --output-file build/coverage.info
         lcov --remove build/coverage.info 'build/*' '/usr/*' '/tmp/*' --output-file build/coverage.info
--- a/.github/workflows/ci-sharkfuser.yml
+++ b/.github/workflows/ci-sharkfuser.yml
@@ -112,12 +112,8 @@ jobs:
       working-directory: ${{ env.SHARKFUSER_DIR }}
       run: |
         ctest --test-dir build -T test -T coverage --timeout 30
-        # LLVM includes some broken .gcda files to test llvm-cov. There doesn't
-        # seem to be a way to convince lcov to not scan the broken files, only
-        # to ignore them after the scan, hence the `--ignore-errors` flag as
-        # well as the `--exclude`.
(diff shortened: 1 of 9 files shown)


.github/workflows/ci-sharkfuser.yml
```diff
@@ -108,8 +112,12 @@ jobs:
...
       run: |
         ctest --test-dir build -T test -T coverage --timeout 30
-        lcov --capture --directory build --output-file build/coverage.info
-        lcov --remove build/coverage.info 'build/*' '/usr/*' --output-file build/coverage.info
+        # LLVM includes some broken .gcda files to test llvm-cov. There doesn't
+        # seem to be a way to convince lcov to not scan the broken files, only
+        # to ignore them after the scan, hence the `--ignore-errors` flag as
+        # well as the `--exclude`.
+        lcov --capture --directory build --ignore-errors gcov,version --exclude "*/_deps/*" --output-file build/coverage.info
+        lcov --remove build/coverage.info 'build/*' '/usr/*' --ignore-errors unused --output-file build/coverage.info
```
AaronStGeorge, line 120: @sjain-stanford the `.gcda` files in llvm reak gcov-viewer as well.

AaronStGeorge: We could do something like delete them all from the LLVM checkout in CMake... but that sounds a little error prone.

sjain-stanford: We have some directory exclusions for code coverage in `build_tools/cmake/CTestCustom.cmake.in` - this is specifically to avoid generating `.gcda` / `.gcno` files. Does adding to it help? Or are you saying these files are already available in LLVM so they always get picked by lcov?

sjain-stanford: Also, is gcov-viewer not usable after bringing in LLVM? We should find a way to make it work.

AaronStGeorge: > We have some directory exclusions for code coverage in build_tools/cmake/CTestCustom.cmake.in - this is specifically to avoid generating .gcda / .gcno files. Does adding to it help? Or are you saying these files are already available in LLVM so they always get picked by lcov?

They are test fixtures in LLVM for testing `llvm-cov` that are there in the repo on checkout, not generated by the coverage run, so I don't think the exclusions in CTestCustom.cmake.in do anything in this case.

AaronStGeorge: > Also, is gcov-viewer not usable after bringing in LLVM? We should find a way to make it work.

Yes it's broken after bringing in LLVM. I'm not sure how to fix it besides deleting the text fixtures from the LLVM checkout. I tried a `.lcovrc` file and it seems to just be ignored.

sharkfuser/tests/CMakeLists.txt
```diff
@@ -47,3 +47,8 @@ add_sharkfuser_test(
   SRCS
     test_logging.cpp
 )
+
+add_sharkfuser_lit_test(
+	SRC example_lit_test.cpp
```
sjain-stanford, line 52: Not too important rn but eventually we might want to have this take in multiple SRCS.

sharkfuser/build_tools/cmake/CTestMacros.cmake
```diff
@@ -5,7 +5,21 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 
-function(_add_sharkfuser_target)
+# Creates and executable target for use in a test.
```
sjain-stanford, line 8: typo: an

sharkfuser/build_tools/cmake/CTestMacros.cmake
```diff
@@ -82,10 +139,84 @@ function(add_sharkfuser_sample)
...
+    return()
+  endif()
+
+  cmake_parse_arguments(
+    _RULE               # prefix
+    ""                  # options
+    "SRC"               # one value keywords
+    "TOOLS;DEPS"        # multi-value keywords
+    ${ARGN}
+  )
```
sjain-stanford, line 178: minor: if early return works here, could we change the other functions to do the same for consistency? For some reason I thought the `cmake_parse_arguments` needs to happen first, maybe I was wrong.

sharkfuser/tests/CMakeLists.txt
AaronStGeorge: 👍, when we need to support multiple I think we would need a `SRCS` + `TEST_FILE` argument or something, otherwise I'm not sure how the function would know which file contained the lit directives.

sharkfuser/build_tools/cmake/CTestMacros.cmake
```diff
@@ -29,6 +43,33 @@ function(_add_sharkfuser_target)
...
+  # Place executable in the build/bin sub-directory
+  set_target_properties(
+      ${_RULE_NAME} PROPERTIES
+      RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin/${_RULE_BIN_SUBDIR}
+  )
+endfunction()
+
+
+# Creates and executable target + test.
+function(_add_sharkfuser_test)
```
sjain-stanford, line 55: As this is mainly adding a ctest target, call this `_add_sharkfuser_ctest_target`?

sharkfuser/build_tools/cmake/CTestMacros.cmake
```diff
@@ -82,10 +139,84 @@ function(add_sharkfuser_sample)
...
+  )
+
+  # Apparently this flag helps FileCheck spit out nicer error messages.
+  set_tests_properties(${_TEST_NAME} PROPERTIES ENVIRONMENT
+    "FILECHECK_OPTS=--enable-var-scope")
+
+  # Dependencies for the test.
+  if(_RULE_TOOLS)
+    set_tests_properties(${_TEST_NAME} PROPERTIES DEPENDS "${_RULE_TOOLS}")
+  endif()
```
sjain-stanford, line 221: If not too much trouble, recommend moving these into a separate function to keep `add_sharkfuser_lit_test` easy to follow.

sharkfuser/tests/CMakeLists.txt
```diff
@@ -47,3 +47,8 @@ add_sharkfuser_test(
...
+	TOOLS FileCheck
```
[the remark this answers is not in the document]
AaronStGeorge: Personally I prefer the uniformity + explicitness of all tools being provided through `TOOLS`. But I don't feel strongly about it if you do.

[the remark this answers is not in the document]
sjain-stanford: Mainly worried about the repetition when we have ~100s of lit tests, each will need the same `TOOLS FileCheck` in its target definition. This makes sense for when there's a one-off tool that only some tests need.

sharkfuser/build_tools/cmake/CTestMacros.cmake
```diff
@@ -5,7 +5,21 @@
...
+# SRCS
+#  Source files to compile into the executable (required)
+#
+# DEPS
+#  Additional library dependencies beyond the standard ones
+#  (libfusili and Catch2::Catch2WithMain are always linked)
+#
+# BIN_SUBDIR
+#  Subdirectory under build/bin/ where the executable will be placed
+function(_add_sharkfuser_test_target)
```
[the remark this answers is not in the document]
AaronStGeorge: What do you think of `_add_sharkfuser_executable_for_test`? You wouldn't want the coverages stuff for a general executable.

sharkfuser/build_tools/cmake/CTestMacros.cmake
AaronStGeorge: Which parts would you recommend moving into a separate function? Too me all of this stuff in this function  hangs together pretty well.

sjain-stanford: I forget why I said this. I think I was trying to keep the same pattern across the user-facing macros, but let's leave it how you have it now.
```

- add_sharkfuser_test: parse cmake args, early return, invoke `_add_sharkfuser_ctest_target` with args
- add_sharkfuser_sample: parse cmake args, early return, invoke `_add_sharkfuser_ctest_target` with args
- add_sharkfuser_lit_test: parse cmake args, early return, invoke `_add_sharkfuser_lit_test_target` with args 

```

sharkfuser/build_tools/cmake/CTestMacros.cmake
```diff
@@ -82,10 +139,84 @@ function(add_sharkfuser_sample)
...
+
+  get_filename_component(_ABSOLUTE_RULE_SRC ${_RULE_SRC} ABSOLUTE)
+  get_filename_component(_TEST_NAME ${_RULE_SRC} NAME_WE)
+
+  # The executable who's output is being lit tested.
+  _add_sharkfuser_test_target(
+    NAME ${_TEST_NAME}
+    SRCS ${_RULE_SRC}
+    DEPS ${_RULE_DEPS}
+    BIN_SUBDIR samples
```
sjain-stanford, line 191: Should this be in a separate sub dir for the lit tests?

sharkfuser/tests/CMakeLists.txt
sjain-stanford: Let's just do this: https://sourcegraph.com/github.com/iree-org/iree/-/blob/build_tools/cmake/iree_lit_test.cmake?L102

PR #1882, opened 2025-07-18
Add an MLIR kernel for properly handling FP4 casting

KyleHerndon:
Requires this patch:
https://github.com/iree-org/iree/pull/21413

### DIFF
--- a/sharktank/tests/models/llama/sharded_llama_test.py
+++ b/sharktank/tests/models/llama/sharded_llama_test.py
@@ -459,7 +459,7 @@ class TestTensorParallelFp4QuantizedLlama:
     def testTensorShardFp4QuantizedLlama(self, deterministic_random_seed):
         """Tensor-shard a FP4 quantized Llama 3 model and roundtrip check it is the same."""
         dtype = torch.float32
-        quantization_block_size = 4
+        quantization_block_size = 32
         tensor_parallelism_size = 2
 
         sharded_config = toy_llama.make_config2(
@@ -491,7 +491,7 @@ class TestTensorParallelFp4QuantizedLlama:
             return res
 
         quantized_theta = quantize_theta_to_fp4(
-            theta, quantizer=DynamicFp4BlockQuantizer(block_size=4)
+            theta, quantizer=DynamicFp4BlockQuantizer(block_size=32)
         )
         dequantized_theta = quantized_theta.transform(unbox_transform)
         quantized_sharded_theta = shard_theta(quantized_theta, sharded_config)


KyleHerndon:
Failing checks are real but passed locally when built with the above IREE patch

sharktank/sharktank/types/ocp_floats.py
```diff
@@ -273,11 +273,48 @@ def fp4_e2m1_to_float32(fp4_indices: torch.Tensor) -> torch.Tensor:
...
+        indexing_maps = [
+          affine_map<(d0) -> (d0)>,
+          affine_map<(d0) -> (d0)>
+        ],
+        iterator_types = ["parallel"]
+      } ins(%x : !x) outs(%empty : !result) {
+      ^bb0(%in: f32, %out: i8):
+        %fp4_val = arith.truncf %in : f32 to f4E2M1FN
+        %i4_val = arith.bitcast %fp4_val : f4E2M1FN to i4
+        %result_i8 = arith.extui %i4_val : i4 to i8
```
Groverkss, line 305: Is this correct? Shouldn't we be packing the bits?

Groverkss (commented):
Looks good, I'm just not sure if that's how we want to do the truncf though. Maybe someone else knows better.

MaheshRavishankar (dismissed):
I dont think this in line with expectations. I think the fp4 input is expected to be packed, while this is outputing in unpacked manner. So its doubling the footprint of the result matrix.

sharktank/sharktank/types/ocp_floats.py
```diff
@@ -299,3 +297,94 @@ def float32_to_fp4_e2m1(values: torch.Tensor) -> torch.Tensor:
...
+      %quantized = linalg.generic {
+        indexing_maps = [
+          affine_map<(d0, d1) -> (d0, d1 * 2)>,
+          affine_map<(d0, d1) -> (d0, d1 * 2 + 1)>,
+          affine_map<(d0, d1) -> (d0)>,
+          affine_map<(d0, d1) -> (d0, d1)>
+        ],
+        iterator_types = ["parallel", "parallel"]
+      } ins(%values, %values, %scales : !values, !values, !result_scales) outs(%empty_packed : !result_quantized) {
+      ^bb0(%val_low: f32, %val_high: f32, %scale: i8, %out: i8):
```
krzysz00, line 367: This won't pattern-match right for the AMDGPU backend

To use device-side instructions we need:

1. A _vector_ scaling_truncf - that is, taking a `vector<2xf32>`
2. The scale argument should be a broadcast vector (so a `vector<2xf8E8M0FNU>`)
3. ... except that the hardware intrinsics actually take a `f32` value and do the conversion to f8E8M0FNU implicitly, so you don't need all the casting. (Except that the MFMAs do need the casting)

KyleHerndon: Fixed after extensive discussion elsewhere.

sharktank/sharktank/types/ocp_floats.py
```diff
@@ -299,3 +300,80 @@ def float32_to_fp4_e2m1(values: torch.Tensor) -> torch.Tensor:
...
+    inputs=(MLIRTensor[N, M32, I_DTYPE],),
+    results=(MLIRTensor[N, U8_DTYPE], MLIRTensor[N, M16, U8_DTYPE]),
+)
+def dynamic_quantize_to_fp4(values, result_scales=None, result_quantized=None):
+    mlir = """
+    module {
+    util.func private @{{kernel_name}}(%values : !values) -> (!result_scales, !result_quantized) {
+      %c0 = arith.constant 0 : index
+      %c1 = arith.constant 1 : index
+      %c4_f32 = arith.constant 0.25 : f32
```
krzysz00, line 327: Nit: name of constant is misleading

sharktank/sharktank/types/ocp_floats.py
```diff
@@ -299,3 +300,80 @@ def float32_to_fp4_e2m1(values: torch.Tensor) -> torch.Tensor:
...
+      %neg_inf = arith.constant -3.402823e+38 : f32
```
krzysz00, line 328: Nit: this isn't neg_inf? neg_inf is `0xff800000`

PR #1884, opened 2025-07-18
Expand .create() and .add_to_archive() to handle ShardedTensors with InferenceTensor shards

Alex-Vasile:
Current implementation assumed DefaultPrimitiveTensor shards, this needed to be expanded to work with general inference tensors to support QuantizedTensor shards.

### DIFF
--- a/sharktank/sharktank/types/tensors.py
+++ b/sharktank/sharktank/types/tensors.py
@@ -761,7 +761,13 @@ class DefaultPrimitiveTensor(PrimitiveTensor):
 
     def add_to_archive(self, builder: ShardedArchiveBuilder) -> InferenceTensorMetadata:
         """Adds this tensor to the global archive."""
-        builder.add_tensor(self.name, self._data)
+        if ".shard." in self.name:
+            name, rank = self.name.rsplit(".shard.", 1)
+            rank = int(rank)
+            builder.for_rank(rank).add_tensor(self.name, self._data)
+        else:
+            name = self.name
+            builder.add_tensor(self.name, self._data)
         return InferenceTensorMetadata(self.serialized_name(), {"": self.name})
 
     def _clone_with_subtensors(
@@ -930,10 +936,20 @@ class PlanarQuantizedTensor(QuantizedTensor):
         """Adds this tensor to the global archive."""
         root_name = self.name
         layout = self.unpack()
+
+        is_shard = ".shard." in root_name
+        if is_shard:
+            name, rank = root_name.rsplit(".shard.", 1)
+            rank = int(rank)
+
         name_map: dict[str, str] = {}
+
         for suffix, plane in layout.planes.items():
             irpa_name = f"{root_name}:{suffix}"
-            builder.add_tensor(irpa_name, plane)
+            if is_shard:
+                builder.for_rank(rank).add_tensor(irpa_name, plane)
+            else:
+                builder.add_tensor(irpa_name, plane)
             name_map[suffix] = irpa_name
         extra_properties = {
             "shape": [int(d) for d in self.shape],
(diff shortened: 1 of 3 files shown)


sharktank/sharktank/types/tensors.py
```diff
@@ -761,7 +761,13 @@ def subtensors(self) -> dict[str, torch.Tensor]:
...
     def add_to_archive(self, builder: ShardedArchiveBuilder) -> InferenceTensorMetadata:
         """Adds this tensor to the global archive."""
-        builder.add_tensor(self.name, self._data)
+        if ".shard." in self.name:
+            name, rank = self.name.rsplit(".shard.", 1)
+            rank = int(rank)
+            builder.for_rank(rank).add_tensor(self.name, self._data)
+        else:
+            name = self.name
+            builder.add_tensor(self.name, self._data)
```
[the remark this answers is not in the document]
Alex-Vasile: Okay, I have removed the dependence on the name and have changed how ShardedTensors call add_to_archive on their shards so the shards no longer need if-statements for sharding.

sharktank/sharktank/types/tensors.py
```diff
@@ -1098,17 +1116,28 @@ def serialized_name(cls) -> str:
...
     def add_to_archive(self, builder: ShardedArchiveBuilder) -> InferenceTensorMetadata:
-        for i, pt in enumerate(self._shards):
-            builder.for_rank(i).add_tensor(pt.name, pt._data)
+
         extra_properties = {
-            "shard_count": len(self._shards),
             "shape": list(self.shape),
+            "shard_count": len(self._shards),
+            "shard_type": self._shards[0].serialized_name(),
+            "extra_properties_shards": {},
```
sogartar, line 1124: `extra_properties_shards` seems too long as we are already in the context of extra properties.
Maybe just `shards` or `shard_properties` is OK.

sharktank/sharktank/types/tensors.py
```diff
@@ -1157,22 +1186,55 @@ def create(
...
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
+                raw_tensors_for_rank[int(k)][""] = v
+        shards = []
+        for i, raw_tensors in raw_tensors_for_rank.items():
+            shard = shard_clazz.create(
+                name=f"{name}.shard.{i}",
+                raw_tensors=raw_tensors,
+                extra_properties=extra_properties_shards.get(str(i), {}),
```
sogartar, line 1227: It seems this together with the `shard_type` default would give us backward compatibility to be able to load old archives.

sharktank/tests/types/dataset_test.py
```diff
@@ -143,6 +144,61 @@ def testDatasetRoundtrip(self):
...
+    def testRoundTripDefaultPrimitiveTensor(self):
+        t_orig = DefaultPrimitiveTensor(
+            name="primitive", data=torch.Tensor([1.0, 2.0, 3.0])
+        )
+        ds_orig = Dataset({}, Theta({t_orig.name: t_orig}))
+        ds_orig.save(self.temp_dir / "p_ds.irpa")
+
+        ds_load = Dataset.load(self.temp_dir / "p_ds.irpa", mmap=False)
+        t_load = ds_load.root_theta.tensor(t_orig.name)
+        print(t_load)
```
sogartar, line 156: Here and on other places better to log with a level to not pollute the console aways.

sharktank/sharktank/types/tensors.py
Alex-Vasile: Yes, I was going for that. All of the current tests that use existing IRPAs (e.g. the sharded perplexity tests run without issues).

sharktank/tests/types/dataset_test.py
Alex-Vasile: Done

sharktank/sharktank/types/tensors.py
```diff
@@ -1157,22 +1186,55 @@ def create(
...
-            try:
-                t = raw_tensors[t_name]
-                ts.append(t)
-                # TODO: this should be changed to tracked device affinity
-                DeviceTensorTrait(i).set(t)
-            except KeyError as e:
-                raise IOError(
-                    f"Missing component tensor '{t_name}' in {raw_tensors.keys()}"
-                ) from e
+        shard_type = extra_properties.get(
```
rsuderman, line 1189: Seeing the scale of this change all hardcoded feels wrong. You should be creating / expanding general functionality not hardcoding behavior through.

Look at how tensors are added / embedding. Find encapsulations for parts. There should be grouped together concepts (e.g. add_tensor should know how to handle subtypes). Same with assigning the device tensor traits. The realize is that a tensor should be able to have its placement set with each type handling it individually.

Overall I recommend enumerating the behaviors you need and land independently instead of straight shotting a single overarching implementation.

rsuderman (changes requested):

PR #1900, opened 2025-07-23
[sharktank] make fp4 block-quantized have scales with trailing singleton dimension

sogartar:
To align BlockScaledFp4Layout with the base class BlockScaledLayout make the per-block scale tensor have a shape blocked_shape + [1]. Fixes https://github.com/nod-ai/shark-ai/issues/1833.

This change breaks existing IRPA files that hold FP4 quantized tensors. They will need to be regenerated.

### DIFF
--- a/sharktank/sharktank/types/layouts.py
+++ b/sharktank/sharktank/types/layouts.py
@@ -598,6 +598,10 @@ class BlockScaledFp4Layout(BlockScaledPackedLayout):
         block_size: int = 32,
         use_fe8m0_scale: bool = True,
     ):
+        if len(qs.shape) == len(d.shape) + 1:
+            # Legacy scale format with no trailing singleton dimension to match qs.
+            # This is here to avoid breaking existing IRPA files.
+            d = d.unsqueeze(-1)
         assert iterables_equal(qs.shape[:-1], d.shape[:-1])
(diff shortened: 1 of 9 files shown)


Alex-Vasile (commented):
Looks good to me, I think we should hold off until we coordinate with others because it will break IRPAs

sharktank/sharktank/types/layouts.py
```diff
@@ -598,14 +598,14 @@ def __init__(
         block_size: int = 32,
         use_fe8m0_scale: bool = True,
     ):
-        assert iterables_equal(
-            qs.shape[:-1], d.shape
-        ), "TODO: remove when this class is refactored to comply with BlockScaledLayout"
+        if len(qs.shape) == len(d.shape) + 1:
+            # Legacy scale format with no trailing singleton dimension to match qs.
+            # This is here to avoid breaking existing IRPA files.
+            d = d.unsqueeze(-1)
```
Alex-Vasile, line 604: Can you put a warning in the console or the logger to let people know they are using an irpa format that they should regenerate. Gives them warning ahead of time because we don't want the extra code to stick around forever.

sogartar: I added a warning.

PR #1902, opened 2025-07-24
[sharktank] propagate transpose_rhs into the shards on sharded matmumls

sogartar:
The FP4 block quantized tensor does not support transposition of the blocked dimension and it is not clear how to support this. This forces us to not use transposition to simplify matmuls, but to propagate the bool arg transpose_rhs to the multiplication of the shards.

Rework the tests to cover more cases.


dan-garvey (approved):
LGTM, but I don't see anything in your code specific to FP4? Does this limit functionality for sharded tensors that aren't fp4?

PR #1906, opened 2025-07-24
[fusili] MLIR Assembly Emitter for Fusili Graph with ConvFProp Node

sjain-stanford:
This PR is the first stab at a **textual MLIR ASM emitter** - this forms the bridge between Fusili graph frontend and the IREE compiler.

At a high-level, we recurse through the nodes in a Fusili graph to programmatically:
- generate `func.func` signature for the main graph (populating operand names, types, result types from input and output tensors)
- generate asm for each subNode (in this case conv_fprop):
  - generate list construct ops for constant attributes to conv_fprop (stride, padding, dilation) preserving SSA naming for values
  - generate conv_fprop op taking in various ssa values as inputs, return types
  - no further subNodes (but can be for composite ops)
- generate return signatures for main graph

It is done in a DFS allowing for composite ops to expand into their own regions (containing sub ops) and general future extensibility. 

We use pre-defined op schemas with basic multi-line `R"(...)"` strings and `std::format` (from C++20) for template materializations.

Here's the emitted MLIR assembly for reference: https://gist.github.com/sjain-stanford/bb5975a32f66683210a9fc43e63a08e5

Reviewers: Bulk of the asm emitter is colocated at `asm_emitter.h`. Rest includes interfaces, tests, and docstrings.

Thanks @MaheshRavishankar , @AaronStGeorge , @zjgarvey and @stellaraccident for the insightful discussions leading up to this.


sharkfuser/include/fusili/asm_emitter.h
```diff
@@ -0,0 +1,420 @@
...
+    if (i > 0)
+      oss << ", ";
+    oss << ssaValueNames[i];
+  }
+  oss << " : (";
+  for (size_t i = 0; i < ssaValueNames.size(); ++i) {
+    if (i > 0)
+      oss << ", ";
+    oss << "!torch.int";
+  }
```
AaronStGeorge, line 103: You might want a utility function similar to LLVM's [interleave](https://github.com/llvm/llvm-project/blob/028f70d2524be56bb5d191ec3d7022b40bc4859b/llvm/include/llvm/ADT/STLExtras.h#L2167-L2192).

sharkfuser/include/fusili/asm_emitter.h
```diff
@@ -0,0 +1,420 @@
...
+// For example if `getListOfIntOpsAsm` is called on these inputs:
+//    listOfInts: {1, 2}
+//    prefix: "stride"
+//    suffix: "conv"
+//
+// It generates the following MLIR assembly:
+//
+//   %stride_val_0_conv = torch.constant.int 1
+//   %stride_val_1_conv = torch.constant.int 2
+//   %stride_conv = torch.prim.ListConstruct
```
AaronStGeorge, line 69: We had discussed strategies for removing potential name conflicts, unless I'm missing something this doesn't include anything for that? Have you decided on a strategy?

sharkfuser/include/fusili/asm_emitter.h
```diff
@@ -0,0 +1,420 @@
...
+// `foo_Bar::X0` would become `%foo_BarX0`
+inline std::string TensorAttr::getMlirSSAValueNameAsm() const {
+  assert(!getName().empty() &&
+         "TensorAttr name must not be empty for `getMlirSSAValueNameAsm`");
+
+  std::string filtered;
+  for (char c : getName()) {
+    if (std::isalnum(static_cast<unsigned char>(c)) || c == '_')
+      filtered += c;
+  }
```
AaronStGeorge, line 167:
```suggestion
  std::string filtered;
  filtered.erase(remove_if(filtered.begin(), filtered.end(), [](char c) {
    return !isalnum(c) && c != '_';
  }), filtered.end());
```
If you feel like channeling Sean Parent 🤷‍♂️

AaronStGeorge (approved):
💯 great to see everything taking shape!

sharkfuser/include/fusili/asm_emitter.h
sjain-stanford: A few lines down there's a comment about that:
```
// The prefix is generally what attribute this refers to (e.g.
// padding, stride, dilation etc.) and the suffix is the node's
// unique name (for SSA disambiguation).
```
With this setup we basically only need to ensure the node attributes use unique names - by default this should be the case ([because](https://sourcegraph.com/github.com/nod-ai/shark-ai@d470b828374f121483a16589026cd263306f38fe/-/blob/sharkfuser/include/fusili/graph.h?L158)), but in the rare case when user inadvertently provides say two nodes with the same name, we will weed that out in the `Graph::validate()` call.

sharkfuser/tests/lit/test_lit_example.cpp
```diff
@@ -1,3 +1,9 @@
+// Copyright 2025 Advanced Micro Devices, Inc.
```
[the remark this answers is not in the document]
sjain-stanford: Good point, might as well do it now.

sharkfuser/include/fusili/asm_emitter.h
sjain-stanford: I had to squint at this, but isn't `filtered` an empty string to begin with, so what are we erasing?

Did you mean something like this?
```cpp
std::string filtered = getName();
filtered.erase(
    std::remove_if(filtered.begin(), filtered.end(),
        [](char c) {
            return !(std::isalnum(c) || c == '_');
        }
    ),
    filtered.end()
);
```

I didn't worry much about efficiency (wrt string reallocations) since we're only dealing with relatively small names here so went with code that was easy to follow, but happy to adopt this here.

sharkfuser/include/fusili/asm_emitter.h
sjain-stanford: Nice! Done.

sjain-stanford: I went ahead and added an overload of `interleave` as well, which accepts a `SkipFunctor` to conditionally skip elements based on a predicate. This is useful for cases where I needed to check `input->isScalar()` or `output->isVirtual()` and skip those from the result.

sharkfuser/tests/lit/test_conv_asm_emitter.cpp
```diff
@@ -0,0 +1,72 @@
...
+  // CHECK:       %padding_conv_fprop = torch.prim.ListConstruct %padding_val_0_conv_fprop, %padding_val_1_conv_fprop : (!torch.int, !torch.int) -> !torch.list<int>
+  // CHECK:       %dilation_val_0_conv_fprop = torch.constant.int 1
+  // CHECK:       %dilation_val_1_conv_fprop = torch.constant.int 1
+  // CHECK:       %dilation_conv_fprop = torch.prim.ListConstruct %dilation_val_0_conv_fprop, %dilation_val_1_conv_fprop : (!torch.int, !torch.int) -> !torch.list<int>
+  // CHECK:       %transposed_conv_fprop = torch.constant.bool false
+  // CHECK:       %output_padding_conv_fprop = torch.prim.ListConstruct  : () -> !torch.list<int>
+  // CHECK:       %groups_conv_fprop = torch.constant.int 1
+  // CHECK:       %result = torch.aten.convolution %arg0_image, %arg1_filter, %bias_conv_fprop, %stride_conv_fprop, %padding_conv_fprop, %dilation_conv_fprop, %transposed_conv_fprop, %output_padding_conv_fprop, %groups_conv_fprop : !torch.vtensor<[16,128,64,64],f32>, !torch.vtensor<[256,128,1,1],f32>, !torch.none, !torch.list<int>, !torch.list<int>, !torch.list<int>, !torch.bool, !torch.list<int>, !torch.int -> !torch.vtensor<[16,256,64,64],f32>
```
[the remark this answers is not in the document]
sjain-stanford: We already auto-populate a sensible name (e.g. from the producing op as you suggest). This is done when the op and output tensors are being built ([here](https://sourcegraph.com/github.com/nod-ai/shark-ai@cefd8a91bd75a9fca87aa8dc4b9df3efbfd9be0b/-/blob/sharkfuser/include/fusili/graph.h?L195)). 

However, in some cases the user might want to override it and explicitly specify tensor names, especially for main inputs/outputs. In this case the test explicitly names the output tensor `result`:
```
Y->setName("result")...
```
This overrides the default name. If you remove the `->setName("result")` from the test, this is what we get:

```
%conv_fprop_Y = torch.aten.convolution ...

return %conv_fprop_Y : ...
```

PR #1907, opened 2025-07-24
[Fusili] Use pre-built IREE rather than pip dependency

AaronStGeorge:
When generating a .vmfb we'll need to store it in a type that IREE runtime expects; I believe that's `iree_const_byte_span_t`. We'll need IREE cmake config + headers to able to use said type in sharkfuser. In preparation for .vmfb generation, this PR moves towards pulling in IREE tools in away that allows us to use cmake + headers + pre-built `.so`s.


.github/workflows/ci-sharkfuser.yml
```diff
@@ -90,17 +91,27 @@ jobs:
...
+        tar -xf iree-dist.tar.xz -C iree-dist
+        rm iree-dist.tar.xz
+        echo "$SHARKFUSER_DIR/iree-dist/bin" >> $GITHUB_PATH
+
     - name: Build sharkfuser
       working-directory: ${{ env.SHARKFUSER_DIR }}
       run: |
         source .venv/bin/activate
         cmake -GNinja -S. -Bbuild \
+          -DIREE_DIR=$SHARKFUSER_DIR/iree-dist/lib/cmake \
```
sjain-stanford, line 114: Is this really needed? I see an `INSTALL` target, can we use that to have it installed (system-wide) and used in Fusili through `find_package` / `find_program`? 

https://sourcegraph.com/github.com/iree-org/iree/-/blob/build_tools/cmake/CMakeLists.txt?L28

AaronStGeorge: If I'm not mistaken we'll need `-DIREE_DIR` here unless we move the downloaded package into system locations where `find_package` would look by default.  IIUC that `install` command will decide what's in the bundle we're downloading, I don't think we have a cmake install command available in what we get. But, if we add
```sh
sudo mv iree-dist/* /usr/local/
```
to the install block `find_package` should work without this `-D` variable (I think [CMAKE_SYSTEM_PREFIX_PATH docs](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PREFIX_PATH.html). Would you prefer that?

.github/workflows/ci-sharkfuser.yml
```diff
@@ -90,12 +91,22 @@ jobs:
...
         pip install -r ./test_requirements.txt
 
+    - name: Install prebuilt iree
+      working-directory: ${{ env.SHARKFUSER_DIR }}
+      run: |
+        wget --no-verbose -O iree-dist.tar.xz $IREE_RELEASE_URL
+        mkdir iree-dist
+        tar -xf iree-dist.tar.xz -C iree-dist
+        rm iree-dist.tar.xz
+        sudo mv iree-dist/* /usr/local/
```
sjain-stanford, line 107: Will this work? `iree-dist` when unpacked has its own `bin`, `include`, and `lib` dirs.

AaronStGeorge: No this doesn't work it tries to overwrite the files which isn't correct.

sjain-stanford:
Let's just have it unpacked in a `./third_party/` directory inside sharkfuser (have it gitignored etc) and then use `-DCMAKE_PREFIX_PATH`. Much better than having to fiddle with system installation. This is also nice because we might want to just swap the CMAKE_PREFIX_PATH to a local source checkout for picking bugfixes etc.

sjain-stanford (approved):

.github/workflows/ci-sharkfuser.yml
```diff
@@ -90,12 +91,22 @@ jobs:
...
         pip install -r ./test_requirements.txt
 
+    - name: Install prebuilt iree
+      working-directory: ${{ env.SHARKFUSER_DIR }}
+      run: |
+        wget --no-verbose -O iree-dist.tar.xz $IREE_RELEASE_URL
+        mkdir iree-dist
+        tar -xf iree-dist.tar.xz -C iree-dist
+        rm iree-dist.tar.xz
+        sudo mv iree-dist /usr/local/stow
```
sjain-stanford, line 107: Might as well untar at the location you want them:
```
    mkdir -p /usr/local/stow/iree-dist
    tar -xf /tmp/iree-dist.tar.xz -C /usr/local/stow/iree-dist \
    rm /tmp/iree-dist.tar.xz \
    stow -d /usr/local/stow -t /usr/local iree-dist
```

sharkfuser/test_requirements.txt
```diff
@@ -1,3 +1,2 @@
 lit
 filecheck
```
sjain-stanford, line 2: Forgot to remove this in the patch I shared earlier. Here's the revised one: 
[filecheck.patch](https://github.com/user-attachments/files/21435975/filecheck.patch)

sharkfuser/test_requirements.txt
```diff
@@ -1,3 +1,2 @@
...
-iree-base-compiler
```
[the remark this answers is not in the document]
sjain-stanford: This section in particular needs updating:

**Test Requirements:**
- catch2
- lit
- FileCheck
- iree-opt

Easiest way to get [`lit`](https://llvm.org/docs/CommandGuide/lit.html), [`filecheck`](https://github.com/AntonLydike/filecheck), and `iree-opt` without depending on LLVM or building IREE from source is through Python (pip install). One may either use system Python or create a virtual environment (preferred)
like so:
```shell
python -m venv --prompt fusili .venv
source .venv/bin/activate
python -m pip install --upgrade pip
pip install -r ./test_requirements.txt
```

sjain-stanford:
We could do the following to have a default without needing `-DCMAKE_PREFIX_PATH` provided externally. It also allows `SHARKFUSER_IREE_CMAKE_DIR to be optionally specified if the user wants it picked at a different path:
```
set(SHARKFUSER_IREE_CMAKE_DIR "/usr/local/lib/cmake/IREE" CACHE PATH "Path to IREE CMake directory")
list(APPEND CMAKE_PREFIX_PATH ${SHARKFUSER_IREE_CMAKE_DIR})
```

sharkfuser/test_requirements.txt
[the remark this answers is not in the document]
AaronStGeorge: I've updated the readme in a more recent commit, let me know what you think!

Issue #1769, opened 2025-07-07
[sharktuner] Add Support for Attention Op


bangtianliu:
To support attention op in the sharktuner, we need to make several changes: 

- **Constraint generation for attention to generate config**
   - parser for attention op: https://github.com/nod-ai/shark-ai/pull/1733
       - This relies on Python bindings added in IREE to extract root op informatio: https://github.com/iree-org/iree/pull/21170 and https://github.com/iree-org/iree/pull/21216
   - Defining the search space and strategies for generating decomposition control configs specific to attention ops. 
https://github.com/nod-ai/shark-ai/pull/1772

  The constraints for selecting MMA instructions for these matmuls are determined by layout matching logic: https://github.com/iree-org/iree/blob/9e8691fdab06ee5c432bbf40ea70883b4b78d8bf/compiler/src/iree/compiler/Codegen/LLVMGPU/LLVMGPUConfigureTensorLayouts.cpp#L414-L422. 
So here are some todos after meeting with kunwar:
      1. IREE Python binding PR: Expose Python bindings for virtual MMA, which is currently missing and needed by the tuner. https://github.com/iree-org/iree/pull/21403
      2. Tuner PR: Understand and incorporate layout constraints when selecting MMAs in the attention tuner.
 Required python binding to get single subgroup layout: https://github.com/iree-org/iree/pull/21454
The updated constraints taking layout into consideration: https://github.com/nod-ai/shark-ai/pull/1909
The logic is original from: https://github.com/iree-org/iree/pull/21448. 
      4. Tuner PR: Extend the search space to include virtual MMA options for all ops, including matmul and conv.Here are some steps for generating correct td specs for attention op
- **TD spec generation for attention ops**
   - Extended constraint generators to support a shared config list mechanism across contraction, convolution, and attention ops: https://github.com/nod-ai/shark-ai/pull/1641
   - Added support to build TD specs from these config lists: https://github.com/nod-ai/shark-ai/pull/1602

bangtianliu:
close as all the tasks have been done.

PR #1909, opened 2025-07-25
[tuner] improve support for attention op

bangtianliu:
This PR enhances the tuner’s support for the Attention operation by building on the constraint generation logic (mainly about layout matching, searching two mmas independently) introduced in https://github.com/iree-org/iree/pull/21448. 
- Adds the layout matching constraints based on the python binding exposed from https://github.com/iree-org/iree/pull/21454
- Extends the tuner to handle virtual MMA intrinsics according to python binding exposed from https://github.com/iree-org/iree/pull/21403

For reference, the attached log link demonstrates tuning runs for the Attention op using this PR.: https://gist.github.com/bangtianliu/ce6b925a679a2072b2ceba1bfdfe094c. 
Issue: #1769

### DIFF
--- a/sharktuner/tests/dispatch_constraints_test.py
+++ b/sharktuner/tests/dispatch_constraints_test.py
@@ -359,6 +359,67 @@ def test_is_valid_mma_schedule(
     assert solver.check() == z3.unsat
 
 
+def test_get_mfma_intrinsic_constraints(
+    tuner_ctx: common.TunerContext,
+) -> None:
+    lhs_type = common.ShapedType([16, 16], tuner_ctx.type.f16)
+    rhs_type = common.ShapedType([16, 16], tuner_ctx.type.f16)
+    res_type = common.ShapedType([16, 16], tuner_ctx.type.f32)
+
+    intrinsic_m = z3.Int("intrinsic_m")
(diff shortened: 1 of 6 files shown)


sharktuner/tests/dispatch_constraints_test.py
```diff
@@ -359,6 +359,67 @@ def test_is_valid_mma_schedule(
...
+
+    solver = z3.Solver()
+    solver.add(constraints)
+    assert solver.check() == z3.sat
+
+    model = solver.model()
+    m_val = model[intrinsic_m].as_long()
+    n_val = model[intrinsic_n].as_long()
+    k_val = model[intrinsic_k].as_long()
+    assert (m_val, n_val, k_val) in [(16, 16, 16), (32, 32, 8)]
```
Max191, line 395: I don't fully understand why the result is expected to be one of these options. Why are we able to select `(32, 32, 8)` and why are we not able to select `(32, 32, 16)`? (mostly just for my own understanding)

Max191 (commented):
This overall LGTM, although I'm not an expert on the constraints here. Would be good to get approval from @Groverkss as well.

Groverkss (approved):
Thank you!

PR #1910, opened 2025-07-25
Adds missing float8_e4m3fn type mappings

IanNod:


sharktank/sharktank/types/tensors.py
```diff
@@ -1698,6 +1698,7 @@ def serialized_short_name_to_dtype(dtype_name: str) -> torch.dtype:
     "int64": torch.int64,
     "bool": torch.bool,
     "float8_e4m3fnuz": torch.float8_e4m3fnuz,
+    "float8_e4m3fn": torch.float8_e4m3fnuz,
```
archana-ramalingam, line 1701: Is this wrongly mapped?

PR #1923, opened 2025-07-28
Ensure error message is properly logged when `kvcache OOM` occurs

stbaione:
Ensure KVCache OOM (page exhaustion) failure is properly logged when it occurs server-side.

Will do a follow up to close the client connection, instead of just leaving it hanging.

For now, this will at least make the error obvious server-side.

Output example:

```bash
[2025-07-28 15:21:47.187] [error] [base_attention_cache.py:182] FATAL CacheAllocationFailure: Failed to allocate 6 pages from `PagePool`.
Required pages: 6, Available pages: 2, Total pages: 2
Consider re-exporting the model with a higher `--device-block-count` value.
```


amd-vivekag (approved):
Error message from server is definitely better than current behavior.

PR #1928, opened 2025-07-28
Remove write_range

rsuderman:
Write range was intended for offset prefill however this should just be expanded into the behavior of `write`. Removing to trip down the implementation of kv cache for cleanup.


IanNod (approved):
lgtm

PR #1929, opened 2025-07-29
Cleanup ops.attention overrides for uniform behavior

rsuderman:
Attention overrides should not be special casing arguments.

### DIFF
--- a/sharktank/sharktank/layers/paged_attention.py
+++ b/sharktank/sharktank/layers/paged_attention.py
@@ -1060,7 +1060,7 @@ class PagedAttention:
         elif attention_kernel == "sharktank":
             if mask is not None:
                 attn_output = ops.attention_impls.masked_flash_attention(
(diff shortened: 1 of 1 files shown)


sharktank/sharktank/ops/attention_impls.py
```diff
@@ -38,7 +38,13 @@ def _extract_linear_scale(t):
     return unbox_tensor(t), None
 
 
-def masked_flash_attention(q, k, v, a):
+def masked_flash_attention(q, k, v, a, is_causal, scale):
+    if is_causal:
+        return NotImplemented
+
+    if scale is None:
+        return NotImplemented
```
archana-ramalingam, line 46: If the `scale` is set to `None` in `sharktank/sharktank/layers/paged_attention.py`, will this not always return `NotImplemented`? 
This is breaking ppl tests for fp8. https://github.com/nod-ai/shark-ai/pull/1937 fixes it.

PR #1936, opened 2025-07-30
[Fusilli] Generate compiled artifacts from Fusilli Graph

AaronStGeorge:
This PR updates `Fusilli` to emit VMFB for a graph. Note that the `std::system` approach taken to running `iree-compile` is non-optimal. `std::system` doesn't provide a way to capture `stderr` and `stdout` so potential "logs" will end up on the users terminal rather than in our log file. From what I can tell the C++ std lib doesn't provide anything more full featured (similar to python's `subprocess`) and we need to roll our own using subprocesses. As this work is in the way of runtime integration, I've created https://github.com/nod-ai/shark-ai/issues/1934 to address this as a follow up.

### DIFF
--- a/sharkfuser/include/fusilli/extras.h
+++ b/sharkfuser/include/fusilli/extras.h
@@ -95,11 +95,40 @@ public:
     return ok(CacheFile(path, false));
   }
 
+  // Move constructors
   CacheFile(CacheFile &&other) noexcept
       : path(std::move(other.path)), remove_(other.remove_) {
     other.path.clear();
     other.remove_ = false;
   }
+  CacheFile &operator=(CacheFile &&other) noexcept {
+    if (this == &other) {
+      return *this;
+    }
+
+    // If ownership of the cache file is simply changing, we aren't creating a
+    // dangling resource that might to be removed.
+    bool samePath = path == other.path;
+
+    // Remove current resource if needed
+    if (remove_ && !path.empty() && !samePath) {
+      std::filesystem::remove(path);
+    }
+
+    // Move from other
+    path = std::move(other.path);
+    remove_ = other.remove_;
+    other.path.clear();
+    other.remove_ = false;
+
+    return *this;
+  }
+
+  // Delete copy constructors. A copy constructor would likely not be safe, as
+  // the destructor for a copy could remove the underlying file while the
+  // original is still expecting it to exist.
+  CacheFile(const CacheFile &) = delete;
+  CacheFile &operator=(const CacheFile &) = delete;
 
   ~CacheFile() {
     if (remove_ && !path.empty()) {
@@ -109,13 +138,6 @@ public:
 
   // Path of file this class wraps.
   std::filesystem::path path;
-
-  // Delete copy constructor + assignment operators. A copy constructor would
-  // likely not be safe, as the destructor for a copy could remove the
-  // underlying file while the original is still expecting it to exist.
-  CacheFile(const CacheFile &) = delete;
-  CacheFile &operator=(const CacheFile &) = delete;
-  CacheFile &operator=(CacheFile &&other) noexcept = delete;
 
   // Write to cache file.
   ErrorObject write(const std::string &content) {
--- a/sharkfuser/include/fusilli/extras.h
+++ b/sharkfuser/include/fusilli/extras.h
@@ -106,7 +106,7 @@ public:
       return *this;
     }
 
-    // If ownership of the cache file is simply changing, we aren't creating a
+    // If ownership of the cached file is simply changing, we aren't creating a
     // dangling resource that might to be removed.
     bool samePath = path == other.path;
 
--- a/sharkfuser/include/fusilli/extras.h
+++ b/sharkfuser/include/fusilli/extras.h
@@ -4,8 +4,14 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-#ifndef FUSILLI_UTILS_H
-#define FUSILLI_UTILS_H
+//===----------------------------------------------------------------------===//
+//
+// This file contains utility methods and classes.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef FUSILLI_EXTRAS_H
+#define FUSILLI_EXTRAS_H
 
 #include "fusilli/logging.h"
 
@@ -53,12 +59,41 @@ public:
   // Factory constructor that creates file, overwriting an existing file, and
   // returns an ErrorObject if file could not be created.
   static ErrorOr<CacheFile> create(const std::string &graphName,
-                                   const std::string &filename, bool remove);
+                                   const std::string &fileName, bool remove) {
+    std::filesystem::path path = CacheFile::getPath(graphName, fileName);
+    FUSILLI_LOG_LABEL_ENDL("Creating Cache file");
+    FUSILLI_LOG_ENDL(path);
+
+    // Create directory $HOME/.cache/fusilli/<graphName>
+    std::filesystem::path cacheDir = path.parent_path();
+    std::error_code ec;
+    std::filesystem::create_directories(cacheDir, ec);
+    FUSILLI_RETURN_ERROR_IF(ec, ErrorCode::FileSystemFailure,
+                            "Failed to create cache directory: " +
+                                cacheDir.string() + " - " + ec.message());
+
+    // Create file $HOME/.cache/fusilli/<graphName>/<fileName>
+    std::ofstream file(path);
+    FUSILLI_RETURN_ERROR_IF(!file.is_open(), ErrorCode::FileSystemFailure,
+                            "Failed to create file: " + path.string());
+    file.close();
+
+    return ok(CacheFile(path, remove));
+  }
 
   // Factory constructor that opens an existing file and returns ErrorObject if
   // the file does not exist.
   static ErrorOr<CacheFile> open(const std::string &graphName,
-                                 const std::string &filename);
+                                 const std::string &fileName) {
+    std::filesystem::path path = CacheFile::getPath(graphName, fileName);
+
+    // Check if the file exists.
+    FUSILLI_RETURN_ERROR_IF(!std::filesystem::exists(path),
+                            ErrorCode::FileSystemFailure,
+                            "File does not exist: " + path.string());
+
+    return ok(CacheFile(path, false));
+  }
 
   CacheFile(CacheFile &&other) noexcept
       : path(std::move(other.path)), remove_(other.remove_) {
@@ -119,7 +154,26 @@ public:
   //
   // Format: $HOME/.cache/fusilli/<sanitized version of graphName>/<fileName>
   static std::filesystem::path getPath(const std::string &graphName,
-                                       const std::string &fileName);
+                                       const std::string &fileName) {
+    // Ensure graphName is safe to use as a directory name, we assume fileName
+    // is safe.
+    std::string sanitizedGraphName = graphName;
+    std::transform(sanitizedGraphName.begin(), sanitizedGraphName.end(),
+                   sanitizedGraphName.begin(),
+                   [](char c) { return c == ' ' ? '_' : c; });
+    std::erase_if(sanitizedGraphName, [](unsigned char c) {
+      return !(std::isalnum(c) || c == '_');
+    });
+
+    // Ensure graphName has a value.
+    if (sanitizedGraphName.empty()) {
+      sanitizedGraphName = "unnamed_graph";
+    }
+
+    const char *homeDir = std::getenv("HOME");
+    return std::filesystem::path(homeDir) / ".cache" / "fusilli" /
+           sanitizedGraphName / fileName;
+  }
 
 private:
   // Class should be constructed using one of the factory functions.
@@ -129,63 +183,6 @@ private:
   // Whether to remove the file on destruction or not.
   bool remove_;
 };
-
-inline ErrorOr<CacheFile> CacheFile::create(const std::string &graphName,
-                                            const std::string &fileName,
-                                            bool remove) {
-  std::filesystem::path path = CacheFile::getPath(graphName, fileName);
-  FUSILLI_LOG_LABEL_ENDL("Createing Cache file");
-  FUSILLI_LOG_ENDL(path);
-
-  // Create directory $HOME/.cache/fusilli/<graphName>
-  std::filesystem::path cacheDir = path.parent_path();
-  std::error_code ec;
-  std::filesystem::create_directories(cacheDir, ec);
-  FUSILLI_RETURN_ERROR_IF(ec, ErrorCode::FileSystemFailure,
-                          "Failed to create cache directory: " +
-                              cacheDir.string() + " - " + ec.message());
-
-  // Create file $HOME/.cache/fusilli/<graphName>/<fileName>
-  std::ofstream file(path);
-  FUSILLI_RETURN_ERROR_IF(!file.is_open(), ErrorCode::FileSystemFailure,
-                          "Failed to create file: " + path.string());
-  file.close();
-
-  return ok(CacheFile(path, remove));
-}
-
-inline ErrorOr<CacheFile> CacheFile::open(const std::string &graphName,
-                                          const std::string &filename) {
-  std::filesystem::path path = CacheFile::getPath(graphName, filename);
-
-  // Check if the file exists.
-  FUSILLI_RETURN_ERROR_IF(!std::filesystem::exists(path),
-                          ErrorCode::FileSystemFailure,
-                          "File does not exist: " + path.string());
-
-  return ok(CacheFile(path, false));
-}
-
-inline std::filesystem::path CacheFile::getPath(const std::string &graphName,
-                                                const std::string &fileName) {
-  // Ensure graphName is safe to use as a directory name, we assume fileName
-  // is safe.
-  std::string sanitizedGraphName = graphName;
-  std::transform(sanitizedGraphName.begin(), sanitizedGraphName.end(),
-                 sanitizedGraphName.begin(),
-                 [](char c) { return c == ' ' ? '_' : c; });
-  std::erase_if(sanitizedGraphName,
-                [](unsigned char c) { return !(std::isalnum(c) || c == '_'); });
-
-  // Ensure graphName has a value.
-  if (sanitizedGraphName.empty()) {
-    sanitizedGraphName = "unnamed_graph";
-  }
-
-  const char *homeDir = std::getenv("HOME");
-  return std::filesystem::path(homeDir) / ".cache" / "fusilli" /
-         sanitizedGraphName / fileName;
-}
 
 // An STL-style algorithm similar to std::for_each that applies a second
 // functor between every pair of elements.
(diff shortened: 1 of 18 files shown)


sharkfuser/include/fusili/graph.h
```diff
@@ -143,6 +149,39 @@ Graph::convFProp(const std::shared_ptr<TensorAttr> &x,
   return y;
 }
 
+inline ErrorOr<std::string> Graph::emitVMFB() {
+  FUSILI_LOG_LABEL_ENDL("INFO: Emitting VMFB for graph");
+
+  // Write input asm to temp file.
+  TempFile input = TRY(TempFile::create("iree_compile_input"));
+  FUSILI_CHECK_ERROR(input.write(this->emitAsm()));
```
AaronStGeorge, line 157: Should `emitAsm()` return an `ErrorOr<T>` type when the graph fails to validate?

AaronStGeorge, line 157: @sjain-stanford  should `emitAsm()` return an `ErrorOr<T>` type when the graph fails to validate?

sharkfuser/include/fusili/utils.h
```diff
@@ -0,0 +1,173 @@
...
+      if (!first)
+        between_fn();
+      first = false;
+      each_fn(*begin);
+    }
+  }
+}
+
+} // namespace fusili
+#endif // FUSILI_UTILS_H
```
AaronStGeorge, line 173: Just moved

sharkfuser/include/fusili/utils.h
```diff
@@ -0,0 +1,173 @@
...
+  std::string tmpFile = tempPath.string();
+  int fileDescriptor = mkstemp(tmpFile.data());
+  FUSILI_RETURN_ERROR_IF(fileDescriptor == -1, ErrorCode::FileSystem,
+                         "failed to create temp file");
+  close(fileDescriptor);
+
+  return ok(TempFile(tmpFile, remove));
+}
+
+// An STL-style algorithm similar to std::for_each that applies a second
```
AaronStGeorge, line 119: `interleave` functions moved from `asm_emitter.h`.

sharkfuser/include/fusili/utils.h
```diff
@@ -0,0 +1,173 @@
...
+#define FUSILI_UTILS_H
+
+#include "fusili/logging.h"
+
+#include <filesystem>
+#include <format>
+#include <fstream>
+#include <iostream>
+#include <string>
+#include <unistd.h>
```
sjain-stanford, line 17: Is this header needed? (I think it is for fork/exec but probably not for now)

sharkfuser/include/fusili/graph.h
```diff
@@ -143,6 +149,39 @@ Graph::convFProp(const std::shared_ptr<TensorAttr> &x,
   return y;
 }
 
+inline ErrorOr<std::string> Graph::emitVMFB() {
```
sjain-stanford, line 152: Please add lit/integration tests for this?

sharkfuser/include/fusili/graph.h
```diff
@@ -143,6 +149,39 @@ Graph::convFProp(const std::shared_ptr<TensorAttr> &x,
...
+  FUSILI_LOG_LABEL_ENDL("INFO: Emitting VMFB for graph");
+
+  // Write input asm to temp file.
+  TempFile input = FUSILI_TRY(TempFile::create("iree_compile_input"));
```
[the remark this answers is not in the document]
AaronStGeorge: > Have you considered just having a compilation_artifacts dir in the same source tree (under ./sharkfuser) to which we dump target-specific folders (e.g. one for each unique graph) that contains all the intermediate outputs like the .mlir, .vmfb etc. That way, a check for "cache hit" could be just checking if the .mlir from g->emitAsm() matches what's dumped, if so, reuse the vmfb (or something simple along those lines)...

If we want to treat these compilation artifacts as a persistent cache, we’d probably need a bit of extra metadata for invalidation. We could probably store the extra information as comments in the .mlir file.

- Version number or sha of the binary of `iree-compile`
- Build sha of sharkfuser 
- Compile command used to generate the .vmfb

Depending on the size of vmfb + mlir we're expecting, we might consider an in memory cache? We could store the generated vmfb on a graph instance and have a dirty bit that's set anytime someone modifies the graph, or alternatively a static map where we could check re-generated MLIR (potentially generated by different graph instances) against already compiled vmfb instances. We could memory map the vmfb as well if we're worried about physical memory use (we probably want files in a `.sharkfuser` directory if we're memory mapping things).

> Are we sure TempFile is the right interface for this? We might want these to persist more often than not (especially with re-invocations over multiple batches).

How are you envisioning this piece to fit into the overall design? I wasn't really thinking of `TempFile` as part of the the interface. Does the caching logic use the `emitVMFB` function, or are the `TempFiles` themselves _the_ cache? I was thinking of the `TempFile`s as an internal implementation detail, any caching logic would sit "above" `emitVMFB`; or maybe "beside" if its a static map holding some memoized results.

Would you have a few minutes to chat today?

sharkfuser/include/fusili/graph.h
AaronStGeorge: What are you thinking here? I added a test to [sharkfuser/samples/convolution/conv_fprop.cpp](https://github.com/nod-ai/shark-ai/pull/1936/files#diff-eafc10905e11974cc5b33d82d295e9db72b3962bb25991d84efd36236e960193).

I was thinking the emit .vmfb logic would mainly be tested through e2e integration tests (that we can't write until the runtime is hooked up). Lit tests checking the exact vmfb would break every time we upgrade IREE. There's probably a magic couple of bytes as part of the header for any vmfb, I could check for that in a standard test.

sharkfuser/include/fusili/graph.h
```diff
@@ -143,6 +149,39 @@ Graph::convFProp(const std::shared_ptr<TensorAttr> &x,
...
+  FUSILI_CHECK_ERROR(input.write(this->emitAsm()));
+
+  // Build up iree-compile command.
+  TempFile output = FUSILI_TRY(TempFile::create("iree_compile_output"));
+  std::vector<std::string> args = {IREE_COMPILE_PATH,
+                                   input.path,
+                                   "--iree-hal-target-backends=llvm-cpu",
+                                   "--iree-llvmcpu-target-cpu=host",
```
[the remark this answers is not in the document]
AaronStGeorge: I made this configurable, I haven't togged any tests on and off though.

sharkfuser/include/fusilli/logging.h
```diff
@@ -29,6 +29,9 @@ enum class [[nodiscard]] ErrorCode {
   AttributeNotSet,
   InvalidAttribute,
   TensorNotFound,
+  CompileFailure,
+  FileSystemFailure,
+  CacheMiss,
```
sjain-stanford, line 34: Add this to the `ErrorCodeToStr` below.

sharkfuser/include/fusilli/graph.h
```diff
@@ -213,6 +284,78 @@ Graph::convFProp(const std::shared_ptr<TensorAttr> &x,
...
+      /*graphName=*/this->getName(),
+      /*filename=*/IREE_COMPILER_OUTPUT_FILENAME));
+  CacheFile compileCommand =
+      FUSILLI_TRY(CacheFile::open(/*graphName=*/this->getName(),
+                                  /*filename=*/IREE_COMPILE_COMMAND_FILENAME));
+
+  // Check for a cache miss on generated assembly.
+  if (FUSILLI_TRY(input.read()) != generatedAsm) {
+    return error(ErrorCode::CacheMiss,
+                 "Cache miss: generated assembly does not match");
```
sjain-stanford, line 344: Remove the `Cache miss:` prefix since it'd come from the `ErrorCodeToStr`

sharkfuser/include/fusilli/graph.h
```diff
@@ -213,6 +284,78 @@ Graph::convFProp(const std::shared_ptr<TensorAttr> &x,
...
+  }
+
+  // Check for a cache miss on compile command.
+  std::string cmd = buildCompileCommand(backend, input, output);
+  if (FUSILLI_TRY(compileCommand.read()) != cmd) {
+    return error(ErrorCode::CacheMiss,
```
sjain-stanford, line 350: same

sharkfuser/include/fusilli/logging.h
```diff
@@ -337,4 +342,36 @@ inline ConditionalStreamer &getLogger() {
...
+  })
+
+// Test version of FUSILLI_TRY that uses Catch2 REQUIREs instead of returning an
+// ErrorObject.
+#define TEST_FUSILLI_TRY(expr)                                                 \
+  ({                                                                           \
+    auto _errorOr = (expr);                                                    \
+    REQUIRE(isOk(_errorOr));                                                   \
+    std::move(*_errorOr);                                                      \
+  })
```
sjain-stanford, line 375: Do we really need this? My worry is divergence between the two over time. Any issue with using FUSILLI_TRY and check for isOk or isError at the callsite (tests) itself?

sharkfuser/include/fusilli/extras.h
```diff
@@ -0,0 +1,244 @@
...
+  std::transform(sanitizedGraphName.begin(), sanitizedGraphName.end(),
+                 sanitizedGraphName.begin(),
+                 [](char c) { return c == ' ' ? '_' : c; });
+  std::erase_if(sanitizedGraphName,
+                [](unsigned char c) { return !(std::isalnum(c) || c == '_'); });
+
+  // Ensure graphName has a value.
+  if (sanitizedGraphName.empty()) {
+    sanitizedGraphName = "unnamed_graph";
+  }
```
sjain-stanford, line 183: This is OK to have here, but let's avoid a can of worms by making a graphName mandatory on every Fusilli graph, so it is always provided (non-empty) to `CacheFile`.

sharkfuser/include/fusilli/graph.h
```diff
@@ -213,6 +284,78 @@ Graph::convFProp(const std::shared_ptr<TensorAttr> &x,
   return y;
 }
 
+inline ErrorOr<Graph::CachedAssets>
+Graph::generateCompiledArtifacts(Backend backend,
+                                 const std::string &generatedAsm, bool remove) {
+  FUSILLI_LOG_LABEL_ENDL("INFO: Generating compiled artifacts");
```
sjain-stanford, line 290: In a real scenario, we might have `g.generateCompiledArtifacts` run multiple times - I don't see us check in `Graph::generateCompiledArtifacts` if assets already exist and return them directly if so? If keeping them separate is preferred, perhaps create a wrapper method called `readOrGenerateCompiledArtifacts` that first tries to read and if not calls the generate method. We also need to test sequential invocations to ensure first time it misses and second time hits.

sharkfuser/tests/test_graph.cpp
```diff
@@ -98,3 +98,55 @@ TEST_CASE("Graph asm_emitter requires validation to be run first", "[graph]") {
...
+    auto Y = g.convFProp(X, W, conv);
+    Y->setDim({n, k, h, w}).setStride({k * h * w, h * w, w, 1});
+    Y->setOutput(true);
+    REQUIRE(isOk(g.validate()));
+
+    // Generate asm.
+    std::string generatedAsm = TEST_FUSILLI_TRY(g.emitAsm());
+
+    // Cache should be empty.
+    REQUIRE(isError(g.readCompiledArtifacts(Backend::CPU, generatedAsm)));
```
sjain-stanford, line 132: Let's also test the `g.readOrGenerateCompiledArtifacts` wrapper here once you have it. This will be easier to call from the main `g.execute` method.

sharkfuser/include/fusilli/graph.h
AaronStGeorge: Design-wise, my intent in this PR was for the two new methods to be low-level utilities that `g.execute` will someday call call. 

I'm picturing `g.execute` would go something like something like:

```c++
std::string asmText = TEST_FUSILLI_TRY(emitAsm());

runtime_module_type rmt;  // I made this up for the exmaple
if (hasCompiled) {
  rmt.set_vmfb(TEST_FUSILLI_TRY(readCompiledArtifacts(asmText).output.read()));
} else {
  rmt.set_vmfb(TEST_FUSILLI_TRY(generateCompiledArtifacts(asmText).output.read()));
  hasCompiled = true;
}
```

Personally, I like that the “read” and “generate” paths are separate, I think it might help us avoid any always-regenerate bugs.

sharkfuser/include/fusilli/logging.h
sjain-stanford: > The intent of FUSILLI_TRY is to propagate an error back to the caller in the case where a callee can't handle it, the intent of TEST_FUSILLI_TRY is to deduplicate an often repeated block of code in tests.

We don't want to take a dep on `Catch2` for our main fusilli lib (this is an optional dep if users build it with tests skipped). Why not just define this macro as a test helper (in the testing script itself, or a common helper that lives in tests if this is used by >1 tests).

AaronStGeorge: Sounds good 👍

sharkfuser/include/fusilli/graph.h
```diff
@@ -101,7 +115,64 @@ class Graph : public INode {
...
+  // `CacheFiles` are set to be removed RAII based removal will be tied to this
+  // object.
+  struct CachedAssets {
+    CacheFile input;
+    CacheFile output;
+    CacheFile compileCommand;
+  };
+
+  // Create compiled artifacts from graph writing results to the cache. Set
+  // `remove = false` to remove cache files when `CachedAssets` goes out of
```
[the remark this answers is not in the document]
AaronStGeorge: Yes definitely 😅

sharkfuser/include/fusilli/graph.h
```diff
@@ -101,10 +121,53 @@ class Graph : public INode {
...
+  // compiled artifact, subsequent invocations may return cached versions
+  // assuming cache invalidation checks pass. Set `remove = true` to remove
+  // cache files when this `Graph` instance goes out of scope.
+  //
+  // `didGenerate` will be set to true if a value is passed and the cache was
+  // (re)generated; the parameter is useful for testing.
+  ErrorOr<std::string>
+  readOrGenerateCompiledArtifact(const std::string &generatedAsm,
+                                 bool remove = false,
+                                 std::optional<bool> *didGenerate = nullptr) {
```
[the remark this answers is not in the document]
sjain-stanford: I guess this is just to test... If so, makes sense to keep it as such...

sharkfuser/include/fusilli/graph.h
```diff
@@ -118,6 +181,122 @@ class Graph : public INode {
...
+    }
+
+    // Check for a cache miss on compile command.
+    std::string cmd = buildCompileCommand(input, output);
+    if (FUSILLI_TRY(compileCommand.read()) != cmd) {
+      FUSILLI_LOG_ENDL("Compile command does not match");
+      return ok(false);
+    }
+
+    return ok(true);
```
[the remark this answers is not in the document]
sjain-stanford: Oh nvm, the `FUSILLI_TRY` could return `ErrorObject`.

sharkfuser/tests/test_graph.cpp
```diff
@@ -98,3 +113,127 @@ TEST_CASE("Graph asm_emitter requires validation to be run first", "[graph]") {
...
+
+    // Cache should hit, no compilation should be required.
+    didGenerate = std::nullopt;
+    REQUIRE(isOk(g.readOrGenerateCompiledArtifact(
+        generatedAsm, /*remove=*/true, /*didGenerate=*/&didGenerate)));
+    REQUIRE(didGenerate.has_value());
+    REQUIRE(!didGenerate.value());
+
+    // Cache should miss based on different compile command.
+    // TODO(#1964): GFX942 compilation seems to be broken
```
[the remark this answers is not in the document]
sjain-stanford: Maybe just share the repro command + inputs from cache to dog-food the debugging workflow ;)

sharkfuser/tests/test_graph.cpp
```diff
@@ -98,3 +113,127 @@ TEST_CASE("Graph asm_emitter requires validation to be run first", "[graph]") {
...
+        CacheFile::open(g.getName(), IREE_COMPILE_INPUT_FILENAME));
+    REQUIRE(FUSILLI_REQUIRE_UNWRAP(asmCache.read()) == generatedAsm);
+
+    // New instance should regenerate cache.
+    std::optional<bool> didGenerate = std::nullopt;
+    REQUIRE(isOk(g.readOrGenerateCompiledArtifact(
+        generatedAsm, /*remove=*/true, /*didGenerate=*/&didGenerate)));
+    REQUIRE(didGenerate.has_value());
+    REQUIRE(didGenerate.value());
+  }
```
sjain-stanford, line 231: also test the error code / message...

sharkfuser/include/fusilli/graph.h
sjain-stanford: Sounds great!

sjain-stanford (approved):
LG(reat)TM! Just some minor nits and then feel free to land.

sharkfuser/tests/test_graph.cpp
```diff
@@ -98,3 +113,127 @@ TEST_CASE("Graph asm_emitter requires validation to be run first", "[graph]") {
...
+    // Cache should miss because graph name change.
+    g.setName("new_graph_name");
+    didGenerate = std::nullopt;
+    FUSILLI_REQUIRE_UNWRAP(g.readOrGenerateCompiledArtifact(
+        generatedAsm + " ", /*remove=*/true, /*didGenerate=*/&didGenerate));
+    REQUIRE(didGenerate.has_value());
+    REQUIRE(didGenerate.value());
+  }
+
+  SECTION("should not read cached items from other/previous Graph instances") {
```
sjain-stanford, line 196: Thanks for adding this test.

sharkfuser/include/fusilli/graph.h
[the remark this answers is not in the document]
AaronStGeorge: Yeah... I don't love this, but it works 🤷‍♂️

sharkfuser/samples/convolution/conv_fprop.cpp
```diff
@@ -40,5 +41,11 @@ TEST_CASE("Convolution fprop", "[conv][graph]") {
   Y->setOutput(true);
 
   REQUIRE(isOk(graph->validate()));
-  REQUIRE(isOk(graph->emitAsm()));
+
+  ErrorOr<std::string> generatedAsm = graph->emitAsm();
+  REQUIRE(isOk(generatedAsm));
+
+  ErrorOr<std::string> vmfb =
+      graph->readOrGenerateCompiledArtifact(*generatedAsm, /*remove=*/true);
```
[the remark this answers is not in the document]
sjain-stanford: Having `remove = true` as default + setting it to `false` when called from within `g.validate()` should give us the best of both worlds right?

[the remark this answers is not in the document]
AaronStGeorge: `g.validate()` -> `g.execute()`? Yeah, that sounds very reasonable to me.

PR #1937, opened 2025-07-30
[sharktank] Fix FP8 perplexity test when scale is None

archana-ramalingam:
Fix FP8 perplexity test when scale is None


archana-ramalingam:
Resolved by https://github.com/nod-ai/shark-ai/pull/1938

PR #1943, opened 2025-07-31
[sharktank] make overridable decorator return function that looks like the original

sogartar:
Refactor the decorator to use functools.wraps such that tools like Pyright are able to deduce the correct signature.


KyleHerndon:
I'm not necessarily against this, but why?

PR #1944, opened 2025-07-31
[sharktank] make quantize a proper op with multiple dispatch

sogartar:
In order to avoid proliferating quantizer types like for split and replicated sharded tensors, QuantizerTensor.quantize and descendants are implemented as an op utilizing the dispatch mechanism.

### DIFF
--- a/sharktank/sharktank/layers/conv.py
+++ b/sharktank/sharktank/layers/conv.py
@@ -58,9 +58,9 @@ class Conv2DLayer(ThetaLayer):
             x = ops.elementwise(torch.mul, x, self.premul_input)
 
         if q_input is not None:
-            x = q_input.quantize(x)
+            x = ops.quantize(x, q_input)
         elif qdq_input is not None:
-            x = qdq_input.quantize(x).unpack().dequant()
+            x = ops.quantize(x, qdq_input).unpack().dequant()
 
         # Primary computation.
         y = ops.conv2d(
@@ -120,9 +120,9 @@ class Conv3DLayer(ThetaLayer):
             x = ops.elementwise(torch.mul, x, self.premul_input)
 
         if q_input is not None:
-            x = q_input.quantize(x)
+            x = ops.quantize(x, q_input)
         elif qdq_input is not None:
-            x = qdq_input.quantize(x).unpack().dequant()
+            x = ops.quantize(x, qdq_input).unpack().dequant()
 
         # Primary computation.
         y = ops.conv3d(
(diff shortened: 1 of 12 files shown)


sharktank/sharktank/ops/quantized_impls.py
```diff
@@ -96,6 +111,180 @@ def wrapper(*args, **kwargs):
     return decorator
 
 
+def verify_quantized_shape(actual: tuple[int, ...], expected: tuple[int, ...]):
```
KyleHerndon, line 114: I'm not sure if this is too cursed, but we could move this into `_quantize_trampoline` and then you don't need to call it at the end of quanizer.

sharktank/sharktank/ops/signatures.py
```diff
@@ -1008,6 +1011,30 @@ def _module_register_buffer_trampoline(
         d.fail(args)
 
 
+@overridable
+def quantize(
+    tensor: AnyTensor, quantizer: AnyTensor, name: str = UnnamedTensorName
```
KyleHerndon, line 1016: quantizer's type signature should be a QuantizerTensor, no?

sharktank/sharktank/ops/signatures.py
```diff
@@ -1008,6 +1011,30 @@ def _module_register_buffer_trampoline(
...
+) -> AnyTensor:
+    """Quantize a tensor using the provided quantizer."""
+    ...
+
+
+@quantize.trampoline
+def _quantize_trampoline(
+    d: SignatureDispatcher,
+    tensor: AnyTensor,
```
KyleHerndon, line 1025: (Same here)

sharktank/sharktank/layers/conv.py
```diff
@@ -58,9 +58,9 @@ def forward(self, input: torch.Tensor) -> torch.Tensor:
             x = ops.elementwise(torch.mul, x, self.premul_input)
 
         if q_input is not None:
-            x = q_input.quantize(x)
+            x = ops.quantize(x, q_input)
```
KyleHerndon, line 61: You shouldn't need to make any changes to the `layers/` directory, Quantizer.quantize should redirect to ops.quantize, and I think that looks nicer too.

KyleHerndon (changes requested):

sharktank/sharktank/ops/signatures.py
sogartar: It can be a sharded tensor containing quantizer tensors.

PR #1951, opened 2025-08-01
Fix error: zero-dimensional arrays cannot be concatenated

dezhiAmd:
Why:
When indices[0] is 0-dimensional NumPy array that wraps the Python object None, the if check below will pass
    _if indices[0] is not None:_

But 
np.concatenate(indices, axis=1)[0] will fail since indices[0] is 0-dimensional

How:
Check whether indices[0].ndim >= 1


dan-garvey (changes requested):
We know indices[0] is an ndarray because the line above explicitly creates it as one, so you only need to check the ndim here

dan-garvey (approved):

PR #1958, opened 2025-08-05
Added LLM testing utilities including decode and perplexity

rsuderman:
Created some generalized testing utilities for running and numerically evaluating LLM models. This includes a tool for performing decode on a export VMFB and perplexity/decode tooling for running tests.

A sample of testing on toy Llama is included to demonstrate how these tools are used.


sharktank/sharktank/tools/run_llm_vmfb.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def main(prompt, steps, vmfb, config, irpa, tokenizer):
+    tokenizer = Tokenizer(tokenizer)
+    ids = tokenizer.encode([prompt])
+    decoder = Decoder(vmfb_fp=vmfb, config_fp=config, irpa_fp=irpa)
+    tokens = ids[0]
+
+    selected = decoder.decode(tokens=tokens, steps=steps)
+    print(tokenizer.decode(selected)[0])
```
archana-ramalingam, line 65: Do we want to use logger instead of print?

sharktank/sharktank/utils/llm_utils.py
```diff
@@ -0,0 +1,370 @@
...
+        logits = self._model.decode(
+            tokens,
+            attention_mask=attention_mask,
+            start_positions=start_positions,
+            seq_block_ids=seq_block_ids,
+            cache_state=cache_state,
+        )
+
+        # TODO: This should be handled by the model
+        logits = torch.nn.functional.softmax(logits, dim=-1)
```
archana-ramalingam, line 160: Same comment as above, regarding reusing this for ppl.

sharktank/sharktank/tools/run_llm_vmfb.py
```diff
@@ -0,0 +1,86 @@
...
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--prompt", help="String to decode", required=True)
+    parser.add_argument("--irpa", help="IRPA parameters file", required=True)
+    parser.add_argument("--vmfb", help="vmfb file path", required=True)
+    parser.add_argument("--config", help="json config file for server", required=True)
+    parser.add_argument("--tokenizer", help="json tokenizer config file", required=True)
+    parser.add_argument(
+        "--steps", help="steps to perform decode", type=int, required=True
+    )
```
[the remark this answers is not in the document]
rsuderman: For now I want to avoid the constant appending. People keep adding more flags to a single centralized file even when they are not used outside of the specific tool. Flags will likely be iterated / improved (e.g. it currently hardcodes to `hip`).

sharktank/sharktank/utils/llm_utils.py
```diff
@@ -0,0 +1,370 @@
...
+
+        logits = self._model.prefill(
+            tokens,
+            attention_mask=attention_mask,
+            seq_block_ids=seq_block_ids,
+            cache_state=cache_state,
+        )
+
+        # TODO: This should be handled by the model
+        logits = torch.nn.functional.softmax(logits, dim=-1)
```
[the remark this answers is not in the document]
rsuderman: Honestly we should likely start exporting only with log_softmax. I looked at cross entropy and the computation is related to log softmax. If we know thats the final value its actually simpler to compute perplexity.

sharktank/sharktank/tools/run_llm_vmfb.py
rsuderman: Logging is used for log output: info, debug, errors. Print (or other output components) are meant for outputting to stdout / to files as results. So in this case we would use print (or another similar output utility).

archana-ramalingam (approved):

PR #1965, opened 2025-08-05
[sharktank] Add a more concise trampoline description

sogartar:
In our multi-dispatch mechanism we replicate essentially the same pattern of trampoline definition on many places.

Here a more general approach is added that removes this boilerplate code. Existing trampolines that are not trivial to remove are left intact.

### DIFF
--- a/sharktank/sharktank/ops/_registry.py
+++ b/sharktank/sharktank/ops/_registry.py
@@ -6,11 +6,13 @@
 
 """Signatures for dynamic dispatch of ops covering our fundamental tensor types."""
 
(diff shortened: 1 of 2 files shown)


dan-garvey (approved):
Love it. Do we have any workloads that depend on torch < 2.7.1? We can probably bump to that version if not.

EDIT:: Meant to leave as a comment, not an approval

sogartar:
> Love it. Do we have any workloads that depend on torch < 2.7.1? We can probably bump to that version if not.
> 
> EDIT:: Meant to leave as a comment, not an approval

PR #1967, opened 2025-08-05
Fix Model Integration Test

dezhiAmd:
**Why:**
  GitHub hosted runner does not have enough disk space in /tmp to store all files. This is the reason that cause this CI test fail
**How:**
  Split "Run punet tests" into two tests, one for fp16 and one for int8

### DIFF
--- a/sharktank/integration/models/punet/integration_test.py
+++ b/sharktank/integration/models/punet/integration_test.py
@@ -13,8 +13,9 @@ from sharktank.utils import testing
 @pytest.fixture(scope="module")
 def temp_dir():
     # Create a persistent temp directory under /mnt for GithHub-hosted runner since tmp folder does not have enough disk space
-    path = Path("/mnt/runner/integration_test_temp")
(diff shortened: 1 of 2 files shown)


sharktank/integration/models/punet/integration_test.py
```diff
@@ -12,8 +12,16 @@
 
 @pytest.fixture(scope="module")
 def temp_dir():
-    with testing.temporary_directory(__name__) as td:
-        yield Path(td)
+    # Create a persistent temp directory under /mnt for GithHub-hosted runner since tmp folder does not have enough disk space
```
IanNod, line 15: I think it would be safer to keep things running under a tmp folder since it would ensure the files are automatically cleaned up and allow possible parallel jobs on the same runner. SDXL should be small enough to fit, we just seem to be downloading and exporting multiple versions including the original fp16 and our mixed precision int8 model. Splitting these up to different tests I think should work

PR #1969, opened 2025-08-06
[sharktank] Fix loading of legacy fp4 block scaled quantized tensors having scale with no external tensor trait

sogartar:
The change https://github.com/nod-ai/shark-ai/pull/1900 introduced a bug where a loaded tensor would have no external trait, which then during export caused the tensor to get inlined into the MLIR.


aviator19941 (approved):
LGTM

PR #1971, opened 2025-08-06
use new decoder as default

dezhiAmd:
**Why:**
If num_beams=2, new decoder does not generate 2 responses  (for one prompt) when use_beam_search is false. 
Instead, new decoder will only generate 1 response. 
This make sense because in deterministic response case, when use_beam_search is false,  greedy sampling method is used.

**How**
Remove all smoke tests in testing matrix when num_beams=2, use_beam_search is false.
This PR depends on https://github.com/nod-ai/shark-ai/pull/2001


app_tests/integration_tests/llm/shortfin/tinystories_llama2_25m_test.py
```diff
@@ -256,7 +256,7 @@ def test_multi_hypothesis_switch(
             "max_completion_tokens": 15,
             "temperature": 0.7,
             "num_beams": num_beams,
-            "use_beam_search": False,
```
dezhiAmd, line 259: Based this [line](https://github.com/nod-ai/shark-ai/blob/810cd21374336eff4adf3d1805ac850460e0a146/app_tests/integration_tests/llm/shortfin/tinystories_llama2_25m_test.py#L269), I am assuming there is mistake here

dezhiAmd:
Close this since it is covered by new PR: https://github.com/nod-ai/shark-ai/pull/2026

PR #1972, opened 2025-08-06
[Fusilli] NFC refactor

sjain-stanford:
Create three new sub-dirs: `support`, `backend` and `graph`. Move top-level headers to the corresponding sub-dirs.

### DIFF
--- a/sharkfuser/include/fusilli.h
+++ b/sharkfuser/include/fusilli.h
@@ -15,8 +15,8 @@
 
 // Support
 #include "fusilli/asm_emitter.h"
-#include "fusilli/external_tools.h"
 #include "fusilli/extras.h"
+#include "fusilli/support/external_tools.h"
 #include "fusilli/support/logging.h"
 #include "fusilli/types.h"
 
--- a/sharkfuser/include/fusilli.h
+++ b/sharkfuser/include/fusilli.h
@@ -14,12 +14,13 @@
 #define FUSILLI_H
 
 // Support
-#include "fusilli/extras.h"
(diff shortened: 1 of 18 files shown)


sharkfuser/include/fusilli.h
```diff
@@ -13,25 +13,28 @@
...
 
-// Utilities
-#include "fusilli/asm_emitter.h"
-#include "fusilli/backend.h"
-#include "fusilli/external_tools.h"
-#include "fusilli/extras.h"
-#include "fusilli/logging.h"
-#include "fusilli/types.h"
+// Backend
+#include "fusilli/backend/backend.h"
```
AaronStGeorge, line 34: Do we want everything in a subdirectory? We could wait to break things into sub-directories until conceptual grouping of things requires more than one file. With everything in a subdirectory we have a lot of `#import backend/backend.h" and `#import graph/graph.h`.

AaronStGeorge (approved):
LGTM, I might not put everything into a subdirectory, but I don't feel strongly about that. Up to you 👍

sharkfuser/include/fusilli.h
sjain-stanford: Shortfin does it too (putting all headers in subdirs): https://sourcegraph.com/github.com/nod-ai/shark-ai/-/tree/shortfin/src/shortfin so was going off of that. Also, right now backend has just one header but I'm anticipating all the runtime C-API interfacing also goes here. Backend here is not the compiler backend, rather fusilli's backend - so anything to do with IREE compiler/runtime could go here.

PR #1977, opened 2025-08-07
[Fusilli] Switch to (pre-) source-built and statically linked IREERuntime, and python packages for `iree-compile`

sjain-stanford:
We were relying on the `iree-dist*` releases to pull in IREECompiler and IREERuntime libs/tools from, but these were being published with certain backends (HIP/ROCM) disabled ([here](https://github.com/iree-org/iree/blob/961706dc037114102f43fc31e4163ad7c7852406/build_tools/github_actions/build_dist.py#L121-L154)), which are otherwise enabled through the python releases ([here](https://github.com/iree-org/iree/blob/961706dc037114102f43fc31e4163ad7c7852406/build_tools/python_deploy/build_linux_packages.sh#L158-L171)).

Even after enabling the said backends, there were still issues with missing headers in the dist. Moreover, the process of changing and validating a fixed dist release is on the order of ~O(hours). We didn't want to take a full source dependency on the IREE compiler (as this is quite heavy), while IREE runtime is best consumed as a source build + static link. To keep things nimble we are moving to a hybrid approach as outlined below:

```
# It is up to the consumer of Fusilli to determine how to bring the IREE dependencies in,
# but here's our recommendation:
# - Get CLI tools (iree-compile, iree-opt) from python packages
# - For C-API interfacing with IREERuntime, have it built from source and statically linked
#
# Once IREERuntime is built, just point to the build location with -DIREERuntime_DIR or
# -DCMAKE_PREFIX_PATH, like so:
#
#     -DIREERuntime_DIR=</path/to/iree/build/lib/cmake/IREE>
#
# The dependency on IREECompiler is through the CLI tool (`iree-compile`) which is easily
# accessible from binary releases so a source build is not necessary here.
```


sharkfuser/CMakeLists.txt
```diff
@@ -48,27 +48,24 @@ list(APPEND CMAKE_MODULE_PATH
...
-find_package(IREECompiler REQUIRED)
+# It is up to the consumer of Fusilli to determine how to bring the IREE dependencies in,
+# but here's our recommendation:
+# - Get CLI tools (iree-compile, iree-opt) from python packages
+# - For C-API interfacing with IREERuntime, have it built from source and statically linked
+#
+# Once IREERuntime is built, just point to the build location with -DIREERuntime_DIR or
+# -DCMAKE_PREFIX_PATH, like so:
+#
+#     -DIREERuntime_DIR=</path/to/iree/build/lib/cmake/IREE>
```
ScottTodd, line 60: The recommended way to use the IREE runtime is via the low level C API, depending on the headers and functions that you need _in your application_: https://iree.dev/reference/bindings/c-api/#runtime-api. This external build in a Linux-only dockerfile (https://github.com/sjain-stanford/docker/blob/cc39889895a2d0e2187d3135d037e7c172ec7263/Dockerfile#L49-L68) hides relevant details from developers like compilation options used and which source files are needed.

ScottTodd (commented):
Thanks, this is a step in the right direction but I still have a few points of design feedback.

PR #1978, opened 2025-08-07
Retool export scripting to be invokable from python

rsuderman:
Existing export-and-run tests rely on invoking a sub-process for llama and then relying on it to export the model. Reworking the export tooling we can more cleanly separate the `cli` from the script, and include some configuration dataclasses to make configuration cleaner.

### DIFF
--- a/sharktank/sharktank/examples/export_paged_llm_v1.py
+++ b/sharktank/sharktank/examples/export_paged_llm_v1.py
@@ -6,143 +6,60 @@
 
 """Export support for the PagedLLMV1 protocol of models."""
 
+import dataclasses
 import os
 import logging
 import json
-from typing import Any, Dict, Tuple, Optional
 import torch
 
 from iree.turbine.aot import *
-
-from sharktank.layers import *
-from sharktank.types import *
-from sharktank.types.pipelining import pipeline_parallelize_theta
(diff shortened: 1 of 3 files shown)


dan-garvey (approved):
LGTM, I assume there is follow up work needed to change the test infra

sharktank/sharktank/examples/export_paged_llm_v1.py
archana-ramalingam, line 1: Do we also want to move this export script into a new folder instead of the generic `examples`?

sharktank/sharktank/models/llm/export.py
```diff
@@ -0,0 +1,186 @@
...
+            seq_block_ids=seq_block_ids,
+            cache_state=cache_tensors,
+        )
+
+        logits = ops.unshard(logits)
+
+        if self.config.logits_normalization == "softmax":
+            logits = ops.softmax(logits, dim=-1)
+
+        if self.config.logits_normalization == "log_softmax":
```
archana-ramalingam, line 82: Should this be an `elif`?

rsuderman: It can be but given its a direct comparison there is not a huge benefit. Eventually we should just have the normalization as a single helper function instead of directly including each option.

PR #1979, opened 2025-08-08
Transpose conv wave kernel integration

badgerbroch:
Added integration of transpose conv wave kernel into sharktank.


aviator19941 (commented):
Looks good, btw, can you test to see if the IR generated from the test compiles and runs in IREE?

PR #1984, opened 2025-08-08
[Fusilli] Enable gfx942 compilation test among other changes

sjain-stanford:
Misc changes:
- s/didGenerate/reCompiled in docstring
- remove default on `generateCompiledAssets` remove (since this private and not user facing so tightly controlled)
- Don't use `FUSILLI_REQUIRE_UNWRAP` unless it's assigned to an lvalue for further checks

@AaronStGeorge I get this compiler warning in VSCode, not sure why but it compiles fine:
image

### DIFF
--- a/sharkfuser/include/fusilli/support/cache.h
+++ b/sharkfuser/include/fusilli/support/cache.h
@@ -96,6 +96,32 @@ public:
     return ok(CacheFile(path, /*remove=*/false));
   }
 
+  // Utility method to build the path to cache file given `graphName` and
+  // `fileName`.
+  //
+  // Format: $HOME/.cache/fusilli/<sanitized version of graphName>/<fileName>
+  static std::filesystem::path getPath(const std::string &graphName,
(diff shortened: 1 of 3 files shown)


sjain-stanford:
>  I get this compiler warning in VSCode, not sure why but it compiles fine:

Switching to clangd instead of intellisense makes this go away...

sharkfuser/include/fusilli/support/cache.h
```diff
@@ -96,6 +96,32 @@ class CacheFile {
     return ok(CacheFile(path, /*remove=*/false));
   }
 
+  // Utility method to build the path to cache file given `graphName` and
+  // `fileName`.
+  //
+  // Format: $HOME/.cache/fusilli/<sanitized version of graphName>/<fileName>
+  static std::filesystem::path getPath(const std::string &graphName,
```
sjain-stanford, line 103: just moved it up from below to keep static methods together

PR #1998, opened 2025-08-12
[sharktank] xfail nightly VAE tests

sogartar:


IanNod:
> > Why are we xfailing instead of fixing?
> 
> There is no fix for the compilation issue yet.

Are you sure? I just ran the vae tests locally at 
```iree-base-compiler==3.7.0.dev0+bd9d4ea01875660c5af431ff80298dd35e1ca7c9```

and all tests pass

PR #2009, opened 2025-08-13
Make 2.6.0 the default PyTorch version

sogartar:


sogartar:
It is desirable that before this the [nightly tests](https://github.com/nod-ai/shark-ai/actions/workflows/ci-sharktank-nightly.yml) pass.
Also this [PR](https://github.com/nod-ai/shark-ai/pull/1998) should be merged first.

IanNod:
I would hesitate to make this change without better regression tests. It's possible IR will change with a different torch version and we will need to check any possible changes will not regress performance

sogartar:
Yes, it is possible that benchmarks regress, I have not checked yet.

PR #2013, opened 2025-08-14
[sharktank] Add chunked prefill to llm

archana-ramalingam:
Add an optional `start_positions` in prefill to enable chunked prefill


sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -82,12 +84,12 @@ def generate_batch_prefill(bs: int):
 
         @fxb.export_program(
             name=f"prefill_bs{bs}",
-            args=(tokens, seq_lens, seq_block_ids, cache),
+            args=(tokens, start_pos, seq_lens, seq_block_ids, cache),
             dynamic_shapes=dynamic_shapes,
             strict=strict,
         )
-        def _(model, tokens, seq_lens, seq_block_ids, cs):
```
rsuderman, line 89: Lets make this optional. We can either have a flag for prefill or make a separate `chunked_fill` function that exports instead. We want to maintain the existing path initially as this will be breaking shortfin side initially.

rsuderman (changes requested):

rsuderman (changes requested):
Can we break this into three PRs with tests:

1. Rotary embedding changes
2. Paged Attention changes
3. E2e model support

We should make sure rotary and attention have tests for the new behavior

PR #2023, opened 2025-08-15
[sharktank] fix test for wave fp4 gemm kernel

sogartar:
Fix the RHS transpose as the kernel does `A x B^T` rather than `A x B`.

During quantization don't take a view, but rashape as the tensor may have layout that can't be reshaped without copying.

Refactor the test to include different matrix sizes and make sure m != n != k.


KyleHerndon (approved):

aviator19941 (approved):
LGTM thanks!

PR #2026, opened 2025-08-15
Enable new decoder - remove multigreedy case

rsuderman:
We should only be performing beam search with multiple hypothesis cases. Removed the corresponding tests and enabled the new decoder. Some fixes to the page manager work better for non topk.


IanNod (approved):
looks good to me

PR #2042, opened 2025-08-18
Add a raw benchmark utility

rsuderman:
Perform a simulated benchmark with a VMFB with a fixed context length and step count. This can be used to roughly estimate the throughput of a given export model.


dan-garvey (approved):
Dataclass > config dict. lgtm

PR #2049, opened 2025-08-19
Refactor `LlmInvoker` and `LlmTask` to no longer reference global `exec_requests`

stbaione:
# Description

Currently, we're using `exec_requests` essentially as global variables, where anything that has to do with invocation has a reference and modifies them.

This too closely ties the invocation of the Llm to the `LlmInferenceExecRequests` themselves, when in reality, we only need a subset of the data from those requests to invoke. 

By decreasing how tightly copied the request objects are with the invocation, it should make it easier in the future to do certain multi-step operations, like chunked prefill, without having to modify `exec_requests` for each invocation.

As a side issue, it can be difficult to debug when there are so many modification, many layers down in the stack.

This makes it so that the `LlmInvoker` and `LlmTask`s don't have any reference to `LlmInferenceExecRequests`. They agnostically perform invocation based on the data that they receive, and leave it to the batcher to update the requests themselves. This should also make it easier to move certain caching related operations to the decoder as well, since it's bubbled up, and, at this point, should just be copy and paste.

# TODO

- Run a quick benchmark
- Double check docstrings
- Write/fix unit tests


shortfin/python/shortfin_apps/llm/components/batcher.py
```diff
@@ -69,6 +72,8 @@ def __init__(
 
         self.program_isolation = program_isolation
 
+        self.active_requests: Dict[str, List[LlmInferenceExecRequest]] = {}
```
rsuderman, line 75: These kinds of stateful components are pretty big red flags to me. We should avoid having under cover mutation where possible. I am pretty sure that by tracking the run command we can handle fine with depending on global mutation.

shortfin/python/shortfin_apps/llm/components/batcher.py
```diff
@@ -125,6 +130,7 @@ def make_invoker(
         page_cache: BasePagedAttentionCache,
         fiber: Fiber,
         exec_requests: list[LlmInferenceExecRequest],
+        group_id: str,
```
rsuderman, line 133: Arbitrary group name string feels bad, we shouldn't have some arbitrary string for managing requests.

shortfin/python/shortfin_apps/llm/components/batcher.py
```diff
@@ -168,13 +174,53 @@ def board(
...
+        group_id = str(uuid4())
+        self.active_requests[group_id] = exec_requests
+        exec_process = self.make_invoker(page_cache, fiber, exec_requests, group_id)
 
         # We've filled our flight. Remove from the boarding area.
         if exec_requests:
             # And takeoff.
             exec_process.launch()
 
+    def completion_callback(
```
rsuderman, line 186: Callback interfaces should be defined within the invoker then overridden / handled. E.g. the responder classes the cli/fastapi server is an example of this. Avoid defining out in a class and forcing the owned class to adopt.

shortfin/python/shortfin_apps/llm/components/invocation.py
```diff
@@ -443,23 +440,23 @@ def __init__(
         self,
         name: str,
         fiber: sf.Fiber,
-        array_cache: DeviceArrayCache,
         llm_task: LlmTask,
         functions: dict[int, sf.ProgramFunction],
-        seq_stride: int,
         page_tables,
         program_isolation: sf.ProgramIsolation,
+        completion_callback: Callable,
```
rsuderman, line 447: See this just defines a callable function with no for what the callback should look like. You need to define something at the LlmInvoker level that is overridden or there is no way for a user to implement.

shortfin/python/shortfin_apps/llm/components/invocation.py
```diff
@@ -326,37 +326,33 @@ def get_args_data(
...
             )
-            # Extend the sequence block IDs data with padded values.
-            seq_block_ids_data.extend(padded)
+            seq_block_ids_data.extend(block_ids)
 
         return (
             token_data,
-            seq_lens_data,
-            start_positions_data,
+            [position + 1 for position in task_input.start_positions],
```
rsuderman, line 346: Don't manipulate types on return. The result should be independently defined. Return is just for returning.

shortfin/python/shortfin_apps/llm/components/invocation.py
```diff
@@ -124,6 +125,25 @@ async def get_result(
             req_count (int): The number of requests in the batch.
         """
 
+    async def invoke(
+        self,
+        page_tables: sfnp.device_array,
+        batch_size: int,
+        invocation_function: Callable[..., Awaitable[sfnp.device_array]],
```
rsuderman, line 132: Another red flag with an arbitrary invocation function.

PR #2050, opened 2025-08-19
Move allocate_cache from btacher.py to decoder.py

dezhiAmd:
1. Rewire batcher.py so that the page_cache no longer allocates in the batcher and instead perform the allocation in decoder.py. 
2. Removing the publish pages function so it occurs post prefill and post all decode steps

### DIFF
--- a/shortfin/python/shortfin_apps/llm/components/decoder/decoder.py
+++ b/shortfin/python/shortfin_apps/llm/components/decoder/decoder.py
@@ -365,15 +365,8 @@ class LlmDecoder:
 
         return decode_reqs
 
-    async def run(self, input_ids):
-        input_length = len(input_ids)
-        prefill_req = LlmInferenceExecRequest(
-            phase=InferencePhase.PREFILL,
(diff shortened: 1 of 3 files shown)


stbaione (approved):

shortfin/python/shortfin_apps/llm/components/decoder/decoder.py
```diff
@@ -384,13 +377,29 @@ async def run(self, input_ids):
...
+                f"Failed to allocate {needed_pages} pages for prefill request."
             )
+
         logger.debug(f"Successfully acquired allocation: {allocation}")
         prefill_req.free_cache_pages()
         prefill_req.allocation = allocation
 
+    def _allocate_decode_cache(self, request: LlmInferenceExecRequest):
+        if request.allocation is not None:
+            request.allocation.extend_allocation(
```
rsuderman, line 391: We should not be using the extend allocation command. This should ultimately be removed from the allocator

PR #2059, opened 2025-08-20
Kvcache Replumbing

Alex-Vasile:
- Instead of passing around a list of tensors for the cache state, we now pass around a KVCache object.
- Remove KVCache from PagedAttention and disentangle the two.
- Renamed model.cache to model.paged_attention, which is what it is.

### DIFF
--- a/sharktank/sharktank/models/llm/llm.py
+++ b/sharktank/sharktank/models/llm/llm.py
@@ -74,7 +74,7 @@ class PagedLlmModelV1(BaseCausalLMModel):
         )
         self.config = config
         self.hp = self.config.hp
-        self.cache = create_paged_kv_cache(self.config)
+        self.paged_attention = create_paged_attention(self.config)
         # TODO: Add inference_norm as an optional value from config
         self.inference_norm = self.config.hp.model_arch == "grok"
 
@@ -106,7 +106,7 @@ class PagedLlmModelV1(BaseCausalLMModel):
                 AttentionFFNBlock(
                     theta("blk", n),
                     block_index=n,
-                    cache=self.cache,
+                    paged_attention=self.paged_attention,
                     config=self.config,
                     fake_quant=self.fake_quant,
                 )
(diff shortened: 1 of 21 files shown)


sharktank/sharktank/utils/create_cache.py
```diff
@@ -5,9 +5,10 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 from sharktank.layers import *
+from sharktank.layers.paged_attention import KVCache
```
archana-ramalingam, line 8: Doesn't `from sharktank.layers import *` already import KVCache?

sharktank/sharktank/models/llm/llm.py
```diff
@@ -13,6 +13,7 @@
 
 from sharktank import ops
 from sharktank.layers import *
+from sharktank.layers.paged_attention import KVCache
```
archana-ramalingam, line 16: Doesn't `from sharktank.layers import *` already import `KVCache`?

sharktank/sharktank/utils/create_cache.py
archana-ramalingam, line 8: same as above

Alex-Vasile: Same as above.

sharktank/sharktank/models/llm/llm.py
Alex-Vasile: I forgot to add the import in `layers.__init__`, fixed.

sharktank/sharktank/utils/create_cache.py
```diff
@@ -7,7 +7,7 @@
 from sharktank.layers import *
 
 
-def create_paged_kv_cache(config: LlamaModelConfig) -> PagedAttention:
+def create_paged_attention(config: LlamaModelConfig) -> PagedAttention:
     if config.kv_cache_type != "paged":
```
archana-ramalingam, line 11: Do we want to verify cache type here as well?

Groverkss (changes requested):
> Remove KVCache from PagedAttention and disentangle the two.

What does this mean? You cannot distangle them. The KVCache layout is entierly decided by the PagedAttention implementation, they are not seperate things. What are you trying to do exactly?

PR #2061, opened 2025-08-20
[sharktank] Add a more concise trampoline description

KyleHerndon:
In our multi-dispatch mechanism we replicate essentially the same pattern of trampoline definition on many places.

Here a more general approach is added that removes this boilerplate code. Existing trampolines that are not trivial to remove are left intact.

### DIFF
--- a/sharktank/sharktank/ops/_registry.py
+++ b/sharktank/sharktank/ops/_registry.py
@@ -6,11 +6,13 @@
 
 """Signatures for dynamic dispatch of ops covering our fundamental tensor types."""
 
+from collections.abc import Iterable
 from inspect import isclass
 from typing import Any, Callable, Iterable, Optional, Tuple
 
 import collections
 import functools
+import inspect
 
 from torch import Tensor
 from sharktank.types import PrimitiveTensor, QuantizedTensor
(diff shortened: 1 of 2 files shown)


sharktank/sharktank/ops/_registry.py
```diff
@@ -386,10 +416,66 @@ def overridable(
...
+    dispatch_args = [
+        a if isinstance(a, str) else signature_arg_names[a] for a in dispatch_args
+    ]
+
+    def trampoline(_signature_dispatcher_: SignatureDispatcher, *args, **kwargs) -> Any:
+        # We need the signature created here and not captured from the parent scope.
+        # Otherwise torch tracing fails.
+        signature = inspect.signature(f)
+        bound_args = signature.bind(*args, **kwargs)
+        bound_args.apply_defaults()
```
sogartar, line 447: Why can't you remove the call to `bound_args.apply_defaults()` and reimplement it to make it tracing friendly?
Probably the tracing problem is that in `bound_args.apply_defaults()` `bound_args.arguments` gets completely replaced with a new value.

KyleHerndon: I removed apply_defaults and all tests are still passing. I'm not sure if this achieves the full desired behavior, but if you send me PR making the desired changes I'll gladly accept it. I want to build off of this PR, so I want to land this now, though.

PR #2070, opened 2025-08-21
[Dataset] Add json load/save support to Dataset

Groverkss:
This patch adds json support to save / load dataset as empty tensors. This is designed to allow exporting models without needing the whole irpa file for repros.

For example, for llama-405b, the entire irpa file is 207GB, but the json file is 500kb (unseralized with indenting).

Main changes:

- Split add_to_archive method on InferenceTensor to get_metadata and add_to_archive, preserving add_to_archive behavior.
- Changes add_to_archive behavior for ReplicatedTensor to save the tensor name as the first shard, this is more consistent with how the raw entries are stored during runtime, and its better to save them in the same way.
- Adds cli dataset support for --json-file = 
- Add a tool to dump a dataset (gguf/irpa/json) to json

### DIFF
--- a/sharktank/tests/tools/convert_to_json_test.py
+++ b/sharktank/tests/tools/convert_to_json_test.py
@@ -60,6 +60,3 @@ def test_import_export_json(tmp_path: Path):
     assert isinstance(inference_tensors["b"], ReplicatedTensor)
     assert "c" in inference_tensors
     assert isinstance(inference_tensors["c"], SplitPrimitiveTensor)
-
-
-test_import_export_json(Path("./"))


rsuderman (approved):

sharktank/sharktank/types/theta.py
```diff
@@ -556,29 +585,66 @@ def load_tensors(self, entries: dict[str, ParameterArchiveEntry]):
...
+                try:
+                    raw_entry = meta[global_name]
+                except KeyError as e:
+                    raise IOError(
+                        f"InferenceTensor missing one of its tensor components"
+                    ) from e
+                # Each raw tensor is structure as Tuple[List, dtype]
+                shape, dtype = raw_entry
+                dtype = serialized_name_to_dtype(dtype)
+                raw_tensor = torch.empty(shape, dtype=dtype)
```
sogartar, line 607: If you read the contents of this tensor afterwards you would be reading uninitialized memory and for example if you try to profile some memory corruption with the address sanitizer it will falsely report these reads as errors.
I understand that you don't want to zero the tensor for performance reasons.

sogartar (approved):
I wonder if we could utilize [fake tensors](https://docs.pytorch.org/docs/stable/torch.compiler_fake_tensor.html) explicitly. They can actually be used even during eager execution, yielding just the resulting tensor metadata.
Internally they are used during tracing, so for export we actually should need only fake tensors.

sharktank/sharktank/types/theta.py
Groverkss: Yeah, this should ideally be a FakeTensor, torch just doesn't have very nice support to create a FakeTensor directly

Groverkss: ok made it a FakeTensor

Groverkss:
Moved to using FakeTensor now

sharktank/tests/tools/convert_to_json_test.py
```diff
@@ -0,0 +1,65 @@
...
+
+    assert "a" in inference_tensors
+    assert isinstance(inference_tensors["a"], DefaultPrimitiveTensor)
+    assert "b" in inference_tensors
+    assert isinstance(inference_tensors["b"], ReplicatedTensor)
+    assert "c" in inference_tensors
+    assert isinstance(inference_tensors["c"], SplitPrimitiveTensor)
+
+
+test_import_export_json(Path("./"))
```
sogartar, line 65: This should not be called here. Let Pytest collect the tests it wants to run.

Groverkss: yes, thank you! I missed this.

PR #2071, opened 2025-08-21
Kernel selection for attention and matmul

KyleHerndon:


jinchen62:
I don't get how to use this. Which kernel is default and how to invoke another one?

PR #2072, opened 2025-08-21
[Sharktank] Pipeline parallelism FFN example refactor

Alex-Vasile:
Instead of #2014.

### DIFF
--- a/sharktank/sharktank/examples/pipeline/export_ppffn_net.py
+++ b/sharktank/sharktank/examples/pipeline/export_ppffn_net.py
@@ -22,7 +22,10 @@ from sharktank.utils import cli
 from sharktank.layers import *
 from sharktank import ops
 from sharktank.types import *
-from sharktank.types.pipelining import pipeline_parallelize_llm_theta
+from sharktank.types.pipelining import (
+    pipeline_parallelize_llm_theta,
+    inter_block_callback,
+)
 
 from iree.turbine.aot import DeviceAffinity, export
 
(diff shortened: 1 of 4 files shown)


sharktank/sharktank/examples/pipeline/export_ppffn_net.py
```diff
@@ -150,23 +163,26 @@ def main(raw_args=None):
...
+        args.pipeline_parallelism_size > 1
+    ), "Pipeline parallelism size must be greater than 1."
 
     bs = 16
     sl = 128
-    primary_dim = 128 * 2**5
-    shard_count = 2
-    num_layers = 40
-    create_theta(primary_dim, shard_count, num_layers, save_path=args.output_irpa_file)
+    dim = 128 * 2**5
```
dan-garvey, line 172: nit: wouldnt hurt to provide context as to why you picked this, even if it isn't new to this patch

dan-garvey (approved):
minor nit, but lgtm

sharktank/sharktank/examples/pipeline/export_ppffn_net.py
```diff
@@ -77,57 +92,55 @@ def create_theta(dim: int, shard_count: int, num_layers: int, save_path):
...
+            shard_count=len(self.pipeline_stage_to_devices[0]),
+            devices=self.pipeline_stage_to_devices[0],
         )
-        for block in range(num_blocks):
-            weight = self.theta.tensor(f"blk.{block}.ffn.weight")
-            x = ops.replicate(ops.linear(x, weight), shard_count)
-            x = self._inter_layer_callback(x, block)
+        for block_idx, block in enumerate(self.blocks):
+            x = block(x)
+            x = transfer_between_blocks_if_needed(
```
rsuderman, line 120: In the follow up - lets do a "transfer_between_blocks_if_needed" where you pass in the `theta` for the block. Then you just process determine the devices to transfer to based on the placement of those tensors. This way we no longer need to compute / pass a staging map.

PR #2073, opened 2025-08-22
Online serving response checks

yash-amd:
Online Serving Response Checks Integration in CI

### DIFF
--- a/.github/workflows/ci_llama_3.1_70b_fp16.yml
+++ b/.github/workflows/ci_llama_3.1_70b_fp16.yml
@@ -8,10 +8,6 @@ name: Release Llama 3.1 70B FP16 Benchmarking Tests
 
 on:
   workflow_dispatch:
-  pull_request:
-  push:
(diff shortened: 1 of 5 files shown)


scripts/run_serving.sh
```diff
@@ -116,3 +116,21 @@ curl http://localhost:$port/generate \
...
+file="$(pwd)/../output_artifacts/online_serving.log"
+if [ -e "$file" ]; then
+    echo "The file '$file' exists."
+else
+    echo "The file '$file' does NOT exist."
+    exit 1
+fi
+
+# Check for the Online Serving Response Match
+if grep -F "\"responses\": [{\"text\": \"assistant\nThe capital of the United States is Washington, D.C.\"}]" $file; then
```
vivekkhandelwal1, line 130: I am not sure if this is the correct way of doing this. Even a small difference would result in failure.

pravg-amd (approved):

PR #2075, opened 2025-08-22
[shortfin][llm] Move batching logic behind a strategy pattern.

vinayakdsci:
Still WIP.

This patch moves all the batching logic behind an opaque facade that is exposed to the user.
Internal implementation details are completely hidden, and new batching engines can be added through the use of a creation factory. 
This ensures that major changes to the batching code do not touch other parts of the code base, and it is easy to plug in new batching algorithms.

### DIFF
--- a/shortfin/python/shortfin_apps/llm/components/batching/facade.py
+++ b/shortfin/python/shortfin_apps/llm/components/batching/facade.py
@@ -1,0 +1,81 @@
+# Copyright 2025 Advanced Micro Devices, Inc.
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+"""
+Implements a unified facade to handle batching.
+"""
+
+import shortfin as sf
+
+from typing import Callable
+
+from ..kvcache.base_attention_cache import BasePagedAttentionCache
+from ..messages import LlmInferenceExecRequest
+from .factory import _BatchingEngineImpl, _create_impl
+from .config import BatchConfig, Phase
+
+
+class UnifiedBatcher:
+    def __init__(self, *, impl: _BatchingEngineImpl):
+        self._impl = impl
+        self._impl_prefill_obj = self._impl.prefill_lane.impl_cls
+        self._impl_decode_obj = self._impl.decode_lane.impl_cls if self._impl.decode_lane is not None else None
+
+    def submit(self, phase: Phase) -> Callable[[LlmInferenceExecRequest], None] | None:
+        fn = None
+        if phase == Phase.PREFILL:
+            fn = self._impl_prefill_obj.submit # type: ignore
+        if phase == Phase.DECODE:
+            fn = self._impl_decode_obj.submit # type: ignore
+        if fn is None:
+            raise ValueError("Unsupported Batching Lane requested")
+
+        return fn
+
+    def launch(self):
+        self._impl.launch()
+
+    def shutdown(self):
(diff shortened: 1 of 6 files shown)


shortfin/python/shortfin_apps/llm/components/batching/factory.py
```diff
@@ -0,0 +1,65 @@
...
+        self.batching_engine.submit(request)
+
+    def reserve_workload(self, *, rid: str, count: int):
+        self.batching_engine.reserve_workload(rid=rid, count=count)
+
+    def model_params(self):
+        return self.batching_engine.get_model_params()
+
+
+def _create_impl(batch_cfg: BatchConfig, page_cache: BasePagedAttentionCache, prefill_fiber: sf.Fiber, decode_fiber: sf.Fiber | None = None):  # type: ignore
```
stbaione, line 53: Should this be public, since it's imported & used in `facade.py`?

stbaione (commented):
I like this interface, just a couple nit picks. Not sure if we should have the empty `maximal_decode.py` file, or if we should cut here, with the new interface, then make another PR that introduced the `MaximalDecode` concept

shortfin/python/shortfin_apps/llm/components/batching/facade.py
```diff
@@ -0,0 +1,57 @@
...
+
+from typing import Callable
+
+from ..kvcache.base_attention_cache import BasePagedAttentionCache
+from ..messages import LlmInferenceExecRequest
+from .factory import _BatchingEngineImpl, _create_impl
+from .config import BatchConfig
+
+
+class BatchingFacade:
```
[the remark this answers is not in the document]
vinayakdsci: It was initially called `UnifiedBatcher`. I renamed it to make the intention and responsibility explicit. What I observed was that UnifiedBatcher could mean unified prefill and decode, which might become confusing, especially for the current default implementation.

shortfin/python/shortfin_apps/llm/components/batching/factory.py
vinayakdsci: We could make it public. I wanted the private signature to mean that this is not supposed to be called outside of the batching. So anything that calls into `BatchingFacade` should know that the factory is not a public API.

stbaione (approved):

vinayakdsci:
@stbaione thanks for the approval!
@rsuderman could I get a review/stamp?

shortfin/python/shortfin_apps/llm/components/batching/batching_trait.py
```diff
@@ -0,0 +1,40 @@
...
+from ..config_struct import ModelParams
+
+
+# BatchingTrait is an interface that ensures that the batching
+# engine always has the methods needed by the facade class.
+# This is used second in inheritance, usually with another base
+# class. Since this is not a concrete class and Python implements
+# a neat MRO, we can safely use multiple inheritance without worrying
+# about complications like the diamond problem.
+class BatchingTrait(ABC):
```
rsuderman, line 17: Name something that lines up with with an abstract class, e.g. `AbstractBatching`. Trait feels more like a parameter than a controlling structure.

rsuderman (approved):
Only question is how we are planning to swap the existing batcher with the new one. This one doesn't appear to delete the original batcher implementation so we should make sure we substitute and delete on the old impl.

shortfin/python/shortfin_apps/llm/components/batching/batching_trait.py
vinayakdsci: @rsuderman will fix this up in a follow up PR that's stacked on top of it.

PR #2076, opened 2025-08-22
Lisal.sep allocate lookup

lisaliu1:
refactor TrieCacheCode to achieve the following:
* allocation returns cache info metadata
* make the allocation class redundant so we can remove it in the end


stbaione:
I think that the way we're releasing pages is incorrect in this PR. We seem to be releasing all of the kvcache pages after a request is finished, instead of retaining them for potential prefix matches from future requests. What I believe should be happening, is that the pages should be retained, unless we run out of pages, at which point, we would evict w/ LRU strategy: see the `pseudo code` on page 15 [here](https://arxiv.org/pdf/2312.07104)

Let's say that we have the following requests that we send one at a time:

```python
req_1 = ' '.join(str(i) for i in range(64))
req_2 = ' '.join(str(i) for i in range(32))
```

`req_2` should have a prefix match w/ `req_1` (tokens [0...31]). 
However, since we're releasing all of the kvcache pages for `req_1` as soon as req_1 is finished, it doesn't appear that req_2 is matching with anything, when it should.

## Proof

I added a log statement in `free_cache_pages` ([here](https://github.com/nod-ai/shark-ai/pull/2076/files#diff-97a05f753de963b84211080bfd992240ae736f365fd3880ecbc2849a894f433dR790)), to see which pages are being freed from the cache back to the page pool after the request is finished, and it appears to be all of them:


### Post req_1
```bash
Freeing Cache Pages: [499, 500, 501, 502, 503, 504, 505, 506, 507, 508, 509, 510, 511]
```

### req_2
So, when `req_2` is sent to the server, the page pool is back at full capacity:

```bash
Page Pool State: PagePool(0/512 pages in use: 100.0% free)
```

Which means that we don't match w/ any kvcache pages:

```bash
[trie_attention_cache.py:346] MatchedPages=[]
```

### Allocated pages

Logging the pages that are allocated for each request for prefill, we see the following:

```bash
req_1: Allocated Cache Info: [PageInfo(index=511, pool=PagePool(4/512 pages in use: 99.21875% free)), PageInfo(index=510, pool=PagePool(4/512 pages in use: 99.21875% free)), PageInfo(index=509, pool=PagePool(4/512 pages in use: 99.21875% free)), PageInfo(index=508, pool=PagePool(4/512 pages in use: 99.21875% free))]
```

```bash
req_2: Allocated Cache Info: [PageInfo(index=492, pool=PagePool(2/512 pages in use: 99.609375% free)), PageInfo(index=493, pool=PagePool(2/512 pages in use: 99.609375% free))]
```

So, req_1 has pages: `[511, 510, 509, 508]` and req_2 has pages `[492, 493]`.

What I would expect is:

req_1: `[511, 510, 509, 508]`
req_2: `[511, 510]`

Let me know if I'm missing something or reading incorrectly.

stbaione:
Was finally able to reproduce your results w/ our local batch script. Was due to package version issue w/ Torch on my side. Lemme know when you're done with cleanup and I can review

PR #2080, opened 2025-08-22
Refactor transfer_between_blocks_if_needed

Alex-Vasile:
Use the placement data on the weights rather than needing to read it in from the pipelien_stage mapping, removing the need for the model to have `block_to_pipeline_stage` and `pipeline_stage_to_devices`.

Following up from https://github.com/nod-ai/shark-ai/pull/2072#discussion_r2294192044

### DIFF
--- a/sharktank/sharktank/examples/pipeline/export_ppffn_net.py
+++ b/sharktank/sharktank/examples/pipeline/export_ppffn_net.py
@@ -100,12 +100,8 @@ class PPFFN(ThetaLayer):
         )
 
     def forward(self, x: torch.Tensor):
-        # Note: In export_paged_llm, this replicate happens outside of forward().
-        #       Done here in this example for simplicity.
-        devices_0 = self.blocks[0].weight.devices
-        x = ReplicatedTensor(ts=x, shard_count=len(devices_0), devices=devices_0)
         for block_idx, block in enumerate(self.blocks):
-            x = transfer_between_blocks(x, block_idx, self.theta)
+            x = transfer_between_blocks(x, self.theta.tensor("blk", block_idx))
             x = block(x)
 
         return ops.unshard(x)
--- a/sharktank/sharktank/types/pipelining.py
+++ b/sharktank/sharktank/types/pipelining.py
@@ -16,8 +16,50 @@ from sharktank.types import (
 )
 
 
+def get_devices_from_block_tensors(
+    curr_block_tensors: dict[str, dict[str, AnyTensor]]
+) -> list[int] | None:
+    """
+    Helper function to extract devices from the current block tensors.
+    Ensures that all tensors in the block are on the same devices, raises an error if they're not.
+
+    Args:
+        curr_block_tensors: The tensors associated with the current block. From theta.tensor(...)
+
+    Returns:
+        A list of devices if the block is sharded, None if the block is not sharded.
+    """
+
+    def iter_equal(A, B):
+        return all(a == b for a, b in zip(A, B))
+
+    devices = -1
+    for subdict in curr_block_tensors.values():
+        for tensor in subdict.values():
+            if isinstance(tensor, ShardedTensor):
+                if devices == -1:
+                    devices = tensor.devices
+                assert (
+                    devices is not None
+                ), "Block contains a mix of sharded and unsharded tensors."
+                assert iter_equal(devices, tensor.devices), (
+                    f"All tensors in a block must be on the same devices."
+                    f"Found {devices} and {tensor.devices}."
+                )
+            else:
+                if devices == -1:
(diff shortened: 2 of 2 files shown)


sharktank/sharktank/examples/pipeline/export_ppffn_net.py
```diff
@@ -92,37 +92,21 @@ def create_theta(
...
+        x = ReplicatedTensor(ts=x, shard_count=len(devices_0), devices=devices_0)
         for block_idx, block in enumerate(self.blocks):
             x = block(x)
-            x = transfer_between_blocks_if_needed(
-                x,
-                block_idx,
-                self.block_to_pipeline_stage,
-                self.pipeline_stage_to_devices,
-            )
+            x = transfer_between_blocks_if_needed(x, block_idx, self.theta)
```
rsuderman, line 109: Change transfer to occur at the start of the block.

sharktank/sharktank/types/pipelining.py
```diff
@@ -30,34 +27,24 @@ def transfer_between_blocks_if_needed(
...
-    if curr_stage == next_stage:
+    block_count = len(theta.tensor("blk"))
+    if curr_block == block_count - 1:
         return x
 
-    curr_devices = pipline_stage_to_devices[curr_stage]
-    next_devices = pipline_stage_to_devices[next_stage]
+    curr_devices = x.devices
+    # Grab first weight from next block, all weights in a block are on the same devices.
+    next_weights = list(theta.tensor("blk", curr_block + 1).values())[0]["weight"]
```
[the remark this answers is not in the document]
rsuderman: You can drop the `+ 1` if you transfer at the start.

sharktank/sharktank/types/pipelining.py
```diff
@@ -30,34 +27,24 @@ def transfer_between_blocks_if_needed(
...
+    next_devices = next_weights.devices
 
     # If the current and next devices are the same, nothing to do.
     if all(d_curr == d_next for d_curr, d_next in zip(curr_devices, next_devices)):
```
rsuderman, line 50: Drop the check and just always transfer. The compiler should remove the redunant transfers.

sharktank/sharktank/types/pipelining.py
```diff
@@ -30,34 +27,24 @@ def transfer_between_blocks_if_needed(
...
     """
-    assert (block_to_pipeline_stage is None) == (pipline_stage_to_devices is None), (
-        "Either both block_to_pipeline_stage and pipline_stage_to_devices are None, "
-        "or both are not None."
-    )
-
-    # No pipeline parallelism, nothing to do.
-    if block_to_pipeline_stage is None:
+    # Unsharded tensors do not need to be moved.
+    if not isinstance(x, ShardedTensor):
```
rsuderman, line 36: Perform the transfer on unsharded versions too (it should be useful for replicated tensors as we don't need to handle transfers separately).

sharktank/sharktank/types/pipelining.py
```diff
@@ -30,34 +27,24 @@ def transfer_between_blocks_if_needed(
...
-    if curr_block == len(block_to_pipeline_stage) - 1:
-        return x
-
-    curr_stage = block_to_pipeline_stage[curr_block]
-    next_stage = block_to_pipeline_stage[curr_block + 1]
-
-    # If the current and next stages are the same, nothing to do.
-    if curr_stage == next_stage:
+    block_count = len(theta.tensor("blk"))
+    if curr_block == block_count - 1:
```
rsuderman, line 41: We can remove this if we do the start of the block

sharktank/sharktank/types/pipelining.py
rsuderman: Also below - we shouldn't need the old  devices for the transfer.

Alex-Vasile: `move_shards_to_new_devices` uses the old devices to insert a barrier instead of a transfer if doing on the same device. I was following the pattern used in our scatters.

sharktank/sharktank/types/pipelining.py
rsuderman: So some extra changes would be needed - namely move to new device always transfers. For the initial PP we can have the redundant transfers of deviceN -> deviceN then rely on the compiler to clean them up. Overall we should using information that shouldn't be actually required (e.g. where its coming from).

Without tensor parallelism we are less worried about device placement confusion.

sharktank/sharktank/types/pipelining.py
Alex-Vasile: Discussed offline and addressed comments with newest commits.

PR #2081, opened 2025-08-22
Replace KVCache list of tensors with simple CacheAllocation wrapper object

Alex-Vasile:
Instead of plumbing through a list of tensors everywhere, use a small object. No functionality changes for now, just initial signature changes.

### DIFF
--- a/sharktank/sharktank/examples/export_paged_llm_v1.py
+++ b/sharktank/sharktank/examples/export_paged_llm_v1.py
@@ -14,6 +14,7 @@ import torch
 
 from iree.turbine.aot import *
 from sharktank.layers.configs import LlamaModelConfig, LlamaHParams
+from sharktank.layers.paged_attention import CacheAllocation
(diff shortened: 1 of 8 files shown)


sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -90,8 +93,18 @@ def generate_batch_prefill(bs: int):
...
-                return model.prefill(tokens, start_pos, seq_lens, seq_block_ids, cs)
+            def _(
+                model: ServicePagedLlmModelV1,
+                tokens,
+                start_pos,
+                seq_lens,
+                seq_block_ids,
+                cs,
+            ):
+                cache_state = CacheAllocation(allocation=cs)
```
IanNod, line 104: Does this end up changing the exported IR at all?

Groverkss (approved):

rsuderman (approved):

PR #2082, opened 2025-08-22
Llm task remove exec requests

stbaione:
Final refactor for `invocation.py. Stacked on #2077 and #2078. Review those first.

- Removes references to `exec_requests` in `LlmTask`
- Introduces `LlmTaskInput` dataclass to feed into `LlmTask`
- Use a `build_invocation_process` function to hide the initialization logic from `batcher.py`
- Moves `responder` to `LlmInvocationProcess`
- Does a little more general cleanup on the interface

Draft because I need to fix up the unit tests, but have a couple items to get to while I wait for review in previous PRs.

### DIFF
--- a/shortfin/tests/apps/llm/components/invocation_test.py
+++ b/shortfin/tests/apps/llm/components/invocation_test.py
@@ -28,8 +28,8 @@ from shortfin_apps.llm.components.invocation import (
     DecodeTask,
     _pad_list,
 )
-from shortfin_apps.llm.components.kvcache.base_attention_cache import (
-    BasePagedAttentionCacheAllocation,
+from shortfin_apps.llm.components.kvcache.attention_cache_abstract import (
+    CacheInfo,
 )
 from shortfin_apps.llm.components.kvcache.page_pool import (
     PageInfo,
@@ -104,8 +104,12 @@ def staggered_exec_req_list(cache_ref_count, page_pool):
                 PageInfo(index=page_offset + i, pool=page_pool)
                 for i in range(len(req.input_token_ids) // 2 + 1)
             ]
-            allocation = BasePagedAttentionCacheAllocation(pages, cache=cache_ref_count)
-            req.allocation = allocation
+            req.allocated_cache_info = CacheInfo(
+                num_tokens=len(req.input_token_ids),
+                pages=pages,
+                pool=page_pool,
+            )
+            req.page_ids = [page.index for page in pages]
             page_offset += len(pages)
 
         yield exec_reqs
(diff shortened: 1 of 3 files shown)


shortfin/tests/apps/llm/components/invocation_test.py
```diff
@@ -178,7 +179,7 @@ def result_logits_none_indices_decode(decode_task, fiber):
     """Fixture to create a result logits device array."""
     vocab_size = 16
     batch_size = decode_task.req_count
-    seq_len = max(len(req.input_token_ids) for req in decode_task._exec_requests)
+    seq_len = decode_task._task_input.batch_seq_len
```
rsuderman, line 182: Its usually a pretty  bad sign to have a test reaching into classes to grab implementation specific members.  Is there a better way to do this?

shortfin/python/shortfin_apps/llm/components/invocation.py
```diff
@@ -1,29 +1,78 @@
...
+    start_positions: Optional[List[int]] = None
+
+    @property
+    def batch_seq_len(self):
+        seq_stride = self.seq_stride
+        bsl = max(len(tokens) for tokens in self.input_tokens)
+        return int(math.ceil(bsl / seq_stride) * seq_stride)
+
+    @staticmethod
+    def from_exec_requests(
```
rsuderman, line 38: This construction should be in `batcher` not `invocation`. If you are breaking the `exec_req` requirement we don't want to continue its use in `invocation.py`.

shortfin/python/shortfin_apps/llm/components/invocation.py
```diff
@@ -355,35 +335,67 @@ async def run(self):
...
+    functions: dict[int, sf.ProgramFunction],
+    program_isolation: sf.ProgramIsolation,
+    responder: LlmTaskResponder,
+) -> LlmInvocationProcess:
+    task_inputs = LlmTaskInput.from_exec_requests(
+        exec_requests=exec_requests,
+        seq_stride=page_cache.tokens_per_page,
+    )
+
+    llm_task_cls = PrefillTask if phase == InferencePhase.PREFILL else DecodeTask
```
rsuderman, line 387: This feels wrong, having an enum just for invoking the wrong contruction means you are probably making this decision lower in the stack than you should. If you create the `LlmTask` in `batcher` you won't have to worry about this selection logic.

rsuderman (changes requested):

rsuderman (approved):

PR #2085, opened 2025-08-22
Add Wave gemm optimizations

aviator19941:
Adds the Wave gemm optimizations of SchedulingType.PREFETCH and updates the WaveCompileOptions to use `use_buffer_load_ops=True` and `waves_per_eu=2`.


jinchen62 (approved):
Nice!

aviator19941:
> Does this affect numerics at all? Also can you share the expected performance delta?

No it shouldn't, I ran PPL and run_llm_vmfb on Friday and got the same numerics as the old wave kernel. I will check expected performance delta.

PR #2086, opened 2025-08-22
Migrate reshaping ops and test each

KyleHerndon:
The patch may look big, but actually, all the op implementations are exact copies, with only exceptions for addressing uncovered issues, such as with the type signature inconsistency with view.

I intend to repeat this migration for all the ops.

Core improvements are:
1) that the documentation (i.e. the signature) is right next to the implementation which makes understanding the code easier
2) the override order is now implicit in a single file, as opposed to across files which consequently means the order of imports in __init__ does not embed any logic itself, making the override ordering behavior easier to track
3) Organizing ops by logical category in directories rather than alphabetically in a monolithic file.
3) Tests!


sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -181,7 +123,9 @@ def should_override(*types: tuple[type]) -> bool:
             return False
         return all(replicated_if_tensor(t) for t in types)
 
-    for func_name in signatures.__all__:
+    all_signatures = list(signatures.__all__)
+    all_signatures.extend(_trivially_replicable_ops)
+    for func_name in all_signatures:
```
sogartar, line 128: This is not the only place using this pattern. I kind of expected that this may be a problem if we stop putting all ops in signatures. The other place is [here](https://github.com/nod-ai/shark-ai/blob/main/sharktank/sharktank/ops/sharded_impls.py#L144). This patching checks if device placement is compatible. Meaning, sharded tensors span the same devices. We probably lack tests for these particular ops that are going to be left out, where we test that we catch wrong device placement errors.

I was thinking that it may be a design flaw to assume that all ops declarations are at `signatures.__all___`. We need some other way to query all ops. As an example due to this requirement [this test](https://github.com/nod-ai/shark-ai/blob/3f48cabc0ed8bc0a8836ac47e2aca3922ef2e546/sharktank/tests/ops/sharded_test.py#L2100-L2114) is xfailed. We probably should be able to ask the ops registry to give us the registered ops.

sharktank/sharktank/ops/sharding/__init__.py
sogartar, line 1: This file is not needed.

sharktank/sharktank/ops/sharding/utils.py
sogartar, line 1: There another file added in this PR `sharding_utils.py`. Why isn't there only one such file?

sogartar (commented):
It would be nice to not jump across files to explore implementations and this gives us the ability to reorder resolution order for each op individually.
I am curious if having the ops declarations and implementations always constructed together would be a problem at some point. Would we ever want all declarations constructed first before the implementations?

Some multiple dispatch frameworks provide a priority ordinal for each implementation to establish a resolution order. Then you can decouple implementation order from resolution order. But then you would have to manage the priority numbers which would be another hassle.

sharktank/sharktank/ops/sharding/utils.py
KyleHerndon: Fixed.

sharktank/tests/ops/shaping/__init__.py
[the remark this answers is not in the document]
KyleHerndon: It changes imports from shaping.view.view to shaping.view, which I prefer, but maybe this can be simplified in ops/__init__.py when everything is migrated?

sharktank/sharktank/ops/sharded_impls.py
KyleHerndon: Good idea! I added a global registry. I think this is better as well, but let me know if you have any other ideas.

Alex-Vasile (changes requested):
Current failures are due to error somewhere in `.view` for `ReplicatedTensor`.

Please split the tests out into their own PR to land first.

PR #2100, opened 2025-08-26
[Sharktank] Improve and use transfer_between_blocks

Alex-Vasile:
1. Improve `transfer_between_blocks` to support var args
2. Add transfer call to llm.py to transfer the required variables between blocks

### DIFF
--- a/sharktank/sharktank/types/pipelining.py
+++ b/sharktank/sharktank/types/pipelining.py
@@ -59,8 +59,8 @@ def get_devices_from_block_tensors(
 
 
 def transfer_between_blocks(
-    x: AnyTensor, curr_block_tensors: dict[str, dict[str, AnyTensor]]
-) -> ShardedTensor:
+    *xs: AnyTensor | None, curr_block_tensors: dict[str, dict[str, AnyTensor]]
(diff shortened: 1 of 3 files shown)


sharktank/sharktank/types/pipelining.py
```diff
@@ -59,8 +59,8 @@ def iter_equal(A, B):
 
 
 def transfer_between_blocks(
-    x: AnyTensor, curr_block_tensors: dict[str, dict[str, AnyTensor]]
-) -> ShardedTensor:
+    *xs: AnyTensor | None, curr_block_tensors: dict[str, dict[str, AnyTensor]]
```
rsuderman, line 62: Drop the `None` support. Its not really useful and makes the usecase more complicated.

rsuderman (approved):

sharktank/sharktank/types/pipelining.py
Alex-Vasile: I have the None support since certain things in lllm.prefill are labelled as optional (e.g. `start_position` and `attention_mask`). I can handle those at the call site, but that will make it even longer there.

PR #2105, opened 2025-08-27
Add --iree-hip-enable-tensor-ukernels to compiler flags to enable tensor ukernels

jtuyls:
The `--iree-hip-enable-tensor-ukernels` flag is required after https://github.com/iree-org/iree/pull/21794 to enable the ping-pong ukernels. This replaces the `--iree-codegen-enable-default-tuning-specs=true` flag, however the latter is kept around because it includes tuned attention configs as well.


jtuyls:
https://github.com/iree-org/iree/pull/21794 git merged, so this PR is now needed to avoid performance regressions. Could someone help review this? @IanNod @jinchen62 @aviator19941

PR #2107, opened 2025-08-27
[Sharktank] Replumb KV Quantizers

Alex-Vasile:
- Move quantizer from arg to state inside KVCache.
- Move packing and unpacking of quantized tensors inside kvcache read and writes.
- Remove sharding related checks.

### DIFF
--- a/sharktank/sharktank/layers/paged_attention.py
+++ b/sharktank/sharktank/layers/paged_attention.py
@@ -418,12 +418,6 @@ class PagedAttention:
             device=device,
         )
 
-    def shard_state(self, state: List[torch.Tensor]) -> List[torch.Tensor]:
-        return self.kv_cache.shard_state(state=state)
-
-    def unshard_state(self, state: List[torch.Tensor]) -> List[torch.Tensor]:
-        return self.kv_cache.unshard_state(state=state)
-
     @property
     def pad_sequence_stride(self) -> int:
         return self.block_seq_stride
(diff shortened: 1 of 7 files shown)


sharktank/sharktank/layers/paged_llama_attention_block.py
```diff
@@ -99,8 +97,8 @@ def __init__(
                     matmul_kernel=matmul_kernel,
                 ),
             )
-            self.k_quantizer = self.attn_k.q_output
-            self.v_quantizer = self.attn_v.q_output
+            self.paged_attention.kv_cache.k_quantizer = self.attn_k.q_output
+            self.paged_attention.kv_cache.v_quantizer = self.attn_v.q_output
```
Alex-Vasile, line 101: Nevermind, this won't work since these are theoretically different per stage, and we are passing around a single PagedAttention object to all of them.

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -143,7 +143,9 @@ def unpack_to_raw_tensor(tensor: AnyTensor) -> AnyTensor:
     return tensor
 
 
-def pack_raw_tensor(tensor, quantizer):
+def pack_raw_tensor(
```
archana-ramalingam, line 146: Is there a better place for the `pack_raw_tensor` /`unpack_to_raw_tensor` ? 
Probably `shark-ai/sharktank/sharktank/types/tensors.py` containing `unbox_tensor` or entirely a separate `tensor_utils.py` ?

IanNod (approved):
Test failure is unrelated to this PR

PR #2108, opened 2025-08-27
[Tuner] Add time limit to benchmark phases

RattataKing:
Add time limit options (in minutes) for dispatch benchmark and model benchmark phases.
When a phase’s deadline is reached, benchmarking stops early, returns the results collected so far, and the pipeline proceeds to the next phase. Baseline benchmarking is excluded from the time limit.

common.py:
- Import time module to use `time.monotonic`
- Add `TimeBudget` class

libtuner.py:
- `benchmark()` will pass `benchmark_time` to `benchmark_candidates()`, which will then create `time_budget = TimeBudget.for_minutes(benchmark_time)`
- Pass `time_budget` to multiprocessing progress wrapper

model_tuner.py:
- Add `--dispatch-benchmark-timeout-mins`
- Add `--model-benchmark-timeout-mins`

common_test.py:
- Add unit test for `time_budget`

### DIFF
--- a/sharktuner/sharktuner/common.py
+++ b/sharktuner/sharktuner/common.py
@@ -71,16 +71,22 @@ class TimeBudget:
     deadline: Optional[float] = None  # Absolute monotonic time (seconds).
 
     @classmethod
-    def for_minutes(cls, minutes: Optional[float]):
-        """Create a budget that lasts 'minutes' from now."""
+    def for_minutes(cls, minutes: Optional[float], now: Optional[float] = None):
+        """Create a budget that lasts 'minutes' from a given 'now' (monotonic seconds)."""
         if minutes is None or minutes <= 0:
             return None
-        return cls(time.monotonic() + (minutes * 60.0))
+        if now is None:
+            now = time.monotonic()
+        return cls(now + (minutes * 60.0))
 
-    def expired(self, current_time: float = time.monotonic()) -> bool:
+    def expired(self, current_time: Optional[float] = None) -> bool:
+        if current_time is None:
+            current_time = time.monotonic()
         return self.deadline is not None and current_time >= self.deadline
 
-    def remaining(self, current_time: float = time.monotonic()) -> Optional[float]:
+    def remaining(self, current_time: Optional[float] = None) -> Optional[float]:
+        if current_time is None:
+            current_time = time.monotonic()
         if self.deadline is None:
             return None
         return max(0.0, self.deadline - current_time)
--- a/sharktuner/tests/common_test.py
+++ b/sharktuner/tests/common_test.py
@@ -438,11 +438,16 @@ def test_time_budget():
     assert time_budget == None
     time_budget = common.TimeBudget.for_minutes(0)
     assert time_budget == None
-    time_budget = common.TimeBudget.for_minutes(0.1)  # Set a 6s timer.
+
+    base = 1.234
+    time_budget = common.TimeBudget.for_minutes(
+        minutes=0.1, now=base
+    )  # Set a 6s timer from timestamp `base`.
     assert time_budget != None
-    assert time_budget.remaining() > 1
-    assert time_budget.expired() == False
-    current_time = time.monotonic()
-    assert time_budget.remaining(current_time + 3) > 2
-    assert time_budget.expired(current_time + 3) == False
-    assert time_budget.expired(current_time + 6) == True
+    assert time_budget.remaining(base) > 1
+    assert time_budget.expired(base) == False
+
+    assert time_budget.remaining(base + 3) > 2
+    assert time_budget.expired(base + 3) == False
+    assert time_budget.remaining(base + 6) == 0
+    assert time_budget.expired(base + 6) == True


sharktuner/sharktuner/libtuner.py
```diff
@@ -106,6 +107,28 @@ def get_candidate_vmfb_filename(self, candidate_id: int) -> str:
         return f"{candidate_id}.vmfb"
 
 
+@dataclass
```
bangtianliu, line 110: Consider moving this class to common.py for better code organization, and add unit tests to verify its behavior.

bangtianliu (changes requested):

sharktuner/sharktuner/libtuner.py
RattataKing: moved to common.py, thanks

sharktuner/sharktuner/libtuner.py
```diff
@@ -345,6 +345,12 @@ def parse_arguments(
             "if fully covered."
         ),
     )
+    general_args.add_argument(
+        "--dispatch-benchmark-time-mins",
```
bangtianliu, line 349: We have two benchmarking phases: first phase is for benchmarking selected dispatch using generated td specs, second one is for benchmarking the whole model using selected td specs. it seems that this parameter applied to both two phases since it is just added to benchmark function directly. Update the naming and also add explanation as code comment here.

RattataKing: moved arg to `model_tuner.py`, thanks

bangtianliu (approved):
Thanks!

sharktuner/sharktuner/common.py
```diff
@@ -63,6 +64,28 @@ def __exit__(
         return self.mlir_ctx.__exit__(exc_type, exc_value, traceback)
 
 
+@dataclass
+class TimeBudget:
+    """Wall-clock deadline helper based on time.monotonic()."""
+
+    deadline: Optional[float] = None  # Absolute monotonic time (seconds)
```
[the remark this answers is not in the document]
RattataKing: updated, thanks

sharktuner/sharktuner/common.py
```diff
@@ -63,6 +64,28 @@ def __exit__(
...
+    def for_minutes(cls, minutes: Optional[float]):
+        """Create a budget that lasts 'minutes' from now."""
+        if minutes is None or minutes <= 0:
+            return None
+        return cls(monotonic() + (minutes * 60.0))
+
+    def expired(self) -> bool:
+        return self.deadline is not None and monotonic() >= self.deadline
+
+    def remaining(self) -> Optional[float]:
```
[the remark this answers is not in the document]
RattataKing: updated, thanks

sharktuner/tests/common_test.py
```diff
@@ -430,3 +431,17 @@ def test_determine_td_specs_to_link(
...
+    time_budget = common.TimeBudget.for_minutes(-5)
+    assert time_budget == None
+    time_budget = common.TimeBudget.for_minutes(0)
+    assert time_budget == None
+    time_budget = common.TimeBudget.for_minutes(0.1)  # set a 6s timer
+    assert time_budget != None
+    time.sleep(3)
+    assert time_budget.remaining() > 1
+    assert time_budget.expired() == False
+    time.sleep(3)
```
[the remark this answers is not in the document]
RattataKing: updated, thanks

sharktuner/tests/common_test.py
```diff
@@ -438,10 +438,11 @@ def test_time_budget():
...
     assert time_budget == None
-    time_budget = common.TimeBudget.for_minutes(0.1)  # set a 6s timer
+    time_budget = common.TimeBudget.for_minutes(0.1)  # Set a 6s timer.
     assert time_budget != None
-    time.sleep(3)
     assert time_budget.remaining() > 1
     assert time_budget.expired() == False
-    time.sleep(3)
-    assert time_budget.expired() == True
+    current_time = time.monotonic()
```
kuhar, line 445: you don't need to query the current time here, just any float value as the initial time

sharktuner/sharktuner/common.py
```diff
@@ -63,6 +64,28 @@ def __exit__(
...
+    """Wall-clock deadline helper based on time.monotonic()."""
+
+    deadline: Optional[float] = None  # Absolute monotonic time (seconds).
+
+    @classmethod
+    def for_minutes(cls, minutes: Optional[float]):
+        """Create a budget that lasts 'minutes' from now."""
+        if minutes is None or minutes <= 0:
+            return None
+        return cls(time.monotonic() + (minutes * 60.0))
```
kuhar, line 78: I think we also need to make this a parameter for the purpose of testing

sharktuner/sharktuner/common.py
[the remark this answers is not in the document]
RattataKing: We shouldn't do this:
```def expired(self, current_time: float = time.monotonic()) -> bool:
def remaining(self, current_time: float = time.monotonic()) -> Optional[float]:
```
> The default values are evaluated at the point of function definition in the _**defining**_ scope, see [this](https://docs.python.org/3/tutorial/controlflow.html).

PR #2110, opened 2025-08-27
[Sharktank] Pipeline parallelism version of rotary embedding layer

Alex-Vasile:


sharktank/sharktank/layers/rotary_embedding.py
```diff
@@ -66,24 +69,96 @@ def apply_batched_mask(
         return self._rotary_layer(q=xt, sincos_cache=mask)
 
 
+class ReplicatedRotaryLayer(CachedRotaryLayer):
```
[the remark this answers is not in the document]
dezhiAmd: In my humble opinion, a new class may be a better choice for adding more feature down the road.
Today, ReplicatedRotaryLayer is a subclass of CachedRotaryLayer that handles replicated tensors across devices.
It helps rotary embeddings are computed and applied consistently across shards.

In class ReplicatedRotaryLayer , compute_batch_mask() is Overload. Accepts ReplicatedTensor and devices to compute sine/cosine masks per shard. Handles both sharded and replicated input scenarios. Without the new class, there might be many if/esle

sharktank/sharktank/layers/rotary_embedding.py
```diff
@@ -66,24 +69,96 @@ def apply_batched_mask(
...
+        assert (start_positions is not None) or (
+            devices is not None
+        ), "Either start_positions or devices must be provided to properly place the replicated tensor"
+
+        if devices is None:
+            devices = start_positions.devices
+
+        if start_positions is None:
+            # Create from a single shard and introduce transfers since there's no start positions to hint what device they should go on
+            # TODO: Is this needed, or will IREE figure out what device to put it on based downstream uses?
```
dezhiAmd, line 101: Last time, I wrote a TODO. One reviewer' comment is that I should create an issue. Is creating an issue for TODOs a common practice in shark-ai?

dezhiAmd:
Try to verify my understanding here:
Stage 0 (Device 0):
Applies token embedding and rotary embedding.
Produces hidden states (h) and passes them to Stage 1.

Stage 1 (Device 1):
Runs attention block with rotary embeddings.
Uses transfer_between_blocks() to move tensors and metadata.
Produces updated hidden states and passes them to Stage 2.

Stage 2 (Device 2):
Runs another attention block and output layers.
Produces final logits for inference.

If that is correct, how do we test this PR? Do we need a test to make sure the newly added feature is not broken?

sharktank/sharktank/layers/rotary_embedding.py
Alex-Vasile: In this case I don't want to create an issue to resolve later, I would like to get an answer for it before landing this PR. Special casing this adds a lot of complexity and I would like to not do it if possible.

Alex-Vasile:
> Try to verify my understanding here: Stage 0 (Device 0): Applies token embedding and rotary embedding. Produces hidden states (h) and passes them to Stage 1.
> 
> Stage 1 (Device 1): Runs attention block with rotary embeddings. Uses transfer_between_blocks() to move tensors and metadata. Produces updated hidden states and passes them to Stage 2.
> 
> Stage 2 (Device 2): Runs another attention block and output layers. Produces final logits for inference.
> 
> If that is correct, how do we test this PR? Do we need a test to make sure the newly added feature is not broken?

I can add tests for it, but what we're expecting is that the output for unsharded tensors remains completely unchanged. For pipelined usage, the resulting tensor should have the same values as the unsharded variant except that it will be a ReplicatedTensor with 1 shard that's placed on the provided devices.

sharktank/tests/layers/rotary_embedding_test.py
```diff
@@ -4,14 +4,28 @@
...
 
-from sharktank.layers import build_rotary_layer
+from parameterized import parameterized_class
 
+from sharktank import ops
+from sharktank.layers import CachedRotaryLayer, build_rotary_layer
+from sharktank.types import AnyTensor, ReplicatedTensor
+
+
+def validate(
```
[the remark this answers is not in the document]
IanNod: The export itself shouldn't matter if it is individual layer vs full model, so something like [this](https://github.com/nod-ai/shark-ai/blob/c37e31d9c7e27def163d42c1389dd865b9c8f6f6/sharktank/sharktank/utils/export.py#L182C5-L182C22) should work, with the exception that it currently expects the "model" to have a sample_inputs method. We really should update that to allow sample from args instead of the "model" to supply them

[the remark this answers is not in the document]
Alex-Vasile: Added a test for iree execution.

sharktank/tests/layers/rotary_embedding_test.py
```diff
@@ -66,52 +85,168 @@ def validate(xq, em, rope_dims, rope_freq_base, interleaved):
...
+    def test_rotary_table_eager_replicated(self):
+        self.pipeline_stage_to_device_map = [[0], [1], [2], [3], [4], [5], [6], [7]]
+        rotary_layer = self.create_rotary_layer()
+        for devices in self.pipeline_stage_to_device_map:
+            xq = ReplicatedTensor(ts=[self.xq], devices=devices)
+            em = rotary_layer(xt=xq)
+            assert em.devices == xq.devices
+            self.validate(em, xq)
+
+    @is_mi300x
```
IanNod, line 215: Can we also run on MI35X? Can enable with @is_mi350x and would need to update iree-hip-target flag accordingly, I'm not positive if that runner is multi-gpu or not though

sharktank/tests/layers/rotary_embedding_test.py
```diff
@@ -66,52 +85,168 @@ def validate(xq, em, rope_dims, rope_freq_base, interleaved):
...
         )
-        em = default_layer(xt=xq)
-        validate(
-            xq=xq,
-            em=em,
-            rope_dims=rope_dims,
-            rope_freq_base=rope_freq_base,
-            interleaved=False,
+
+    def export_compile_run_layer(self, layer: CachedRotaryLayer, xq: AnyTensor) -> None:
```
IanNod, line 128: The export process for tests really needs some cleanup (which is outside scope of this PR). Can you add a todo to clean up later once that is resolved?

IanNod (commented):
Couple nit comments but otherwise looks good to me once failing tests are fixed

sharktank/tests/layers/rotary_embedding_test.py
Alex-Vasile: Yes, that's something could really use. Will helper greatly with bisecting model issues and with bringup.

Alex-Vasile:
@IanNod current test failure is unrelated, it's the same paged_attention one we discussed yesterday.

PR #2111, opened 2025-08-27
Add model based VMFB tool and perplexity option

rsuderman:
Reworked tooling to have a torch backed perplexity tool this allows us to test llama numerics without requiring iree execution.

### DIFF
--- a/sharktank/sharktank/tools/perplexity_llm_model.py
+++ b/sharktank/sharktank/tools/perplexity_llm_model.py
@@ -12,7 +12,7 @@ import tokenizers
 from datasets import load_dataset
 
 from sharktank.models.llm.config import ServiceConfig
-from sharktank.utils.llm_utils import IreeInstance, LlmInstance
+from sharktank.utils.llm_utils import TorchInstance, LlmInstance, llama_config_page_size
 
 
 class Tokenizer:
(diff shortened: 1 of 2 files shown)


sharktank/sharktank/tools/perplexity_llm_model.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def decode(self, sequences) -> list[str]:
+        """Decodes a batch of sequences to text."""
+        return self.t.decode_batch(sequences)
+
+
+def main(dataset, irpa, tokenizer, perplexity):
+    tokenizer = Tokenizer(tokenizer)
+
+    with open(dataset, "r") as dataset:
```
IanNod, line 36: Should this be moved to an llm_utils dataset load function instead?

sharktank/sharktank/tools/perplexity_llm_model.py
```diff
@@ -0,0 +1,87 @@
...
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import argparse
+import dataclasses
+import json
+import tokenizers
+
+from datasets import load_dataset
+
+from sharktank.models.llm.config import ServiceConfig
```
IanNod, line 15: I don't see ServiceConfig used in this script. Remove here?

rsuderman: Removed along with dataclasses

PR #2114, opened 2025-08-27
[Sharktank] Remove old unused code

Alex-Vasile:
- `static_tables` from `LlamaModelConfig`
- `_assert_device` from `BaseCausalLMModel`

### DIFF
--- a/sharktank/sharktank/utils/tokenizer.py
+++ b/sharktank/sharktank/utils/tokenizer.py
@@ -108,12 +108,3 @@ def _create_transformers_tokenizer(model_path: os.PathLike) -> InferenceTokenize
             return t.vocab_size
 
     return _TransformersTokenizer(t)
-
-
-if __name__ == "__main__":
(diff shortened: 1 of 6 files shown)


sharktank/sharktank/utils/tokenizer.py
```diff
@@ -108,12 +108,3 @@ def _vocab_size(self):
...
     return _TransformersTokenizer(t)
-
-
-if __name__ == "__main__":
-    t = load_tokenizer("/home/stella/tmp/downloaded_open_llama_3b")
-    enc, lens = t.encode(["Hi there", "who are you?"], pad_to_multiple_of=16)
-    print(enc)
-    print(lens)
-    dec = t.decode(enc, lens)
-    print(dec)
```
archana-ramalingam, line 119: Since we don't have an independent tokenizer test, this has been useful before to verify the tokenizer's function independently, without writing much extra code. Could enhance this with flexible `args` or add a test.

PR #2119, opened 2025-08-28
Fix tensor tracing for multiple tensors and add tracing to load_iree_module by default

KyleHerndon:
I was unable to export a full model without these changes


sharktank/sharktank/ops/default_impls.py
```diff
@@ -752,10 +752,13 @@ def to_default(tensor: Tensor, *args, **kwargs) -> PrimitiveTensor:
...
-        raise ValueError("Tracing more than one tensor at a time is not supported.")
-    tensor = unbox_tensor(unshard(tensors[0]))
-    iree.turbine.ops.iree.trace_tensor(key, tensor)
+    # Skip tracing during export
+    if torch.compiler.is_compiling():
+        return
+    
+    for i, tensor in enumerate(tensors):
+        tensor = unbox_tensor(unshard(tensor))
+        iree.turbine.ops.iree.trace_tensor(f"{key}.{i}", tensor)
```
sogartar, line 761: The problem is that this will cause tensors trace in eager and IREE to have different keys.

sogartar (approved):
Looks OK.
The PR title and description needs to be updated to reflect the actual change.

sharktank/sharktank/utils/iree.py
```diff
@@ -30,6 +30,8 @@
 )
 from sharktank.utils import verify_exactly_one_is_not_none
 from .tree import Tree
+from .logging import get_logger
+from .debugging import flags as debug_flags
```
sogartar, line 34: I think this import may interfere with the [import below](https://github.com/nod-ai/shark-ai/blob/944d32e3213324b9b9327b45133edaf063f7f3b7/sharktank/sharktank/utils/iree.py#L630) when constructing the callback.
Somehow [here](https://github.com/nod-ai/shark-ai/blob/944d32e3213324b9b9327b45133edaf063f7f3b7/sharktank/sharktank/utils/iree.py#L642) `debugging` is not defined. I guess it tries to run the import lazily when calling the callback and it fails.
This is the Python function that gets called by IREE during the module execution.
Quite wired.

sharktank/sharktank/utils/iree.py
```diff
@@ -360,6 +362,20 @@ def load_iree_module(
     parallel_size = tensor_parallel_size * pipeline_parallel_size
     assert parallel_size == len(devices)
 
+    # Setup debug sink for tensor tracing if enabled and not provided by caller
+    if debug_sink is None:
+        if debug_flags.enable_tensor_trace and debug_flags.trace_path is not None:
+            if len(devices) > 1:
+                logger = get_logger("sharktank.utils.iree")
```
sogartar, line 369: The customary is to have
```
logger = get_logger(__name__)
```
at module scope.

PR #2132, opened 2025-08-28
[Sharktank] Replumb Attention and embedding masks

Alex-Vasile:
Instead of creating the masks outside of the model, passing them around, and relying on IREE to fuse them for us, this PR changes it so that each mask is created right before being used. This will make it easier to fuse down the line and requires less state to be passed around.

### DIFF
--- a/sharktank/sharktank/layers/paged_attention.py
+++ b/sharktank/sharktank/layers/paged_attention.py
@@ -575,7 +575,6 @@ class PagedAttention:
             scale=scale,
             sliding_window=sliding_window,
             sink=sink,
-            is_prefill=False,
         )
 
     def paged_attention(
@@ -594,7 +593,6 @@ class PagedAttention:
         fake_quant: Optional[bool],
         softcap: Optional[float],
         scale: Optional[float],
-        is_prefill: bool,
         sliding_window: Optional[int] = None,
         sink: Optional[torch.Tensor] = None,
     ):
@@ -606,6 +604,7 @@ class PagedAttention:
                 page_ids=seq_block_ids,
             )
 
+        is_prefill = q.shape[1] != 1
         if is_prefill:
             # q, k, v, x, and h all have the same .shape[1] (batch_seqlen)
             input_mask = create_input_mask(seq_lens, q.shape[1])
@@ -682,5 +681,4 @@ class PagedAttention:
             scale=scale,
             sliding_window=sliding_window,
             sink=sink,
-            is_prefill=True,
         )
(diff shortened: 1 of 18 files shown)


sharktank/tests/layers/paged_llama_attention_block_test.py
```diff
@@ -171,15 +171,15 @@ def forward(self, h, seq_block_ids, cache_state: list[torch.Tensor]):
 ]
 _CONTEXT_LEN = [2048]
 _DT_CASES = [
-    (torch.float32, 1e-4, 1e-4),
+    # (torch.float32, 1e-4, 1e-4),  # TODO: Re-enable after fixing.
```
oyazdanb, line 174: instead of removing the test for float32, we need to triage and see why it's failing with the new changes.

sharktank/tests/layers/paged_llama_attention_block_test.py
```diff
@@ -179,7 +179,7 @@ def forward(self, h, seq_block_ids, cache_state: list[torch.Tensor]):
 
 _SINK_CASES = [  # sliding_window, sink_scale
     (None, None),  # base path
-    (19, 0.25),  # sink path enabled
+    # (19, 0.25),  # sink path enabled  # TODO: Re-enable after fixing.
```
archana-ramalingam, line 182: Can we file and link to an issue?

Alex-Vasile:
There are iree compile failures when running on CPUs. They look like a duplicate of: https://github.com/iree-org/iree/issues/21828

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -386,6 +387,9 @@ def __init__(
         block_seq_stride: int = 16,
         cache_dtype: torch.dtype = torch.float32,
         attn_dtype: torch.dtype = torch.float32,
+        activation_dtype: torch.dtype = torch.float32,
+        use_rope: bool,
+        attention_chunk_size: int | None,
```
dezhiAmd, line 392: Is this related to chunk prefill? If yes, I wonder whether it worthwhile to add one sentence in PR description?

sharktank/tests/layers/paged_llama_attention_block_test.py
Alex-Vasile: #2156

sharktank/sharktank/layers/paged_attention.py
rsuderman: No this is related to chunked attention which is a feature used by Llama4

sharktank/tests/models/llama/attention_test.py
```diff
@@ -101,7 +102,7 @@ def test(self, prefill_offset: bool):
             input_tensor,
             start_positions=start_positions,
             embedding=attention_embedding,
-            attention_mask=torch.zeros(1, seq_len, seq_len, dtype=torch.float32),
```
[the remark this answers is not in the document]
Alex-Vasile: Addressed in #2160.

rsuderman (approved):

PR #2139, opened 2025-08-30
Refactor paged attention

dezhiAmd:


sharktank/sharktank/layers/paged_llama_attention_block.py
```diff
@@ -71,47 +76,93 @@ def __init__(
...
+        )
+        self.add_module(
+            "attn_v",
+            LinearLayer(
+                theta("attn_v"),
+                fake_quant=self.fake_quant,
+                matmul_kernel=matmul_kernel,
+            ),
+        )
+        self.paged_attention = create_paged_attention(
```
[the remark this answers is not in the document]
dezhiAmd: Agree. But please allow me to do that in a separate PR. Otherwise, if there is a regression, there is a big revert

sharktank/sharktank/layers/paged_llama_attention_block.py
```diff
@@ -182,33 +232,208 @@ def gqa_attention(
...
+        h = h + attn_output.to(dtype=h.dtype)
+        return h
+
 
-        elif self.attn_type == "mla":
-            xq, xk, xv = self.latent_attn(
-                x,
-                embedding=embedding,
-                embedding_batch_mask=embedding_batch_mask,
+class PagedLlamaAttentionBlockMla(PagedLlamaAttentionBlock):
```
dan-garvey, line 335: nit: prefer PagedLlamaAttentionMLABlock

dan-garvey (changes requested):
Why is this refactor valuable? 

Are we planning to add many more attn_types? I feel like this just increases the code bloat without providing a meaningful mental abstraction?

sharktank/sharktank/layers/paged_llama_attention_block.py
dezhiAmd: This refactor separates gqa from mla, make it easier to add pipeline parallelism.

dan-garvey (approved):

PR #2140, opened 2025-08-31
Integrate wave extend attention

aviator19941:
This PR uses the extend attention kernel in Wave and integrates it into sharktank as a CustomOp. There are a few things which need to be fixed:

1. Test needs to be updated to include compile, run, and numerics checks

Run test:
```
pytest -v -s sharktank/tests/kernels/wave/extend_attention_test.py --iree-hip-target=gfx950 --iree-device='hip://1'
```


sharktank/tests/kernels/wave/extend_attention_test.py
```diff
@@ -0,0 +1,133 @@
...
+            "--iree-dispatch-creation-enable-early-trunc-fusion=true",
+        ]
+
+    @pytest.mark.parametrize(
+        "query_seq_len, kv_seq_len, s, num_query_heads, head_size, num_kv_heads, head_size_kv, max_len_extend",
+        [
+            (512, 512, 3, 16, 128, 1, 128, 458),
+        ],
+    )
+    def test_extend_attention_export_compile_run(
```
archana-ramalingam, line 52: Do we want to run this on cpu as well? If not we need a @mi300 fixture here.

sharktank/tests/kernels/wave/extend_attention_test.py
```diff
@@ -0,0 +1,133 @@
...
+from sharktank.utils.testing import assert_cosine_similarity_close
+import iree.compiler as ireec
+import iree.runtime as ireert
+from pathlib import Path
+import numpy as np
+from sharktank.utils.testing import is_mi350x, IreeFlags
+
+
+@pytest.mark.usefixtures("iree_flags")
+class TestExtendAttention:
```
archana-ramalingam, line 30: Do we want to run this on cpu as well? If not we need a `@is_mi300x` fixture here.

archana-ramalingam: Fixed

sharktank/tests/kernels/wave/extend_attention_test.py
```diff
@@ -0,0 +1,138 @@
...
+@is_mi300x
+@pytest.mark.usefixtures("iree_flags")
+@pytest.mark.skipif(
+    torch.__version__ >= (2, 6),
+    reason="Wave extend attention kernel requires torch version >= 2.6 ",
+)
+class TestExtendAttention:
+    def hip_flags(self):
+        return [
+            "--iree-hip-target=gfx942",
```
IanNod, line 38: Shouldn't hardcode this for MI300 but instead use {hip-target}

PR #2146, opened 2025-09-02
[Sharktank] Remove model.paged_attention

Alex-Vasile:
Instead of using one of the block's PagedAttention object directly, this wraps it in a function to allow for pipeline support when creating the allocation.

Will add the pipeline support in a follow up PR to limit this one to refactors

### DIFF
--- a/sharktank/sharktank/evaluate/perplexity_iree.py
+++ b/sharktank/sharktank/evaluate/perplexity_iree.py
@@ -174,9 +174,11 @@ class PerplexityIree:
 
         pp = self.pipeline_parallelism_size
         tp = self.tensor_parallelism_size
+        assert pp == 1
+        assert tp == 1
         block_count = hp.block_count
-        block_to_pipeline = [i * pp // block_count for i in range(block_count)]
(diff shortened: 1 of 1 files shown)


sharktank/sharktank/models/llm/llm.py
```diff
@@ -115,7 +115,13 @@ def __init__(self, theta: Theta, config: LlamaModelConfig):
...
             ]
         )
-        self.paged_attention = self.attn_blocks[0].attn.paged_attention
+
+    def allocate_cache(self, page_count: int) -> CacheAllocation:
+        pipeline_to_device_map = self.config.pipeline_to_device_map
+        if pipeline_to_device_map is None:
+            return self.attn_blocks[0].attn.paged_attention.allocate(page_count)
+        else:
+            raise NotImplementedError
```
IanNod, line 124: Can you add an error message so any user hitting this can more easily triage?

IanNod (commented):
This seems to cause a lot of test failures with NotImplementedError. Is there a clean way to do this PR without breaking or xfailing all those tests?

IanNod (approved):

PR #2161, opened 2025-09-04
[sharktank] Add extend attention to llm

archana-ramalingam:
Add extend attention to prefill in llm

### DIFF
--- a/sharktank/sharktank/examples/export_paged_llm_v1.py
+++ b/sharktank/sharktank/examples/export_paged_llm_v1.py
@@ -74,6 +74,8 @@ def export_llm_v1(
         )
         start_pos = torch.empty(bs, dtype=torch.int64)
         seq_lens = torch.empty(bs, dtype=torch.int64)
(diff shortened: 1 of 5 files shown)


sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -66,7 +66,12 @@ def generate_batch_prefill(bs: int):
         block_dim = torch.export.Dim("block", min=block_dim_min, max=block_dim_max)
 
         sl_dim = llama_config.block_seq_stride * block_dim
-        seq_block_ids = torch.empty(bs, block_dim_min, dtype=torch.int64)
+
+        bs_min = 1
+        bs_max = ceildiv(bs, llama_config.block_seq_stride) - 1
```
rsuderman, line 71: I don't think this would be `- 1` as its the shape not final index.

rsuderman (changes requested):

PR #2163, opened 2025-09-04
[Sharktank] Remove block_index from PagedAttention.forward

Alex-Vasile:
Each PagedAttention object exists on a single block. The block_index is known when the object is instantiated.

Raw read, write, write_timestep functions were left unmodified to allow for easier testing.


IanNod (approved):
LGTM

PR #2167, opened 2025-09-04
Refactor Class PagedLlamaAttentionBlock

dezhiAmd:
The single purpose of this PR is to split class PagedLlamaAttentionBlock into two classses:
One for Gqa which is used by LLama
One for Mla which is used by deepseek
The goal is to make it easier to focus on LLama for the moment. 

There are follow-up PRs in draft status to make it better:
1) [[Sharktank] PagedAttentionBlock cleanup by Alex-Vasile · Pull Request #2162 · nod-ai/shark-ai](https://github.com/nod-ai/shark-ai/pull/2162)
2) https://github.com/nod-ai/shark-ai/pull/2157

### DIFF
--- a/sharktank/sharktank/examples/export_paged_llm_v1.py
+++ b/sharktank/sharktank/examples/export_paged_llm_v1.py
@@ -44,21 +44,6 @@ def export_llm_v1(
 
     fxb = FxProgramsBuilder(model)
 
-    def setup_cache(
-        model: ServicePagedLlmModelV1,
-    ) -> tuple[list[torch.Tensor], list[dict[int, torch.export.Dim]]]:
-        if not model.is_paged:
-            raise NotImplementedError(f"Unsupported KV cache type")
-
-        device_block_count = export_config.device_block_count
-        cache_state = model.allocate_cache(page_count=device_block_count)
-        page_dim = torch.export.Dim("page")
-
-        unpacked = cache_state.allocation
-        dynamic_shapes = [{0: page_dim} for _ in range(len(unpacked))]
-
-        return unpacked, dynamic_shapes
-
     def generate_batch_prefill(bs: int):
         # torch.export.Dim would make min at least 2
         block_dim_min = 2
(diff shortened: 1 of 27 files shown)


rsuderman (changes requested):

sharktank/sharktank/layers/paged_llama_attention_block.py
```diff
@@ -1,313 +1,542 @@
...
+        attention_scale: Optional[float] = None,
+        softcap: Optional[float] = None,
+        fake_quant: Optional[bool] = True,
+        use_rope: bool = True,
+        use_qk_norm: bool = False,
+        attn_temperature_tuning: bool = False,
+        floor_scale: Optional[float] = None,
+    ):
+        super().__init__(
+            theta,
```
rsuderman, line 63: Make this take one line, no reason to breakout across.

sharktank/sharktank/layers/paged_llama_attention_block.py
```diff
@@ -1,313 +1,542 @@
...
+from .latent_attention_block import LatentAttentionBlock
+from .paged_attention import CacheAllocation, KVCache, attn_type_map
+from sharktank import ops
+
+
+class PagedLlamaAttentionBlock(ThetaLayer):
+    """Implements a self attention layer in the style of Llama using a
+    paged cache."""
+
+    def __init__(
```
rsuderman, line 28: This can be collapsed down to one line

sharktank/sharktank/layers/paged_llama_attention_block.py
rsuderman, line 1: Why are all of these lines showing up as different in the diff? Did a property of the file change because this shouldn't be completely different. E.g. the imports are largely the same so we shouldn't see them showing up as different.

sharktank/sharktank/layers/paged_llama_attention_block.py
```diff
@@ -1,313 +1,542 @@
...
+            matmul_kernel=config.matmul_kernel,
+            fake_quant=fake_quant,
+            softcap=softcap,
+            use_rope=use_rope,
+            use_qk_norm=use_qk_norm,
+            attn_temperature_tuning=attn_temperature_tuning,
+            floor_scale=floor_scale,
+            attention_scale=attention_scale,
+        )
+    elif attn_type == "mla":
```
rsuderman, line 519: Just do `if` as the previous statement includes a return.

rsuderman (changes requested):
Something is wrong with the line diff. Identical lines are showing up as different which means it will wipe out blame history. Can you determine why this is happening? We can't land as it will pollute git history.

rsuderman (approved):
Approving but have Alex look over to make sure this doesn't collide with the pipelining work.

Alex-Vasile (approved):
Two minor nits, but the rest looks good.

sharktank/sharktank/layers/paged_llama_attention_block.py
```diff
@@ -294,21 +453,88 @@ def forward(
...
+    attention_scale: Optional[float] = None,
+    softcap: Optional[float] = None,
+    fake_quant: Optional[bool] = True,
+    use_rope: bool = True,
+    use_qk_norm: bool = False,
+    attn_temperature_tuning: bool = False,
+    floor_scale: Optional[float] = None,
+):
+    attn_type = attn_type_map[model_arch]
+    if attn_type == "gqa":
```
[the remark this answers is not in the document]
dezhiAmd: Will do that in follow up PR

PR #2169, opened 2025-09-04
Users/kyle/preshuffle

KyleHerndon:


KyleHerndon:
@jinchen62 The problem is that while it is faster, it also requires regenerating all IRPAs that want to use it, which would be nice to avoid. There is some work Max is doing that I'll integrate when its available with IREE that will make it so that we don't need to regenerate IRPAs.

I'll update the [related issue](https://github.com/nod-ai/shark-ai/issues/2253) to also note to make it the default.

PR #2178, opened 2025-09-05
[Sharktank] Move implementation of attention masks into ops folder to handle parallelism

Alex-Vasile:
Will replace current calls separate PR to limit line noise.

### DIFF
--- a/sharktank/sharktank/ops/default_impls.py
+++ b/sharktank/sharktank/ops/default_impls.py
@@ -158,19 +158,21 @@ def cat_default(tensors: Sequence[Tensor | PrimitiveTensor], dim: int):
 def chunked_attention_mask_default(
     attention_mask: torch.Tensor, attention_chunk_size: int
 ) -> torch.Tensor:
-    """
-    Apply a chunked attention mask onto a mask.
+    assert attention_mask.dim() == 4, "Attention mask must be 4-dimensional"
+    assert (
+        attention_mask.shape[1] == 1
+    ), f"Attention mask shape[1] ({attention_mask.shape[1]}) must be 1"
+    s2 = attention_mask.shape[2]
+    s3 = attention_mask.shape[3]
+    assert (
+        s2 == s3
+    ), f"Attention mask must be square in the last two dimensions ({s2} != {s3})"
 
-    This is a convenience function that combines the creation of the boolean
-    chunked attention mask and its application to the provided attention mask.
+    sl = attention_mask.shape[2]
+    assert (
+        sl % attention_chunk_size == 0
+    ), f"Sequence length ({sl}) must be divisible by attention chunk size ({attention_chunk_size})"
 
-    Args:
-        attention_mask: The original attention mask of shape [bs, 1, sl, sl].
-        attention_chunk_size: The size of each attention chunk.
-
-    Returns:
-        A new attention mask with chunked masking applied.
-    """
     attention_mask = unbox_tensor(attention_mask)
 
     device = attention_mask.device


sharktank/sharktank/ops/default_impls.py
```diff
@@ -100,6 +154,48 @@ def cat_default(tensors: Sequence[Tensor | PrimitiveTensor], dim: int):
...
+    chunked attention mask and its application to the provided attention mask.
+
+    Args:
+        attention_mask: The original attention mask of shape [bs, 1, sl, sl].
+        attention_chunk_size: The size of each attention chunk.
+
+    Returns:
+        A new attention mask with chunked masking applied.
+    """
+    attention_mask = unbox_tensor(attention_mask)
```
oyazdanb, line 174: add assertion for attention_mask shape and attention_chunk_size something like 
```
assert attention_mask.dim() == 4 and attention_mask.shape[1] == 1, "Expected [B,1,L,L]"
 L = attention_mask.shape[2]
  if L % attention_chunk_size != 0:
      raise ValueError(f"L={L} not divisible by chunk size {attention_chunk_size}")
```

sharktank/sharktank/utils/attention.py
```diff
@@ -28,40 +28,22 @@ def create_attention_mask(
     Since this is a bool tensor of context_length^2, different deployment
     scenarios can benefit from managing this in different ways.
     """
-    device = boolean_input_mask.device
+    import sharktank.ops as ops
 
-    # Combine the causal context mask and input mask.
-    dtype = (
-        torch.float32 if attention_dtype == torch.float8_e4m3fnuz else attention_dtype
+    return ops.attention_mask(
```
IanNod, line 33: since all these functions are doing is calling ops the appropriate ops.attention_mask/attention_mask_for_decode/chunked_attention_mask, can we remove these functions and directly call the ops?

sharktank/sharktank/ops/default_impls.py
```diff
@@ -92,6 +97,55 @@ def _split_argmax(input_tensor, dim, keepdim: bool = False, chunk_size: int = 12
...
+        dtype
+    )
+    return numeric_mask.to(device)
+
+
+attention_mask.override(Tensor, Tensor)(attention_mask_default)
+attention_mask.override(Tensor)(attention_mask_default)
+
+
+@attention_mask_for_decode.override(Tensor)
```
IanNod, line 131: instead of doing these overrides you may want to consider passing an impl flag to select the proper op, like done in this commit for attention and matmuls: https://github.com/nod-ai/shark-ai/commit/53d7be40e3b5a245a4a88c04cfb31023f8d7e77f

sharktank/sharktank/utils/attention.py
IanNod: Sure, that is fine to do in a follow up

sharktank/sharktank/utils/attention.py
```diff
@@ -176,7 +140,6 @@ def create_input_mask(seq_lens: torch.Tensor, batch_seqlen: int) -> torch.Tensor
         seq_lens: [bs] tensor of integers representing the sequence lengths.
         batch_seqlen: The maximum sequence length in the batch.
     """
-    range_vector = torch.arange(0, batch_seqlen, 1, device=seq_lens.device)
-    matrix = seq_lens.unsqueeze(dim=-1)
-    mask = range_vector >= matrix
-    return mask
+    import sharktank.ops as ops
```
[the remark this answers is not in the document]
Alex-Vasile: Removed functions from here, no longer need to do the import.

sharktank/sharktank/ops/default_impls.py
Alex-Vasile: The issue I see is that the two implementations (decode vs prefill) take different args (the `start_position` is missing for decode, and it's optional for prefill so we can't use it's presence/absence as a marker). I'm not sure if we'll come ahead on complexity once we handle the extra arg correctly.

IanNod (approved):

PR #2184, opened 2025-09-05
[Sharktank] Add a ParallelismConfig for PP/TP related settings

Alex-Vasile:


sharktank/sharktank/layers/configs/llm_configs.py
```diff
@@ -362,6 +368,68 @@ def _optional_int_prop(
...
 
 
+@dataclass
+class ParallelismConfig:
+    # Mapping between a transformer block and the corresponding pipeline
+    block_to_pipeline_map: tuple[int, ...] | None = None
+
+    # Mapping between a pipeline and the corresponding devices
+    pipeline_to_device_map: tuple[tuple[int, ...], ...] | None = None
+
```
rsuderman, line 378: Make a `@staticmethod` constructor that automagically makes the right pipeline and block maps for the device. You can just pass in `tp` and `pp` args and construct a default version.

rsuderman (changes requested):

PR #2196, opened 2025-09-08
[Sharktank] Unpack to qs

Alex-Vasile:
We will want to deal with this TODO sooner rather than later.


dan-garvey:
Also @IanNod if we don't care about models that use cache_quantizer anymore, we can drop this functionality entirely.

PR #2202, opened 2025-09-08
Cast dtype before calling attention

KyleHerndon:
We could do the cast inside the attention op itself, but we should not, so that we can maintain parity with torch scaled_dot_product_attention, IMO


dan-garvey (commented):
In what case would we have a different dtype for q and k than the attention dtype?

dan-garvey (commented):
What do you think about adding the dtype attribute to packed tensors? Is the meaning too vague?

PR #2204, opened 2025-09-09
Autodetect KV Cache dtype

rsuderman:
The KV cache dype can be detected from the IRPA file. Changed to detect and set the dtype when not manually specified.

### DIFF
--- a/sharktank/sharktank/layers/configs/llm_configs.py
+++ b/sharktank/sharktank/layers/configs/llm_configs.py
@@ -23,6 +23,7 @@ from transformers import T5Config as T5ConfigHf
 from iree.turbine.aot import DeviceAffinity
 from .config import ModelConfig
 from sharktank.utils import parse_version
+from sharktank.types import Dataset
 from sharktank.types.tensors import serialized_name_to_dtype, dtype_to_serialized_name
 
 if TYPE_CHECKING:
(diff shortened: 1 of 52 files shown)


KyleHerndon (approved):

sharktank/sharktank/utils/cli.py
```diff
@@ -105,7 +105,7 @@ def add_model_options(parser: argparse.ArgumentParser):
     parser.add_argument(
         "--attention-dtype",
         help="DType to use for attention in the model",
-        default="float16",
+        default=None,
```
archana-ramalingam, line 108: Should the `--activation-dtype` default be `None` too?

sharktank/sharktank/layers/configs/llm_configs.py
```diff
@@ -607,6 +608,24 @@ def from_properties(properties: "PropertyValueType") -> "LlamaModelConfig":
             kwargs["chunked_attention_layers"] = set(kwargs["chunked_attention_layers"])
         return LlamaModelConfig(**kwargs)
 
+    @staticmethod
```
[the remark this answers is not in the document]
rsuderman: Moved the arg puller into a helper. Still has some replication but it avoids the specialization per flex. Should make removal easier as well.

sharktank/sharktank/utils/cli.py
rsuderman: Changed back to `float16`. Did some discussion and the `kv_dtype` path is not guaranteed to quite work yet so regressing a bit.

PR #2208, opened 2025-09-09
[LLM Server] Refactor Scheduler Interface

stbaione:
Stacked on: #2195 
- Refactor as a step to enable a ChunkedScheduler for `ChunkedPrefill`
- Introduce a basic `AbstractScheduler` interface

The goal is to implement a design like this:

chunked_prefill_scheduler drawio

### DIFF
--- a/shortfin/tests/apps/llm/components/batcher_test.py
+++ b/shortfin/tests/apps/llm/components/batcher_test.py
@@ -133,12 +133,12 @@ class TestLlmBatcherProcess:
         llm_batcher_process.board = MagicMock()
 
         ## Empty
-        llm_batcher_process.pending = set()
+        llm_batcher_process.scheduler.pending = []
         await llm_batcher_process.board_flights()
         assert llm_batcher_process.board.call_count == 0
         llm_batcher_process.board.reset_mock()
 
-        assert llm_batcher_process.pending == set()
+        assert llm_batcher_process.scheduler.pending == []
 
         ## Non-empty
         task_inputs = []


shortfin/python/shortfin_apps/llm/components/scheduler.py
```diff
@@ -119,31 +122,60 @@ def available(self):
...
+
+
+class Scheduler(AbstractScheduler):
     def __init__(self, *, ideal_batch_size):
         self._ideal_batch_size = ideal_batch_size
         self._unreserved_strobe = None
         self._wid = 0
         self._preferred_groups = 1
 
+        self.pending: set[LlmTaskInput] = set()
```
zeeshanhaque21, line 154: I do not think pending should be a set. for the base case at least, it needs to be two queues, one for prefill task inputs, and one for decode task inputs. 

For the base case, You would probably just create workgroups with either only prefills or only decodes. And then move on to combining both in a workgroup once we have extend attention working.

At least that is how I understand it. @rsuderman, please let me know if this makes sense.

stbaione: `PrefillBatcherProcess` and `DecodeBatcherProcess` each get their own `Scheduler` instance already, and maintaining that here.

So, when we enable `ChunkedScheduler`, we will pass a `ChunkedScheduler` to `PrefillBatcherProcess` and `Scheduler` to `DecodeBatcherProcess`. Currently, there isn't a need to combine prefill & decode under one scheduler instance.

This sounds like changes for `DecodeMaximalBatching`, but that would be an added feature separate to `ChunkedPrefill`. That's where we can piggy back chunked prefill requests with decode requests. That's a separate effort though, that will also require an entirely new batcher being added, which was the motivation for #2075.

With the current `Vanilla` Scheduler, we don't really need to make any changes, like to store the workgroups ahead of time. It works the same way as it did before. The important aspect of this PR is that it implements a shared interface, which will make it possible for us to enabled `ChunkedPrefill`.

In order to implement `ExtendedAttention`, we would first need to land `ChunkedPrefill`.

stbaione: Verified w/ @rsuderman that `ExtendedAttention` and `DecodeMaximalBatcher` are separate, so we don't need to put both prefill and decode sets in the batcher

stbaione: Updated `pending queue` to be a list, instead of a set

PR #2211, opened 2025-09-09
[Sharktank] Export paged_kv_block_size_elements_per_device 

Alex-Vasile:
Depends on #2209.

Re-enables the export of ` paged_kv_block_size_elements_per_device` for running pipelined models through shortfin.


sharktank/sharktank/layers/paged_attention.py
```diff
@@ -167,12 +167,12 @@ def __init__(
             self.attn_head_dim,
         ]
 
-        self.page_slab_flat_dims = math.prod(self.sub_page_dims)
+        self.page_slab_flat_dims = [math.prod(self.sub_page_dims)]
```
rsuderman, line 170: Why is this in a list when we just grab the first element in allocate anyway.

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -167,12 +167,12 @@ def __init__(
...
 
     def allocate(self, page_count: int) -> CacheAllocation:
         tensors = [
             torch.zeros(
-                [page_count, self.page_slab_flat_dims],
+                [page_count, self.page_slab_flat_dims[0]],
```
rsuderman, line 175: Weird to just grab the first element.

rsuderman (changes requested):

PR #2223, opened 2025-09-10
[Sharktank] adding sliding_w and sink to paged attention

oyazdanb:
Reworking sliding_window and sink for gpt-oss to be compatible with new changes in the paged attetion

### DIFF
--- a/sharktank/sharktank/ops/attention_impls.py
+++ b/sharktank/sharktank/ops/attention_impls.py
@@ -25,38 +25,47 @@ from .signatures import (
 from ._registry import AnyType
 
 
-def build_causal_and_sw_prefill(n_tokens, sliding_window, dtype, device):
-    mask_2d = torch.triu(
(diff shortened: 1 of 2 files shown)


sharktank/sharktank/ops/attention_impls.py
```diff
@@ -25,38 +25,47 @@
...
         # decode path
         if sliding_window > 0 and kv_size > sliding_window:
-            start = kv_size - sliding_window
-            neq_inf = float("-inf")
-            a[..., :start] = neq_inf
+            start_idx = kv_size - sliding_window
+            neg_inf = float("-inf")
+            a[..., :start_idx] = neg_inf
+
     if a is not None:
```
Alex-Vasile, line 69: I think a is always going to be not None at this point. We don't run decode without a mask.

Alex-Vasile (approved):

PR #2236, opened 2025-09-11
[Sharktank] Irpa for gpt-oss

oyazdanb:
Creating irpa file for gpt-oss


sharktank/sharktank/models/gpt_oss/import_gpt-oss.py
```diff
@@ -0,0 +1,202 @@
+import argparse
```
IanNod, line 1: We shouldn't need an import script specific to gpt-oss. @sogartar is working on adding a more generic safetensor->irpa converter that shark-ai can consume. Would be good to move what features we need in this to there

PR #2239, opened 2025-09-12
[sharktank] improve module patching for tensor tracing

sogartar:
Make patching be able to patch/wrap not just the forward method but select what methods to patch based on a regex or fnmatch-like match on the fully qualified name of submodule methods.

When using TraceTensorModulePatch always trace with a key like SaveModuleResultTensorsPatch. This makes both patchers consistent.
We probably  should retire `SaveModuleResultTensorsPatch` at some point.

### DIFF
--- a/.github/workflows/ci-fusilli-plugin.yml
+++ b/.github/workflows/ci-fusilli-plugin.yml
@@ -55,13 +55,10 @@ jobs:
         include:
           - name: gfx942_clang20_debug
             runs-on: linux-mi325-2gpu-ossci-nod-ai
-            # TODO(aaron): Debug CI failure and enable `-DFUSILLI_BUILD_BENCHMARKS=ON`
             fusilli-plugin-cmake-options:
               -DCMAKE_C_COMPILER=clang-20
               -DCMAKE_CXX_COMPILER=clang++-20
               -DCMAKE_BUILD_TYPE=Debug
-              -DFUSILLI_SYSTEMS_AMDGPU=ON
-              -DFUSILLI_BUILD_BENCHMARKS=OFF
               -DIREERuntime_DIR=/workspace/.cache/docker/iree/build/lib/cmake/IREE
 
     steps:
--- a/.github/workflows/ci-sharktuner.yml
+++ b/.github/workflows/ci-sharktuner.yml
@@ -57,4 +57,4 @@ jobs:
         run: pytest sharktuner/
 
       - name: Run mypy type checker
-        run: mypy sharktuner/sharktuner
+        run: mypy sharktuner --python-version 3.11
--- a/.github/workflows/cleanup_releases.yml
+++ b/.github/workflows/cleanup_releases.yml
@@ -21,7 +21,7 @@ on:
       dry_run:
         description: 'Perform a dry run (show what would be deleted without actually deleting)'
         type: boolean
-        default: false
+        default: true
         required: false
 
 permissions:
@@ -48,7 +48,7 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         RETENTION_DAYS: ${{ inputs.retention_days || 90 }}
-        DRY_RUN: ${{ inputs.dry_run || true }}
+        DRY_RUN: ${{ inputs.dry_run || false }}
       run: |
         python3 << 'EOF'
         import os
--- a/app_tests/integration_tests/llm/model_management.py
+++ b/app_tests/integration_tests/llm/model_management.py
@@ -286,7 +286,7 @@ class ModelStageManager:
                 return self.base_dir / self.config.dataset_name.replace("/", "_")
             return self.base_dir / self.config.repo_id.replace("/", "_")
         elif self.config.source == ModelSource.LOCAL:
-            return self.base_dir / "local" / self.config.irpa_path.stem
+            return self.base_dir / "local" / self.config.name
         elif self.config.source == ModelSource.AZURE:
             return (
                 self.base_dir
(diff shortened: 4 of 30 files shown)


sharktank/sharktank/utils/patching.py
```diff
@@ -18,85 +20,176 @@
...
+        one-by-one. The first successful match terminates the iteration.
+        A successful match on filter element of kind PatchFilterElement.INCLUDE would cause
+        the final match to succeeded.
+        A successful match on filter element of PatchFilterElement.EXCLUDE inverts
+        the meaning and would cause the final match to fail.
+        If no filter element is a match then the method is rejected and is not patched.
+        E.g.
+        ```
+        [
+            PatchFilterElement(regex=".*\.layer\.0\.attention\..+$", kind=FilterKind.EXCLUDE)
```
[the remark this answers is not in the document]
sogartar: I added [fnmatch-like](https://docs.python.org/3/library/fnmatch.htm) matching.
If this matching is not enough the user can always use `patch_method` directly.

sharktank/sharktank/utils/patching.py
```diff
@@ -138,6 +267,9 @@ def _add_nested_tensors(
         tensors: list[Any] | dict[str, Any] | torch.Tensor,
         name_delimiter: str,
     ):
+        if isinstance(tensors, str):
```
archana-ramalingam, line 270: `tensors` type hint above doesn't include a standalone `str` as a type, does this condition still occur?

sharktank/tests/utils/patching_test.py
```diff
@@ -35,16 +40,18 @@ def config_tracing(tmp_path: Path):
...
+@pytest.fixture
+def module_for_patching() -> BaseLayer:
     class Inner(BaseLayer):
         def forward(
             self, arg0: torch.Tensor, arg1: torch.Tensor
         ) -> tuple[torch.Tensor, torch.Tensor]:
+            self.other_method(arg0)
             return arg0, arg1
 
+        def other_method(self, arg0: torch.Tensor):
```
archana-ramalingam, line 52: More descriptive naming for `other_method`?

sharktank/tests/utils/patching_test.py
```diff
@@ -56,25 +63,110 @@ def forward(
...
+
+
+def test_trace_tensor_module_patch(config_tracing, module_for_patching: BaseLayer):
+    tensor0 = torch.arange(1, 3, dtype=int)
+    tensor1 = torch.arange(3, 6, dtype=int)
+
+    patcher = TraceTensorModulePatch(with_before_call=True)
+    patcher.patch_child_modules(module_for_patching)
+
+    module_for_patching(tensor0, arg1=tensor1)
```
archana-ramalingam, line 76: Mention both function vars or we can skip both altogether. 
eg: `module_for_patching(arg0=tensor0, arg1=tensor1)`

sharktank/tests/utils/patching_test.py
```diff
@@ -56,25 +63,110 @@ def forward(
...
+    }
+    for path, expected_value in path_expected_value_map.items():
+        with safetensors.safe_open(path, framework="pt", device="cpu") as f:
+            assert len(f.keys()) == 1
+            recorded_tensor = f.get_tensor("")
+            torch.testing.assert_close(recorded_tensor, expected_value, rtol=0, atol=0)
+
 
-    outer(tensor0, arg1=tensor1)
+def test_trace_tensor_module_patch_with_more_complex_fully_qualified_name_regex_filter(
```
archana-ramalingam, line 95: Given that regex and other alternatives have been elaborated in the functions earlier, can we shorten these test names to `test_trace_tensor_module_regex`, etc?

sharktank/sharktank/utils/patching.py
sogartar: I changed the name of the arg from `tensors` to `maybe_tensors` and added `Any` to the type hints.

This function is meant to handle only the tensors inside trees. For other more exotic types there is a warning at the end if not handled. There is a general problem that iterables should be traversed to see if there are tensors inside. `str` is kinda nasty because it is `Iterable` but when you iterate on it, it again returns the characters as strings causing an infinite recursion.

sharktank/tests/utils/patching_test.py
sogartar: It is deliberate to check that positional args get correct file names. `tensor0` would get a filename `.forward.arg%0.safetensors` while `tensor1` would get a filename `.forward.arg%arg1.safetensor`. Traced tensors get names based on how the function is called.

There is another approach which would inspect the signature of the underlying callable to determine what is the name of the argument but it is not implemented like that now.

sharktank/tests/utils/patching_test.py
archana-ramalingam: Got it, can we have a better name for this function?

sharktank/sharktank/utils/patching.py
```diff
@@ -18,85 +23,209 @@
...
-    """Patches calls to forward, allowing various forms of interception."""
+    """Patches calls to methods, allowing various forms of interception.
+
+    Can patch the pre and post calling submodule methods method."""
+
+    def patch_child_modules(
+        self,
+        module: torch.nn.Module,
+        *,
+        filter: list[PatchFilterElement] = get_default_patch_filter(),
```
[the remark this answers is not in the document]
sogartar: I changed it to a global var.

sharktank/tests/utils/patching_test.py
sogartar: I renamed it to `some_other_method`.

PR #2245, opened 2025-09-14
[Fusilli] Support user-allocated output buffers (destination passing style ABI)

sjain-stanford:
Rewrites the Graph ASM emitter and `Graph::execute()` to support destination passing style ABI (i.e., have output buffers pre-allocated and provided alongside inputs). Uses `torch.overwrite.tensor.contents` op for implementing aliasing semantics in the emitted MLIR assembly. 

One nice side effect of this change is it let's us retire some of our not-so-ideal workarounds for `Buffer::reset()` and empty `Buffer()` constructors that were previously needed to support output buffer placeholders to be created that are then lazily reset to the runtime returned (internally allocated) output buffer views after call invocation.


zjgarvey (approved):
This looks good to me. 

The IR seems to match what I would expect for the in-place torch-ir.

PR #2256, opened 2025-09-15
GEMM Sharktank kernel updates to optimize shuffle_scales

KyleHerndon:
Prefill (down from ~4300):
```
BM_prefill_bs4/process_time/real_time_mean         4193 ms         7846 ms           10 items_per_second=0.238474/s
```
Decode (down from ~210):
```
BM_decode_bs4/process_time/real_time        161 ms          287 ms            4 items_per_second=6.20525/s
```


sharktank/sharktank/kernels/gemm_fp4_asm.py
```diff
@@ -65,27 +65,35 @@ def _build_mlir_spec(bitstring):
 
     mlir = f"""
 #rocm_target = #hal.executable.target<"rocm", "rocm-hsaco-fb", {{target_arch = "gfx950", ukernels = "none"}}>
+#scales_shuffle_encoding = #iree_encoding.layout<[#iree_gpu.gpu_encoding_resolver<configuration =
+    {{encoding_info = {{
+        innerDimsPos = [0, 1],
+        innerTileSizes = [32, 8],
```
IanNod, line 71: Where do all these values come from?

IanNod (commented):
Can you leave a message in your PR for how much "faster" this shuffle_scales is (or expected to be with iree changes)?

sharktank/sharktank/kernels/gemm_fp4_asm.py
KyleHerndon: This whole iree_encoding.layout thing is directly from Max, the rest is integration to make it work. IIUC, it essentially tells the tensor how it is going to be reshaped, so that it can be const-eval'd.

PR #2263, opened 2025-09-17
[sharktank] add tool to import/convert LLM Hugging Face model parameters to sharktank

sogartar:
This would allow us to stop using llama.cpp's `convert_hf_to_gguf.py` tool.

### DIFF
--- a/sharktank/sharktank/utils/hf.py
+++ b/sharktank/sharktank/utils/hf.py
@@ -14,7 +14,11 @@ import torch
 from pathlib import Path
 
 from huggingface_hub import snapshot_download
-from sharktank.layers.configs import LlamaHParams, LlamaModelConfig
+from sharktank.layers.configs import (
+    LlamaHParams,
+    LlamaModelConfig,
+    is_hugging_face_llama3_config,
+)
 from sharktank.types import *
 from sharktank.utils.functools import compose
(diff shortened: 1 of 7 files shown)


sharktank/sharktank/utils/hf.py
```diff
@@ -80,3 +127,79 @@ def import_hf_dataset_from_hub(
...
+    if re_match:
+        layer_idx = re_match.group(2)
+        name_to_map = re.sub(layer_idx_pattern, r"\1{layer_idx}\3", tensor.name)
+
+    sharktank_name = _llama3_hf_to_sharktank_tensor_name_map[name_to_map]
+    sharktank_name = sharktank_name.format(layer_idx=layer_idx)
+    return DefaultPrimitiveTensor(data=tensor.as_torch(), name=sharktank_name)
+
+
+def llama3_hf_config_to_sharktank(hf_config: dict[str, Any]) -> dict[str, Any]:
```
KyleHerndon, line 164: This functionality should either be in `sharktank/sharktank/layers/configs/llm_configs.py`, I think.

I would like to prevent duplication of this code; if they get out of sync, it will be really annoying to maintain/debug.

sogartar: I moved the config related stuff there.

PR #2265, opened 2025-09-17
[Sharktank] Rename test files to remove duplicates

Alex-Vasile:
Pytest throws errors about duplicate file names.


sogartar (approved):
Here is some more information https://docs.pytest.org/en/latest/explanation/pythonpath.html.
TLDR
To properly allow this we need to add `__init__.py` files to directories containing tests so that pytest treats them as packages.

PR #2268, opened 2025-09-17
[Sharktank] Simplify dequantize_raw_tensor

Alex-Vasile:
`.unpack()` simply returns the layout property, there doesn't seem to be a point to build the PlanarQuantized object only to immediately discard it.


dan-garvey (approved):
lol

PR #2272, opened 2025-09-17
[tuner] use subgroup basis for lowering config

bangtianliu:
This PP makes the change from tuner side to sync with IREE changes from https://github.com/iree-org/iree/pull/21912, also depending on the updated python binding from IREE PR: https://github.com/iree-org/iree/pull/22027.

### DIFF
--- a/sharktuner/sharktuner/common.py
+++ b/sharktuner/sharktuner/common.py
@@ -358,7 +358,8 @@ def link_tuning_specs(tuner_ctx: TunerContext, td_specs: list[ir.Module]) -> ir.
     into one tuning spec.
     """
     module = combine_tuning_specs(tuner_ctx, td_specs)
-    iree_opt = ireec.binaries.find_tool("iree-opt")
+    iree_opt = ireec.binaries.find_tool("iree-opt")  # type: ignore
+    assert iree_opt, "iree-opt tool not found"
 
     if len(td_specs) == 1:
         # avoid unnessary link overhead.
--- a/sharktuner/sharktuner/constraint_generator.py
+++ b/sharktuner/sharktuner/constraint_generator.py
@@ -227,6 +227,7 @@ def generate_generic_contraction_solutions(
                 ),
             ]
         # Setting subgroup basis.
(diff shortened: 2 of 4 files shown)


sharktuner/sharktuner/common.py
```diff
@@ -256,10 +259,17 @@ def get_lowering_config(
                     assert (
                         False
                     ), f"Unsupported type for key '{key}': {type(value).__name__}"
-            case "subgroup_m_count" | "subgroup_n_count":
-                if isinstance(value, int):
-                    promoted_value = tuner_ctx.type.getI64(value)
-                elif not isinstance(value, tuner_ctx.type.i64):
+            case "subgroup_basis":
+                if isinstance(value, list) and len(value) == 2:
```
kuhar, line 263: Wasn't this a `std::tuple` on the iree side? I'm surprised value became a list.

bangtianliu: Good catch! This is about assembling the input from the Python side, which then gets used to create `iree_gpu.LoweringConfigAttr`. The Python input format is `[[counts], [mapping]]` (list of two lists), which gets converted to the appropriate MLIR attributes internally (I think).

sharktuner/sharktuner/constraint_generator.py
```diff
@@ -219,6 +226,13 @@ def set_cdim_tile_sizes(tile_sizes, contraction_dims, csizes):
                     for d in contraction_dims.k
                 ),
             ]
+        # Setting subgroup basis.
+        subgroup_basis_counts = [1] * num_loops
+        m_dim = contraction_dims.m[-1]
+        subgroup_basis_counts[m_dim] = lookup(sg_m_cnt)
+        n_dim = contraction_dims.n[-1]
+        subgroup_basis_counts[n_dim] = lookup(sg_n_cnt)
```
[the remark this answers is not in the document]
bangtianliu: https://github.com/iree-org/iree/pull/22000 has been landed, so I will also sync in this PR directly after I understand how it works @kuhar, also cc @Groverkss here.

PR #2275, opened 2025-09-18
[FusilliPlugin] Basic CI, Build, and README (`fusilli-plugin` 1 of N)

AaronStGeorge:
Adds basic CI, CMake build scrips, and Readme.

In the (very messy) git history you can see a passing CI run for a full (also messy) hipDNN plugin + integration test https://github.com/nod-ai/shark-ai/pull/2275/commits/e36508648718aed569269d044e12b102469a78ef . Most of the implementation is removed from the final PR, but the passing test should show confidence in the build + CI setup (which did not change).

### DIFF
--- a/.github/workflows/ci-fusilli-plugin.yml
+++ b/.github/workflows/ci-fusilli-plugin.yml
@@ -41,7 +41,7 @@ jobs:
         # -exo pipefail:
         #   -e exit immediately if any command fails
         #   -x print each command before executing
-        #   -o pipefail ensure pipeline fail sif any command in it fails
+        #   -o pipefail ensure pipeline fails if any command in it fails
         # {0} github actions placeholder for the command to run
         shell: bash --noprofile --norc -exo pipefail {0}
     strategy:
--- a/fusilli-plugin/README.md
+++ b/fusilli-plugin/README.md
@@ -10,12 +10,12 @@ The plugin builds as a shared library (`fusilli_plugin.so`) providing a `hipDNN`
 
 ### Setup
 
-For the time being, `fusilli-plugin` setup relies on / builds on [Fusilli setup](../sharkfuser/#setup).
+For the time being, `fusilli-plugin` setup relies on / builds on [Fusilli setup](../sharkfuser/README.md#setup).
 Keeping the projects in sync prevents "works on my machine" style bugs.
 Requirements that are unique to `fusilli-plugin`, `hipDNN` and `googletest` for
 example, are fetched configured and built as part of `fusilli-plugin` build.
 
-After following steps in [Fusilli Setup](../sharkfuser/#setup), build and test
+After following steps in [Fusilli Setup](../sharkfuser/README.md#setup), build and test
 `fusilli-plugin` as follows:
 ```shell
 $ cmake -GNinja -S. -Bbuild \
--- a/fusilli-plugin/build_tools/cmake/FusilliPluginDependencyUtils.cmake
+++ b/fusilli-plugin/build_tools/cmake/FusilliPluginDependencyUtils.cmake
@@ -44,44 +44,34 @@ include(FetchContent)
 # <dependency-specific args>
 #   The `_fetch_X` macro for dependency X defines the available options.
 #   Examples: GTEST_VERSION for GTest, HIP_DNN_HASH for hipdnn_frontend
-function(fusilli_plugin_dependency dep_name)
-    # Route to appropriate `_fetch_X` macro.
-    if(COMMAND _fetch_${dep_name})
-        _with_logging(_fetch_${dep_name} ${dep_name} ${ARGN})
-    else()
-        message(FATAL_ERROR "Unknown dependency: ${dep_name}")
-    endif()
-endfunction()
-
-# Think of this as a python decorator, wraps a `_fetch_X` macro with some
-# logging conveniences.
 #
-# MACRO_NAME
-#   Name of the `_fetch_X` macro.
-#   NOTE: CMake macros execute in current scope, CMake functions create a new
-#         scope. `_with_logging` checks variables it expects a `_fetch_X` macro
-#         to set _in_ `_with_logging`s scope, requiring that `_fetch_X` is a
-#         macro and not a function.
-#
-# DEP_NAME
-#   Dependency name for logging
-function(_with_logging MACRO_NAME DEP_NAME)
-    # Set indent for logging, any logs from dep "X" will be prefixed with [x].
+function(fusilli_plugin_dependency DEP_NAME)
+    # Set indent for logging, any logs from dep "X" will be prefixed with [X].
     set(CMAKE_MESSAGE_INDENT "[${DEP_NAME}] ")
 
-    # CMake macros aren't textual expansions like C preprocessor macros, so a
-    # dynamic call (like below) to a macro isn't a problem.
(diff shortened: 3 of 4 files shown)


fusilli-plugin/.gitignore
```diff
@@ -0,0 +1,8 @@
+# CMake build
```
AaronStGeorge, line 1: The Apache OSS header is apparently actually optional https://softwareengineering.stackexchange.com/a/357796. Seemed a little silly on a 1 line configuration file 🤷‍♂️ . I don't feel strongly if it's preferable to add it.

.github/workflows/ci-fusilli-plugin.yml
```diff
@@ -0,0 +1,100 @@
...
+    defaults:
+      run:
+        # --noprofile skips loading ~/.bash_profile
+        # --norc      skips loading ~/.bashrc
+        # -exo pipefail:
+        #   -e exit immediately if any command fails
+        #   -x print each command before executing
+        #   -o pipefail ensure pipeline fail sif any command in it fails
+        # {0} github actions placeholder for the command to run
+        shell: bash --noprofile --norc -exo pipefail {0}
```
sjain-stanford, line 46: This is useful :) Mind adding the comments to `ci-sharkfuser.yml` as well?

.github/workflows/ci-fusilli-plugin.yml
```diff
@@ -0,0 +1,100 @@
...
+    name: "Build & Test :: ${{ matrix.name }}"
+    runs-on: ${{ matrix.runs-on }}
+    defaults:
+      run:
+        # --noprofile skips loading ~/.bash_profile
+        # --norc      skips loading ~/.bashrc
+        # -exo pipefail:
+        #   -e exit immediately if any command fails
+        #   -x print each command before executing
+        #   -o pipefail ensure pipeline fail sif any command in it fails
```
sjain-stanford, line 44: nit: `fails if`

.github/workflows/ci-fusilli-plugin.yml
```diff
@@ -0,0 +1,100 @@
...
+      working-directory: ${{ github.workspace }}
+      run: |
+        ${{ env.FUSILLI_DIR }}/build_tools/docker/exec_docker_ci.sh \
+        bash -c "export FUSILLI_CACHE_DIR=/tmp && \
+                 cd fusilli-plugin && \
+                 cmake -GNinja -S. -Bbuild \
+                   -DIREERuntime_DIR=/workspace/.cache/docker/iree/build/lib/cmake/IREE \
+                   ${{ matrix.fusilli-plugin-cmake-options }} && \
+                 cmake --build build --target all &&
+                 ctest --test-dir build --output-on-failure --extra-verbose --timeout 30"
```
sjain-stanford, line 81: Now I know where it was picking the timeout of 30 seconds for Fusilli! When it was failing due to timeout in CI, I instead increased the TIMEOUT in `FusilliTestUtils.cmake` to 60 seconds.

fusilli-plugin/build_tools/cmake/FusilliPluginDependencyUtils.cmake
```diff
@@ -0,0 +1,204 @@
...
+# MACRO_NAME
+#   Name of the `_fetch_X` macro.
+#   NOTE: CMake macros execute in current scope, CMake functions create a new
+#         scope. `_with_logging` checks variables it expects a `_fetch_X` macro
+#         to set _in_ `_with_logging`s scope, requiring that `_fetch_X` is a
+#         macro and not a function.
+#
+# DEP_NAME
+#   Dependency name for logging
+function(_with_logging MACRO_NAME DEP_NAME)
```
AaronStGeorge, line 68: I think I can collapse what this function is doing into `fusilli_plugin_dependency` actually. I don't think extra "hop" is necessary.

fusilli-plugin/README.md
```diff
@@ -0,0 +1,26 @@
...
+## Developer Guide
+
+### Setup
+
+For the time being, `fusilli-plugin` setup relies on / builds on [Fusilli setup](../sharkfuser/#setup).
+Keeping the projects in sync prevents "works on my machine" style bugs.
+Requirements that are unique to `fusilli-plugin`, `hipDNN` and `googletest` for
+example, are fetched configured and built as part of `fusilli-plugin` build.
+
+After following steps in [Fusilli Setup](../sharkfuser/#setup), build and test
```
sjain-stanford, line 18: Curious, where does `../sharkfuser/#setup` point to? Don't you need the `README.md` in there somewhere?

AaronStGeorge: Hmm yeah, it seemed like this worked in vscode but it doesn't look like it does.

sharkfuser/CMakeLists.txt
```diff
@@ -78,6 +78,10 @@ add_library(libfusilli INTERFACE)
 # Consumers of exported/installed target use fusilli::fusilli.
 set_target_properties(libfusilli PROPERTIES EXPORT_NAME fusilli)
 
+# Create alias for consistent naming in consuming projects building locally (via
+# FetchContent) and against an installed version (via find_package).
+add_library(fusilli::fusilli ALIAS libfusilli)
```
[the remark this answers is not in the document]
AaronStGeorge: We could do that, but we would need `add_library(fusilli::libfusilli ALIAS libfusilli)`. The issue is the `fusilli` namespace, not the name. When consuming project uses the installed version, the `fusilli::` namespace is applied
```
# Create FusilliTargets.cmake in the install tree from FusilliTargets export set
install(EXPORT FusilliTargets
  FILE FusilliTargets.cmake
  NAMESPACE fusilli::
  DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/fusilli
)
```
but when build with `FetchContent` (as we do in `fusilli-plugin`) we essentially have a subdirectory build, so the namespace won't be applied. The alias allows consuming projects to refer to the library the same way irrespective of how they got the dependency.

Here's an excerpt from "Professional CMake: A Practical Guide" by Craig Scott about it.
 
> ```cmake
>add_library(MyShared SHARED ...)
>add_library(BagOfBeans::MyShared ALIAS MyShared)
>install(TARGETS MyShared
>  EXPORT BagOfBeans
>  DESTINATION ${CMAKE_INSTALL_LIBDIR}
>)
>install(EXPORT BagOfBeans
>  DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/BagOfBeans
>  NAMESPACE BagOfBeans::
>)
>```
>The above example follows the advice from Section 19.4, “Recommended Practices” where each
regular target also has a namespaced ALIAS associated with it. When installing the export for the
non-alias MyShared target, the same namespace is used as for the alias target (i.e. BagOfBeans::). This
allows projects that consume the exported details to refer to the target in the same way as this
project can refer to the alias (BagOfBeans::MyShared). Consuming projects can then elect to add this
project directly via add_subdirectory() or pull in the export file via find_package(), yet still use the
same BagOfBeans::MyShared target name regardless of which method was chosen. This important
pattern is emerging as a fairly common expectation on projects among the CMake community, so it
is in most projects’ interests to try to follow it.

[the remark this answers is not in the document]
sjain-stanford: I see. Makes sense.

PR #2281, opened 2025-09-18
[Fusilli] CI timeout fixes

sjain-stanford:
Remove the ctest timeout from the CMake target and update the ctest command in GHA workflow instead, since local runs shouldn't need a timeout.


AaronStGeorge (approved):
🚀

PR #2283, opened 2025-09-18
[Fusilli] Refactor stride utilities to support ambiguous (unit dim) cases

sjain-stanford:
We can't always calculate the stride order from stride, especially in ambiguous (unit dim) cases. For example a tensor of shape `(256, 128, 1, 1)` would have contiguous stride `(128, 1, 1, 1)` but we can't certainly say the stride order for this is `(3, 2, 1, 0)` as it could also be `(3, 0, 1, 2)`.

A more robust way of getting the stride order is to reconstruct the stride from the dims and compare it with the known stride order for contiguous vs channels last.

Apart from this change, the PR also removes the contiguity check from TensorAttr (letting it be more relaxed) and moves it to the pre/post validate checks on the ConvFPropNode itself. This makes sense because certain layout restrictions are op-semantics driven and we should model those in the op specific validations rather than on the generic TensorAttr.

Tests updated accordingly.

### DIFF
--- a/sharkfuser/include/fusilli/attributes/tensor_attributes.h
+++ b/sharkfuser/include/fusilli/attributes/tensor_attributes.h
@@ -110,6 +110,89 @@
 
 namespace fusilli {
 
+// Generates stride order for a contiguous (channels-first) tensor. For a 4D
+// tensor, this would return {N: 3, C: 2, H: 1, W: 0} to represent an NCHW
+// in-memory layout. Here N is the slowest changing and W is the fastest
+// changing dimension.
+inline std::vector<size_t> getContiguousStrideOrder(size_t numDims) {
+  assert(numDims >= 3 &&
+         "Contiguous (channels-first) layout requires at least 3 dimensions");
+
+  std::vector<size_t> strideOrder(numDims);
+  size_t order = 0;
+  // Caution: Reverse iteration with size_t (unsigned) can lead to underflow
+  // when `i == 0` is decremented due to wrap-around to SIZE_MAX (which is
+  // positive and doesn't prevent control from entering the loop as intended).
+  // This can cause heap corruption when accessing strideOrder[SIZE_MAX].
+  // So we instead gate loop entry with `i > 0`.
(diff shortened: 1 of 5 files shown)


sharkfuser/include/fusilli/attributes/tensor_attributes.h
```diff
@@ -99,17 +99,74 @@
...
 #include <memory>
-#include <numeric>
 #include <optional>
 #include <string>
 #include <variant>
 #include <vector>
 
 namespace fusilli {
 
+// Generates stride order for a contiguous tensor. For a 4D tensor, this would
```
sjain-stanford, line 111: These are merely moved from below (not real diffs)

sharkfuser/include/fusilli/attributes/tensor_attributes.h
```diff
@@ -289,89 +341,6 @@ struct TensorAttrSortByName {
...
-  int64_t order = 0;
-  strideOrder[1] = order++;
-  for (size_t i = numDims - 1; i > 1; --i)
-    strideOrder[i] = order++;
-  strideOrder[0] = order;
-  return strideOrder;
-}
-
-inline std::vector<int64_t>
-getStrideOrderFromStride(const std::vector<int64_t> &stride) {
```
sjain-stanford, line 328: This is removed now.

sjain-stanford:
> Trying to understand why you need to get the stride order ever? Isnt that effectively input?

Op inputs need have user-defined dims and strides, but op outputs may be underspecified, in which case we need to infer the output strides. To do this, we need to know the stride order of the inputs.

I now have this nice utility method on the TensorAttr itself to query the layout of any tensor easily:
```c++
  bool isContiguous() const {
    std::vector<int64_t> expectedStride =
        generateStrideFromDim(dim_, getContiguousStrideOrder(dim_.size()));
    return expectedStride == stride_;
  }
```

AaronStGeorge (approved):
LGTM

PR #2288, opened 2025-09-19
Lisal.fix batch trie

lisaliu1:
fix trie cache issue with batch size > 1


shortfin/python/shortfin_apps/llm/components/kvcache/trie_attention_cache.py
```diff
@@ -201,14 +201,18 @@ def match(self, tokens: List[int]) -> Tuple[TrieNode, List[PageInfo]]:
...
 
-        for i in range(0, len(tokens), self.tokens_per_page):
-            token_block = tokens[i : i + self.tokens_per_page]
-
-            if token_block not in cur.children:
-                break
-            cur = cur.children[token_block]
-            cur.access_time = time.monotonic()
-            matched_pages.append(cur.page)
+        nodes_to_search = [cur]
```
stbaione, line 204: This seems like we're traversing the `trie structure` incorrectly/inefficiently.

See the attached diagram:
trie_structure_traverse

We should be iterating block-by-block for the `passed in tokens`, and only `stepping` when we find a child that matches the current block, starting from root.

This is taking a node that doesn't match, adding it to a queue, and checking if we match that nodes children, which just doesn't seem right, and causes us to check a ton of paths/nodes that we know aren't matches from the start. If, for example, no children of `root` match the first block of tokens, then we don't traverse it's children, because we don't have a prefix match.

shortfin/python/shortfin_apps/llm/components/decoder/decoder.py
```diff
@@ -205,47 +210,48 @@ def update_decode_reqs(
...
-            decode_reqs[i].page_ids = self._shared_pages + new_beam_page_ids[i]
-        return decode_reqs[: len(tokens)]
+            # setup decode_reqs
+            for i, ids in enumerate(next_token_ids):
+                decode_reqs[i].input_token_ids = ids
+                decode_reqs[i].start_position = position
+                decode_reqs[i].page_ids = self._shared_pages + new_beam_page_ids[i]
+            return decode_reqs[: len(tokens)]
 
     def release_pages(self):
```
stbaione, line 252: This feels like we're repeating a previous mistake, when we would call free on `cache._allocated_pages`.

If we're freeing the pages from each of our decode requests when we're done w/ our decode loop, then they should already be freed. We shouldn't have to call our `cache` again w/ what should be the same pages.

shortfin/python/shortfin_apps/llm/components/kvcache/base_attention_cache.py
```diff
@@ -231,3 +241,6 @@ def publish_pages_for_tokens(
     def release_pages(self, cache_info: CacheInfo):
         if cache_info is not None:
             self.free_pages(cache_info.pages)
+
+    def free_allocated_pages(self, page_ids: List[int]):
```
stbaione, line 245: Why do we need `free_allocated_pages` and `release_pages`?

shortfin/python/shortfin_apps/llm/components/kvcache/trie_attention_cache.py
```diff
@@ -402,6 +406,56 @@ def extend_allocation(
...
+            return TrieCacheInfo(
+                num_tokens=num_tokens,
+                tokens=tokens,
+                pages=cache_info.pages,
+                pool=cache_info.pool,
+                last_cached_node=cache_info.last_cached_node,
+                number_of_published_pages=cache_info.number_of_published_pages,
+            )
+
+    def update_allocated_pages(
```
stbaione, line 424: Shouldn't need this function

stbaione (changes requested):
I don't think we can/should have to extend our interface the way we do in this PR to fix our Trie accuracy. It worked previously with far less functions and steps, and this makes the workflow cloudy.

Our implementation is getting away from us. We should be reviewing our existing workflow and finding/fixing the bugs in our current interface. This seems like we're "adding bandaids", by adding functions that shouldn't be needed.

We need to `simplify`, not `extend`. I think our extending has been a reason for why we've been seeing more bugs in our caches lately.

shortfin/python/shortfin_apps/llm/components/kvcache/base_attention_cache.py
lisaliu1: _free_allocated_pages are allocated but not used in a single decoder.run, they are shared across different decode_requests.
release_pages release them after a decodeer.run so we don't pin the memory that is not gonna to be used by next decoder.run

shortfin/python/shortfin_apps/llm/components/kvcache/trie_attention_cache.py
lisaliu1: We need this function to to keep cache_info.pages and page_ids consistent.

PR #2297, opened 2025-09-22
[Sharktank] Changes for eval_llm_vmfb and run_llm_vmfb for PP

Alex-Vasile:


sharktank/sharktank/tools/eval_llm_vmfb.py
```diff
@@ -47,58 +53,70 @@ def export_ir(irpa):
...
             raise ValueError("Config found without corresponding vmfb")
-        ir, config = export_ir(irpa)
-        vmfb = compile_ir(ir, iree_hal_target_device, iree_hip_target)
+        ir, config = export_ir(irpa, pipeline_parallelism_size)
+        vmfb = compile_ir(ir, iree_hal_target_devices, iree_hip_target)
 
     if isinstance(config, str):
-        config = ServiceConfig.load(config)
+        config: ServiceConfig = ServiceConfig.load(config)
+    assert pipeline_parallelism_size == len(
```
rsuderman, line 90: You should assume that if the config is built and passed through it returns a valid config.

sharktank/sharktank/tools/eval_llm_vmfb.py
```diff
@@ -132,9 +150,27 @@ def main(
...
+    parser.add_argument(
+        "--pipeline-parallelism-size",
+        help="Pipeline parallelism size",
+        type=int,
+        default=1,
+    )
     args = parser.parse_args()
+
+    # TODO: This is deceiving, it's the tokenizer directory
+    if args.tokenizer.endswith("tokenizer.json"):
```
rsuderman, line 169: Better to throw errors for the right path rather than assuming json name. This just leads to weird tokenizer naming behavior.

rsuderman (changes requested):

PR #2307, opened 2025-09-23
[sharktank] Fix chunked prefill export issue with dynamic shapes

archana-ramalingam:
For chunked prefill, the default `seq_len_blocks_dim` range causes numeric regression and is updated here.


sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -67,15 +67,19 @@ def generate_batch_prefill(bs: int):
...
             bs_min = 2
             bs_max = ceildiv(llama_config.block_seq_stride, bs)
             extend_bs = torch.export.Dim("extend_bs", min=bs_min, max=bs_max)
             dynamic_shapes["tokens"][0] = extend_bs
             dynamic_shapes["seq_lens"][0] = extend_bs
             dynamic_shapes["seq_block_ids"][0] = extend_bs
-        else:
-            bs_min = bs
+        elif export_config.has_prefill_position:
+            seq_block_dim = torch.export.Dim("seq_block_dim", max=block_dim_max)
```
sogartar, line 80: I think this dimension should be called something like `tokens_seq_len_block`.

Also the above dimension
```
```
Should be `seq_len_block` or `seq_len_block_id` instead of `block`. `block` is too ambiguous as we already use block to also mean a layer. It would be confusing if export fails due to some reason saying something about this dimension.

sogartar (approved):
On the WIP [branch](https://github.com/nod-ai/shark-ai/tree/users/sogartar/chunked-prefill-debugging) (on [snapshot](https://github.com/nod-ai/shark-ai/commit/f18e860d0f37103e09c108b723f41f06dc1a0674)) with the original/unmodified tracing args I saw the export error.
```
- The values of 32*sl_dim_ceildiv_block_seq_stride = L['args'][0][0].size()[1] must always be related to the values of seq_block_count = L['args'][0][3].size()[1] by 32*sl_dim_ceildiv_block_seq_stride = 32*seq_block_count.
````

@archana-ramalingam I suppose you do not see it when exporting on this branch?
If this exports file then probably the attention mask butchering on the WIP branch caused the bogus constraint. But then it worked in eager, so still strange.

This PR needs a better title.

It is worth thinking about how this will be used and if we are going to need later a variant where in chunked prefill the tokens sequence length is a static dimension. Would this yield better optimization in IREE?
Also we need tests in a subsequent PR.

PR #2312, opened 2025-09-23
[Sharktank] Auto-detect cache_quantizer dtype

Alex-Vasile:


rsuderman (approved):

dan-garvey (commented):
I encourage you to just drop support for the cache_quantizer route, it won't be used for newer generations of models, and the one that does use it isn't on the critical path afaik. I think doing a bunch of gymnastics to maintain support for it is probably not worth it, especially given it shouldn't have been implemented this way in the first place.

PR #2319, opened 2025-09-24
[Fusilli] Bump docker; add security-opt option for memory mapping

sjain-stanford:
New docker bumps IREE and TheRock to a more recent nightly release (useful for perf benchmarking). Also adds `--security-opt seccomp=unconfined` for memory mapping / perf as recommended by https://rocm.docs.amd.com/projects/install-on-linux/en/latest/how-to/docker.html#accessing-gpus-in-containers.

### DIFF
--- a/sharkfuser/build_tools/docker/exec_docker_ci.sh
+++ b/sharkfuser/build_tools/docker/exec_docker_ci.sh
@@ -17,7 +17,8 @@ done
 # Bind mounts for the following:
 # - current directory to /workspace in the container
 docker run --rm \
-           ${DOCKER_RUN_DEVICE_OPTS} \
            -v "${PWD}":/workspace \
-           ghcr.io/sjain-stanford/compiler-dev-ubuntu-24.04:main@sha256:5d70805daf1c9e89c3515e888fa54282982ce868784b13bb9c98f09d4fbd1c5b \
+           ${DOCKER_RUN_DEVICE_OPTS} \
+           --security-opt seccomp=unconfined \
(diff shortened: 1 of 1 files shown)


sharkfuser/build_tools/docker/exec_docker_ci.sh
```diff
@@ -17,7 +17,8 @@ done
 # Bind mounts for the following:
 # - current directory to /workspace in the container
 docker run --rm \
-           ${DOCKER_RUN_DEVICE_OPTS} \
            -v "${PWD}":/workspace \
-           ghcr.io/sjain-stanford/compiler-dev-ubuntu-24.04:main@sha256:5d70805daf1c9e89c3515e888fa54282982ce868784b13bb9c98f09d4fbd1c5b \
+           ${DOCKER_RUN_DEVICE_OPTS} \
+           --security-opt seccomp=unconfined \
```
AaronStGeorge, line 22: How was it working without this?

sjain-stanford: This is optional as stated in the ROCm guide but good for perf: 

> --security-opt seccomp=unconfined (optional)

> This option enables memory mapping, and is recommended for containers running in HPC environments. See [Optional security options (–security-opt)](https://docs.docker.com/reference/cli/docker/container/run/#security-opt).
> The performance of an application can vary depending on the assignment of GPUs and CPUs to the task. Typically, numactl is installed as part of many HPC applications to provide GPU/CPU mappings. This Docker runtime option supports memory mapping and can improve performance.

PR #2321, opened 2025-09-24
Update KVCache type for llama 70B FP8

pravg-amd:


dan-garvey (approved):
Fine with this, but I thought I saw a PR a few days ago that dropped use-hf for something that instead just specified the rotary embedding type, might be worth updating the invocations here as well.

scripts/export_and_compile.sh
```diff
@@ -88,7 +88,7 @@ elif [[ $DTYPE = "fp8" ]]; then
         --output-mlir=$OUTPUT_DIR/output.mlir \
         --output-config=$OUTPUT_DIR/config_attn.json \
         --bs-prefill=$PREFILL_BS --bs-decode=$DECODE_BS --attention-kernel sharktank \
-        --use-hf --device-block-count 8043
+        --use-hf --device-block-count 8043 --kv-cache-dtype=$KV_CACHE_DTYPE
```
IanNod, line 91: Why are we using 8043 for device-block-count? Seems like a strange number

PR #2327, opened 2025-09-24
[Sharktank] Fix ParallelismConfig.num_blocks_per_pipeline to produce correct result for PP1

Alex-Vasile:


sharktank/sharktank/layers/configs/llm_configs.py
```diff
@@ -476,7 +478,13 @@ def from_properties(properties: "PropertyValueType") -> "ParallelismConfig":
         kwargs = dict(properties)
         fields_name_set = set(field.name for field in fields(ParallelismConfig))
         kwargs = {k: v for k, v in kwargs.items() if k in fields_name_set}
-        return ParallelismConfig(**kwargs)
+
+        if len(kwargs) != 0:
+            return ParallelismConfig(**kwargs)
+        else:
```
rsuderman, line 484: You return in the if so you might as well drop the `else`

rsuderman (approved):

PR #2328, opened 2025-09-24
[Fusilli] Add lit tests with padding

sjain-stanford:
This includes the problematic case of `T -> Pad -> Conv -> T` which iree-compile doesn't optimize well today. Will update the lit test once the compiler pass is available. 

Also removes unnecessary checks from linalg input IR to make them less fragile. Eventually we'll want to clean them up further and possibly relocate to the IREE repo.

### DIFF
--- a/sharkfuser/tests/lit/test_conv_asm_emitter_nchw_kcrs.cpp
+++ b/sharkfuser/tests/lit/test_conv_asm_emitter_nchw_kcrs.cpp
@@ -84,17 +84,11 @@ int main() {
   // TORCH-CHECK:     }
   // TORCH-CHECK:   }
   //
-  // LINALG-CHECK:    util.func public @main$async(%arg0: !hal.buffer_view, %arg1: !hal.buffer_view, %arg2: !hal.buffer_view, %arg3: !hal.fence, %arg4: !hal.fence) attributes {inlining_policy = #util.inline.never, iree.abi.model = "coarse-fences", iree.abi.stub} {
-  // LINALG-CHECK:      %cst = arith.constant 0.000000e+00 : f32
-  // LINALG-CHECK:      %0 = hal.tensor.import wait(%arg3) => %arg1 : !hal.buffer_view -> tensor<16x128x64x32xf32>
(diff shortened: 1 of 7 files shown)


sharkfuser/tests/lit/test_conv_asm_emitter_nchw_kcrs.cpp
```diff
@@ -84,7 +84,6 @@ int main() {
   // TORCH-CHECK:     }
   // TORCH-CHECK:   }
   //
-  // LINALG-CHECK:  module @module {
   // LINALG-CHECK:    util.func public @main$async(%arg0: !hal.buffer_view, %arg1: !hal.buffer_view, %arg2: !hal.buffer_view, %arg3: !hal.fence, %arg4: !hal.fence) attributes {inlining_policy = #util.inline.never, iree.abi.model = "coarse-fences", iree.abi.stub} {
   // LINALG-CHECK:      %cst = arith.constant 0.000000e+00 : f32
   // LINALG-CHECK:      %0 = hal.tensor.import wait(%arg3) => %arg1 : !hal.buffer_view -> tensor<16x128x64x32xf32>
```
AaronStGeorge, line 89:
```
// LINALG-CHECK:      %[[CST:.+]] = arith.constant 0.000000e+00 : f32
```
Do you think it's worth capturing the variables in lit variables to avoid breakages due to reordering and mlir emission differences? 

Alternatively could we cut these down even more? Maybe just
```
 // LINALG-CHECK:   linalg.conv_2d_nchw_fchw {dilations = dense<1> : vector<2xi64>, strides = dense<1> : vector<2xi64>} ins(%{{.+}}, %{{.+}} : tensor<16x128x64x32xf32>, tensor<256x128x1x1xf32>) outs(%{{.+}} : tensor<16x256x64x32xf32>) -> tensor<16x256x64x32xf32>
```
and/or a `LINALG-CHECK-NOT` for ops that are expected to be folded away.

AaronStGeorge (approved):
LGTM though if there's something to be done to cut the linag tests down more that would be great

sharkfuser/tests/lit/test_conv_asm_emitter_nchw_kcrs.cpp
sjain-stanford: The lazy me did not want to do it but you persevered so.... :)

PR #2333, opened 2025-09-24
[FusilliPlugin] Stub plugin (`fusilli-plugin` 2 of N)

AaronStGeorge:
This PR adds "all the boring bits" of the fusilli plugin. It adds a basic engine plugin that hipDNN can load, which ensures that all of the required API definitions are present, and those APIs called as part of load function. The PR also tests for plugin loading, logging, and basic functionality.

### DIFF
--- a/.github/workflows/ci-fusilli-plugin.yml
+++ b/.github/workflows/ci-fusilli-plugin.yml
@@ -55,7 +55,7 @@ jobs:
               -DCMAKE_C_COMPILER=clang-18
               -DCMAKE_CXX_COMPILER=clang++-18
               -DFUSILLI_SYSTEMS_AMDGPU=ON
-              -DFUSILLI_DEBUG_BUILD=ON
+              -DCMAKE_BUILD_TYPE=Debug
               -DIREERuntime_DIR=/workspace/.cache/docker/iree/build/lib/cmake/IREE
 
     steps:


fusilli-plugin/src/fusilli_plugin.cpp
```diff
@@ -0,0 +1,407 @@
...
+
+hipdnnPluginStatus_t hipdnnPluginGetName(const char **name) {
+  LOG_API_ENTRY("name_ptr={:p}", static_cast<void *>(name));
+  FUSILLI_PLUGIN_CHECK_NULL(name);
+
+  *name = FUSILLI_PLUGIN_NAME;
+
+  LOG_API_SUCCESS_AUTO("pluginName={}", *name);
+  return HIPDNN_PLUGIN_STATUS_SUCCESS;
+}
```
AaronStGeorge, line 67: `hipDNN` uses a try catch pattern 
```C++
    return hipdnn_plugin::tryCatch([&, apiName = __func__]() {
        throwIfNull(name);

        *name = pluginName;

        LOG_API_SUCCESS(apiName, "pluginName={:p}", static_cast<void*>(name));
    });
```
fusilli (aka sharkfuser), and fusilli-plugin are using the LLVM style guide which discourages (well technically outright bans) exceptions, so I went with this slightly less exceptional (🤣) pattern.

sjain-stanford (approved):
LGTM other than nits. Let's also get a review from the hipDNN team if possible.

fusilli-plugin/src/fusilli_plugin.cpp
sjain-stanford: SGTM

fusilli-plugin/CMakeLists.txt
BrianHarrisonAMD, line 1: You probably don't need it yet, but there is some path variables that can be used for setting the plugin output dirs to match where hipDNN is expecting them.

Sample from MIOpen plugin [here](https://github.com/ROCm/hipDNN/blob/develop_deprecated/plugins/miopen_legacy_plugin/CMakeLists.txt#L98-L114).

fusilli-plugin/src/hipdnn_engine_plugin_execution_context.h
```diff
@@ -0,0 +1,29 @@
...
+// in hipdnnEnginePluginExecutionContext_t which is a pointer to the undefined
+// HipdnnEnginePluginExecutionContext. Each plugin must define
+// HipdnnEnginePluginExecutionContext in order to create something when hipDNN
+// asks for an execution context.
+//
+// The execution context should store what's needed to execute a given kernel
+// (plan in hipDNN parlance) in a hot loop without any overhead. For fusilli
+// plugin, that maps to constructing and storing a fusilli::Graph based on
+// hipDNN graph. When an execution is requested, it should be a simple lookup
+// for uid -> tensor attribute, then a graph execution.
```
BrianHarrisonAMD, line 20: 👍

fusilli-plugin/test/integration/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+    hipdnn_sdk
+    Threads::Threads
+)
+target_compile_definitions(fusilli_plugin_integration_tests PRIVATE
+    FUSILLI_PLUGIN_DIR="${FUSILLI_PLUGIN_DIR}"
+    FUSILLI_PLUGIN_NAME="${FUSILLI_PLUGIN_NAME}"
+    FUSILLI_PLUGIN_ENGINE_ID="${FUSILLI_PLUGIN_ENGINE_ID}"
+)
+# Register with CTest
+gtest_discover_tests(fusilli_plugin_integration_tests)
```
BrianHarrisonAMD, line 29: Heads up, maybe you won't hit the same issue, but we found in a couple cases now that gtest_discover_tests can add a ton of overhead to our tests due to each individual case being treated as a separate executable.

We swapped to adding the [test executable instead](https://github.com/ROCm/hipDNN/blob/ebba3d34660ed206491f33a7dc1825efe0571780/cmake/Tests.cmake#L198-L241).

BrianHarrisonAMD (approved):
LGTM!

fusilli-plugin/src/fusilli_plugin.cpp
```diff
@@ -0,0 +1,408 @@
...
+
+  LOG_API_SUCCESS_AUTO("engineDetails->ptr={:p}", engineDetails->ptr);
+  return HIPDNN_PLUGIN_STATUS_SUCCESS;
+}
+
+hipdnnPluginStatus_t
+hipdnnEnginePluginGetWorkspaceSize(hipdnnEnginePluginHandle_t handle,
+                                   const hipdnnPluginConstData_t *engineConfig,
+                                   const hipdnnPluginConstData_t *opGraph,
+                                   size_t *workspaceSize) {
```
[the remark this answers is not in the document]
AaronStGeorge: I'll be on the lookout for when that lands!

PR #2334, opened 2025-09-24
Add tests for all the shaping ops

KyleHerndon:
#2086 used to have both a migration and add tests. I split out the part with tests into a separate PR, this one, and 2086 will now be for the migration of the ops.


Alex-Vasile (approved):
Looks good, only minor nit is that not of the copyright dates are for this year even though the files are new.

PR #2336, opened 2025-09-25
[Fusilli] Fix release builds and refactor cmake flags

sjain-stanford:
There is a bug in this code which caused the release builds to never be exercised before:
```cmake
if(FUSILLI_DEBUG_BUILD OR FUSILLI_CODE_COVERAGE)
  message(STATUS "Setting CMAKE_BUILD_TYPE to Debug")
  set(CMAKE_BUILD_TYPE "Debug" CACHE STRING "Debug build" FORCE)
endif()
```

When `-DFUSILLI_DEBUG_BUILD=OFF` (and `-DFUSILLI_CODE_COVERAGE=OFF`), the `CMAKE_BUILD_TYPE` is left unset. If not, it is set to `Debug`. So it is never being set to `Release`.

We probably want something like this:

```cmake
if(NOT CMAKE_BUILD_TYPE)
  message(STATUS "Setting CMAKE_BUILD_TYPE to Release")
  set(CMAKE_BUILD_TYPE Release CACHE STRING "Release build" FORCE)
endif()
if(FUSILLI_CODE_COVERAGE)
  message(STATUS "Setting CMAKE_BUILD_TYPE to Debug")
  set(CMAKE_BUILD_TYPE "Debug" CACHE STRING "Debug build" FORCE)
endif()
```

With this fixed, it exposed a latent bug that made the lit test drivers to be unusable for release builds, because it did validate/emitAsm as part of the assert which are optimized away in release builds:
```c++
  assert(isOk(graph->validate()) && "Graph is invalid");
  ErrorOr<std::string> errorOrAsm = graph->emitAsm();
  assert(isOk(errorOrAsm) && "Graph ASM emission failed");
  std::cout << *errorOrAsm << std::endl;
```

This led to the following (somewhat cryptic) error:

```
13: /home/srajeshk/code/shark-ai/sharkfuser/build/bin/lit/test_conv_asm_emitter_nhwc_krsc_with_pad | iree-opt --verify-roundtrip
13: # executed command: /home/srajeshk/code/shark-ai/sharkfuser/build/bin/lit/test_conv_asm_emitter_nhwc_krsc_with_pad
13: # .---command stderr------------
13: # | terminate called after throwing an instance of 'std::bad_variant_access'
13: # |   what():  std::get: wrong index for variant
13: # `-----------------------------
13: # error: command failed with exit status: -6
13: # executed command: iree-opt --verify-roundtrip
13: # .---command stdout------------
13: # | module {
13: # | }
13: # | 
13: # `-----------------------------
```

This PR 
- fixes the lit drivers to use `FUSILLI_CHECK_ERROR` and `FUSILLI_TRY` macros instead of the raw asserts,
- removes the redundant `FUSILLI_DEBUG_BUILD ` flag
- adds a new `FUSILLI_ENABLE_LOGGING` flag to set the logging env vars for tests/samples automatically.

### DIFF
--- a/.github/workflows/ci-fusilli-plugin.yml
+++ b/.github/workflows/ci-fusilli-plugin.yml
@@ -54,6 +54,7 @@ jobs:
             fusilli-plugin-cmake-options:
               -DCMAKE_C_COMPILER=clang-18
               -DCMAKE_CXX_COMPILER=clang++-18
+              -DCMAKE_BUILD_TYPE=Debug
               -DFUSILLI_SYSTEMS_AMDGPU=ON
 
     steps:
--- a/.github/workflows/ci-sharkfuser.yml
+++ b/.github/workflows/ci-sharkfuser.yml
@@ -53,31 +53,29 @@ jobs:
             cmake-options:
               -DCMAKE_C_COMPILER=clang-18
               -DCMAKE_CXX_COMPILER=clang++-18
-              -DFUSILLI_DEBUG_BUILD=OFF
-              -DFUSILLI_CODE_COVERAGE=OFF
+              -DCMAKE_BUILD_TYPE=Release
               -DFUSILLI_SYSTEMS_AMDGPU=ON
           - name: cpu_clang18_debug
             runs-on: azure-cpubuilder-linux-scale
             cmake-options:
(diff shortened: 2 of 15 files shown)


AaronStGeorge (approved):
LGTM thanks!

PR #2344, opened 2025-09-25
Add dataset for 405b eval_llm_vmfb

pravg-amd:
1. Add dataset for 405B FP4 model
2. Enable eval_llm_vmfb check in CI for 405B FP4 model
3. Export model with and without topk
4. Enable run_llm_vmfb tests for 405B FP4 model


IanNod (approved):
Overall I think this looks fine, but I do see a failing test. It does not look related to this PR but please verify before merging

PR #2352, opened 2025-09-26
Add MLIR mxfp4 matmul kernel

jtuyls:
Adds a pure MLIR-based mxfp4 matmul kernel to be compiled through IREE.


sharktank/sharktank/ops/custom_impls.py
```diff
@@ -112,6 +113,37 @@ def matmul_generic_tensor_block_scaled_i4(
...
+@matmul.override(Tensor, QuantizedTensor, impl_name="sharktank.iree")
+def matmul_generic_tensor_block_scaled_fp4_iree(
+    lhs, rhs: QuantizedTensor, *, transpose_rhs: bool
+):
+    """Generic kernel for FP4 E2M1 block scaled layouts."""
+
+    if rhs.layout_type is not BlockScaledFp4Layout:
+        return NotImplemented
+
+    if not torch.compiler.is_compiling():
```
KyleHerndon, line 125: Can IREE not compile this to CPU?

If IREE *can* compile it to CPU, we can take out this if statement. Otherwise, leave it in.

PR #2362, opened 2025-09-26
[Tuner] use custom transform op for matching contraction op

bangtianliu:
This PR updates the contraction matching logic to use the newly defined transform ops `transform.iree.match.contraction` and `transform.iree.match.dims_equal` instead of relying on `transform.iree.match.cast_compatible_dag_from_root`.

The current change is intended as a temporary solution to demonstrate the functionality of the new transform ops. Once similar transform ops are defined for convolution and attention (following the same design), and the custom ops are properly registered with the Python bindings, the `build_td_spec `function will be fully rewritten using python bindings instead of textual strings.

FYI, tuning log for matmul based on this PR: https://gist.github.com/bangtianliu/0c33db3225757ec722d6eb7d03b2d989.

### DIFF
--- a/sharktuner/sharktuner/spec_builder.py
+++ b/sharktuner/sharktuner/spec_builder.py
@@ -89,10 +89,9 @@ def build_td_spec(
         ]
 
         dims_equal_checks = []
-        if batch_dims:
-            dims_equal_checks.append(
-                f"transform.iree.match.dims_equal %batch, {batch_dims} : !transform.param<i64>"
-            )
+        dims_equal_checks.append(
+            f"transform.iree.match.dims_equal %batch, {batch_dims} : !transform.param<i64>"
+        )
         dims_equal_checks.append(
             f"transform.iree.match.dims_equal %m, {m_dims} : !transform.param<i64>"
         )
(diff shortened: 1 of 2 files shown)


sharktuner/sharktuner/spec_builder.py
```diff
@@ -57,25 +57,82 @@ def build_td_spec(
...
+        ]
+        n_dims = [
+            rhs_type_shape.shape[rhs_dims.index(dim)] for dim in contraction_dims.n
+        ]
+        k_dims = [
+            lhs_type_shape.shape[lhs_dims.index(dim)] for dim in contraction_dims.k
+        ]
+        batch_dims = [
+            lhs_type_shape.shape[lhs_dims.index(dim)] for dim in contraction_dims.batch
+        ]
```
kuhar, line 89: Are you sure the lhs always knows about the batch dim? In existing tuning specs for Unet, we sometimes had the RHS being broadcast: https://github.com/nod-ai/sdxl-scripts/blob/4fa7ccbc3de4873c0751a48ef9b4b86a7f24428e/int8-model/specs/attention_and_matmul_spec.mlir#L632-L648

I think it could also be that LHS gets broadcast.

kuhar (approved):
LGTM. Just checking, have you run the tuner and confirmed it works?

bangtianliu:
> LGTM. Just checking, have you run the tuner and confirmed it works?

Yes, I did.

PR #2365, opened 2025-09-27
[FusilliPlugin] Graph conversion (`fusilli-plugin` 3 of N)

AaronStGeorge:
This PR implements [`hipdnnEnginePluginGetApplicableEngineIds`](https://github.com/ROCm/rocm-libraries/blob/7e29d65a8db488a5e038c3616398144d7c2289a5/projects/hipdnn/sdk/include/hipdnn_sdk/plugin/EnginePluginApi.h#L90-L115) and [`hipdnnEnginePluginCreateExecutionContext`](https://github.com/ROCm/rocm-libraries/blob/7e29d65a8db488a5e038c3616398144d7c2289a5/projects/hipdnn/sdk/include/hipdnn_sdk/plugin/EnginePluginApi.h#L170-L191), which together completes `hipDNN` -> `fusilli` graph translation for the initial plugin goal of e2e run of single node ConvFprop graph.


fusilli-plugin/include/graph_import.h
```diff
@@ -0,0 +1,247 @@
...
+  // (inputs and outputs). Used by hipdnnEnginePluginExecuteOpGraph to match
+  // incoming device buffers (identified by UID) to their corresponding
+  // fusilli::TensorAttr.
+  //
+  // All tensors in this map should be non-virtual boundary tensors---a virtual
+  // tensor is an internal intermediate---internal tensors will be tracked
+  // elsewhere. Currently we only support single node graphs, so there are no
+  // virtual tensors to track.
+  std::unordered_map<int64_t, std::shared_ptr<fusilli::TensorAttr>>
+      uidToIOTensor;
```
sjain-stanford, line 83: This is maybe a question for the hipDNN folks (@BrianHarrisonAMD) - How prevalent is the UID based variantPack compared to the name based variantPack? I see both being exposed in cudnn frontend. In fusilli, we're only supporting name based. I guess the plugin here is basically trying to map from UID -> name (which is not too bad) but if this is something we can do without, we should clean it up.

fusilli-plugin/include/graph_import.h
```diff
@@ -0,0 +1,247 @@
...
+    std::shared_ptr<fusilli::TensorAttr> x =
+        FUSILLI_TRY(importNodeInput(hipDnnConvFwdAttr->x_tensor_uid(), "x"));
+    std::shared_ptr<fusilli::TensorAttr> w =
+        FUSILLI_TRY(importNodeInput(hipDnnConvFwdAttr->w_tensor_uid(), "w"));
+
+    // Import node.
+    auto fusilliConvFwdAttr =
+        fusilli::ConvFPropAttr()
+            // TODO(#2369): pre_padding and post_padding vs padding
+            .setPadding(*hipDnnConvFwdAttr->post_padding())
```
sjain-stanford, line 147: We could probably assert here that post_padding == pre_padding, since Fusilli only supports `padding` (so symmetric).

AaronStGeorge: +1 we should add a check to `hipdnnEnginePluginGetApplicableEngineIds` (and a test for said check) to ensure we never get to the point where we are attempting to create an execution context for a conv node with asymmetric padding as well.
https://github.com/nod-ai/shark-ai/blob/a56c94ba8f9e48df14439955db1b208dc3ce188d/fusilli-plugin/src/fusilli_plugin.cpp#L210-L226

fusilli-plugin/include/graph_import.h
```diff
@@ -0,0 +1,247 @@
...
+// organization and is used by importGraph.
+//
+// Graph import is designed around individual Node import functions (such as
+// importConvFPropAttr) which convert a given node type, and track input and
+// output tensors in shared state (via importNodeInput and importNodeOutput
+// functions). Graph nodes are processed in topological order to ensure that
+// outputs of producer nodes are tracked and available for consuming nodes.
+//
+// NOTE: topological ordering not yet implemented. As we're only handling
+// single-node graphs currently, that's not a problem.
```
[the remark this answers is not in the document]
BrianHarrisonAMD: For now you can likely just assume it's sorted, and eventually there will be a method on the graph wrapper for getting the topological order of nodes.

fusilli-plugin/include/graph_import.h
AaronStGeorge: > I am not 100% sure how this will work on your side with the intermediates and the TensorAttrs for Fusilli, but for the hipDNN schema, virtual tensors will also have Ids, but just won't have a variant pack device buffer.

I think we're aligned well here, any intermediates will be tracked for graph translation, but when it comes to execution we will only care about the id <> `fusilli::TensorAttr` mapping so we can properly map input and output buffers.

PR #2367, opened 2025-09-28
[CI] Run fewer pull_rquest checks on unrelated changes

kuhar:
The intention is to stop running sharktank, sharkfuser/fusilli, and shortfin CI checks on sharktuner pull requests.

Tuner changes should never affect the other projects.

### DIFF
--- a/.github/workflows/ci-fusilli-plugin.yml
+++ b/.github/workflows/ci-fusilli-plugin.yml
@@ -11,6 +11,10 @@ on:
   pull_request:
     branches:
       - main
+    paths:
+      - '.github/workflows/ci-sharfuser.yml'
+      - 'sharkfuser/**'
+      - 'fusilli-plugin/**'
   push:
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-fusilli-plugin.yml
```diff
@@ -11,6 +11,10 @@ on:
   pull_request:
     branches:
       - main
+    paths:
+      - '.github/workflows/ci-sharkfuser.yml'
```
sjain-stanford, line 15: should perhaps also include `.github/workflows/ci-fusilli-plugin.yml` here?

sjain-stanford (approved):
Nice. I always wanted to do this but didn't know it was this simple. In Bazel we could simply define a condition based on the dependency graph to determine which changes should trigger/not trigger a certain test suite. 

LGTM besides one comment.

PR #2378, opened 2025-09-29
[sharktank] make IreeInstance expose LlamaModelConfig to be consistent with TorchInstance

sogartar:
Although these two classes do not inherit an ABC explicitly now, they need to have the same interface.


Alex-Vasile (approved):
They should probably inherent from the same class (or equivalent) if we expect them to be interchangeable.

PR #2393, opened 2025-09-30
[sharktank] Remove redundant parallelism fields in LlamaModelConfig.to_properties

sogartar:
We are already saving these in `parallelism_config`.


sharktank/sharktank/layers/configs/llm_configs.py
```diff
@@ -633,9 +633,6 @@ def to_properties(self) -> "PropertyValueType":
         res["activation_dtype"] = dtype_to_serialized_name(self.activation_dtype)
         res["attention_dtype"] = dtype_to_serialized_name(self.attention_dtype)
         res["fake_quant"] = self.fake_quant
-        res["tensor_parallelism_size"] = self.tensor_parallelism_size
```
Alex-Vasile, line 636: `tensor_parallelism_size` is not being saved by parallelism_config

sogartar: Nevermind, its not that.

PR #2395, opened 2025-09-30
[Fusilli] Picks IREE compile fixes for pad/transpose propagation and check num dispatches

sjain-stanford:
Since we have different dispatch counts based on the backend / compiler flags, we now configure LIT check prefixes based on the configured backend.


IanWood1 (approved):
Nice!

PR #2399, opened 2025-09-30
Add page allocation to llm_utils

rsuderman:
LLM Utils had a limited page allocation system. Updated so that page allocation is performed in a more intelligent way.

PR #2407, opened 2025-10-02
[Sharktank] vmfb-runner `LlmTask`

stbaione:
Stacked on #2399 

- This is a basic PR that introduces `LlmTask` and `LlmTaskInput` components to the `vmfb-runner`
    - First cut is just introducing and using the classes
    - Follow up PR will introduce a basic `Scheduler`, where we will queue-up all of the `LlmTasks`, then pop and run from that queue. 
    - After that, we will enable a basic `ChunkScheduler`, so that we can do chunked-prefill inside of the `vmfb-runner`
- Rename `LlmBatch` -> `LlmBatcher`

### DIFF
--- a/app_tests/integration_tests/llm/shortfin/tinystories_llama2_25m_test.py
+++ b/app_tests/integration_tests/llm/shortfin/tinystories_llama2_25m_test.py
@@ -102,12 +102,19 @@ class TestLLMServer:
     def test_basic_generation(
         self,
         server: tuple[Any, int, ServerConfig],
+        request: pytest.FixtureRequest,
     ) -> None:
         """Tests basic text generation capabilities.
 
         Args:
             server: Tuple of (process, port) from server fixture
         """
+        test_id = request.node.callspec.id
+        if "trie" in test_id:
+            pytest.skip(
+                reason="TrieAttentionCache APIs are under development, skip it for now."
+            )
+
         process, port, config = server
         assert process.poll() is None, "Server process terminated unexpectedly"
         dataset = (
@@ -138,12 +145,19 @@ class TestLLMServer:
     def test_multi_page_generation(
         self,
         server: tuple[Any, int, ServerConfig],
+        request: pytest.FixtureRequest,
     ) -> None:
         """Tests multi-page text generation capabilities.
 
         Args:
             server: Tuple of (process, port) from server fixture
         """
+        test_id = request.node.callspec.id
+        if "trie" in test_id:
+            pytest.skip(
+                reason="TrieAttentionCache APIs are under development, skip it for now."
+            )
+
         process, port, config = server
         assert process.poll() is None, "Server process terminated unexpectedly"
         dataset = (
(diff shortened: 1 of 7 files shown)


sharktank/sharktank/utils/llm_utils.py
```diff
@@ -266,92 +268,160 @@ def allocate(self, *shape, dtype, device_index: int):
...
+        cache: iree.runtime.DeviceArray | torch.Tensor,
+    ) -> List[numpy.ndarray | iree.runtime.DeviceArray | torch.Tensor]:
+        block_stride = self._block_stride
+        bs = self._batch_size
+
+        tokens = [task_input.tokens for task_input in task_inputs]
+        page_ids = [task_input.pages for task_input in task_inputs]
+
+        max_len = max(len(input_tokens) for input_tokens in tokens)
+        blocks = math.ceil(max_len / block_stride)
```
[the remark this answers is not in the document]
stbaione: Updated

sharktank/sharktank/utils/llm_utils.py
```diff
@@ -266,6 +268,172 @@ def allocate(self, *shape, dtype, device_index: int):
...
+    ) -> List[numpy.ndarray | iree.runtime.DeviceArray | torch.Tensor]:
+        pass
+
+    @abstractmethod
+    def _process_results(
+        self, results
+    ) -> Tuple[numpy.ndarray, Optional[numpy.ndarray]]:
+        pass
+
+    def _fill_page_table(
```
rsuderman, line 312: Can we move this to util function instead? I am being a hypocrite because I added it but seeing it as part of tasks is a bit weird considering its a static function anyway.

sharktank/sharktank/utils/llm_utils.py
```diff
@@ -266,6 +268,172 @@ def allocate(self, *shape, dtype, device_index: int):
...
+
+
+class LlmTask(ABC):
+    def __init__(
+        self,
+        task_type: LlmTaskType,
+        instance: IreeInstance | TorchInstance,
+        llm_task_inputs: List[LlmTaskInput],
+        batch_size: int,
+        block_stride: int,
```
rsuderman, line 292: Add an invocation function to the constructor

sharktank/sharktank/utils/llm_utils.py
```diff
@@ -266,6 +268,172 @@ def allocate(self, *shape, dtype, device_index: int):
...
+
+    def run(
+        self, cache_state: iree.runtime.DeviceArray | torch.Tensor
+    ) -> Tuple[numpy.ndarray, Optional[numpy.ndarray]]:
+        task_type = self._task_type
+        task_inputs = self._task_inputs
+
+        args = self._prepare_args(task_inputs, cache_state)
+        instance = self._instance
+        invocation_fn = getattr(instance, task_type.name.lower())
```
rsuderman, line 330: Pass this in as a construciton instead. You won't need `task_type` then.

rsuderman (approved):

PR #2408, opened 2025-10-02
Update BaseAttentionCache APIs

lisaliu1:
This PR includes the following work:
1. update BaseAttentionCache APIs to work with decoder
2. update related tests.

### DIFF
--- a/shortfin/python/shortfin_apps/llm/components/kvcache/base_attention_cache.py
+++ b/shortfin/python/shortfin_apps/llm/components/kvcache/base_attention_cache.py
@@ -54,8 +54,12 @@ class BasePagedAttentionCache:
     ):
         self.page_pool = page_pool
         self.tokens_per_page = tokens_per_page
+        self._allocated_pages: List[
+            PageInfo
+        ] = []  # global allocated page pool that contains all un-tracked pages
 
     def shutdown(self):
+        self.page_pool.free_pages(self._allocated_pages)
         available = self.page_pool.available_page_count()
         total = self.page_pool.total_page_count()
         if available != total:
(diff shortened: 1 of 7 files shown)


zeeshanhaque21 (approved):

shortfin/python/shortfin_apps/llm/components/kvcache/base_attention_cache.py
```diff
@@ -77,12 +81,27 @@ def fork_pages(self, tokens: list[int], cache_info: CacheInfo) -> CacheInfo:
...
+        )
+
+    def get_allocated_pages(self, page_ids: List[int]) -> List[PageInfo]:
+        pages = []
+        for page in self._allocated_pages:
+            if page.index in page_ids:
+                pages.append(page)
+        return pages
+
     def allocate(
```
stbaione, line 100: Do we need to return a `CacheInfo` object from `allocate`, or could we just return `List[PageInfo]`, and add the acquired pages to the `CacheInfo` object stored on `LlmInferenceExecRequest`? It seems like the page_ids is the only piece of data we need from this func. Could simplify things going forward

shortfin/python/shortfin_apps/llm/components/kvcache/base_attention_cache.py
```diff
@@ -77,12 +81,27 @@ def fork_pages(self, tokens: list[int], cache_info: CacheInfo) -> CacheInfo:
...
+    def lookup(self, tokens: List[int]) -> CacheInfo:
+        return CacheInfo(
+            num_tokens=0,
+            tokens=[],
+            pages=[],
+            pool=self.page_pool,
+            last_cached_node=None,
+        )
+
+    def get_allocated_pages(self, page_ids: List[int]) -> List[PageInfo]:
```
stbaione, line 93: This doesn't seem to be used anywhere

stbaione (changes requested):

shortfin/python/shortfin_apps/llm/components/kvcache/base_attention_cache.py
lisaliu1: will be used in next PR.

PR #2410, opened 2025-10-02
[Fusilli] Benchmark Driver 1/N

sjain-stanford:
Adds a master benchmark driver (with CLI11 for arg parsing) that supports subcommands for benchmarking different ops (with different set of flags). Also adds CMake support to easily add benchmark suites to CI. 

Example invocation:
```
./driver conv -n 128 -c 384 -H 24 -W 48 -k 384 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 
```
(in the future)
```
./driver sdpa <flags>
```

In a follow-on (2/N...), I'll add:
- [ ] support for choosing layouts
- [ ] support for 3D spatial dims
- [ ] support for varying types - `convfp16`, `convbf16`

### DIFF
--- a/sharkfuser/benchmarks/driver.cpp
+++ b/sharkfuser/benchmarks/driver.cpp
@@ -20,7 +20,7 @@ ErrorObject benchmark_conv_fprop(int64_t n, int64_t c, int64_t h, int64_t w,
                                  int64_t k, int64_t r, int64_t s, int64_t u,
                                  int64_t v, int64_t p, int64_t q, int64_t l,
                                  int64_t j) {
-#ifdef FUSILLI_SYSTEMS_AMDGPU
+#ifdef FUSILLI_ENABLE_AMDGPU
   Handle handle = FUSILLI_TRY(Handle::create(Backend::GFX942));
 #else
   Handle handle = FUSILLI_TRY(Handle::create(Backend::CPU));
--- a/sharkfuser/benchmarks/driver.cpp
+++ b/sharkfuser/benchmarks/driver.cpp
@@ -10,11 +10,18 @@
 
 #include <CLI/CLI.hpp>
 #include <cstdint>
+#include <limits>
 #include <memory>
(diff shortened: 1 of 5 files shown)


sharkfuser/benchmarks/driver.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+int main(int argc, char **argv) {
+  CLI::App mainApp{"Fusilli Benchmark Driver"};
+  mainApp.require_subcommand(1);
+
+  // Conv flags are kept in sync with MIOpen's ConvDriver:
+  // https://github.com/ROCm/rocm-libraries/blob/db0544fb61f2c7bd5a86dce98d4963420c1c741a/projects/miopen/driver/conv_driver.hpp#L878
+  CLI::App *convApp =
+      mainApp.add_subcommand("conv", "Fusilli Benchmark Forward Convolution");
+  int64_t n = 100, c = 3, h = 32, w = 32, k = 32, r = 3, s = 3, u = 1, v = 1,
```
MaheshRavishankar, line 100: Is this the max value? Do we need to even specify this? Quick scan of things, we should probably make all of the required?

MaheshRavishankar (approved):
Nice! Looks good. Just one comment.

sharkfuser/benchmarks/driver.cpp
sjain-stanford: These are defaults from MIOpen's driver: https://github.com/ROCm/rocm-libraries/blob/develop/projects/miopen/driver/conv_driver.hpp#L915
So we're at parity with it when invoked without any args. They can be made required as well if needed.

sjain-stanford: Removed defaults and made all options required: https://github.com/nod-ai/shark-ai/pull/2410/commits/402c6258c079cb51db208c8d28746637ec600d39

Also added range checks on options.

PR #2418, opened 2025-10-03
[Sharktank] Llm Task Scheduler

stbaione:
## Description

- Introduces a `Scheduler` into the `vmfb-runner`
- Adds a `submit + run` flow to the `Batcher`
    - Each request is given a single `LlmTaskInput`, which gets added to the `Scheduler`, with a `count` variable.
    - This `count` var allows the scheduler to determine how long each request should be tracked. For example, the `decode_scheduler` will track each task for at most `steps` number of schedules.
 - The `Decoder` submits the tasks to the `Batcher`, then returns the results from calling `run`
 - Currently, keeping the standalone prefill/decode functions for the other components using `Batcher` (i.e. perplexity). Would need to think of a way to handle the fact that PPL needs all of the raw logits, not just the selections, but that should be possible to do. Maybe by moving the `selection` logic back outside of `Batcher`, and just providing it a callback for where it should submit the logits + indices.
 - Splits components into `llm_scheduler.py`, `llm_task.py` and `llm_utils.py`, because the file was getting complex

### DIFF
--- a/sharktank/sharktank/utils/llm_scheduler.py
+++ b/sharktank/sharktank/utils/llm_scheduler.py
@@ -1,0 +1,53 @@
+from dataclasses import dataclass
+from sharktank.utils.llm_task import LlmTaskInput
+from typing import List, Dict
+
+
+@dataclass
+class SchedulerEntry:
+    task: LlmTaskInput
+    remaining_count: int
+
+
+class Scheduler:
+    def __init__(self, batch_size: int):
+        self._batch_size = batch_size
+        self._queue: List[LlmTaskInput] = []
+        self._schedule_count: Dict[str, SchedulerEntry] = {}
+
+    def schedule_task(self, task: LlmTaskInput, count: int):
+        self._queue.append(task)
+        self._schedule_count[task.task_id] = SchedulerEntry(
+            task=task, remaining_count=count
+        )
+
+    def get_scheduled_tasks(self) -> List[LlmTaskInput]:
+        batch_size = self._batch_size
+        schedule_tasks = self._queue[:batch_size]
+        self._queue = self._queue[batch_size:]
+
+        for task in schedule_tasks:
+            self._schedule_count[task.task_id].remaining_count -= 1
+
+        return schedule_tasks
+
+    def has_pending_tasks(self) -> bool:
+        return len(self._queue) > 0
+
+    def on_task_complete(self, task_id: str, last_token: int) -> bool:
+        task = self._schedule_count[task_id].task
+        task.tokens.append(last_token)
+        task.seq_len = len(task.tokens)
+        task.start_position = task.seq_len - 1
+
+        if self._schedule_count[task_id].remaining_count == 0:
+            del self._schedule_count[task_id]
+            return True
+
+        self._queue.append(self._schedule_count[task_id].task)
+        return False
+
+    def remove_task(self, task_id: str):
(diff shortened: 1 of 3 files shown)


sharktank/sharktank/utils/llm_task.py
```diff
@@ -0,0 +1,167 @@
...
+class LlmTaskInput:
+    task_id: str
+    tokens: List[int]
+    seq_len: int
+    pages: List[int]
+
+    start_position: Optional[int] = None
+
+
+class LlmTask(ABC):
```
rsuderman, line 30: I would move this to the scheduler and generalize it.  Most of the basics are pretty use case vague and it would be easy to write some tests on the scheduler + task

sharktank/sharktank/utils/llm_utils.py
```diff
@@ -538,6 +391,83 @@ def decode(
...
+
+        for i, request in enumerate(requests):
+            task_input = LlmTaskInput(
+                task_id=f"req_{i}",
+                tokens=request,
+                seq_len=len(request),
+                pages=page_ids[i],
+            )
+            # Submit prefill task
+            self._prefill_scheduler.schedule_task(task_input, 1)
```
rsuderman, line 410: This feels wrong - why are you submitting to both prefill and decode?

sharktank/sharktank/utils/llm_utils.py
```diff
@@ -538,6 +391,83 @@ def decode(
...
+            if not self._decode_scheduler.has_pending_tasks():
+                break
+            task_inputs = self._decode_scheduler.get_scheduled_tasks()
+            decode_task = DecodeTask(
+                invocation_fn=self._instance.decode,
+                llm_task_inputs=task_inputs,
+                batch_size=self._decode_bs,
+                block_stride=self._block_stride,
+            )
+            logits, indices = decode_task.run(*self._cache)
```
rsuderman, line 458: This looks wrong.  You make a task then call run on it then  call the scheduler? You should be making a job and submitting it. The scheduler should basically have a `schedule` command for each task (a task being a separate request). Then a call for competion. Something is off on this.

sharktank/sharktank/utils/llm_utils.py
```diff
@@ -538,6 +391,83 @@ def decode(
...
+            last = selection_fn(logits, indices, [0] * len(task_inputs))
+            for task_input, token in zip(task_inputs, last):
+                selections_map[task_input.task_id].append(token)
+                self._decode_scheduler.on_task_complete(
```
rsuderman, line 462: This on task completion callback is wrong. You should submit then process on the returned results.

sharktank/sharktank/utils/llm_utils.py
```diff
@@ -538,6 +391,83 @@ def decode(
...
+                    task_input.task_id,
+                    token,
+                )
+                if token == eos_token:
+                    self._decode_scheduler.remove_task(task_input.task_id)
```
rsuderman, line 467: Schedulers should take requests, process them, then be done. You shouldn't be externally managing.

sharktank/sharktank/utils/llm_scheduler.py
rsuderman, line 1: Make enough parts in scheduler and write some tests for it. I think you will understand the issues with your API if you attempt to use it standalone vs retrofitting it into the old process.

rsuderman (changes requested):

PR #2425, opened 2025-10-06
[fusilli] Add asm emitter for pointwise node

IanWood1:
Implements ASM emitter methods for `PointwiseNode`. This is a followup to https://github.com/nod-ai/shark-ai/pull/2416.


sjain-stanford (commented):
I haven't looked at this in depth yet, but just from a cursory look this doesn't include the "fusions" with convFProp right? Is that to come in a follow-on PR?

sjain-stanford:
> Would you suggest adding a new lit test file for testing fusion with convFProp?

Sure. Looks like you already have them in separate lit tests, which is how I'd have done it too. There's only so much you can cram in a single test, and they're not that expensive to have multiple of :)

PR #2427, opened 2025-10-06
[Sharktank] Scheduler for `vmfb-runner`

stbaione:
# Description

- Introduce very simple `Scheduler` component
- Split components into separate files
- Modify `LlmDecoder` to use new flow

This also enables us to run any size dataset. Previously, we would throw in error if, for example, `bs_prefill == 4` and we tried to run with `5 prompts`.

With these changes, you can request any number of prompts, and it works correctly.

### DIFF
--- a/sharktank/sharktank/utils/llm_utils.py
+++ b/sharktank/sharktank/utils/llm_utils.py
@@ -19,6 +19,7 @@ This module is not intended to be run directly, but is imported by other compone
 """
 
 import dataclasses
+from typing import Callable, List
 import iree.runtime
 import math
 import numpy
@@ -36,7 +37,9 @@ from sharktank.models.llm.config import ServiceConfig
 from sharktank.models.llm import PagedLlmModelV1
 from sharktank.types import Dataset, Theta
 from sharktank.utils.attention import *
-from typing import Callable, List, Optional, Tuple
+from sharktank.utils.llm_tasks import LlmTaskInput, PrefillTask, DecodeTask
+from sharktank.utils.llm_scheduler import Scheduler
+from uuid import uuid4
 
 np_dtype_to_torch_dtype = {
     # This torch-to-torch map is an abuse to circumvent that numpy does not have bf16.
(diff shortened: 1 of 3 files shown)


sharktank/sharktank/utils/llm_utils.py
```diff
@@ -525,13 +354,93 @@ def allocate(
...
+            self._decode_scheduler.schedule_task(task_input)
+
+    def run_prefill(
+        self,
+        selection_fn: Callable[
+            [numpy.ndarray, Optional[numpy.ndarray], List[int]], List[int]
+        ],
+    ):
+        selections = []
+        while self._prefill_scheduler.has_pending_tasks():
```
rsuderman, line 389: You should just do something like `self._prefill_scheduler.run` as it already has the task inputs. Everything else being passed is static so the invoker can run. Having an external loop like this is not really a behavior of the class.

sharktank/sharktank/utils/llm_utils.py
```diff
@@ -525,13 +354,93 @@ def allocate(
...
+        return selections
+
+    def run_decode(
+        self,
+        selection_fn: Callable[
+            [numpy.ndarray, Optional[numpy.ndarray], List[int]], List[int]
+        ],
+    ):
+        selections = []
+        while self._decode_scheduler.has_pending_tasks():
```
rsuderman, line 415: Ditto.

sharktank/sharktank/utils/llm_utils.py
stbaione: Updated

sharktank/sharktank/utils/llm_utils.py
```diff
@@ -525,13 +362,63 @@ def allocate(
...
+            self._decode_scheduler.schedule_task(task_input)
+
+    def run_prefill(
+        self,
+        selection_fn: Callable[
+            [numpy.ndarray, Optional[numpy.ndarray], List[int]], List[int]
+        ],
+    ):
+        return self._prefill_scheduler.run(
+            invocation_fn=self._instance.prefill,
```
rsuderman, line 397: These functions could be passed in on construction.

sharktank/sharktank/utils/llm_utils.py
```diff
@@ -525,13 +362,63 @@ def allocate(
...
+        )
+
+    def run_decode(
+        self,
+        selection_fn: Callable[
+            [numpy.ndarray, Optional[numpy.ndarray], List[int]], List[int]
+        ],
+    ):
+        return self._decode_scheduler.run(
+            invocation_fn=self._instance.decode,
```
rsuderman, line 409: Ditto.

rsuderman (changes requested):

sharktank/sharktank/utils/llm_utils.py
stbaione: Moved the `invocation_fn` to the `Scheduler` construction, but kept the `selection_fn` here. Since `LlmRunner` is owned by `LlmDecoder` and `LlmDecoder` defines the `selection_fn`, I'm not able to move that to the constructor

rsuderman (approved):

PR #2429, opened 2025-10-06
[Fusilli] Update architecture diagram for README

sjain-stanford:
As titled.


MaheshRavishankar:
So `g->compile()` and `g->execute()` both go to cache. My mental model was that `g-compile()` just goes to cache and recompiles on cache miss. But `g->execute()` just picks up either the newly generated kernel or just what was retrieved on cache hit. So `g->execute()` "reads from the cache" instead of "writing to cache".

sjain-stanford:
`g->execute()` just checks that the graph was compiled at least once and a cache is populated. So in the true sense it doesn't go to the cache. 

Do you prefer this version instead?


Untitled-2025-06-24-1340

PR #2432, opened 2025-10-06
Make pipeline parallelism size configurable

dezhiAmd:
Why:
This is the first step for verifying pp size 4, 8 beside 2. The change in conftest.py can also be useful for verification using ireeInstance
How:
To test different parallelism, use command:
pytest ./integration/models/llama/llama_integration_test.py --test-data-dir /shark-dev --pipeline-parallelism-size 8 --log-cli-level=INFO -m "expensive"

### DIFF
--- a/sharktank/conftest.py
+++ b/sharktank/conftest.py
@@ -239,7 +239,7 @@ def pytest_addoption(parser):
         "--pipeline-parallelism-size",
         action="store",
         type=int,
-        default=2,
+        default=1,
         help="Number of devices for pipeline parallel sharding",
     )
 


sharktank/integration/models/llama/llama_integration_test.py
```diff
@@ -25,7 +25,9 @@
 
 @pytest.mark.expensive
 def test_pruned_llama3_405b_f4_pipeline_parallel_eager_vs_eager_perplexity(
-    deterministic_random_seed, test_data_dir: Path
+    deterministic_random_seed,
+    test_data_dir: Path,
+    pipeline_parallelism_size: int,
```
Alex-Vasile, line 30: I don't see the flag hooked up to the class.

sharktank/conftest.py
```diff
@@ -234,6 +234,15 @@ def pytest_addoption(parser):
         default=1,
         help="Number of devices for tensor parallel sharding",
     )
+
+    parser.addoption(
+        "--pipeline-parallelism-size",
+        action="store",
+        type=int,
+        default=2,
```
Alex-Vasile, line 242: Default should be 1 for consistency with other usages of the flag.

Alex-Vasile (changes requested):

sharktank/conftest.py
dezhiAmd: But in this case, Default should be 2 to because the goal is to verify perplexity score is the same between with and without pipeline-parallelism

sharktank/conftest.py
```diff
@@ -234,6 +234,15 @@ def pytest_addoption(parser):
         default=1,
         help="Number of devices for tensor parallel sharding",
     )
+
+    parser.addoption(
+        "--pipeline-parallelism-size",
```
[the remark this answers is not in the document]
dezhiAmd: I agree that 3 layers. So you really can't do a PP8 test. And this PR is not enough. Allow me to improve that in follow-up PRs

PR #2433, opened 2025-10-06
Add initial BOO tuner script

rkayaith:


sharktuner/model_tuner/model_tuner.py
```diff
@@ -41,7 +42,7 @@ def read_flags_file(flags_file: str) -> list[str]:
         return file.read().splitlines()
 
 
-def arg_parse() -> argparse.Namespace:
+def arg_parse(parser_args: Sequence[str] | None = None) -> argparse.Namespace:
```
kuhar, line 45: I'd rather make it a sequence and disallow None -- we can modify users to pass in an empty list instead

sharktuner/model_tuner/model_tuner.py
```diff
@@ -93,12 +94,12 @@ def arg_parse() -> argparse.Namespace:
...
     )
     # Remaining arguments come from libtuner
-    args = libtuner.parse_arguments(parser)
+    args = libtuner.parse_arguments(parser, parser_args)
     return args
 
 
-def main() -> None:
-    args = arg_parse()
+def main(parser_args: Sequence[str] | None = None) -> None:
```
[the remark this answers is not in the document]
kuhar: also here -- I'd disallow `None`

sharktuner/sharktuner/libtuner.py
```diff
@@ -248,6 +248,7 @@ class CodegenPipelines(str, Enum):
 
 def parse_arguments(
     initial_parser: Optional[argparse.ArgumentParser] = None,
+    parser_args: Sequence[str] | None = None,
```
kuhar, line 251: also here

sharktuner/boo_tuner.py
```diff
@@ -0,0 +1,124 @@
+"""
+Generate BOO dispatches for the specified kernels and tune them. Kernel
+configurations follow the MIOpen driver format.
+
+Example usage:
+python boo_tuner.py \
```
kuhar, line 6:
```suggestion
python -m boo_tuner \
```

sharktuner/boo_tuner.py
```diff
@@ -0,0 +1,124 @@
...
+import traceback
+
+from iree.turbine.kernel.boo import runtime as boo_runtime
+from iree.turbine.kernel.boo.driver.launch import get_launchable
+from iree.turbine.kernel.boo.op_exports.conv import ConvParser as mio
+
+from model_tuner.model_tuner import main as tuner_main
+
+
+def main():
```
kuhar, line 27:
```suggestion
def main() -> None:
```

sharktuner/boo_tuner.py
```diff
@@ -0,0 +1,124 @@
...
+    parser = argparse.ArgumentParser()
```
kuhar, line 28: You can add the top-level docstring to the parser help: https://stackoverflow.com/questions/18106327/display-pydocs-description-as-part-of-argparse-help

sharktuner/boo_tuner.py
```diff
@@ -0,0 +1,124 @@
...
+    parser.add_argument("--commands-file", type=str, help="read commands from file")
```
kuhar, line 29: Can you specify what commands you have in mind? 

Also, we use full sentences for `help` across sharktank
```suggestion
    parser.add_argument("--commands-file", type=str, help="Read commands from file.")
```

sharktuner/boo_tuner.py
```diff
@@ -0,0 +1,124 @@
...
+    parser.add_argument("--output-td-spec", type=Path, default="tuning-spec.mlir")
+    parser.add_argument("--starter-td-spec", type=Path)
+    parser.add_argument("--num-candidates", type=int, default=100)
+    parser.add_argument("--devices", type=str, default="hip://0")
+    parser.add_argument(
+        "--tmp-dir", type=str, default="", help="directory to save temporary files"
```
kuhar, line 35:
```suggestion
        "--tmp-dir", type=str, default="", help="Directory to save temporary files."
```

sharktuner/boo_tuner.py
```diff
@@ -0,0 +1,124 @@
...
+    )
+    args, extra_cli_args = parser.parse_known_args()
+
+    if args.commands_file:
+        with open(args.commands_file) as f:
+            mio_file_args = [
+                shlex.split(s) for s in f.readlines() if not s.startswith("#")
+            ]
+    else:
+        mio_file_args = [[]]  # use CLI arguments
```
kuhar, line 45: What does `mio` stand for?

```suggestion
    mio_file_args = [[]]  # Use CLI arguments.
    if args.commands_file:
        with open(args.commands_file) as f:
            mio_file_args = [
                shlex.split(s) for s in f.readlines() if not s.startswith("#")
            ]
```

sharktuner/boo_tuner.py
```diff
@@ -0,0 +1,124 @@
...
+
+        # Re-compile to dump dispatches.
+        bench_dir = tmp_dir / "bench"
+        compile_args = shlex.split(compile_command) + [
+            "--iree-codegen-tuning-spec-path=",  # disable tuning spec flag as it prevents 'iree-config-add-tuner-attributes' from working.
+            "--iree-config-add-tuner-attributes",
+            "--iree-hal-dump-executable-benchmarks-to",
+            str(bench_dir),
+            "-o",
+            "/dev/null",
```
kuhar, line 84: Does this work on windows? We don't currently have any CI runners to check, but eventually we would like to support it too.

sharktuner/boo_tuner.py
```diff
@@ -0,0 +1,124 @@
...
+            for f in os.listdir(boo_cache_dir / op_cache_dir)
+            if f.startswith("compile_command")
+        ]
+        with open(boo_cache_dir / op_cache_dir / compile_command_path) as f:
+            compile_command = f.read().strip()
+
+        # Re-compile to dump dispatches.
+        bench_dir = tmp_dir / "bench"
+        compile_args = shlex.split(compile_command) + [
+            "--iree-codegen-tuning-spec-path=",  # disable tuning spec flag as it prevents 'iree-config-add-tuner-attributes' from working.
```
kuhar, line 79:
```suggestion
            "--iree-codegen-tuning-spec-path=",  # Disable tuning spec flag as it prevents 'iree-config-add-tuner-attributes' from working.
```

sharktuner/boo_tuner.py
```diff
@@ -0,0 +1,124 @@
...
+            bench_path = bench_dir / bench_file
+
+            # Run tuner.
+            tuner_args = [
+                "model-unused",
+                str(bench_path),
+                *(
+                    ("--starter-td-spec", str(starter_td_spec))
+                    if args.starter_td_spec is not None
+                    else ()
```
kuhar, line 101: Can you hoist this to a local variable?

sharktuner/boo_tuner.py
```diff
@@ -0,0 +1,124 @@
...
+                ),
+                *("--output-td-spec", str(args.output_td_spec)),
```
kuhar, line 103:
```suggestion
                f"--output-td-spec={args.output_td_spec}",
```

sharktuner/boo_tuner.py
```diff
@@ -0,0 +1,124 @@
...
+import tempfile
+import subprocess
+import shutil
+import traceback
+
+from iree.turbine.kernel.boo import runtime as boo_runtime
+from iree.turbine.kernel.boo.driver.launch import get_launchable
+from iree.turbine.kernel.boo.op_exports.conv import ConvParser as mio
+
+from model_tuner.model_tuner import main as tuner_main
```
kuhar, line 24: This should use the dispatch tuner

sharktuner/boo_tuner.py
```diff
@@ -0,0 +1,124 @@
...
+            try:
+                tuner_main(tuner_args)
+                starter_td_spec = args.output_td_spec
+            except Exception as err:
+                # Print error and continue.
+                traceback.print_exception(err)
+                # Retain temporary directory for debugging.
+                should_cleanup_tmp_dir = False
+        if should_cleanup_tmp_dir:
+            shutil.rmtree(tmp_dir)
```
kuhar, line 120: Can we use a with statement to take care of the cleanup instead of relying on the control flow reaching this?

rkayaith:
> Can you put this in its own directory and invoke it as a python module, similar to the model_tuner and dispatch_tuner?

Sure I can move it to its own directory. But rather than invoking it as a module, I think it'd make more sense to add it as an entrypoint to the `sharktuner` package, that way it can be invoked from any directory (I think we should do this for the other tuners as well). I'll expose it as `shark-boo-tuner` unless anyone has other suggestions.

sharktuner/boo_tuner/boo_tuner.py
```diff
@@ -0,0 +1,304 @@
...
+
+    # Extract MIOpen operation arguments (parser now knows all BOO + libtuner arguments).
+    _, miopen_op_args = parser.parse_known_args()
+
+    return args, miopen_op_args
+
+
+def main() -> None:
+    args, miopen_op_args = arg_parse()
+    path_config = libtuner.PathConfig()
```
kuhar: also here

sharktuner/boo_tuner/boo_tuner.py
```diff
@@ -0,0 +1,304 @@
...
+    root_logger = libtuner.setup_logging(args, path_config)
+    print(path_config.run_log, end="\n\n")
+
+    if not args.dry_run:
+        print("Validating devices")
+        libtuner.validate_devices(args.devices)
+        print("Validation successful!\n")
+
+    # These imports are slow due to a pytorch dependency. Keeping them local.
+    # helps get fast '--help' output.
```
kuhar:
```suggestion
    # This helps make '--help' fast.
```

sharktuner/boo_tuner/boo_tuner.py
```diff
@@ -0,0 +1,304 @@
...
+    if commands_file:
+        splitter: Callable[[str], list[str]] = lambda s: (
+            s.strip().split("\t") if commands_file.endswith(".tsv") else shlex.split(s)
+        )
+        with open(commands_file) as f:
+            mio_args = [
+                splitter(s) + miopen_op_args
+                for s in f.readlines()
+                if s.strip() and not s.startswith("#")
+            ]
```
kuhar: Should we hoist this to a helper function and add tests?

sharktuner/boo_tuner/boo_tuner.py
```diff
@@ -0,0 +1,304 @@
...
+        print(f"source_mlir_path: {source_mlir_path}")
+
+        # Find the compile command file.
+        [compile_command_file] = [
+            f for f in os.listdir(op_cache_path) if f.startswith("compile_command")
+        ]
+        with open(op_cache_path / compile_command_file) as f:
+            compile_command = f.read().strip()
+
+        # Use all turbine compile flags except the output file (-o *.vmfb).
```
kuhar:
```suggestion
        # Use all turbine compile flags except the output file (`-o *.vmfb`).
```

sharktuner/boo_tuner/boo_tuner.py
```diff
@@ -0,0 +1,304 @@
...
+        benchmarks_dir = tmp_dir / "benchmarks"
+        compile_args.extend(
+            [
+                "--iree-config-add-tuner-attributes",
+                "--iree-hal-dump-executable-benchmarks-to",
+                str(benchmarks_dir),
+                "-o",
+                os.devnull,
+            ]
+        )
```
kuhar: We should probably also test this

sharktuner/boo_tuner/example_configs.txt
```diff
@@ -0,0 +1 @@
+convbfp16 -n 3 -c 2016 -H 1 -W 1 -k 224 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC -m conv -g 1 -F 1 -t 1
```
kuhar: Maybe also add a sample gemm?

sharktuner/boo_tuner/README.md
```diff
@@ -0,0 +1,76 @@
+# Getting Started with BOO Tuner
+
+Example of tuning BOO (Bag of Ops) kernels from MIOpen driver commands.
```
kuhar:
```suggestion
Example of tuning BOO (Bag of Ops) kernels with MIOpen driver commands.
```

sharktuner/boo_tuner/README.md
```diff
@@ -0,0 +1,76 @@
...
+source ../iree-build/.env && export PYTHONPATH
+```
+
+Install PyTorch for ROCm ([IREE Turbine ROCm requirements](https://github.com/iree-org/iree-turbine/blob/main/pytorch-rocm-requirements.txt)):
+
+```shell
+pip install --index-url https://download.pytorch.org/whl/rocm6.3 'torch>=2.5,<2.8'
+```
+
+Install IREE Turbine (see [IREE Turbine README](https://github.com/iree-org/iree-turbine/blob/main/README.md)) - there are two ways:
```
kuhar:
```suggestion
Install IREE Turbine (see the [IREE Turbine README](https://github.com/iree-org/iree-turbine/blob/main/README.md)) - there are two ways:
```

sharktuner/boo_tuner/boo_tuner.py
```diff
@@ -0,0 +1,304 @@
...
+    )
+    boo_args.add_argument(
+        "--boo-dispatch-benchmark-timeout-mins",
+        type=float,
+        default=None,
+        help="Timeout in minutes for dispatch benchmarking.",
+    )
+
+    # Insert placeholder input_file for libtuner (BOO generates files internally).
+    sys.argv = [sys.argv[0], "boo.mlir"] + sys.argv[1:]
```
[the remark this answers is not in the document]
bangtianliu: Python slicing is safe here.  `sys.argv[1:]` returns `[]` if no arguments are provided, resulting in `['boo_tuner.py', 'boo.mlir']`.

sharktuner/boo_tuner/boo_tuner.py
```diff
@@ -0,0 +1,333 @@
...
+    logging.getLogger("turbine").setLevel(logging.WARNING)
+
+    assert not (
+        args.commands_file and miopen_op_args
+    ), "Cannot specify both --commands-file and MIOpen operation arguments"
+
+    mio_args = load_commands_from_file_or_args(args.commands_file, miopen_op_args)
+
+    starter_td_spec: Path | None = args.starter_td_spec
+    for idx, cli_args in enumerate(mio_args):
```
kuhar: Can we turn the body of this loop into a helper function? I'm concerned we are putting so much logic directly in `main`.

sharktuner/tests/boo_tuner_test.py
```diff
@@ -0,0 +1,131 @@
...
+
+from boo_tuner.boo_tuner import (
+    load_commands_from_file_or_args,
+    insert_placeholder_input_file,
+    build_compile_args,
+)
+
+
+@pytest.fixture
+def tmp_file() -> Generator[Callable[[str, str], Path], None, None]:
```
kuhar: Can you add a comment explaining what is the meaning of these three arguments: `(str, str) -> Path`? I think one of the inputs is a filename, the other is the contents, and the path is the output of the temp file with those contents?

sharktuner/tests/boo_tuner_test.py
```diff
@@ -0,0 +1,131 @@
...
+        f.write(content)
+        f.close()
+        temp_path = Path(f.name)
+        temp_files.append(temp_path)
+        return temp_path
+
+    yield _create
+
+    for temp_file in temp_files:
+        temp_file.unlink(missing_ok=True)
```
kuhar: What's the difference between this and the default tmp file fixtures from https://docs.pytest.org/en/stable/how-to/tmp_path.html#the-tmp-path-factory-fixture ?

I think the main question is if we want to delete temporary files or not -- I'd think that keeping them around is useful in case the tests fail and we want to inspect them later.

The default retention specified by pytest makes sense for me:
> By default, the last 3 temporary directories are kept, but this behavior can be configured with [tmp_path_retention_count](https://docs.pytest.org/en/stable/reference/reference.html#confval-tmp_path_retention_count) and [tmp_path_retention_policy](https://docs.pytest.org/en/stable/reference/reference.html#confval-tmp_path_retention_policy).

it allows us to inspect temp files, but eventually deletes them.

kuhar: Actually, I think I see what you are doing -- you want the fixture to give you a callable that creates temporary files but handle the cleanup for you. I think you can achieve something equivalent through a parameterized fixture that uses TmpPathFactory with the contents as a parameter: https://docs.pytest.org/en/latest/example/parametrize.html

sharktuner/boo_tuner/boo_tuner.py
```diff
@@ -0,0 +1,371 @@
...
+) -> Path | None:
+    """Process a single BOO command through compilation and tuning."""
+    sig = BooOpRegistry.parse_command(cli_args, ignore_unhandled_args=True)
+    if sig is None:
+        raise ValueError(f"Boo op registry failed to parse '{shlex.join(cli_args)}'.")
+
+    # Setup temporary directory.
+    if args.tmp_dir:
+        tmp_dir = Path(args.tmp_dir)
+        shutil.rmtree(tmp_dir, ignore_errors=True)
```
RattataKing: Is this safe? In the real use cases do we need to consider path like `--tmp-dir=/home/user`?
I think it's also good to add some logging/print message when clearing the dir

RattataKing (commented):
Consider create a `class BooPathConfig`: https://github.com/nod-ai/shark-ai/blob/main/sharktuner/sharktuner/libtuner.py#L89-L120

bangtianliu:
> Consider create a `class BooPathConfig`: https://github.com/nod-ai/shark-ai/blob/main/sharktuner/sharktuner/libtuner.py#L89-L120

Good suggestion! Although current one using tuning_xx_xx is fine, maybe I can further improve along this direction.

sharktuner/boo_tuner/boo_tuner.py
```diff
@@ -0,0 +1,371 @@
...
+        candidates = libtuner.generate_candidate_specs(args, path_config, boo_tuner)
+        print(f"Stored candidate tuning specs in {path_config.specs_dir}\n")
+
+        print("Compiling dispatch candidates...")
+        boo_tuner.compile_flags = ["--compile-from=executable-sources"]
+        compiled_candidates = libtuner.compile(args, path_config, candidates, boo_tuner)
+
+        message = "Benchmarking compiled dispatch candidates..."
+        print(message)
+        logging.info(message)
```
RattataKing: Consider setup a `logging.StreamHandler()` if needed, see: https://docs.python.org/3/library/logging.handlers.html

sharktuner/boo_tuner/boo_tuner.py
bangtianliu: FYI: The --tmp-dir option specifies where to store temporary MLIR artifacts from IREE Turbine BOO compilation.
By default (without --tmp-dir), a temporary directory is created under ./boo_tuner/ with a random suffix (`~/shark-ai/sharktuner/boo_tuner/boo-tuner-xxx`), so it will safe. 
When user specify --tmp-dir, users are responsible for choosing a safe location.

sharktuner/boo_tuner/boo_tuner.py
```diff
@@ -0,0 +1,410 @@
...
+        # BooTuner has only one phase, so prune candidates if all are slower than baseline.
+        return True
+
+
+def insert_placeholder_input_file(argv: list[str]) -> list[str]:
+    """Insert a placeholder input file for libtuner compatibility."""
+    return [argv[0], "boo.mlir"] + argv[1:]
+
+
+def arg_parse() -> tuple[argparse.Namespace, list[str]]:
```
kuhar:
```suggestion
def parse_args() -> tuple[argparse.Namespace, list[str]]:
```

We start function names with action verbs

sharktuner/boo_tuner/boo_tuner.py
```diff
@@ -0,0 +1,410 @@
...
+    boo_runtime,
+    get_launchable,
+    BooOpRegistry,
+) -> Path | None:
+    """Process a single BOO command through compilation and tuning."""
+    sig = BooOpRegistry.parse_command(cli_args, ignore_unhandled_args=True)
+    if sig is None:
+        raise ValueError(f"Boo op registry failed to parse '{shlex.join(cli_args)}'.")
+
+    # Setup temporary directory.
```
kuhar:
```suggestion
    # Set up temporary directory.
```

setup is a noun, set up is a phrasal verb

sharktuner/boo_tuner/boo_tuner.py
```diff
@@ -0,0 +1,410 @@
...
+            if benchmark_path_config.run_log is not None:
+                print(f"\nCheck the detailed execution logs in:")
+                print(benchmark_path_config.run_log.resolve())
+            print(f"Check the summary in:")
+            print(summary_log_file.resolve())
+
+        except Exception as err:
+            traceback.print_exception(err)
+        finally:
+            summary_handler.close()
```
kuhar: Can we use a `with` statement to avoid this manual close?

sharktuner/boo_tuner/README.md
```diff
@@ -0,0 +1,79 @@
+# Getting Started with BOO Tuner
+
+Example of tuning BOO (Bag of Ops) kernels with MIOpen driver commands.
+
+**MIOpen Command Format**: This tuner accepts convolution parameters in MIOpen driver format (e.g., `convbfp16 -n 128 -c 128 -H 24 -W 48 -k 384 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 1 -t 1 --in_layout NHWC --out_layout NHWC --fil_layout NHWC`). For detailed explanations of MIOpen command parameters, usage examples, and environment variables (like `BOO_TUNING_SPEC_PATH` and `TURBINE_LOG_LEVEL`), see the [IREE Turbine BOO README](https://github.com/iree-org/iree-turbine/blob/main/iree/turbine/kernel/boo/README.md).
```
kuhar: Can you reflow this and try to fit in the column limit? There are markdown extensions that do that automatically.

sharktuner/tests/boo_tuner_test.py
```diff
@@ -0,0 +1,147 @@
...
+
+@pytest.fixture
+def tmp_file(tmp_path: Path) -> Callable[[str, str], Path]:
+    """Factory fixture for creating temporary files.
+
+    Returns a callable that takes (content: str, suffix: str) and returns a Path
+    to a temporary file with that content and suffix. Cleanup is handled by pytest's
+    tmp_path fixture.
+    """
+    counter = [0]
```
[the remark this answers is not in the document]
bangtianliu: The counter is a one-element list to work around Python's scoping rules. 

If use variable instead of list, the code should be like this: 

```
    counter = 0

    def _create(content: str, suffix: str = ".txt") -> Path:
        nonlocal counter
        counter += 1
        temp_file = tmp_path / f"test_file_{counter}{suffix}"
        temp_file.write_text(content)
        return temp_file
```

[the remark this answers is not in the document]
kuhar: You mean python's mutability rules? If this is the case, you add a comment explaining why we need this hack?

kuhar (approved):

PR #2439, opened 2025-10-07
[LLM Server] Trie Cache w/ Offset Prefill

stbaione:
## Description

- Enable `trie cache` w/ `offset prefill`
- Small cleanup
- Add test case to `accuracy_test`
- Add test case to `smoke tests`

Follow up will enable it w/ chunked prefill as well

### DIFF
--- a/.github/workflows/cleanup_releases.yml
+++ b/.github/workflows/cleanup_releases.yml
@@ -21,7 +21,7 @@ on:
       dry_run:
         description: 'Perform a dry run (show what would be deleted without actually deleting)'
         type: boolean
-        default: false
+        default: true
         required: false
 
 permissions:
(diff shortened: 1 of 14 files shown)


shortfin/python/shortfin_apps/llm/components/decoder/decoder.py
```diff
@@ -459,21 +460,38 @@ def create_decode_reqs(self, prefill_req: LlmInferenceExecRequest):
...
+
+        return prefill_req, cached_allocation
+
     def create_prefill_req(self, input_ids):
         prefill_req = LlmInferenceExecRequest(
             phase=InferencePhase.PREFILL, input_token_ids=input_ids, rid=self._rid
         )
 
         cached_allocation = self._page_cache.lookup(input_ids)
+        if self._prefill_config.has_prefill_position:
```
rsuderman, line 484: Just lookup for fewer tokens rather than doing a larger lookup and rewinding.

rsuderman (changes requested):

stbaione:
Closed, because included in #2443

PR #2443, opened 2025-10-07
[LLM Server] Trie Chunked Prefill

stbaione:
Stacked on #2439 
- Enabled `chunked_prefill` for `TrieAttentionCache`
- Fix the calculation for `_get_block_size` in the chunked scenario
- Adds `accuracy_test` case
- Adds `smoke_test` case

### DIFF
--- a/shortfin/python/shortfin_apps/llm/components/decoder/decoder.py
+++ b/shortfin/python/shortfin_apps/llm/components/decoder/decoder.py
@@ -14,23 +14,22 @@ from typing import Callable, Dict, List, Optional, Tuple, Union
 
 from _shortfin import lib as _sfl
 
-from ..batching.facade import BatchingFacade
-from ..decode_config import (
+from shortfin_apps.llm.components.batching.facade import BatchingFacade
+from shortfin_apps.llm.components.decode_config import (
     DecodeConfig,
     LogitsNormalization,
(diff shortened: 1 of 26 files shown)


shortfin/python/shortfin_apps/llm/components/decoder/decoder.py
```diff
@@ -4,32 +4,33 @@
...
-from shortfin_apps.llm.components.messages import (
-    LlmInferenceExecRequest,
-    InferencePhase,
-)
-from typing import Callable, List, Optional, Tuple, Union
-
-from _shortfin import lib as _sfl
-from shortfin_apps.llm.components.kvcache.attention_cache_abstract import CacheInfo
-from shortfin_apps.llm.components.kvcache.base_attention_cache import (
+from ..kvcache.attention_cache_abstract import CacheInfo
```
rsuderman, line 22: Try to use full path imports - its just better for when files are moves as you can find / replace more easily.

rsuderman (changes requested):

shortfin/python/shortfin_apps/llm/components/decoder/decoder.py
stbaione: Updated

PR #2447, opened 2025-10-07
[sharktank] add sliding window in export

oyazdanb:
Optimize export for sliding window models

### DIFF
--- a/sharktank/sharktank/examples/export_paged_llm_v1.py
+++ b/sharktank/sharktank/examples/export_paged_llm_v1.py
@@ -53,12 +53,9 @@ def export_llm_v1(
         # torch.export.Dim would make min at least 2
         block_dim_min = 2
 
-        if hp.sliding_window > 0:
-
-            # For sliding window models, we only need enough blocks for the window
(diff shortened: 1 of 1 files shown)


sharktank/sharktank/examples/export_paged_llm_v1.py
```diff
@@ -144,7 +152,15 @@ def _(model: ServicePagedLlmModelV1, tokens, seq_lens, seq_block_ids, cs):
...
         # torch.export.Dim would make min at least 2
         block_dim_min = 2
-        block_dim_max = ceildiv(hp.context_length, llama_config.block_seq_stride) - 1
+
+        if hp.sliding_window > 0:
+
+            # For sliding window models, we only need enough blocks for the window
+            effective_context = hp.sliding_window
+        else:
+            effective_context = hp.context_length
```
Alex-Vasile, line 161:
```suggestion
        # For sliding window models, we only need enough blocks for the window
        effective_context = hp.sliding_window if hp.sliding_window else hp.context_length
```

Same for above

Issue #2494, opened 2025-10-13
When evaluating LLMs we should stop using cross-entropy against the prompt


sogartar:
Right now when [evaluating](https://github.com/nod-ai/shark-ai/blob/0c254262c8c7d0d1643d418a65cc2e9822071a48/sharktank/sharktank/utils/llm_utils.py#L739) we are using the average `ln(softmax(logits))` across all tokens.
The problem with cross-entropy is that when there are unlikely tokens in the prompt sequence then they will dominate the result. They will have low probability predicted by the model. Then their cross-entropy will tend to `-inf` as they approach 0. Then small variations of the predicted probability will yield large changes in the cross-entropy.
E.g.
```
ln(1e-5) = -11.512925464970229
ln(1e-10) = -23.025850929940457
```

From a practical point of view when interpreting the results a probability of `1e-5` is almost the same as `1e-10`. Its like noise. But this difference will dominate the average cross-entropy giving you a large error. We would be largely measuring the noise.

PR #2499, opened 2025-10-14
[sharktank] Make LLM utils support chunked prefill for included Llama 8b eager quick evaluation tests

sogartar:
Make LlmRunner.prefill use the LLM scheduler mechanism and handle chunked prefill.

Adds pre-submit tests for Llama 3.1 8b f16 and fp8_e4m3fnuz.

For the fp8_e4m3fnuz evaluation tests the accepted tolerance is higher than the
non-chunked test due to different tiling for some ops and brittle error metric
sensitive to outliers. See https://github.com/nod-ai/shark-ai/issues/2494.
We need to refactor this metric in the future.

### DIFF
--- a/sharktuner/dispatch_tuner/dispatch_tuner.py
+++ b/sharktuner/dispatch_tuner/dispatch_tuner.py
@@ -1,4 +1,4 @@
-# Copyright 2025 Advanced Micro Devices, Inc
+# Copyright 2025 Advanced Micro Devices, Inc.
 #
 # Licensed under the Apache License v2.0 with LLVM Exceptions.
 # See https://llvm.org/LICENSE.txt for license information.
@@ -85,8 +85,8 @@ def arg_parse() -> argparse.Namespace:
         "--dispatch-benchmark-timeout-mins",
         type=float,
         default=None,
-        help="Time budget in minutes for disptach benchmark phase.",
-    ),
+        help="Time budget in minutes for dispatch benchmark phase.",
+    )
     # Remaining arguments come from libtuner.
     args = libtuner.parse_arguments(parser)
     return args
@@ -97,7 +97,6 @@ def main() -> None:
 
     path_config = libtuner.PathConfig()
     path_config.base_dir.mkdir(parents=True, exist_ok=True)
-    stop_after_phase: str = args.stop_after
 
     print("[WARNING] SHARK Tuner is still experimental")
     root_logger = libtuner.setup_logging(args, path_config)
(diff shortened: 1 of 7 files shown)


sharktank/sharktank/utils/llm_utils.py
```diff
@@ -494,28 +522,45 @@ def run_decode(
...
 
-        prefill_task = PrefillTask(
-            invocation_fn=self._instance.prefill,
-            llm_task_inputs=task_inputs,
-            batch_size=self._prefill_bs,
-            block_stride=self._block_stride,
-        )
-        logits, indices = prefill_task.run(*self._cache)
+        logits = [r[0][0] for r in request_results.values()]
+        indices = [r[0][1] for r in request_results.values()]
```
archana-ramalingam: Can be rewritten as `logits, indices = zip(*[(r[0][0], r[0][1]) for r in request_results.values()])`

sharktank/sharktank/utils/llm_utils.py
```diff
@@ -494,28 +522,45 @@ def run_decode(
             cache=self._cache,
         )
 
-    def prefill(self, requests: list[list[int]], page_ids: list[list[int]]):
+    def prefill(
```
archana-ramalingam: Can we move the chunk-handling logic into `PrefillTask` so the `LlmRunner` stays minimal. Prefill and decode must do similar minimal steps inside the runner class.

sogartar: In the current object model the `LlmRunner` seems the appropriate place to orchestrate this.
The `Scheduler` is concerned with batching and running tasks and the tasks are pieces of work description that should not implement interaction between each other.
The `LlmRunner` already orchestrates this. [Here](https://github.com/nod-ai/shark-ai/blob/7ac5ac6e4c76a7e4e97f59e5f12d6514e14e106f/sharktank/sharktank/utils/llm_utils.py#L373) and [here](https://github.com/nod-ai/shark-ai/blob/7ac5ac6e4c76a7e4e97f59e5f12d6514e14e106f/sharktank/sharktank/utils/llm_utils.py#L432). It breaks up the work into multiple pieces/tasks.

The thing that it did not do is handle the chunking configuration correctly in `prefill`. As it would act as if there is no chunking.

For decode there is no chunking. I think its functionality should not change.

archana-ramalingam: Since the `LlmTaskInput` and `PrefillTask` already have chunking related variables, would it fit better in `PrefillTask`, as we can branch in `_prepare_args` and `_process_results`?

archana-ramalingam: Yes, `Scheduler` shouldn't do chunking. Linking `LlmDecoder.greedy_decode` and `LlmRunner.prefill` could be a TODO for future PR.

archana-ramalingam (approved):

PR #2506, opened 2025-10-14
[sharktank] Change LLM scheduler to allow for more broad uses of chunked prefill

sogartar:
Make the selection function return `list[Any]` for a batch. This for example allows us to propagate the logits instead of doing a selection.

Make the run `Scheduler.run` function return a list of selections for each request id. This is needed to return the chunk results. It is not appropriate to do collection and stitching of results in the scheduler as its purpose is to run the tasks. It is not concerned with breaking up a request into tasks, so it should not be concerned of doing any sort of reduction.


sharktank/sharktank/utils/llm_scheduler.py
```diff
@@ -120,10 +123,7 @@ def __init__(
         batch_size: int,
         block_seq_stride: int,
         llm_task_class: type[LlmTask],
-        invocation_fn: Callable[
-            [List[numpy.ndarray | iree.runtime.DeviceArray | torch.Tensor]],
-            Tuple[numpy.ndarray, Optional[numpy.ndarray]],
-        ],
```
archana-ramalingam: All 3 `Scheduler`'s `init()` have a different Callable format than `run()`. Revert this and refactor this out like the other Callable.

sharktank/sharktank/utils/llm_scheduler.py
```diff
@@ -195,16 +193,17 @@ def run(
...
             )
 
             for i, task in enumerate(task_inputs):
-                if len(self._pending_tasks.get(task.request_id, [])) == 0 and (
-                    task not in self._ready_tasks
-                ):
-                    selections[task.request_id] = last[i]
-
+                if task.request_id not in selections:
+                    selections[task.request_id] = []
```
archana-ramalingam: Can we replace this with  `selections: defaultdict[str, list[Any]] = defaultdict(list)` at initialization?

sharktank/sharktank/utils/llm_scheduler.py
sogartar: I added `InvocationFn` as named type hint.

sharktank/sharktank/utils/llm_scheduler.py
```diff
@@ -44,11 +48,12 @@ def _get_next_batch(self) -> list[LlmTaskInput]:
...
-        selection_fn: Callable[
-            [numpy.ndarray, Optional[numpy.ndarray], List[int]], List[int]
-        ],
+        selection_fn: SelectionFn,
         cache: List[iree.runtime.DeviceArray | torch.Tensor],
-    ) -> Dict[str, int]:
+    ) -> Dict[str, list[Any]]:
+        """Returns a map from request ID to a list of all selections for
+        tasks/chunks associated with a request.
+        The tasks results for a request have the same ordered as scheduled."""
```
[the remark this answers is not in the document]
sogartar: Fixed it.

archana-ramalingam:
Failing tests are unrelated to this PR.

PR #2507, opened 2025-10-14
[sharktank] Materialize indices in PrefillTask to ease branching logic in downstream code

sogartar:
Handling the the None special case is error prone and adds branching complexity. We don't care that much about performance when doing evaluation.

Allow compute_cross_entropy to accept sequences of logits and indicies instead of single ndarray. This makes the function symmetric with respect to the requests arg. Which is a sequence.
This will help with chunked prefill and other methods of constructing the results where they are stitched from multiple requests. We don't need to burden the caller to make a single ndarray when we don't need it. But a single ndarray is still supported.

### DIFF
--- a/sharktank/sharktank/utils/llm_utils.py
+++ b/sharktank/sharktank/utils/llm_utils.py
@@ -27,6 +27,7 @@ import pathlib
 import time
 import torch
 
+from collections.abc import Sequence
 from copy import deepcopy
 from datasets import load_dataset
 from iree.runtime import ParameterIndex
@@ -712,10 +713,10 @@ class LlmPerplexityEval:
 
     def compute_cross_entropy(
         self,
-        logits: list[numpy.ndarray],
-        indices: list[numpy.ndarray],
+        logits: Sequence[numpy.ndarray],
+        indices: Sequence[numpy.ndarray],
         requests: list[list[int]],
-        min_context=0,
+        min_context: int = 0,
     ) -> list["LlmPerplexityEval.Result"]:
         results = []
         for i, req in enumerate(requests):


sharktank/sharktank/utils/llm_utils.py
```diff
@@ -702,19 +702,19 @@ def __init__(self, batch: LlmRunner, logits_normalization: str):
         self._logits_normalization = logits_normalization
 
     def compute_cross_entropy(
-        self, logits, indices, requests, min_context=0
+        self,
+        logits: list[numpy.ndarray],
+        indices: list[numpy.ndarray],
+        requests: list[list[int]],
+        min_context=0,
```
archana-ramalingam: Add type hint to last arg too.

sharktank/tests/utils/llm_prefill_tasks_test.py
```diff
@@ -201,7 +201,7 @@ def _invocation_fn(*args):
             logits,
             numpy.array([i for i in range(self._batch_size)], dtype=self._int_dtype),
         )
-        assert indices is None
+        assert numpy.array_equal(indices, [0, 1, 2, 3])
```
archana-ramalingam: Same comment as above.

sharktank/tests/utils/llm_prefill_tasks_test.py
```diff
@@ -335,4 +335,4 @@ def _invocation_fn(*args):
...
```
archana-ramalingam: Same comment as above.

sharktank/tests/utils/llm_prefill_tasks_test.py
```diff
@@ -123,7 +123,7 @@ def _invocation_fn(*args):
...
```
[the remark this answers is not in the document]
sogartar: Done.

sharktank/tests/utils/llm_prefill_tasks_test.py
sogartar: Done.

sharktank/tests/utils/llm_prefill_tasks_test.py
sogartar: Done.

PR #2522, opened 2025-10-15
[sharktank] Make trivial replication place results on the correct devices

sogartar:
Instead of placing results on devices [0, 1, 2, ...], place them based on the devices of the first replicated arg.
This allows the `trivially_replicable` decorator to be used outside the context of an overridable ops. Ops have their own mechanism of fixing/patching of wrongly placed results.

Also fix `is_deep_equal` for sharded tensors to also compare devices. Make `ReplicatedTensor` use `is_deep_equal` from its base. It does not make sense to have a custom one.

Fixes trivially replicable tests that actually had no asserts. Ops.

### DIFF
--- a/sharktank/tests/ops/sharded_test.py
+++ b/sharktank/tests/ops/sharded_test.py
@@ -2028,7 +2028,7 @@ class TriviallyReplicableTest(unittest.TestCase):
             ts=arg, shard_count=shard_count, devices=(1, 2)
         )
         replicated_result = fn(replicated_arg)
-        replicated_arg.is_deep_equal(replicated_result)
+        assert replicated_arg.is_deep_equal(replicated_result)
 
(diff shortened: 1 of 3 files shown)


sharktank/tests/ops/sharded_test.py
```diff
@@ -2028,7 +2028,7 @@ def fn(a: torch.Tensor) -> torch.Tensor:
             ts=arg, shard_count=shard_count, devices=(1, 2)
         )
         replicated_result = fn(replicated_arg)
-        replicated_arg.is_deep_equal(replicated_result)
+        assert replicated_arg.is_deep_equal(replicated_result)
```
archana-ramalingam: Is there a reason the asserts are added now?

PR #2525, opened 2025-10-15
[Fusilli] Fix to re-enable concurrent tests in CI

sjain-stanford:
Use a unique tempdir for each test to prevent interference when tests are run concurrently (using `ctest -j N`), as they all try to read/write to the same `.lit_test_times.txt` causing malformed text. I also started seeing this issue with local development today. This change makes it so every test gets it's own tempdir to run in. Re-enabling parallel tests in CI (was disabled in #2485).

### DIFF
--- a/.github/workflows/ci-sharkfuser.yml
+++ b/.github/workflows/ci-sharkfuser.yml
@@ -100,7 +100,7 @@ jobs:
       working-directory: ${{ env.FUSILLI_DIR }}
       run: |
         build_tools/docker/exec_docker_ci.sh \
-        bash -c "ctest --test-dir build --output-on-failure --extra-verbose --timeout 120 -j 16 && \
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-sharkfuser.yml
```diff
@@ -100,7 +100,7 @@ jobs:
       working-directory: ${{ env.FUSILLI_DIR }}
       run: |
         build_tools/docker/exec_docker_ci.sh \
-        bash -c "ctest --test-dir build --output-on-failure --extra-verbose --timeout 120 && \
+        bash -c "ctest --test-dir build --output-on-failure --extra-verbose --timeout 120 -j 16 && \
```
AaronStGeorge: `nproc`?

AaronStGeorge (approved):
🥳

PR #2526, opened 2025-10-16
Add utilities for automatically converting a torch module to a IREE-backed module-type

KyleHerndon:


sharktank/sharktank/utils/iree.py
```diff
@@ -72,39 +84,213 @@ def oneshot_iree_run(
...
+    """
 
-    return with_iree_device_context(run, iree_devices)
+    def __call__(self, *args: Any, **kwargs: Any) -> Any:
+        """Execute the module's forward pass."""
+        ...
+
+    def forward(self, *args: Any, **kwargs: Any) -> Any:
+        """Execute the module's forward pass explicitly."""
+        ...
```
sogartar: We could have other functions, not just `forward`. Also there my be no `forward` at all.
The `@runtime_checkable` decorator would allow us to use `isinstance`, but if a module lacks a `forward` method, `isinstance` would return `False`.

sharktank/tests/utils/iree_test.py
```diff
@@ -115,3 +119,139 @@ def roundtrip(iree_devices: list[iree.runtime.HalDevice]):
...
+        )
+
+        result = iree_module.forward(example_input)
+        assert isinstance(result, torch.Tensor)
+        assert result.shape == (2, 32)
+        assert not torch.isnan(result).any()
+
+    def test_output_matches_torch(self):
+        """Test that IREE output matches torch output."""
+        torch.manual_seed(42)
```
sogartar: You could use the [deterministic_random_seed](https://github.com/nod-ai/shark-ai/blob/b65fdae35ede615932734d6c204771a33cfdc7ec/sharktank/conftest.py#L449) pytest fixture. After the test it will return the RNG state to what was before the test.

sharktank/tests/utils/iree_test.py
```diff
@@ -115,3 +119,139 @@ def roundtrip(iree_devices: list[iree.runtime.HalDevice]):
...
+
+        result = oneshot_iree_run(
+            model,
+            args=(example_input,),
+            device="local-sync",
+            compile_args=COMPILE_FLAGS,
+        )
+        assert isinstance(result, torch.Tensor)
+        assert result.shape == (2, 32)
+        assert not torch.isnan(result).any()
```
sogartar: This check depends on non-deterministic values.

sharktank/sharktank/utils/iree.py
```diff
@@ -72,39 +84,213 @@ def oneshot_iree_run(
...
+            parameters_path=parameters_path,
+            tensor_parallel_size=len(devices),
         )
-        results = getattr(torch_like_iree_module, function)(*args, **kwargs)
-        # Clone to avoid leaking IREE-backed torch tensors.
-        results = tuple(t.clone() for t in results)
-        return results
+        return TorchLikeIreeModule(vm_module, vm_context, devices)
+
+    return with_iree_device_context(load_fn, iree_devices)
```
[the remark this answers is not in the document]
KyleHerndon: > I think you could remove the usage of with_iree_device_context and just return the TorchLikeIreeModule.

This part, done.

sharktank/tests/utils/iree_test.py
KyleHerndon: Uses integers and deterministic_random_seed now, done.

sharktank/sharktank/utils/iree.py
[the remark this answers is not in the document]
KyleHerndon: I appreciate the explanation, but its unclear to me if the rest of this is actionable, so I'm currently leaving this thread open.

PR #2532, opened 2025-10-16
[Sharktank] Add ops.arange

Alex-Vasile:


sharktank/sharktank/ops/sharded_impls.py
```diff
@@ -52,6 +52,21 @@
...
+    *args,
+    devices: Sequence[int] | None = None,
+    **kwargs,
+):
+    if devices is None:
+        return NotImplemented
+
+    # Do not use `tranfer_to_logical_device` here.
+    # Adding the transfer op would prevent the arange result from being fused.
+    shards = [arange(*args, **kwargs) for _ in devices]
```
sogartar: Would we support at some point here calling a custom arange function that constructs on the required device explicitly. Something that exports to an MLIR op that has the device affinity set.

sharktank/sharktank/ops/signatures.py
```diff
@@ -150,6 +151,20 @@ def _all_reduce_trampoline(d: SignatureDispatcher, tensor: AnyTensor):
         d.fail(tensors)
 
 
+@overridable(dispatch_args=(), is_trivially_replicable=False)
+def arange(
+    *args,
+    devices: Sequence[int] | None = None,
```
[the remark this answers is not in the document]
sogartar: It seems using `devices` would not allow for constructing a non-replicated tensor on a given device.

sharktank/sharktank/ops/sharded_impls.py
Alex-Vasile: Being able to tag these 0 input ops and explicitly placing them is a good idea and would help not rely as strongly on IREE's affinity analysis. I don't know if that's possible at present, but if it is, it'll slot in easily here.

PR #2534, opened 2025-10-16
Modify sharktank KVCache and local Wave kernel copy for extend attention

aviator19941:
This PR modifies the sharktank KVCache by reordering the block_seq_stride to before num_attn_heads in order to allow flattening the first 4 dimensions of the sharktank KVCache to match the Wave KVCache layout. The Wave kernel also expects separate k_cache and v_cache, which sharktank does not currently support, so the local sharktank copy of the Wave extend attention kernel is updated to use a single KVCache block and separate k_indices and v_indices instead of separate k_cache and v_cache blocks and a single kv_indices. In the future, sharktank will need to split the KVCache into k_cache and v_cache. There is also an extend attention test for a single request with 2 chunks that uses the updated KVCache and updated kernel.


Groverkss (changes requested):
Putting a block right now, need to review more carefully, i see some things that need to be resolved before landing, but need to do a careful review.

sharktank/sharktank/layers/paged_attention.py
```diff
@@ -231,11 +312,24 @@ def unwrap_args(*ts):
...
+        else:
+            key = kv_cache_gather(*unwrap_args(page_table, page_ids, t_id, key_p_id))
+            value = kv_cache_gather(
+                *unwrap_args(page_table, page_ids, t_id, value_p_id)
+            )
 
-        key = key.transpose(2, 3).flatten(1, 2)
-        value = value.transpose(2, 3).flatten(1, 2)
+            key = key.transpose(2, 3).flatten(1, 2)
+            value = value.transpose(2, 3).flatten(1, 2)
```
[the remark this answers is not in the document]
aviator19941: Done

sharktank/tests/kernels/wave/extend_attention_test.py
```diff
@@ -209,13 +216,13 @@ def forward(
         assert_close(iree_results, ref_output, rtol=1e-3, atol=1e-3, check_dtype=False)
 
 
-@is_mi300x
+# @is_mi300x
```
archana-ramalingam: Is this meant to run on gfx942?

aviator19941: Yeah good catch, fixed it.

sharktank/sharktank/kernels/wave/extend_attention.py
```diff
@@ -65,8 +65,8 @@ def get_wave_extend_attention_asm(
         q_extend_shape,
         k_extend_shape,
         v_extend_shape,
-        k_cache_shape,
-        v_cache_shape,
+        kv_cache_1_shape,
```
rsuderman: Don't do kv_cache_#.  This is ambiguous. It should be clear what these buffer / shape represents (e.g. the k or v component). Its especially important when invoking for wave.

sharktank/sharktank/kernels/wave/templates/extend_attention_kernel.py
```diff
@@ -152,20 +152,22 @@ def get_extend_attention_kernel(
     k_layout = tkl.MemoryLayout(shape=set_dynamic_dim(k_shape))
     v_layout = tkl.MemoryLayout(shape=set_dynamic_dim(v_shape))
     o_layout = tkl.MemoryLayout(shape=set_dynamic_dim(o_shape))
-    k_cache_layout = tkl.MemoryLayout(shape=set_dynamic_dim(k_cache_shape))
-    v_cache_layout = tkl.MemoryLayout(shape=set_dynamic_dim(v_cache_shape))
+    kv_cache_1_layout = tkl.MemoryLayout(shape=set_dynamic_dim(kv_cache_1_shape))
```
rsuderman: Same with layout.

sharktank/tests/kernels/wave/extend_attention_test.py
```diff
@@ -261,9 +295,167 @@ def test_no_cache(
...
+    )
+    def test_extend_kv_cache_single_request_two_chunks(
+        self,
+        batch,
+        heads,
+        seq_len,
+        head_dim,
+        attn_dtype,
+        device,
+    ):
```
rsuderman: When you are writing this much boiler plate for a test its a sign that either helpers / cleanup should be included. The transposition, data setup, etc inherently means its too complex for clearly describe the test.

sharktank/tests/kernels/wave/extend_attention_test.py
```diff
@@ -261,9 +295,167 @@ def test_no_cache(
...
+        kv_cache_extend = build_cache(
+            transformer_block_count=transformer_block_count,
+            attn_head_count=heads,
+            attn_head_dim=head_dim,
+            block_seq_stride=block_seq_stride,
+            cache_dtype=attn_dtype,
+            extend_attention=True,
+        )
+
+        cache = PagedGQAttention(
```
rsuderman: This is inverted - the subcomponent will be invoking the kernel so they test should be using it externally.

rsuderman (changes requested):

sharktank/tests/kernels/wave/extend_attention_test.py
```diff
@@ -261,9 +295,167 @@ def test_no_cache(
...
+            end = (chunk_id + 1) * chunk_size
+
+            q = q_full[:, start:end, :, :]
+            k = k_full[:, start:end, :, :]
+            v = v_full[:, start:end, :, :]
+
+            write_page_ids = page_ids[
+                :, start // block_seq_stride : end // block_seq_stride
+            ]
+            cache_partitions = [k.cpu(), v.cpu()]
```
archana-ramalingam: Is `.cpu()` required?

sharktank/tests/kernels/wave/extend_attention_test.py
```diff
@@ -261,9 +295,167 @@ def test_no_cache(
...
+        k_sdpa = k_full.transpose(1, 2)
+        v_sdpa = v_full.transpose(1, 2)
+        input_mask = ops.input_mask(torch.tensor([seq_len]), seq_len)
+        a = ops.attention_mask(
+            input_mask,
+            source_len=seq_len,
+            target_len=seq_len,
+            attention_dtype=attn_dtype,
+        ).to(device)
+        sdpa_ref = ops.scaled_dot_product_attention(q=q_sdpa, k=k_sdpa, v=v_sdpa, a=a)
```
archana-ramalingam: The `sdpa_ref` and `extend_attention_output` calculation can be extracted to be separate functions under `sharktank/tests/utils/`

PR #2543, opened 2025-10-17
[tuner] use python binding to build td specs for contraction

bangtianliu:
This PR uses python binding to build td specs for contraction op.
The python binding is exposed from IREE PR https://github.com/iree-org/iree/pull/22227 (MatchContractionOp binding).

Right now, the code still contains two paths for building TD specs:
- Contraction ops: Use new Python bindings via `build_contraction_td_spec()`
- Convolution/Attention ops: Still use old string template approach via `build_td_spec()`

After this PR, I will also send PRs for convolution and attention ops to migrate them to Python bindings.  Once all operations are supported, some code relevant to temporary solutions and string-based method will be fully cleaned up.

### DIFF
--- a/sharktuner/sharktuner/candidate_gen.py
+++ b/sharktuner/sharktuner/candidate_gen.py
@@ -157,7 +157,7 @@ class ConvolutionOpInterfaceTuner(
         config_list: list[common.TuningConfiguration],
     ) -> ir.Module:
         conv_op = self.get_root_op()
-        func_name = spec_builder.get_matcher_function_name(conv_op)
+        func_name = spec_builder.get_matcher_named_sequence_name(conv_op)
         return spec_builder.build_td_spec(
             conv_op.context, conv_op, config_list, func_name
         )
@@ -193,7 +193,7 @@ class AttentionOpInterfaceTuner(
         config_list: list[common.TuningConfiguration],
     ) -> ir.Module:
         attention_op = self.get_root_op()
-        func_name = spec_builder.get_matcher_function_name(attention_op)
+        func_name = spec_builder.get_matcher_named_sequence_name(attention_op)
         return spec_builder.build_td_spec(
             attention_op.context, attention_op, config_list, func_name
         )
@@ -239,9 +239,6 @@ def set_dispatch_tuner(
 
     dispatch_tuner: Optional[DispatchTuner] = None
     for tuner_class in dispatch_tuners:
-        # Check if the tuner class can handle this root op type (e.g., contraction,
-        # convolution, attention) before instantiation, since the constructor will
-        # extract op_info and will assert if the op is not supported.
         if tuner_class.supports_root_op(root_op):
             tuner = tuner_class(root_op, tuner_ctx)
             dispatch_tuner = tuner
--- a/sharktuner/sharktuner/spec_builder.py
+++ b/sharktuner/sharktuner/spec_builder.py
@@ -20,9 +20,9 @@ from .dispatch_parser import *
 ROOT_OP_ATTR_NAME = "root_op"
 
 
-def get_matcher_function_name(root_op: ir.Operation) -> str:
+def get_matcher_named_sequence_name(root_op: ir.Operation) -> str:
     """
-    Adds the "match_" prefix to the parent function name.
+    Returns the symbol name for a transform dialect named sequence matchger.
     """
     return f"match_{get_parent_function_name(root_op)}"
 
@@ -81,9 +81,9 @@ def build_td_spec(
                 f"Root op has repeated operand. This can cause failure to match in the resulting TD spec at compile time."
             )
             continue
-        ssa_name = operand.get_name()
+        operand_name = operand.get_name()
         operand_type = operand.type
-        bbargs.append(f"{ssa_name}: {operand_type}")
+        bbargs.append(f"{operand_name }: {operand_type}")
         captured_values.add(operand)
     bbargs_str = ", ".join(bbargs)
     matcher_block = f"""%ins, %outs = transform.iree.match.cast_compatible_dag_from_root %cont {{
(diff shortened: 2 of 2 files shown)


sharktuner/sharktuner/candidate_gen.py
```diff
@@ -69,30 +76,51 @@ def find_handler(self, op_name: str) -> DispatchTuner:
...
+    def __init__(self, root_op: ir.Operation, tuner_ctx: common.TunerContext):
+        super().__init__(root_op, tuner_ctx)
+
+    @classmethod
+    def supports_root_op(cls, root_op: ir.Operation) -> bool:
+        if not linalg.isa_convolution_op(root_op):
+            return False
+        convolution_dims = linalg.infer_convolution_dimensions(root_op)
+        if not convolution_dims:
+            return False
```
kuhar: When can this fail?

sharktuner/sharktuner/dispatch_parser.py
```diff
@@ -28,40 +30,115 @@ def parse_mlir(mlir_text: str, ctx: common.TunerContext) -> ir.Module:
     return mlir_module
 
 
+@dataclass
+class OpInfo:
+    # Name of parent func.FuncOp with "match_" prefix.
+    parent_function_name: str
```
kuhar: Can we remove this member and query the nearest parent of type funcopinterface instead?

sharktuner/sharktuner/candidate_gen.py
bangtianliu: `linalg.isa_convolution_op` goes to `mlir::linalg::detail::isConvolutionInterfaceImpl` in low level, 
And `linalg.infer_convolution_dimensions` goes to `inferConvolutionDimsImpl` in low level.

isConvolutionInterfaceImpl does contain below code block:
https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/Linalg/IR/LinalgInterfaces.cpp#L1006-L1012
```C++
if (dimensions) {
    FailureOr<ConvolutionDimensions> res = inferConvolutionDimsImpl(
        linalgOp, inputExprWalker, allowEmptyConvolvedDims);
    assert(succeeded(res) && "unexpected failure to infer convolution dims");
    *dimensions = *res;
  }
```
However, When `linalg.isa_convolution_op` is called, it passes dimensions=nullptr, so this block is not executed since dimension is nullptr. 

So from my understanding,` linalg.infer_convolution_dimensions` could fail even after `linalg.isa_convolution_op` succeeds, because it's the first time the actual dimension inference logic runs.

sharktuner/sharktuner/spec_builder.py
```diff
@@ -184,3 +134,220 @@ def build_td_spec(
...
+        Creates a constant parameter for each configuration.
+        Parameters can contain #iree_codegen.compilation_info or other configuration attributes.
+        """
+        config_params = []
+        for config in config_list:
+            config_param = transform.ParamConstantOp(
+                transform.AnyParamType.get(),
+                config.configuration,
+            ).result
+            config_params.append(config_param)
```
kuhar:
```suggestion
            config_params.append(ransform.ParamConstantOp(
                transform.AnyParamType.get(),
                config.configuration
            ).result)
```

sharktuner/sharktuner/dispatch_parser.py
```diff
@@ -28,40 +30,139 @@ def parse_mlir(mlir_text: str, ctx: common.TunerContext) -> ir.Module:
...
+
+    @property
+    def parent_function_name(self) -> str:
+        """
+        Extracts the parent function name and prefixes it with "match_".
+        """
+        func_op = self.root_op.parent.opview
         assert isinstance(
             func_op, func.FuncOp
         ), f"Expected func.func, got {func_op.name}"
```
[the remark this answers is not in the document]
kuhar: ok fine to leave func.func for the time being

sharktuner/sharktuner/spec_builder.py
```diff
@@ -157,11 +164,12 @@ def create_config_params(
         """
         config_params = []
         for config in config_list:
-            config_param = transform.ParamConstantOp(
-                transform.AnyParamType.get(),
-                config.configuration,
-            ).result
-            config_params.append(config_param)
+            config_params.append(
```
kuhar: actually you can turn this whole loop into list comprehension

sharktuner/sharktuner/dispatch_parser.py
```diff
@@ -8,14 +8,24 @@
...
+from iree.compiler.dialects import func, iree_codegen, linalg  # type: ignore
 
 from . import common
 
 
+def get_parent_function_name(root_op: ir.Operation) -> str:
+    """
+    Returns the parent function's symbol name from a root operation.
+    """
+    func_op = root_op.parent.opview
```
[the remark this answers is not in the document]
bangtianliu: This part of the code wasn’t originally added in this PR and I just moved it here. But thanks for pointing it out. I can update it to make the check more robust.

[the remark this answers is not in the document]
kuhar: We already have an assertion, so I'd leave it as `# FIXME: ...` if we can't easily query this from python

bangtianliu:
> tests are failing

Yes, just fixed it.

sharktuner/sharktuner/spec_builder.py
```diff
@@ -184,3 +141,219 @@ def build_td_spec(
...
+        Creates a transform.named_sequence that matches the operation and returns
+        the matched operation handle along with configuration parameters.
+        """
+        input_types = [transform.AnyOpType.get()]
+        output_types = [transform.AnyOpType.get()] + [
+            transform.AnyParamType.get()
+        ] * len(config_list)
+
+        named_seq = transform.NamedSequenceOp(
+            get_matcher_function_name(self.op_info.root_op),
```
kuhar: I would only add the `match_` prefix here, not in the helper

PR #2550, opened 2025-10-17
[Sharktank] Use .to() instead of .bool(), .float(), or .int(). Remove .bool() from InferenceTensors

Alex-Vasile:
Adding `.int()` and `.float()` to InferenceTensor would mean that we clobber python's built in int and float. It breaks out type hints and would cause strange issues if we actually try calling the functions expecting python's implementation. 

Also removing `.bool()` to make conversions more explicit.

Better to replace them all with the equivalent `.to(dtype)`.


sogartar (approved):
You should say in the description that when delegating to `sharktank.ops` then their declaration would clobber the builtins. The `InferenceTensor` class is fine to have these.

Actually torch also does not have module level int/float. There is no `torch.int`, but there is `torch.Tensor.int`. We could do the same by calling `ops.to(...)` in the implementations.

PR #2560, opened 2025-10-20
[Fusilli] Pick compiler fixes and change dispatch count checks for conv3d

sjain-stanford:
Compiler fix: https://github.com/iree-org/iree/pull/22320
Issue: https://github.com/iree-org/iree/issues/22312


IanWood1:
The reason CPU dispatch count improved is because we are generalizing convs. This is most likely causing really slow perf on CPU. It doesn't seem to be too problematic but this can be disabled with  `--iree-global-opt-experimental-disable-conv-generalization` and there is an open IREE issue https://github.com/iree-org/iree/issues/21955 to fix the bad perf

PR #2574, opened 2025-10-21
[sharktank] adding weight generation in util for gpt-oss toy model

oyazdanb:
adding util for weight generation for gpt-oss toy model

### DIFF
--- a/sharktank/sharktank/utils/random.py
+++ b/sharktank/sharktank/utils/random.py
@@ -26,7 +26,7 @@ def make_random_mask(shape: tuple[int], dtype: Optional[torch.dtype] = None):
 
 
 def make_wide_range_weights(
-    shape: list[int], dtype: Optional[torch.dtype] = None
+    shape: list[int], dtype: Optional[torch.dtype] = None, seed: int = 1234
 ) -> torch.Tensor:
     """Generate weights with proper variance scaling to prevent numerical explosions.
 
@@ -35,7 +35,6 @@ def make_wide_range_weights(
     maintaining numerical stability.
 
     """
-    seed = 12345
     generator = torch.Generator()
     generator.manual_seed(seed)
     fan_in = shape[-1]


sharktank/sharktank/utils/random.py
```diff
@@ -24,6 +25,43 @@ def make_random_mask(shape: tuple[int], dtype: Optional[torch.dtype] = None):
...
+
+    Uses Xavier-like initialization: scale by 1/sqrt(fan_in) to keep output variance
+    stable regardless of layer dimensions. The 0.8 factor provides diversity while
+    maintaining numerical stability.
+
+    """
+    seed = 12345
+    generator = torch.Generator()
+    generator.manual_seed(seed)
+    fan_in = shape[-1] if len(shape) > 1 else shape[0]
```
archana-ramalingam: Isn't shape[0] == shape[-1]? So fan_in will always be shape[-1] and we don't need if-else.

sharktank/sharktank/utils/random.py
```diff
@@ -24,6 +25,43 @@ def make_random_mask(shape: tuple[int], dtype: Optional[torch.dtype] = None):
...
+    std = 0.8 / math.sqrt(fan_in)
+    weights = torch.randn(shape, dtype=dtype, generator=generator) * std
+    return weights
+
+
+def make_simple_calculable_weight_torch(
```
archana-ramalingam: Is there a reason `torch.Tensor.repeat` isn't an option here?

sharktank/sharktank/utils/random.py
```diff
@@ -24,6 +25,43 @@ def make_random_mask(shape: tuple[int], dtype: Optional[torch.dtype] = None):
...
+    shape: list[int], dtype: torch.dtype = torch.bfloat16
```
archana-ramalingam: If we plan to use this for other models as well, is it disruptive to set default to `float16`?

sharktank/sharktank/utils/random.py
oyazdanb: as @dan-garvey  mentioned `torch.tensor.repeat` only covers exact multiples of the pattern, so we'd need extra slicing to get the partial ones.

PR #2583, opened 2025-10-22
[sharktank] gpt-oss_theta_block

oyazdanb:
Adding theta_block for toy model for gpt-oss

### DIFF
--- a/sharktank/sharktank/layers/testing.py
+++ b/sharktank/sharktank/layers/testing.py
@@ -318,53 +318,6 @@ def make_random_ffn_theta(
     )
 
 
-def _build_moe_weight_tensors(
-    *,
-    block_idx: int,
-    hidden_dim: int,
-    expert_ffn_dim: int,
-    num_experts: int,
-    weight_dtype: torch.dtype,
-    gate_inp_dtype: torch.dtype | None = None,
-    gate_inp_prefix: str = "ffn_gate_inp",
-    gate_exp_prefix: str = "ffn_gate_exps",
-    up_exp_prefix: str = "ffn_up_exps",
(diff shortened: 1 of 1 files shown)


sharktank/sharktank/layers/testing.py
```diff
@@ -373,4 +420,131 @@ def make_random_moe_block_theta(
...
+    if with_attention_sinks:
+        res["attn_sinks"] = DefaultPrimitiveTensor(
+            name=f"blk.{block_idx}.attn_sinks",
+            data=make_rand_torch((head_count,), dtype=dtype),
+        )
+
+    return Theta(res)
+
+
+def make_moe_block_with_bias_theta(
```
Alex-Vasile: Why is this a separate function rather than making the existing `make_random_moe_block_theta` more generic?

sharktank/sharktank/layers/testing.py
```diff
@@ -329,35 +376,67 @@ def make_random_moe_block_theta(
...
+
+    Builds the common MoE structure with optional FFN norm, shared experts, layer output norm, and bias terms.
+    """
+    gate_inp_prefix = "ffn_gate_inp"
+    gate_exp_prefix = "ffn_gate_exps"
+    up_exp_prefix = "ffn_up_exps"
+    down_exp_prefix = "ffn_down_exps"
+    ffn_norm_prefix = "ffn_norm"
+    ffn_norm_scale_prefix = "ffn_norm_scale"
+    layer_output_norm_prefix = "layer_output_norm"
```
Alex-Vasile: What do we gain from hardcoding these values here rather where they are used?

Alex-Vasile (approved):

PR #2592, opened 2025-10-23
[tuner] support phase-aware candidate pruning

bangtianliu:
When `libtuner.benchmark()` finds no candidates faster than baseline, downstream code crashes with `IndexError` accessing `top_candidates[0]`. However, simply returning early breaks multi-phase tuners like `ModelTuner`, where slower dispatch candidates may still improve in the full model phase due to different fusions.

This PR introduces phase-aware pruning: adds abstract method `should_prune_slower_candidates()` to `TuningClient`, extends `get_candidates_ordered_by_speedup()` with `prune_slow_candidates` parameter for testable baseline filtering logic, and implements phase-specific behavior in `DispatchTuner` (always prune) and `ModelTuner` (prune only in final phase).

### DIFF
--- a/sharktuner/sharktuner/libtuner.py
+++ b/sharktuner/sharktuner/libtuner.py
@@ -1082,7 +1082,7 @@ class BaselineResultHandler:
     def get_candidates_ordered_by_speedup(
         self,
         candidate_results: list[BenchmarkResult],
-        should_prune: bool = False,
+        prune_slow_candidates: bool = False,
     ) -> list[tuple[BenchmarkResult, float]]:
         """
         Returns a list of tuples (BenchmarkResult, speedup) sorted in ascending order based on speedup
@@ -1098,9 +1098,16 @@ class BaselineResultHandler:
 
         If no valid baseline times are available for a specific device, the fallback baseline is used.
         The fallback baseline is the average of all valid baseline times across devices.
+
+        Args:
+            candidate_results: List of benchmark results to sort.
+            prune_slow_candidates: If True and all candidates are slower than baseline,
+                returns empty list. Otherwise, returns all sorted candidates regardless.
         """
         # Check if all candidates are slower than baseline and should be pruned.
-        if should_prune and not self.is_better_than_baseline(candidate_results):
+        if prune_slow_candidates and not self.is_better_than_baseline(
+            candidate_results
+        ):
             return []
 
         if not self.is_valid():
@@ -1265,7 +1272,7 @@ def benchmark(
 
     all_candidates_with_speedup = baseline_handler.get_candidates_ordered_by_speedup(
         candidate_results,
-        should_prune=tuning_client.should_prune_slower_candidates(),
+        prune_slow_candidates=tuning_client.should_prune_slower_candidates(),
     )
     top_candidates_with_speedup = (
         all_candidates_with_speedup[:num_candidates]
--- a/sharktuner/tests/libtuner_test.py
+++ b/sharktuner/tests/libtuner_test.py
@@ -363,11 +363,13 @@ def test_baseline_result_handler_speedup():
         libtuner.BenchmarkResult(2, 3.0, "hip://0"),
     ]
     assert (
-        handler.get_candidates_ordered_by_speedup(slower_candidates, should_prune=True)
+        handler.get_candidates_ordered_by_speedup(
+            slower_candidates, prune_slow_candidates=True
+        )
         == []
     )
 
     candidates_with_speedup = handler.get_candidates_ordered_by_speedup(
-        slower_candidates, should_prune=False
+        slower_candidates, prune_slow_candidates=False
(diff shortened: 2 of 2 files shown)


bangtianliu:
> This doesn't look like a correct fix -- we can print a warning for sure, but it's still possible that even a slower dispatch candidate is faster than the baseline across the full model. We should still select top N best candidates and continue to model compilation and benchmarking next.

Good suggestion! I realized the empty list can be caused by two reasons:
1. All compilations/benchmarks failed
2. All candidates performed worse than baseline

The early return is still needed to handle the first case. For the second case, I introduced an is_final_phase() method that distinguishes between single-phase and multi-phase tuners:
- Single-phase tuners (DispatchTuner) stop when candidates are slower than baseline
- Multi-phase tuners (ModelTuner) continue to the next phase

This ensures dispatch candidates always get evaluated at the model level, where they may perform better despite being slower at the dispatch level.

Attached tuning log demonstrating the fix works as expected.
- Model tuner: https://gist.github.com/bangtianliu/977c4cc062af65e7c2216847bc6b2808
- Dispatch tuner:  https://gist.github.com/bangtianliu/7a3e7c8863849ee5291dcc56d770126a

sharktuner/model_tuner/model_tuner.py
```diff
@@ -44,6 +45,13 @@ def get_iree_benchmark_timeout_s(self) -> Optional[float]:
     def is_auto_iree_benchmark_timeout(self) -> bool:
         return self.auto_benchmark_timeout
 
+    @override
+    def should_prune_slower_candidates(self) -> bool:
+        # ModelTuner has two phases:
+        # - First phase (dispatch): return False to keep slower candidates for model phase.
+        # - Second phase (model): return True to prune slower candidates.
```
Max191: nit: IMO it's useful to explain why we want to keep slower candidates for the model phase. Maybe include that in the comment? (i.e., because slower dispatch candidates could still give improvement when run in the full model due to different fusions and/or concurrent dispatches).

sharktuner/sharktuner/libtuner.py
```diff
@@ -152,6 +152,19 @@ def is_auto_iree_benchmark_timeout(self) -> bool:
...
+    @abstractmethod
+    def should_prune_slower_candidates(self) -> bool:
+        """
+        Return True to prune all candidates when they are slower than baseline.
+        When True and all candidates are slower, returns empty list.
+        When False, returns top N candidates regardless of baseline comparison.
+
+        Single-phase tuners (e.g., DispatchTuner) should return True.
+        Multi-phase tuners (e.g., ModelTuner) should return False during early phases,
+        True during the final phase.
```
Max191: nit: This suggestion may be slightly misleading. I think the distinction is not about multi-phase vs single phase tuners, but more about whether slower candidates from the first phase could still show improvement in the later phases. Perhaps you can phrase this suggested implementation in this way instead.

kuhar (commented):
Can we implement this in a way that allows us to write unit tests?

bangtianliu:
> Can we implement this in a way that allows us to write unit tests?

Sure, I will think how to do it in that way.

sharktuner/sharktuner/libtuner.py
```diff
@@ -1191,6 +1201,27 @@ def compile(
...
+    """
+    Filters and orders candidates based on baseline comparison.
+    """
+    if not baseline_handler.is_better_than_baseline(candidate_results):
+        if should_prune_slower_candidates:
+            return []
+
+    all_candidates_with_speedup = baseline_handler.get_candidates_ordered_by_speedup(
+        candidate_results
+    )
```
kuhar: Have you considered passing a second parameter to this function, `should_prune`? I'm wondering if we need a new function at all

sharktuner/sharktuner/libtuner.py
```diff
@@ -1191,6 +1201,27 @@ def compile(
...
+    top_candidates_with_speedup = all_candidates_with_speedup[:num_candidates]
+
+    return top_candidates_with_speedup
```
kuhar: Usually there's no need for local variables only used once
```suggestion
    return all_candidates_with_speedup[:num_candidates]
```

sharktuner/sharktuner/libtuner.py
```diff
@@ -1070,7 +1080,9 @@ def is_better_than_baseline(self, candidate_results: list[BenchmarkResult]) -> b
         return False
 
     def get_candidates_ordered_by_speedup(
-        self, candidate_results: list[BenchmarkResult]
+        self,
+        candidate_results: list[BenchmarkResult],
+        should_prune: bool = False,
```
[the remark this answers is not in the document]
kuhar: I would also call it something more descriptive like `prune_slow_candidates`

kuhar (approved):

PR #2595, opened 2025-10-24
[Fusilli] Add .clang-tidy

IanWood1:
Adds .clang-tidy file to help promote a consistent code style and catch bugs. This should only affect local development; it doesn't make any changes to CI.

This has been mostly forked from [IREE's clang-tidy](https://github.com/iree-org/iree/blob/main/compiler/src/.clang-tidy) with the addition of `bugprone`, `performance`, and `readability-identifier-naming` checks. Many Fusilli files have 20+ warnings generated (mostly naming). These should probably be cleared up so that new warnings don't get ignored by authors.

### Changes:
- Adds `.clang-tidy` file.
- Enforces `kGlobalConstant` naming for global constant variables.
- Fixes names that are against the conventions.
- Disables clang-tidy checks for endl.
- Disables `-bugprone-macro-parentheses` since it conflicts with some of the logging macros.
- Makes other small changes suggested by clang tidy (std::move, make functions static, etc).

### DIFF
--- a/sharkfuser/benchmarks/driver.cpp
+++ b/sharkfuser/benchmarks/driver.cpp
@@ -267,7 +267,7 @@ int main(int argc, char **argv) {
     if (I.size() != 4 || F.size() != 4 || O.size() != 4) {
       std::cerr << "Detected at least one invalid {input, filter, output} "
                    "layout for 2D convolution."
-                << '\n';
+                << std::endl;
       return 1;
     }
   }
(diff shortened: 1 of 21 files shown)


sjain-stanford:
> Many Fusilli files have 20+ warnings generated (mostly naming). These should probably be cleared up so that new warnings don't get ignored by authors.

Strongly agree - either fix the issue to resolve the warnings or adjust the checks. Otherwise it defeats the purpose of having something like this.

Might as well do it in this PR.

sharkfuser/benchmarks/driver.cpp
```diff
@@ -181,70 +180,70 @@ int main(int argc, char **argv) {
...
-      ->check(PositiveInteger);
+      ->check(positiveInteger);
   convApp->add_option("--dilation_w,-j", j, "Conv dilation width")
       ->required()
-      ->check(PositiveInteger);
+      ->check(positiveInteger);
   convApp->add_option("--in_layout", I, "Input layout")
       ->required()
-      ->check(ValidConvLayout);
+      ->check(validConvLayout);
```
sjain-stanford: `isValidConvLayout`

PR #2607, opened 2025-10-27
[Fusilli] Add clang-tidy checks to CI

IanWood1:
Adds `FUSILLI_ENABLE_CLANG_TIDY` option to cmake and enables it on CI. Running `clang-tidy` through CLI/cmake performs more checks than running in an editor (uses the clang-tidy integrated into clangd). I disabled some checks that we should reevaluate to see if they should be kept or not. 

For example, I disabled `misc-const-correctness` since it was not showing up as an error in my editor but was erroring out the build. Fixing this would mean editing large portions of almost every file. Then, once added, it would probably cause a headache for devs due to lack of editor support.

I'm suggesting to first enable clang-tidy on CI (with a few checks disabled) so that new PRs can be analyzed against the current checks. And then over time we can adjust the checks as necessary.

Fixes all instances of:
- `misc-include-cleaner`
- `readability-identifier-naming`
- `performance-unnecessary-copy-initialization`
- `misc-use-internal-linkage`

### DIFF
--- a/.github/workflows/ci_llama_3.1_405b_fp4.yml
+++ b/.github/workflows/ci_llama_3.1_405b_fp4.yml
@@ -86,7 +86,6 @@ jobs:
 
           # Install shark deps
           bash scripts/setenv.sh --source
-          pip install -U "huggingface_hub[cli]"
           hf auth login --token ${{ secrets.HF_FLUX_TOKEN }}
 
       - name: Check runner health
--- a/scripts/export_and_compile.sh
+++ b/scripts/export_and_compile.sh
@@ -150,9 +150,8 @@ if [[ $TENSOR_PARALLELISM_SIZE = "8" ]]; then
         --iree-opt-level=O3 \
         --iree-hal-indirect-command-buffers=true \
         --iree-stream-resource-memory-model=discrete \
-        --iree-hal-memoization=true --iree-codegen-enable-default-tuning-specs=true \
+        --iree-hal-memoization=true \
         --iree-hip-enable-tensor-ukernels \
-        --iree-stream-affinity-solver-max-iterations=1024 \
         --iree-llvmgpu-test-combine-layout-transformation=false
 
 elif [[ $DTYPE = "llama-405B-FP4" ]]; then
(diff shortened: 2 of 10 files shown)


sharkfuser/tests/hip_tests/hip_kernels.h
```diff
@@ -4,6 +4,7 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+// NOLINTNEXTLINE(llvm-header-guard)
 #ifndef TESTS_HIP_TESTS_HIP_KERNELS_H
```
IanWood1: Note to reviewers: I'm not 100% sure whats going on here. I think there is some configuration issue with cmake that is causing clang-tidy to not realize that this is a correct header guard.

sharkfuser/tests/hip_tests/test_buffer.cpp
```diff
@@ -178,8 +182,10 @@ TEST_CASE("Buffer read async allocated and populated buffer",
   for (size_t i = 0; i < elementCount; ++i) {
     hostData[i] = static_cast<float>(i);
   }
+  // NOLINTBEGIN(misc-include-cleaner)
```
IanWood1: I think this is the correct usage of `include <hip/hip_runtime.h>`. This is due to hip headers not following the `misc-include-cleaner` rules.

sharkfuser/tests/hip_tests/hip_kernels.h
IanWood1: It suggests to add the full file path to the guard. I figured out how clang-tidy determines what the name should be, its from this line in tests/hip_tests/CMakeLists.txt:
```
  target_include_directories(libhiputils INTERFACE ${CMAKE_CURRENT_SOURCE_DIR})
```

So, `HIP_KERNELS_H` would work.

sjain-stanford: Makes sense. We only really care about main fusilli headers so this is OK.

PR #2617, opened 2025-10-30
[Fusilli] Header cleanup based on misc-include-cleaner from VSCode/Cursor clang-tidy extension

sjain-stanford:
For some reason these are not being surfaced through the CMake / CI invocations of clang-tidy, but these are all true positives (I've checked each one). 

It appears to only be an issue with the headers in include. For the .cpp files in samples/tests/benchmarks, the misc-include-cleaner from CMake / CLI does surface include errors that I see with the VSCode tidy extension.

So most likely how CMake target is setup - it only addresses it when compiling the cpp files but not the standalone headers which are part of the INTERFACE.
 It's not too bad - can live with it. Just one check and just for include/**. Rest of clang-tidy checks are working AFAICT in CI. (edited) 

I think its a limitation of clang-tidy's integration with cmake. clang-tidy gets run on each translation unit so its never directly run on any of the headers. The same thing happens when running the run-clang-tidy tool which also seems to run on each translation unit. But I got the same errors by running on a specific header file. Idk why this changes the results for misc-include-cleaner

### DIFF
--- a/sharkfuser/CMakeLists.txt
+++ b/sharkfuser/CMakeLists.txt
@@ -29,9 +29,9 @@ project(
 set(CMAKE_C_STANDARD 11)
 set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
-set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 # https://discourse.cmake.org/t/cmake-3-28-cmake-cxx-compiler-clang-scan-deps-notfound-not-found/9244/3
 set(CMAKE_CXX_SCAN_FOR_MODULES OFF)
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 # Includes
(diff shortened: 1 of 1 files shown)


sharkfuser/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 # https://discourse.cmake.org/t/cmake-3-28-cmake-cxx-compiler-clang-scan-deps-notfound-not-found/9244/3
 set(CMAKE_CXX_SCAN_FOR_MODULES OFF)
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
IanWood1: This is already set above

```suggestion
```

IanWood1 (approved):

Issue #2528, opened 2025-10-16
[Tuner] improve the boo tuner


bangtianliu:
- I noticed many errors of this type appearing in the tuner logs:
```
2025-10-15 21:18:53,076 - ERROR - Error in candidate_gen.py:
Traceback (most recent call last):
  File "/home/bangtliu/shark-ai/sharktuner/sharktuner/libtuner.py", line 738, in generate_candidate_specs
    config_specs: list[ir.Module] = candidate_gen.generate_configs_and_td_specs(
  File "/home/bangtliu/shark-ai/sharktuner/sharktuner/candidate_gen.py", line 206, in generate_configs_and_td_specs
    for i, config in enumerate(
  File "/home/bangtliu/shark-ai/sharktuner/sharktuner/constraint_generator.py", line 116, in generate_generic_contraction_solutions
    constraints = dispatch_constraints.generate_tile_and_fuse_constraints(
  File "/home/bangtliu/shark-ai/sharktuner/sharktuner/dispatch_constraints.py", line 288, in generate_tile_and_fuse_constraints
    M[-1] = ((M[-1] + intrinsic_mn - 1) / intrinsic_mn) * intrinsic_mn
IndexError: list index out of range
```
Conclusion: this is caused by empty M list caused by mat-vec op. In short term we should send a simple PR to raise unsupport error.

- When setting num_candidate to 5000, only about 50–60 specs are actually generated and used for compilation. This seems suspicious.

- The tuner also fails to handle certain tuning cases. For example, no td specs are generated for the following configuration: `convbfp16 -n 1 -c 224 -H 1 -W 1 -k 8 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC -m conv -g 1 -F 2 -t 1
`

All of the above are just initial observations after setting up the BOO tuner. [WIP]

bangtianliu:
cc @kuhar @MaheshRavishankar

PR #2627, opened 2025-11-01
[tuner] exits gracefully for unsupported cases like mat-vec operations

bangtianliu:
Motivate by issues we encountered here: https://github.com/nod-ai/shark-ai/issues/2528#issuecomment-3417332356

### Problem
The existing tuner cannot support mat-vec operations, which are processed along the contraction code path. This causes crashes when accessing empty dimension arrays:

```
  M[-1] = ((M[-1] + intrinsic_mn - 1) / intrinsic_mn) * intrinsic_mn
IndexError: list index out of range
```
The root cause is that mat-vec operations have an empty M dimension, but the tuner attempts to access `M[-1]` without validating that the dimension exists.

### Solution
This PR improves error handling in the tuner by implementing graceful exits when dispatch tuner setup fails, instead of crashing with assertion errors.


sharktuner/tests/candidate_gen_test.py
```diff
@@ -215,3 +215,144 @@ def test_get_td_spec_convolution(tuner_ctx: common.TunerContext) -> None:
...
+    Once the tuner supports mat-vec, this test can be removed.
+    """
+    context = tuner_ctx.mlir_ctx
+    module_str = """
+        builtin.module{
+            func.func @test(%x: tensor<8xbf16>, %A: tensor<8x224xbf16>) -> tensor<224xf32> {
+                %zero = arith.constant 0.0 : f32
+                %init = tensor.empty() : tensor<224xf32>
+                %y0 = linalg.fill ins(%zero : f32) outs(%init : tensor<224xf32>) -> tensor<224xf32>
+                %y = linalg.generic {
```
kuhar: Can we use named linalg ops like `linalg.matmul` or `linalg.matvec` to simplify this?

kuhar:
Seems like there are some test failures

bangtianliu:
> Yeah, looking into this and it seems that the old iree version is installed.

it seems that wave-lang still uses iree 3.8.0, should I remove it from `requirements-iree-unpinned.txt`. cc @kuhar

PR #2632, opened 2025-11-03
[Fusilli] Add matmul node 2/4

IanWood1:
Second of 4 PRs to add support for matmul operations:
- https://github.com/nod-ai/shark-ai/pull/2631
- https://github.com/nod-ai/shark-ai/pull/2632 (this PR)
- https://github.com/nod-ai/shark-ai/pull/2633
- ... benchmark driver support.

### DIFF
--- a/sharkfuser/include/fusilli/node/matmul_node.h
+++ b/sharkfuser/include/fusilli/node/matmul_node.h
@@ -72,16 +72,6 @@ public:
   MatmulNode(MatmulAttr &&attr, const Context &ctx)
       : NodeCRTP(ctx), matmulAttr(std::move(attr)) {}
 
-  // MLIR assembly emitter helper methods.
-  std::string emitNodePreAsm() const override final;
-  std::string getOperandNamesAsm() const;
-  std::string getOperandTypesAsm() const;
-  std::string getResultNamesAsm() const;
-  std::string getResultTypesAsm() const;
-  std::string getPermuteAOpsAsm() const;
-  std::string getPermuteBOpsAsm() const;
-  std::string getPermuteCOpsAsm() const;
-
   const std::string &getName() const override final {
     return matmulAttr.getName();
   }
@@ -142,21 +132,17 @@ public:
     const std::vector<int64_t> &aDim = aT->getDim();
     const std::vector<int64_t> &bDim = bT->getDim();
 
-    std::vector<int64_t> cDim = cT->getDim();
-    std::vector<int64_t> cStride = cT->getStride();
+    const std::vector<int64_t> &cDim = cT->getDim();
+    const std::vector<int64_t> &cStride = cT->getStride();
 
     // Infer shape of output tensor.
-    if (cDim.empty()) {
-      cDim = getMatmulInferredOutputShape(aDim, bDim);
-      cT->setDim(cDim);
-    }
+    if (cDim.empty())
+      cT->setDim(getMatmulInferredOutputShape(aDim, bDim));
 
     // Infer stride of output tensor.
     if (cStride.empty()) {
-      // When unspecified, output is always contiguous.
-      cStride =
-          generateStrideFromDim(cDim, getContiguousStrideOrder(cDim.size()));
-      cT->setStride(cStride);
+      cT->setStride(
+          generateStrideFromDim(cDim, getContiguousStrideOrder(cDim.size())));
     }
 
     return ok();
--- a/sharkfuser/tests/test_matmul_node.cpp
+++ b/sharkfuser/tests/test_matmul_node.cpp
@@ -101,7 +101,9 @@ TEST_CASE("MatmulNode inferPropertiesNode when C is fully specified",
       TensorAttr().setDim({m, n}).setStride({n, 1})));
 
   MatmulNode node(std::move(attr), ctx);
+  FUSILLI_REQUIRE_OK(node.preValidateNode());
   FUSILLI_REQUIRE_OK(node.inferPropertiesNode());
+  FUSILLI_REQUIRE_OK(node.postValidateNode());
 
   auto cT = node.matmulAttr.getC();
   REQUIRE(cT->getDim() == std::vector<int64_t>{m, n});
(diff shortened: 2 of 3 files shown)


sharkfuser/include/fusilli/node/matmul_node.h
```diff
@@ -0,0 +1,182 @@
...
+
+  MatmulNode(MatmulAttr &&attr, const Context &ctx)
+      : NodeCRTP(ctx), matmulAttr(std::move(attr)) {}
+
+  const std::string &getName() const override final {
+    return matmulAttr.getName();
+  }
+  Type getType() const override final { return Type::Matmul; }
+
+  ErrorObject preValidateNode() const override final {
```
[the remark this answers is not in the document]
sjain-stanford: [cudnn matmul ](https://docs.nvidia.com/deeplearning/cudnn/backend/v9.15.0/api/cudnn-graph-library.html#cudnn-backend-operation-matmul-descriptor)

> The addressing of the matrix elements from a given tensor can be specified using strides in the tensor descriptor. The strides represent the spacing between elements for each tensor dimension. Considering a matrix tensor A (B x M x N) with strides [BS, MS, NS], it indicates that the actual matrix element A[x, y, z] is found at (A_base_address + x * BS + y * MS + z * NS) from the linear memory space allocated for tensor A. With our current support, the innermost dimension must be packed, which requires either MS=1 or NS=1. Otherwise, there are no other technical constraints with regard to how the strides can be specified in a tensor descriptor as it should follow the aforementioned addressing formula and the strides as specified by the user.

> This representation provides support for some common usages, such as leading dimension and matrix transpose as we will explain through the following examples.

> The most basic case is a fully packed row-major batch matrix, without any consideration of leading dimension or transpose. In this case, BS = M*N, MS = N, and NS = 1.

> Matrix transpose can be achieved by exchanging the inner and outer dimensions using strides. Namely:

> To specify a non-transposed matrix: BS = M*N, MS = N, and NS = 1.

> To specify matrix transpose: BS = M*N, MS = 1, and NS = M.

> Leading dimension, a widely used concept in BLAS-like APIs, describes the inner dimension of the 2D array memory allocation (as opposed to the conceptual matrix dimension). It resembles the stride in a way that it defines the spacing between elements in the outer dimension. The most typical use cases where it shows difference from the matrix inner dimension is when the matrix is only part of the data in the allocated memory, addressing submatrices, or addressing matrices from an aligned memory allocation. Therefore, the leading dimension LDA in a column-major matrix A must satisfy LDA >= M, whereas in a row-major matrix A, it must satisfy LDA >= N. To transition from the leading dimension concept to using strides, this entails MS >= N and NS = 1 or MS = 1 and NS >= M. Keep in mind that, while these are some practical use cases, these inequalities do not impose technical constraints with respect to an acceptable specification of the strides.

sharkfuser/include/fusilli/node/matmul_node.h
```diff
@@ -0,0 +1,182 @@
...
+        "Matmul input tensor A must have a rank of at least 2");
+    FUSILLI_RETURN_ERROR_IF(
+        bRank < 2, ErrorCode::InvalidAttribute,
+        "Matmul input tensor B must have a rank of at least 2");
+
+    // Check that inner dimensions match (K dimension).
+    const std::vector<int64_t> &aDim = aT->getDim();
+    const std::vector<int64_t> &bDim = bT->getDim();
+    int64_t aK = aDim[aRank - 1]; // Last dimension of A
+    int64_t bK = bDim[bRank - 2]; // Second-to-last dimension of B
```
sjain-stanford: Does this assume a specific stride order, or is it sufficient to say the logical dims are always "row-major" / "contiguous" for this to be correct?

sharkfuser/include/fusilli/node/matmul_node.h
```diff
@@ -0,0 +1,182 @@
...
+    const std::vector<int64_t> &bDim = bT->getDim();
+
+    const std::vector<int64_t> &cDim = cT->getDim();
+    const std::vector<int64_t> &cStride = cT->getStride();
+
+    // Infer shape of output tensor.
+    if (cDim.empty())
+      cT->setDim(getMatmulInferredOutputShape(aDim, bDim));
+
+    // Infer stride of output tensor.
```
sjain-stanford: nit: Consider revising the comment to reflect that a contiguous / row-major stride is picked when un-specified.

sharkfuser/tests/test_matmul_node.cpp
```diff
@@ -0,0 +1,515 @@
...
+  int64_t m = 16, k = 32, n = 64;
+
+  auto aT = std::make_shared<TensorAttr>(
+      TensorAttr().setDim({m, k}).setStride({k, 1}).setName("A_contiguous"));
+
+  auto bT = std::make_shared<TensorAttr>(
+      TensorAttr()
+          .setDim({k, n})
+          .setStride(
+              {k * n, k}) // Arbitrary layout (not contiguous or channels-last)
```
sjain-stanford: What does it mean to have an arbitrary layout? Does this change how we emit the mlir asm / handle transposes.

sjain-stanford:
Also, something is off with the automatic base branch change - you might need to rebase to remove the commits pertaining to PR 1/4 now that it landed.

sharkfuser/tests/test_matmul_node.cpp
```diff
@@ -0,0 +1,515 @@
...
+  attr.setA(aT).setB(bT).setC(std::make_shared<TensorAttr>());
+
+  MatmulNode node(std::move(attr), ctx);
+
+  FUSILLI_REQUIRE_OK(node.preValidateNode());
+  FUSILLI_REQUIRE_OK(node.inferPropertiesNode());
+  FUSILLI_REQUIRE_OK(node.postValidateNode());
+}
+
+TEST_CASE("MatmulNode postValidate accepts arbitrary output strides",
```
a-sidorova: Could you please add tests with transposed inputs?
```
To specify a non-transposed matrix: A_strides = [M*N,  N, 1].
To specify matrix transpose: At_strides = [M*N, 1, M].
```

sharkfuser/include/fusilli/node/matmul_node.h
```diff
@@ -0,0 +1,182 @@
...
+
+    FUSILLI_RETURN_ERROR_IF(
+        cT->getDim() !=
+            getMatmulInferredOutputShape(aT->getDim(), bT->getDim()),
+        ErrorCode::InvalidAttribute,
+        "Matmul output tensor C dimensions do not match the expected shapes "
+        "inferred based on the input dimensions");
+
+    // No layout restrictions - we support arbitrary strides through
+    // permutations.
```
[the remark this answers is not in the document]
IanWood1: I'm going to remove this comment since it doesn't seem like the right place to put it. The idea is that you can specify matmul transpose B/normal matmul/matmul transpose A via the strides. I'll try to find a place to add examples.

IanWood1:
Moving to https://github.com/iree-org/fusilli

PR #2629, opened 2025-11-03
[Fusilli] Added lit and sample tests for grouped strided ConvWGrad

a-sidorova:
The remaining tests from https://github.com/nod-ai/shark-ai/pull/2599

PR #2646, opened 2025-11-05
[Fusilli] Cleanup LINALG-CHECK from certain lit tests

sjain-stanford:
Context: https://github.com/nod-ai/shark-ai/pull/2629#discussion_r2495582000

> For this case though, I feel we can just drop the LINALG-CHECKS entirely, because:
> 1. The IR is sub-optimal and will go through significant changes with the torch-mlir revamp of backward conv lowering that you and others are working on). 
> 2. Tests on linalg transformations would normally live alongside the sources for those linalg transformations - so torch-mlir or IREE (we don't do any of that in fusilli as we stop at emitting torch dialect). 

> We made an exception to 2 for some lit tests and added the linalg checks to verify certain assumptions we make in fusilli around layout transformations - like emitting transposes for NHWC -> NCHW don't incur an actual cost and are absorbed by transpose propagation/eliminations in the compiler (Torch->Linalg). Sorry for the confusion.

### DIFF
--- a/.github/workflows/ci_boo_conv.yml
+++ b/.github/workflows/ci_boo_conv.yml
@@ -76,16 +76,16 @@ jobs:
        export ROCR_VISIBLE_DEVICES="0,1,2,3,4,5,6,7"
        echo "Running all proxy config without tuning"
        export BOO_TUNING_SPEC_PATH=""
-       python iree/turbine/kernel/boo/driver/driver.py --gpu-id=-1 --commands-file ../shark-ai-reports/boo/all_proxy_config.txt --csv ../output_artifacts/all_proxy_run_without_tuning.csv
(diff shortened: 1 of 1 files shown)


sharkfuser/tests/lit/test_conv_asm_emitter_ndhwc_kdrsc_grouped.cpp
```diff
@@ -9,8 +9,6 @@
 // XFAIL: {{.*}}
 // RUN: %{TEST_EXE} | iree-opt --verify-roundtrip
 // RUN: %{TEST_EXE} | FileCheck %s --check-prefix=TORCH-CHECK
-// RUN: %{TEST_EXE} | iree-compile - --compile-to=input | \
-// RUN:             FileCheck %s --check-prefix=LINALG-CHECK
```
a-sidorova: Do we need then update comment `TODO` above (remove the asking about adding of `LINALG-CHECK section`)?

PR #2650, opened 2025-11-07
[Fusilli] Benchmark runner improvements

sjain-stanford:
This is just a small collection of updates I found useful when running the batch benchmarks. 

1. Remove `use_tempdir` in favor of directly checking if `output_dir` is set (redundant option)
2. Relax verbosity requirements for a few prints (e.g. printing command and stats is now default)
3. Skipped commands will now appear in the resulting csv (this is important to maintain 1:1 ordering with the spreadsheet)
4. Always `--continue-on-error` as there's no reason not to (don't want a long run to be terminated by a failed run after a while)
5. Reduce timeout to 30s (none of the Fusilli ops should take that long anyway)
6. Move benchmark tests into benchmarks/ dir (this should help unify the runner and driver tests using the same command file)
7. Use `$<TARGET_FILE:fusilli_benchmark_driver>` to point the runner tests to the driver
8. Minor syntactical cleanup

### DIFF
--- a/sharkfuser/benchmarks/run_benchmark.py
+++ b/sharkfuser/benchmarks/run_benchmark.py
@@ -290,13 +290,6 @@ The script will:
         help="Timeout in seconds for each command (default: 30, use -1 for no timeout)",
     )
 
-    parser.add_argument(
-        "--continue-on-error",
-        "-C",
-        action="store_true",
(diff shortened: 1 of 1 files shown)


raayandhar:
Thanks for these changes, definitely makes it a bit nicer to use and cleaner to read.

PR #2653, opened 2025-11-07
[Fusilli] Generalize physical tensor layout support

IanWood1:
Currently, Fusilli only supports 2 physical tensor layouts: NCHW (contiguous) and NHWC. However, pointwise ops and matmul ops (https://github.com/nod-ai/shark-ai/pull/2632) require support for any layout that can be represented as a permutation of dimensions.

#### 1. Added new validation method `hasValidPhysicalRepresentation()`
This method checks that the combination of stride & logical dimensions can be represented as a physical tensor and that the logical tensor is a transpose of the physical tensor.

#### 2. Generalized `getPhysicalDim()`
This now supports getting the physical dims for when they are a permutation of the logical dims. This uses `getLogicalToPhysicalPerm()` to get the physical dims by transposing the logical dims.

#### 3. Updated the ASM emitters
Updated the ASM emitters to use the new helper `getLogicalToPhysicalPerm()`. This is the same utility used by `getPhysicalDims()`, which ensures that the transpose vec matches the expected type. This is mostly an NFC and provides a cleaner and more concise way to get the permutation.

Also, pointwise nodes weren't converting the physical tensor to a logical tensor, which was incorrect. This PR breaks pointwise ops without fixing this, so the fix for pointwise ops is included here.

#### 4. Updated CHECK lines in conv tests
`getPhysicalDim` assumed that physical memory would be stored only in NCHW or NHWC format. That meant there was a unique way to go from logical to physical. Now that this is generalized, we run into a problem: unit dims add ambiguity and so there is no longer a unique physical tensor representation of the logical dims. Unit dims can be placed anywhere without changing the meaning of the tensor. However, we still need a single unique representation for the MLIR types to match. When finding the physical tensor, this PR chooses to place the non-unit dims in the same location as the logical tensor. This provides a way to find a unique physical representation of the tensor.

For example:

```
k = 256, c = 128, r = 1, s = 1
dim = [k, c, r, s] = [256, 128, 1, 1]
stride = [c * r * s, r * s, s, 1] = [128, 1, 1, 1]
```

Previously, this would result in a physical tensor of `!torch.vtensor<[256,1,1,128],f32>` but now this results in `!torch.vtensor<[256,128,1,1],f32>`.

### DIFF
--- a/sharkfuser/samples/pointwise/pointwise_add_transposed.cpp
+++ b/sharkfuser/samples/pointwise/pointwise_add_transposed.cpp
@@ -20,18 +20,20 @@
 using namespace fusilli;
 
 TEST_CASE("Pointwise add with transposed operand", "[pointwise][graph]") {
-  const int64_t n = 2, m = 2;
+  const int64_t n = 3, m = 2;
 
   // clang-format off
   const std::vector<float> inputData = {
     1.0f, 2.0f,
-    3.0f, 4.0f
+    3.0f, 4.0f,
+    5.0f, 6.0f
   };
 
   // Result of inputData + transpose(inputData)
   const std::vector<float> expectedResult = {
-    2.0f, 5.0f,
-    5.0f, 8.0f
+    2.0f, 6.0f,
+    5.0f, 9.0f,
+    8.0f, 12.0f
   };
   // clang-format on
 
@@ -54,18 +56,17 @@ TEST_CASE("Pointwise add with transposed operand", "[pointwise][graph]") {
     graph->setName("pointwise_add_transposed");
     graph->setIODataType(DataType::Float).setComputeDataType(DataType::Float);
 
-    // Tensor A: contiguous 2x2 tensor (row-major)
+    // Tensor A: contiguous nxm tensor (row-major)
     auto aT =
         graph->tensor(TensorAttr().setName("input_a").setDim({n, m}).setStride(
             {m, 1})); // Contiguous: stride={2, 1}
 
-    // Tensor B: transposed 2x2 tensor
-    // Logical dim={2, 2}, but stored with transposed strides
-    auto bT =
-        graph->tensor(TensorAttr()
-                          .setName("input_b_transposed")
-                          .setDim({n, m})
-                          .setStride({1, n})); // Transposed: stride={1, 2}
+    // Tensor B: transposed nxm tensor
+    // Logical dim={n, m}, but stored with transposed strides
+    auto bT = graph->tensor(TensorAttr()
+                                .setName("input_b_transposed")
+                                .setDim({n, m})
+                                .setStride({1, n})); // Transposed
 
     // Create Pointwise ADD op
     auto pointwiseAttr = PointwiseAttr()
--- a/sharkfuser/tests/test_tensor_attributes.cpp
+++ b/sharkfuser/tests/test_tensor_attributes.cpp
@@ -397,6 +397,36 @@ TEST_CASE("getLogicalToPhysicalPermuteOrder", "[TensorAttr]") {
           std::vector<int64_t>{0, 1, 2});
 }
 
+TEST_CASE("getPhysicalToLogicalPermuteOrder", "[TensorAttr]") {
+  // Note: This uses the same underlying helper as
+  // getLogicalToPhysicalPermuteOrder, so we only need basic checks.
+  TensorAttr t;
+
+  // Contiguous layout (identity permutation)
+  t.setDim({2, 3, 4}).setStride({12, 4, 1});
+  REQUIRE(t.getPhysicalToLogicalPermuteOrder() ==
+          std::vector<int64_t>{0, 1, 2});
+
+  // 2D transpose
+  t.setDim({10, 20}).setStride({1, 10});
+  REQUIRE(t.getPhysicalToLogicalPermuteOrder() == std::vector<int64_t>{1, 0});
+
+  // Channels-last NHWC
+  t.setDim({2, 3, 4}).setStride({12, 1, 3});
+  REQUIRE(t.getPhysicalToLogicalPermuteOrder() ==
+          std::vector<int64_t>{0, 2, 1});
+
+  // Complex 4D permutation
+  t.setDim({2, 3, 4, 5}).setStride({60, 1, 15, 3});
+  REQUIRE(t.getPhysicalToLogicalPermuteOrder() ==
+          std::vector<int64_t>{0, 3, 1, 2});
+
+  // With unit dimensions
+  t.setDim({1, 64, 1, 128}).setStride({999, 128, 999, 1});
+  REQUIRE(t.getPhysicalToLogicalPermuteOrder() ==
+          std::vector<int64_t>{0, 1, 2, 3});
+}
+
 TEST_CASE("TensorAttr hasValidPhysicalRepresentation", "[TensorAttr]") {
   TensorAttr t;
 


IanWood1:
TODOs
- Use hasValidPhysicalRepresentation to verify nodes (conv already checks layout which is a subset of hasValidPhysicalRepresentation)

sharkfuser/include/fusilli/attributes/tensor_attributes.h
```diff
@@ -257,6 +257,62 @@ getContiguousToChannelsLastPermuteOrder(size_t numDims) {
...
+    inverse[permutation[i]] = static_cast<int64_t>(i);
+  }
+  return inverse;
+}
+
+// This computes the permutation needed to go from logical dims to physical dims
+// based on the dims and stride. Unit length dimensions are kept in their
+// original positions as they don't meaningfully contribute to the layout.
+inline std::vector<int64_t>
+getLogicalToPhysicalPerm(const std::vector<int64_t> &dim,
```
sjain-stanford:
```suggestion
getLogicalToPhysicalPermuteOrder(const std::vector<int64_t> &dim,
```
for consistency with similar utilities that return the permute order

sharkfuser/tests/lit/test_conv_asm_emitter_nchw_krsc.cpp
```diff
@@ -13,7 +13,7 @@
 // clang-format off
 //
 // TORCH-CHECK:   module @module {
-// TORCH-CHECK:     func.func @main(%result_: !torch.tensor<[16,256,64,32],f32>, %arg0_image: !torch.vtensor<[16,128,64,32],f32>, %arg1_filter: !torch.vtensor<[256,1,1,128],f32>) attributes {torch.assume_strict_symbolic_shapes} {
+// TORCH-CHECK:     func.func @main(%result_: !torch.tensor<[16,256,64,32],f32>, %arg0_image: !torch.vtensor<[16,128,64,32],f32>, %arg1_filter: !torch.vtensor<[256,128,1,1],f32>) attributes {torch.assume_strict_symbolic_shapes} {
```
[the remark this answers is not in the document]
IanWood1: No, this comes from the change to `getPhysicalDim()`. Note that MLIR that you linked to is from the Graph node, not the conv node (since it is the function arg). The graph node should have no concept of NCHW/NHWC. It just determines the physical tensor from the logical tensor + permutation obtained from the stride. The changes to the tests are specific to the ambiguity that unit dims introduce. I.e., no changes were made to tests that had no unit dims. From the PR description:

>Now that this is generalized, we run into a problem: unit dims add ambiguity and so there is no longer a unique physical tensor representation of the logical dims. Unit dims can be placed anywhere without changing the meaning of the tensor. However, we still need a single unique representation for the MLIR types to match. When finding the physical tensor, this PR chooses to place the non-unit dims in the same location as the logical tensor.

sharkfuser/tests/test_tensor_attributes.cpp
```diff
@@ -265,6 +265,130 @@ TEST_CASE("TensorAttr isContiguous and isChannelsLast checks", "[TensorAttr]") {
   REQUIRE(t2.getPhysicalDim() == std::vector<int64_t>{2, 4, 3});
 }
 
+TEST_CASE("TensorAttr getPhysicalDim", "[TensorAttr]") {
```
sjain-stanford: Also needs unit tests for `getLogicalToPhysicalPerm`.

sharkfuser/include/fusilli/attributes/tensor_attributes.h
```diff
@@ -467,21 +523,65 @@ class TensorAttr {
...
+  }
+
+  // Convert logical dims + stride into physical dims. The stride
+  // represents how dimensions are accessed in physical memory.
+  //
+  // Dimensions of size 1 don't meaningfully contribute to the physical layout
+  // and are kept in their relative positions to preserve information.
+  std::vector<int64_t> getPhysicalDim() const {
+    assert(hasValidPhysicalRepresentation() &&
+           "Tensor has invalid physical representation");
```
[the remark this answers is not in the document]
IanWood1: Added this, and now `SECTION("Zero dimension tensor")` (at the top of the tensor tests) checks for failure instead.

sjain-stanford (approved):
LGTM, please feel free to land after adding the sample.

sharkfuser/include/fusilli/attributes/tensor_attributes.h
```diff
@@ -467,21 +548,65 @@ class TensorAttr {
...
+    // and the expected product of dimensions should match
+    int64_t expectedStride = 1;
+    for (size_t i = 0; i < numDims; ++i) {
+      int64_t actualStride = strideAndDim[i].first;
+      int64_t dimSize = strideAndDim[i].second;
+
+      // For unit length dimensions, any stride value is technically valid
+      if (dimSize == 1) {
+        continue;
+      }
```
sjain-stanford: nit: remove braces for single line ifs

sharkfuser/include/fusilli/support/asm_emitter.h
```diff
@@ -1090,27 +1052,98 @@ inline std::string ConvDGradNode::emitNodePreAsm() const {
...
+  // Get the input tensor based on index
+  std::shared_ptr<TensorAttr> inputT = pointwiseAttr.inputs.at(
+      static_cast<PointwiseAttr::InputNames>(inputIndex));
+  std::ostringstream oss;
+  std::string prefix = "permute_IN_" + std::to_string(inputIndex);
+  std::string suffix = pointwiseAttr.getName();
+
+  // Emit permute dimensions based on layout.
+  oss << getListOfIntOpsAsm(
+      getLogicalToPhysicalPermuteOrder(inputT->getDim(), inputT->getStride()),
```
sjain-stanford: I'm a little confused by the naming of this. Aren't we going from physical -> logical on the inputs? 

Consider operand `bT` (IN_1) of logical dim `{128, 256}` stride `{1, 128}`.
```c++
  auto bT =
      graph->tensor(TensorAttr()
                        .setName("arg1_add_transposed")
                        .setDim({128, 256})
                        .setStride({1, 128})); // Transposed: stride={1, 128}
```

In the emitted MLIR, transpose ops are inserted to permute operand `bT` from physical dims `{256, 128}` to logical dims `{128, 256}` before applying the pointwise op, right?

sharkfuser/include/fusilli/support/asm_emitter.h
```diff
@@ -477,14 +469,11 @@ inline std::string ConvFPropNode::getPermuteYOpsAsm() const {
...
-    oss << getListOfIntOpsAsm(
-        getPreserveContiguousPermuteOrder(yT->getDim().size()), prefix, suffix);
-  else
-    oss << getListOfIntOpsAsm(
-        getContiguousToChannelsLastPermuteOrder(yT->getDim().size()), prefix,
-        suffix);
+  // Emit permute dimensions based on layout (inverse to convert back from
+  // contiguous).
+  oss << getListOfIntOpsAsm(inversePermutation(getLogicalToPhysicalPermuteOrder(
+                                yT->getDim(), yT->getStride())),
```
sjain-stanford: For readability, consider creating a separate utility `getPhysicalToLogicalPermuteOrder` that internally calls `inversePermutation` on the `getLogicalToPhysicalPermuteOrder`...  

(btw these will get swapped after addressing the naming comment..)

sharkfuser/include/fusilli/support/asm_emitter.h
sjain-stanford: This should be called `getPhysicalToLogicalPermuteOrder` instead ([comment](https://github.com/nod-ai/shark-ai/pull/2653#discussion_r2512981432)). It makes sense because we want the ops to operate on logical dims but the main inputs/outputs are physical, so the translation from inputs -> op is physical->logical and from op -> outputs is logical -> physical (inverse).

sharkfuser/include/fusilli/attributes/tensor_attributes.h
```diff
@@ -467,21 +548,65 @@ class TensorAttr {
...
+  // Check if the stride pattern is valid and can represent a physical tensor.
+  // A valid stride pattern must satisfy:
+  // 1. Each stride is a product of dimensions in faster-changing positions
+  // 2. The strides are consistent with some permutation of dimensions
+  bool hasValidPhysicalRepresentation() const {
     size_t numDims = dim_.size();
+    if (numDims == 0)
+      return true;
+    if (numDims != stride_.size())
+      return false;
```
a-sidorova: Could you please move this check up? It seems like it should be before empty tensor validation

sharkfuser/include/fusilli/attributes/tensor_attributes.h
```diff
@@ -467,24 +486,153 @@ class TensorAttr {
...
+      size_t targetIdx = strideWithIndex[i].second;
+      permuteOrder[logicalIdx] = static_cast<int64_t>(targetIdx);
+    }
+
+    return permuteOrder;
+  }
+
+  // This computes the permutation needed to go from physical dims to logical
+  // dims. This is the inverse of getLogicalToPhysicalPermuteOrder().
+  std::vector<int64_t> getPhysicalToLogicalPermuteOrder() const {
```
sjain-stanford: nit: please unit test this

sharkfuser/include/fusilli/attributes/tensor_attributes.h
```diff
@@ -257,6 +257,82 @@ getContiguousToChannelsLastPermuteOrder(size_t numDims) {
...
+// This computes the permutation needed to go from logical dims to physical dims
+// based on the dims and stride. Unit length dimensions are kept in their
+// original positions as they don't meaningfully contribute to the layout.
+//
+// Note: perm[i] means that the i-th dimension in the logical layout is mapped
+// to the perm[i]-th dimension in the physical layout.
+//
+// Examples:
+//   1. Contiguous NCHW layout: dim={2, 3, 4}, stride={12, 4, 1}
+//      Returns: {0, 1, 2} (identity - already in physical order)
```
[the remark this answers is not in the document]
a-sidorova: Oh I see, thank you for the explanation!

sharkfuser/samples/pointwise/pointwise_add_transposed.cpp
```diff
@@ -20,18 +20,20 @@
...
   };
 
   // Result of inputData + transpose(inputData)
   const std::vector<float> expectedResult = {
-    2.0f, 5.0f,
-    5.0f, 8.0f
+    2.0f, 6.0f,
+    5.0f, 9.0f,
+    8.0f, 12.0f
   };
```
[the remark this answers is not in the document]
IanWood1: The current implementation uses the stride of the first input that matches the output shape. This matches Cudnn's frontend. However, if users want a specific output stride, they should specify it explicitly.

PR #2666, opened 2025-11-12
[Tuner] Expand tuning candidate logging feature

RattataKing:
This PR expands the logging function to support data analysis of the candidate search space and `candidate_ordering` performance evaluation.
Tuner now records candidate info such as tuning knob configurations, position in benchmark queue, final result ranking, etc.
The collected tuning records will be exported to a CSV file under the tuning base folder.

### DIFF
--- a/sharktuner/sharktuner/candidate_ordering.py
+++ b/sharktuner/sharktuner/candidate_ordering.py
@@ -152,7 +152,7 @@ def build_tuning_records_from_order(
 
 def flatten_records(
     tuning_records: list[TuningRecord],
-) -> tuple[list[str], list[dict[str, Any]]]:
+) -> list[dict[str, Any]]:
     """
     Flatten a list of `TuningRecord` objects into CSV headers and rows.
 
@@ -172,15 +172,14 @@ def flatten_records(
                 row[attr] = val
         rows.append(row)
 
-    headers = list(row.keys())
-
-    return headers, rows
+    return rows
 
 
 def export_record_to_csv(tuning_records: list[TuningRecord], dest_file: Path) -> None:
     assert tuning_records
 
-    headers, rows = flatten_records(tuning_records)
+    rows = flatten_records(tuning_records)
+    headers = list(rows[0].keys())
 
     with open(dest_file, "w", newline="", encoding="utf-8") as f:
         writer = csv.DictWriter(f, fieldnames=headers)
--- a/sharktuner/sharktuner/libtuner.py
+++ b/sharktuner/sharktuner/libtuner.py
@@ -1334,16 +1334,17 @@ def benchmark(
         candidate_results,
         prune_slow_candidates=tuning_client.should_prune_slower_candidates(),
     )
-    if all_candidates_with_speedup:
-        for i, handler_res in enumerate(all_candidates_with_speedup, start=1):
-            benchmark_res, speedup = handler_res
-            cid, _, device_id = benchmark_res
-            baseline_res = baseline_handler.get_average_result_us(device_id)
-            tuning_client.tuning_records[cid].baseline_benchmark_time_us = (
-                round(baseline_res, 2) if baseline_res else None
-            )
-            tuning_client.tuning_records[cid].benchmark_speedup = round(speedup, 5)
-            tuning_client.tuning_records[cid].benchmark_rank_order = i
+
+    # Best candidate gets rank 1.
+    for i, handler_res in enumerate(all_candidates_with_speedup, start=1):
+        benchmark_res, speedup = handler_res
+        cid, _, device_id = benchmark_res
+        baseline_res = baseline_handler.get_average_result_us(device_id)
+        tuning_client.tuning_records[cid].baseline_benchmark_time_us = (
+            round(baseline_res, 2) if baseline_res else None
+        )
+        tuning_client.tuning_records[cid].benchmark_speedup = round(speedup, 5)
+        tuning_client.tuning_records[cid].benchmark_rank_order = i
 
     top_candidates_with_speedup = (
         all_candidates_with_speedup[:num_candidates]
--- a/sharktuner/tests/candidate_ordering_test.py
+++ b/sharktuner/tests/candidate_ordering_test.py
@@ -207,39 +207,7 @@ def test_flatten_records(
     )
     sample_tuning_records = [tr1, tr2]
 
-    headers, rows = candidate_ordering.flatten_records(sample_tuning_records)
-
-    expected_headers = [
-        "gen_id",
-        "candidate_id",
-        "knob_M",
-        "knob_N",
-        "knob_K",
-        "knob_tile_m",
-        "knob_tile_n",
-        "knob_tile_k",
-        "knob_wg_x",
-        "knob_wg_y",
-        "knob_wg_z",
-        "knob_subgroup_m_cnt",
-        "knob_subgroup_n_cnt",
-        "knob_intrinsic_mn",
-        "knob_intrinsic_k",
-        "knob_subgroup_m",
-        "knob_subgroup_n",
-        "knob_subgroup_k",
-        "to_compile",
-        "compile_status",
-        "to_benchmark",
-        "benchmark_device_id",
-        "benchmark_queue_position",
-        "benchmark_status",
-        "baseline_benchmark_time_us",
-        "benchmark_time_us",
-        "benchmark_speedup",
-        "benchmark_rank_order",
-    ]
-    assert headers == expected_headers
+    rows = candidate_ordering.flatten_records(sample_tuning_records)
 
     expected_key_rows: list[dict] = [
         {


sharktuner/sharktuner/candidate_ordering.py
```diff
@@ -104,3 +109,86 @@ def reorder_assignments(
...
+def export_record_to_csv(tuning_records: list[TuningRecord], dest_file: Path) -> None:
+    if not tuning_records:
+        return None
+
+    headers, rows = flatten_records(tuning_records)
+
+    with open(dest_file, "w", newline="", encoding="utf-8") as f:
+        writer = csv.DictWriter(f, fieldnames=headers)
+        writer.writeheader()
+        writer.writerows(rows)
```
bangtianliu: I did a local run, it crashed here.

sharktuner/sharktuner/candidate_ordering.py
```diff
@@ -104,3 +109,86 @@ def reorder_assignments(
...
+                    for nk, nv in nested.items():
+                        key = f"{k}.{nk}"
+                        row[key] = nv
+                        if key not in headers:
+                            headers.append(key)
+                else:
+                    continue
+            else:
+                row[k] = v
+                if k not in headers and k != "knob":
```
bangtianliu: Problem: Hardcoded string.

sharktuner/dispatch_tuner/dispatch_tuner.py
```diff
@@ -159,3 +159,10 @@ def main() -> None:
             print(path_config.run_log.resolve())
         print("Check the summary in:")
         print(summary_log_file.resolve())
+
+        output_csv_name = f"tuning_{args.dispatch_file.stem}.csv"
```
bangtianliu: locally what I got is `tuning_mmt_dispatch.csv`, use a more informative name?

sharktuner/dispatch_tuner/dispatch_tuner.py
```diff
@@ -159,3 +159,10 @@ def main() -> None:
...
+        csv_path = Path(path_config.base_dir / output_csv_name)
+        libtuner.candidate_ordering.export_record_to_csv(
+            dispatch_tuner.tuning_records, csv_path
+        )
+        print(f"Wrote tuning records CSV: {csv_path}")
```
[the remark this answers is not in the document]
RattataKing: Users should just invoke dispatch_tuner in purpose of collecting data, since only one round tuning loop is needed for this

sharktuner/dispatch_tuner/dispatch_tuner.py
RattataKing: I think leave naming options to users is better, they can decide how to name the input .mlir file

sharktuner/sharktuner/candidate_ordering.py
```diff
@@ -104,3 +108,95 @@ def reorder_assignments(
...
+
+def init_tuning_records(
+    knobs: list[Optional[common.KnobAssignment]], sorted_order: list[int]
+) -> list[TuningRecord]:
+    tuning_records: list[TuningRecord] = []
+    tuning_records.append(
+        TuningRecord(gen_id=0, candidate_id=0, to_compile=True, to_benchmark=True)
+    )
+
+    for can_idx, gen_idx in enumerate(sorted_order, start=1):
```
kuhar: can you expand these variable names? I think `can_index` is the actual index while `gen_idx` is the original order form z3?

sharktuner/sharktuner/candidate_ordering.py
```diff
@@ -104,3 +108,95 @@ def reorder_assignments(
...
+
+    - Each record becomes one CSV row.
+    - Top-level attributes (e.g., `gen_id`, `benchmark_time_us`) appear as individual columns.
+    - Nested objects (e.g., `knob`) are flattened into columns like `knob.M`, `knob.tile_m`.
+
+    The original top-level attribute (e.g., 'knob') is removed once nesting is flattened.
+    """
+    rows = []
+    headers = []
+    unneeded_headers = []
```
kuhar: Can you add type hints?

sharktuner/sharktuner/libtuner.py
```diff
@@ -1292,6 +1305,17 @@ def benchmark(
         benchmark_time=benchmark_time,  # Only candidate benchmark has time limit.
     )
 
+    for res in candidate_results:
+        tuning_client.tuning_records[
+            res.candidate_id
+        ].benchmark_device_id = res.device_id
```
kuhar: why do we have to keep looking up the same candidate_id? Can we look it up and keep as a variable?

sharktuner/tests/candidate_ordering_test.py
```diff
@@ -152,3 +157,176 @@ def test_reorder_assignments(target_info: iree_gpu.TargetInfo) -> None:
...
+        gen_id=1,
+        candidate_id=3,
+        knob=sample_knobs[1],
+    )
+    sorted_order = [2, 0, 1]
+    tuning_records = candidate_ordering.init_tuning_records(sample_knobs, sorted_order)
+
+    expected = [tr0, tr1, tr2, tr3]
+
+    assert tuning_records == expected
```
kuhar:
```suggestion
    assert tuning_records == [tr0, tr1, tr2, tr3]
```

sharktuner/tests/candidate_ordering_test.py
```diff
@@ -152,3 +157,176 @@ def test_reorder_assignments(target_info: iree_gpu.TargetInfo) -> None:
...
+            "benchmark_queue_position": None,
+            "benchmark_rank_order": None,
+            "benchmark_speedup": None,
+            "benchmark_status": False,
+            "benchmark_time_us": None,
+            "candidate_id": 0,
+            "compile_status": False,
+            "gen_id": 0,
+            "to_benchmark": True,
+            "to_compile": True,
```
kuhar: why do rows contain key-value pairs? I'd expect them to have value only

sharktuner/sharktuner/candidate_ordering.py
```diff
@@ -104,3 +108,95 @@ def reorder_assignments(
...
+            gen_id=gen_idx,
+            candidate_id=can_idx,
+            knob=knobs[gen_idx],
+        )
+        tuning_records.append(tr)
+
+    return tuning_records
+
+
+def flatten_records(
```
kuhar: You can break this down further and create a function to flatten just one Tuning record into a dictionary and then process a list of dictionaries in this functions and create unified header

sharktuner/sharktuner/candidate_ordering.py
```diff
@@ -104,3 +108,95 @@ def reorder_assignments(
...
+    headers = []
+    unneeded_headers = []
+
+    for tuning_record in tuning_records:
+        row = {}
+        for attr, val in vars(tuning_record).items():
+            if hasattr(val, "__dict__"):
+                nested = vars(val)
+                if not nested:
+                    continue
```
kuhar: When does this happen? Why do we need this condition?

sharktuner/tests/candidate_ordering_test.py
RattataKing: it's csv rows

sharktuner/sharktuner/candidate_ordering.py
```diff
@@ -104,3 +108,80 @@ def reorder_assignments(
...
+    benchmark_speedup: Optional[float] = None
+    benchmark_rank_order: Optional[int] = None
+
+
+def build_tuning_records_from_order(
+    knobs: list[Optional[common.KnobAssignment]], sorted_order: list[int]
+) -> list[TuningRecord]:
+    tuning_records: list[TuningRecord] = []
+
+    for sorted_position, original_gen_index in enumerate(sorted_order, start=1):
```
bangtianliu: Maybe add a comment here to explain why it is 1-based, (exclude baseline?)

sharktuner/tests/candidate_ordering_test.py
```diff
@@ -152,3 +157,150 @@ def test_reorder_assignments(target_info: iree_gpu.TargetInfo) -> None:
...
+            "knob_tile_m": 64,
+            "knob_tile_n": 256,
+            "knob_wg_x": 256,
+            "knob_wg_y": 2,
+            "knob_wg_z": 1,
+            "to_benchmark": False,
+            "to_compile": True,
+        },
+        {
+            "baseline_benchmark_time_us": None,
```
bangtianliu: Could we remove these if they are same as default values

sharktuner/tests/candidate_ordering_test.py
```diff
@@ -152,3 +157,150 @@ def test_reorder_assignments(target_info: iree_gpu.TargetInfo) -> None:
...
+        candidate_id=2,
+        knob=sample_knobs[1],
+        to_benchmark=True,
+        benchmark_time_us=153.56,
+    )
+    sample_tuning_records = [tr1, tr2]
+
+    headers, rows = candidate_ordering.flatten_records(sample_tuning_records)
+
+    expected_headers = [
```
bangtianliu: Since you have the dataclass definition right there, you can derive the expected headers directly from it instead of hardcoding them

bangtianliu: The existing tests for sure can be simplified significantly.

RattataKing: The expected headers are not just field names of class `TuningRecord`. The simplest version to retrieve expected headers from data class is going to be something like:
```
    knob_attr_str = "knob"
    record_fields = [
        f.name for f in fields(candidate_ordering.TuningRecord)
    ]
    knob_idx = record_fields.index(knob_attr_str)
    knob_fields = [
        f"{knob_attr_str}_{f.name}"
        for f in fields(type(sample_knobs[0]))
    ]

    # The knob field is flatten to knob_field1, knob_field2 ...
    expected_headers = record_fields[:knob_idx] + knob_fields + record_fields[knob_idx+1:]
```
I think this reduces the confidence of the test, because it ends up validating my reimplementation of the same logic rather than the actual behavior of the function

sharktuner/sharktuner/libtuner.py
```diff
@@ -1315,6 +1334,17 @@ def benchmark(
         candidate_results,
         prune_slow_candidates=tuning_client.should_prune_slower_candidates(),
     )
+    if all_candidates_with_speedup:
+        for i, handler_res in enumerate(all_candidates_with_speedup, start=1):
```
kuhar: Do we need the if condition? I think if the list is empty, the for loop won't execute?

kuhar (approved):

PR #2670, opened 2025-11-13
[boo tuner] Switch to newer BOO backend

rkayaith:
This switches the BOO tuner to use the newer `iree_boo_experimental` BOO backend, which uses `torch.compile` under the hood.

### DIFF
--- a/sharktuner/boo_tuner/boo_tuner.py
+++ b/sharktuner/boo_tuner/boo_tuner.py
@@ -275,6 +275,10 @@ def process_boo_command(
 
     # Run BOO compilation and extract source IR.
     with boo_runtime.use_cache_dir(boo_cache_dir):
+        # The "iree_boo" backend offloads to IREE in cases where we expect
+        # performance to be better, and falls back to pytorch otherwise. We use
+        # the experimental backend here instead, as we want to use IREE in all
+        # cases.
         # Note: device="cuda" is correct for AMD GPUs.
         sig.get_compiled_module(backend="iree_boo_experimental")(
             *sig.get_sample_args(device="cuda", seed=123)


sharktuner/boo_tuner/boo_tuner.py
```diff
@@ -347,20 +352,12 @@ def process_boo_command(
     return args.output_td_spec if best_spec_path else None
 
 
-def load_boo() -> tuple[types.ModuleType, Callable, type]:
```
rkayaith: Returning modules/functions in this way disables type checking. I've inlined the imports into the function that uses them.

sharktuner/boo_tuner/example_configs.txt
```diff
@@ -1,2 +1,2 @@
-convbfp16 -n 3 -c 2016 -H 1 -W 1 -k 224 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC -m conv -g 1 -F 1 -t 1
+convbfp16 -n 3 -c 2016 -H 4 -W 4 -k 224 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 --in_layout NHWC --fil_layout NHWC --out_layout NHWC -m conv -g 1 -F 1 -t 1
```
rkayaith: a slightly more realistic conv config -- a 1x1 image size doesn't exercise the layout-related code paths properly.

bangtianliu:
> should we also update the readme corresponding to using new backend?

ok, it is used inside the code, which will not affect how we run the tuner. So maybe just add code comment there about `iree_boo_experimental` backend.

bangtianliu (approved):
Thanks!

PR #2671, opened 2025-11-14
[boo tuner] Reset compilation and allocation caches for each config

rkayaith:
This ports over all the cache resetting that the BOO driver does: https://github.com/iree-org/iree-turbine/blob/51a22c97945b049fc79816f38108b5a8f12d2610/iree/turbine/kernel/boo/driver/driver.py#L269-L283

### DIFF
--- a/sharktuner/boo_tuner/boo_tuner.py
+++ b/sharktuner/boo_tuner/boo_tuner.py
@@ -6,6 +6,7 @@
 
 import argparse
 import contextlib
+import gc
 import logging
 import os
 import shlex
@@ -14,7 +15,6 @@ import subprocess
 import sys
 import tempfile
 import traceback
-import types
 from collections.abc import Callable
 from datetime import datetime
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


sharktuner/boo_tuner/boo_tuner.py
```diff
@@ -252,9 +252,10 @@ def process_boo_command(
     """Process a single BOO command through compilation and tuning."""
     # These imports are slow due to a pytorch dependency. Keeping them local helps
     # make '--help' fast.
-    from iree.turbine.kernel.boo.driver.launch import get_launchable
+    import torch  # type: ignore
```
rkayaith: pytorch isn't installed in CI, so this is needed for type checking to pass

rkayaith:
> Maybe add a link to that original iree-turbine PR addressing issues in driver end in the PR description?

The reset logic was added to the driver across multiple PRs over time, so I've linked directly to the driver source instead.

PR #2683, opened 2025-11-20
[tuner] use igemm bindings

bangtianliu:
This PR integrates the new `get_igemm_generic_conv_details` Python binding (from [IREE PR #22598](https://github.com/iree-org/iree/pull/22598)) into the Tuner to enable robust convolution tuning.

## Motivation

The current convolution tuner relies on manual dimension inference and flattening, which can be fragile and layout-specific. The new IGEMM binding provides a principled way to:
- Extract accurate M/N/K dimensions for any convolution layout
- Get pre-computed flattened IGEMM loop bounds

## Implementation Plan

This PR is the **first step** in a three-phase rollout:

1. **Phase 1 (This PR)**: Ensure existing `conv_2d_nhwc_hwcf` tuning continues to work correctly using `igemm_details` from the binding
   - Integrate binding into parser and constraint generator
   - Make IGEMM usage pipeline-aware (`LLVMGPUTileAndFuse` only)
   - Add tests to validate the IGEMM transformation path

2. **Phase 2**: Add padding support for unaligned convolution cases

3. **Phase 3**: Validate and extend support for other layouts.

### DIFF
--- a/amdsharktuner/tests/constraint_generator_test.py
+++ b/amdsharktuner/tests/constraint_generator_test.py
@@ -391,9 +391,6 @@ def test_adjust_problem_size_for_pipeline(
         k=[3],
         batch=[0],
     )
-    tile_and_fuse_pipeline = (
-        iree_codegen.DispatchLoweringPassPipeline.LLVMGPUTileAndFuse
-    )
     pipeline_options_space = dispatch_constraints.PipelineOptionsSearchSpace(
         prefetch_shared_memory=[True],
         no_reduce_shared_memory_bank_conflicts=[True, False],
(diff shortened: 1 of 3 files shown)


amdsharktuner/tests/constraint_generator_test.py
```diff
@@ -391,7 +391,9 @@ def test_adjust_problem_size_for_pipeline(
         k=[3],
         batch=[0],
     )
-    taf_pipeline = iree_codegen.DispatchLoweringPassPipeline.LLVMGPUTileAndFuse
+    tile_and_fuse_pipeline = (
+        iree_codegen.DispatchLoweringPassPipeline.LLVMGPUTileAndFuse
+    )
```
kuhar: Can we use this constant directly instead of creating a local variable for it? I think it only hurts readability here

amdsharktuner/tests/constraint_generator_test.py
```diff
@@ -438,8 +440,104 @@ def test_adjust_problem_size_for_pipeline(
...
+            m=list(conv_op_info.dims.m),
+            n=list(conv_op_info.dims.n),
+            k=list(conv_op_info.dims.k),
+            batch=list(conv_op_info.dims.batch),
+        )
+        conv_size = common.ContractionSizes(
+            M=list(conv_op_info.matmul_size.M),
+            N=list(conv_op_info.matmul_size.N),
+            K=list(conv_op_info.matmul_size.K),
+            B=list(conv_op_info.matmul_size.B),
```
kuhar: Would it help to assert that what the returned dimensions and sizes are? I think it won't ever change, so we don't have to worry about these getting out of sync with the compiler.

Max191 (approved):
Nice! LGTM

Issue #2687, opened 2025-11-22
IGEMM Conv Support in Tuner


bangtianliu:
## Motivation: Expanding Tuner Support Beyond NHWC_HWCF Layout

Currently, the tuner only supports the NHWC_HWCF layout. However, there is demand from the BOO to support additional layouts. IREE already can handle more layouts through the TileAndFuse pipeline with the IGEMM option (via `setIGEMMConvolutionLoweringConfig`).

To extend tuner support, we will proceed with the following approach:

## Implementation Plan

- Expose the necessary Python bindings to leverage IREE and LLVM capabilities
   https://github.com/iree-org/iree/pull/22598
   https://github.com/llvm/llvm-project/pull/167587
- Ensure that the existing convolution tuner integrates properly with the new Python bindings
   https://github.com/nod-ai/shark-ai/pull/2683
- Add padding support for unaligned cases (note: there are two existing attributes, "padding" and "padding_conv", that need to be reconciled)
- Implement support for additional layout configurations
- Refactor and remove legacy code specific to single-layout implementations

bangtianliu:
cc @kuhar @Max191

PR #2692, opened 2025-11-25
[tuner] Sync Padding for TileAndFuse with IREE changes

bangtianliu:
**Context**: The current padding support along the TileAndFuse pipeline was added in #1310. That implementation rounded problem sizes to the nearest multiple of the intrinsic size during the constraint generation phase.

This PR revisits and syncs the padding logic for the TileAndFuse pipeline to properly align with IREE's implementation for both matmul and conv operations, while keeping the padding decision logic separate from constraint generation.

Issue: #2687

### DIFF
--- a/amdsharktuner/amdsharktuner/constraint_generator.py
+++ b/amdsharktuner/amdsharktuner/constraint_generator.py
@@ -93,7 +93,7 @@ def generate_generic_contraction_solutions(
     )
 
     # Apply padding for TileAndFuse pipeline to get better tile sizes.
-    padding_applied = False
+    overpadding_applied = False
     if codegen_pipeline == iree_codegen.DispatchLoweringPassPipeline.LLVMGPUTileAndFuse:
         # Use IGEMM maps if available (dimensions were restructured), otherwise use original indexing maps.
         padding_maps = indexing_maps
@@ -105,13 +105,15 @@ def generate_generic_contraction_solutions(
         (
             matmul_size.M,
             matmul_size.N,
-            padding_applied,
+            overpadding_applied,
         ) = common.calculate_padded_dimensions(
             matmul_size.M, matmul_size.N, contraction_dims, padding_maps
         )
 
     M, N, K = matmul_size.M, matmul_size.N, matmul_size.K
-    tuner_ctx.logger.debug(f"M={M}, N={N}, K={K}, padding_applied={padding_applied}")
+    tuner_ctx.logger.debug(
+        f"M={M}, N={N}, K={K}, overpadding_applied={overpadding_applied}"
+    )
 
     m_vars = [z3.Int(f"m{i}") for i in range(len(M))]
     n_vars = [z3.Int(f"n{i}") for i in range(len(N))]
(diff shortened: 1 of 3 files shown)


amdsharktuner/amdsharktuner/common.py
```diff
@@ -523,3 +523,60 @@ def get_target_info(input_module: ir.Module) -> iree_gpu.TargetInfo:
...
+            continue
+
+        if dim > 128:
+            padded, was_padded = maybe_padded_bounds(dim, 128)
+            result.append(padded)
+            any_padding_applied = any_padding_applied or was_padded
+        elif dim > 32:
+            padded, was_padded = maybe_padded_bounds(dim, 32)
+            result.append(padded)
+            any_padding_applied = any_padding_applied or was_padded
```
Max191: FYI, this logic is to allow over-padding to get larger tile sizes, which may result in better performance despite doing more padded computation. We may want to try making this overpadding a tunable parameter eventually. Can you add a TODO comment?

I wouldn't do it right now because it may blow up the tuning space, but we should be getting better candidate ordering soon, which will help with that, so the TODO comment is good for now.

amdsharktuner/amdsharktuner/common.py
```diff
@@ -523,3 +523,62 @@ def get_target_info(input_module: ir.Module) -> iree_gpu.TargetInfo:
...
+    remainder = original_bound % alignment
+    if remainder == 0:
+        return original_bound, False
+    return original_bound + alignment - remainder, True
+
+
+def get_dim_bounds(
+    dims: list[int],
+    padding_can_be_expensive: bool,
+) -> tuple[list[int], bool]:
```
kuhar: Also here: what does this do?

kuhar (commented):
Is this something that we would rather keep entirely in IREE and expose as new bindings?

bangtianliu:
> Is this something that we would rather keep entirely in IREE and expose as new bindings?

All of these functions are actually lightweight lambda functions. If we want to expose them, we would first need to rewrite them as proper utility functions and then expose those. Also, these functions are quite simple, and they just compute padding based on alignment. So it might be better to stick with the current approach and implement this logic directly on the Python side.

Max191 (changes requested):

amdsharktuner/amdsharktuner/constraint_generator.py
```diff
@@ -238,23 +257,20 @@ def set_cdim_tile_sizes(tile_sizes, contraction_dims, csizes):
             [lookup(v) for v in k_vars],
         )
 
-        required_padding = any(
-            p[-1] % i != 0 for p, i in zip((M, N, K), intrinsic_mnk_shape, strict=True)
-        )
```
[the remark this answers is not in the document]
bangtianliu: The current changes in this PR follow the existing logic in KernelConfig.

I did notice that small-dimension cases aren’t producing any solutions. If we want to support those including the transposed and small case, I can update the logic and add corresponding tests.

PR #2737, opened 2025-12-19
[sharktank] Read num hidden layers from config

jtuyls:
A config file can contain the number of hidden layers, so we can retrieve them if available.


jtuyls:
Could someone help review? @dan-garvey @KyleHerndon

PR #2743, opened 2025-12-22
Support for candidate generation and tuning attention

keshavvinayak01:
- Add support for tuning (SDPA) operations, ensuring `vector_distribute` is used when the dispatch is attention.
- Also relaxed the MMA intrinsics compatibility check to allow different accumulator/result dtypes.
- Added denormal flush support as a tunable option.

### DIFF
--- a/amdsharktuner/amdsharktuner/rocm/rocm_dispatch_constraints.py
+++ b/amdsharktuner/amdsharktuner/rocm/rocm_dispatch_constraints.py
@@ -728,7 +728,7 @@ def _build_compilation_infos(
         iree_codegen.DispatchLoweringPassPipelineAttr.get(codegen_pipeline)
     )
     pipeline_options_list: list[
-        iree_gpu.PipelineOptionsAttr
+        tuple[iree_gpu.PipelineOptionsAttr, bool]
     ] = generate_allowed_pipeline_options(pipeline_options_search_space)
     wg_x, wg_y, wg_z = workgroup_sizes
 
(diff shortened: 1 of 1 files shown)


Groverkss:
Nice! I like the changes. Will redirect to @bangtianliu for review.

amdsharktuner/amdsharktuner/common.py
```diff
@@ -296,21 +296,33 @@ def is_compatible(
...
+            if str(res_type.element_type) == 'bf16' or str(res_type.element_type) == 'f32':
+                return True
+        
+        # For f16 inputs with f32 accumulator, allow f16 result (hardware can cast)
+        if str(a_type) == 'f16' and str(b_type) == 'f16' and str(c_type) == 'f32':
+            if str(res_type.element_type) == 'f16' or str(res_type.element_type) == 'f32':
+                return True
+        
+        # Otherwise, result type must match accumulator type
+        return res_type.element_type == c_type
```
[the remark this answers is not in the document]
bangtianliu: Also, could you also add a test for this new type support.

amdsharktuner/amdsharktuner/dispatch_constraints.py
```diff
@@ -660,24 +669,37 @@ class PipelineOptionsSearchSpace:
         default_factory=lambda: [None]
     )
     use_igemm_convolution: list[Optional[bool]] = field(default_factory=lambda: [None])
+    denorm_flushing: list[Optional[bool]] = field(default_factory=lambda: [True])
```
bangtianliu: Should it be False by default?

Also we do we need Optional[bool], my understanding is that it is enough to use bool as a button to turn on the denorm_flusing,  unless you need to distinguish between "explicitly disabled" vs "not specified".

amdsharktuner/amdsharktuner/libtuner.py
```diff
@@ -702,8 +702,8 @@ def run_iree_benchmark_module_command(benchmark_pack: BenchmarkPack):
         )
 
     mean_benchmark_time = sum(times) / float(len(times))
-    logging.debug(
-        f"Benchmark time of candidate {candidate_id}: {mean_benchmark_time:.2f} us"
+    logging.info(
+        f"Candidate {candidate_id:3d}: {mean_benchmark_time:10.2f} us  (device: {device_id})"
```
bangtianliu: Changing benchmark output from debug to info level adds verbosity, this could flood summary.log. We should keep it at debug still.

bangtianliu (changes requested):

PR #2744, opened 2025-12-22
[tuner] ensure correct number of TD specs are generated

bangtianliu:
## Problem 

When specifying `--num-candidates=30`, only 30 TD specs were generated instead of the expected 31 (1 baseline + 30 candidates).

## Root Cause

The knob_assignments list contained only 30 entries (one per candidate solution), but config_specs contained 31 entries (baseline + 30 candidates). When zipping these lists together:

```python
for candidate_num, (spec, knob) in enumerate(zip(config_specs, knob_assignments)):
```
The zip() function stopped after 30 iterations (the length of the shorter list), causing the 31st spec (the last candidate) to be silently dropped.

## Solution

This PR ensures `knob_assignments` has the same length as `config_specs` by explicitly including `None` for the baseline at index 0

### DIFF
--- a/amdsharktuner/amdsharktuner/libtuner.py
+++ b/amdsharktuner/amdsharktuner/libtuner.py
@@ -939,6 +939,7 @@ def generate_candidate_specs(
             candidate_ordering.build_tuning_records_from_order(knobs, sorted_order)
         )
 
+        # Prepend None for the baseline config (no knob assignment) at index 0.
         knob_assignments = [None] + [
             dispatch_tuner.get_knob_assignment(s) for s in solutions
         ]
--- a/amdsharktuner/tests/libtuner_test.py
+++ b/amdsharktuner/tests/libtuner_test.py
@@ -425,28 +425,3 @@ def test_compute_rocprof_avg_kernel_time(caplog):
     trace_rows = [drop_row] * 10 + [cal_row] * 5 + [cal_row_2] * 5
     avg_us = libtuner.compute_rocprof_avg_kernel_time(trace_rows)
     assert avg_us == pytest.approx(1.75)
-
-
-def test_knob_assignments_with_baseline():
-    """Test that knob assignments list is correctly constructed with baseline at index 0."""
-
-    knob1 = {"tile": 64}
-    knob2 = {"tile": 128}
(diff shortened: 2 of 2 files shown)


amdsharktuner/amdsharktuner/libtuner.py
```diff
@@ -939,7 +952,10 @@ def generate_candidate_specs(
             candidate_ordering.build_tuning_records_from_order(knobs, sorted_order)
         )
 
-        knob_assignments = [dispatch_tuner.get_knob_assignment(s) for s in solutions]
+        knob_assignments = build_knob_assignments_with_baseline(
+            dispatch_tuner, solutions
+        )
+        assert len(config_specs) == len(knob_assignments)
```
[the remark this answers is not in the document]
bangtianliu: The helper function was originally added to make testing easier, but I agree it can be simplified to just inline code.

amdsharktuner/amdsharktuner/libtuner.py
```diff
@@ -939,7 +939,10 @@ def generate_candidate_specs(
             candidate_ordering.build_tuning_records_from_order(knobs, sorted_order)
         )
 
-        knob_assignments = [dispatch_tuner.get_knob_assignment(s) for s in solutions]
+        knob_assignments = [None] + [
```
kuhar: Can you add a one-line comment explaining the None element is for the baseline?

amdsharktuner/tests/libtuner_test.py
```diff
@@ -425,3 +425,28 @@ def test_compute_rocprof_avg_kernel_time(caplog):
...
+
+    knob_assignments = [None] + [
+        dispatch_tuner.get_knob_assignment(s) for s in solutions
+    ]
+    assert knob_assignments == [None, knob1, knob2, knob3]
+
+    empty_knob_assignments = [None] + [
+        dispatch_tuner.get_knob_assignment(s) for s in []
+    ]
+    assert empty_knob_assignments == [None]
```
kuhar: This doesn't test anything

bangtianliu: In this test, I just replicated what I did in the tuner. This test can be removed.

PR #2747, opened 2025-12-26
[tuner] use virtual mma only for attention op

bangtianliu:
While investigating a low compilation ratio issue in the tuner, I discovered a discrepancy between IREE and the tuner: 
1. IREE: Uses virtual MMA intrinsics only for attention ops.
2. Tuner: Was searching virtual MMA intrinsics for all ops.

This PR addresses the discrepancy by adding an allow_virtual_mma flag to `get_compatible_mma_intrinsics`, which defaults to False and filters out virtual intrinsics for non-attention ops.

Results: After this change, the compilation ratio for matmul (the example from readme) along the TileAndFuse pipeline improved from 17/30 to 30/30.

### DIFF
--- a/amdsharktuner/amdsharktuner/dispatch_constraints.py
+++ b/amdsharktuner/amdsharktuner/dispatch_constraints.py
@@ -63,7 +63,7 @@ def match_layout(
     )
 
 
-def get_mfma_intrinsic_constraints(
+def get_mma_intrinsic_constraints(
     lhs_type: common.ShapedType,
     rhs_type: common.ShapedType,
     res_type: common.ShapedType,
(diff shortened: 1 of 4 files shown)


amdsharktuner/amdsharktuner/dispatch_constraints.py
```diff
@@ -74,10 +74,18 @@ def get_mfma_intrinsic_constraints(
...
     compatible_intrinsics = common.get_compatible_mfma_intrinsics(
         lhs_type, rhs_type, res_type, mma_intrinsics
     )
+    # Filter out virtual intrinsics unless explicitly allowed (e.g., for attention ops).
+    if not allow_virtual_mma:
+        compatible_intrinsics = [
+            instr
+            for instr in compatible_intrinsics
+            if isinstance(instr, iree_gpu.MMAIntrinsic)
+        ]
```
kuhar:
```suggestion
        compatible_intrinsics = filter(lambda x: isinstance(x, iree_gpu.MMAIntrinsic), compatible_intrinsics)
```

kuhar:
> Why not handle this in `get_compatible_mfma_intrinsics`? It should be easier to test.

also, we should rename it to `get_compatible_mma_intrinsics`, since we use wmma on gfx11+.

amdsharktuner/tests/dispatch_constraints_test.py
```diff
@@ -408,6 +408,34 @@ def test_get_mfma_intrinsic_constraints(
...
+        rhs_type=rhs_type,
+        res_type=res_type,
+        intrinsic_m=intrinsic_m,
+        intrinsic_n=intrinsic_n,
+        intrinsic_k=intrinsic_k,
+        mma_intrinsics=[
+            iree_gpu.MMAIntrinsic.MFMA_F32_16x16x16_F16,
+            iree_gpu.VirtualMMAIntrinsic.VMFMA_F32_32x32x16_F8E4M3FNUZ,
+        ],
+        allow_virtual_mma=True,
```
kuhar: Do we have any tests that exercises `allow_virtual_mma=False`? IE, does anything fail if you always pass `allow_virtual_mma=False`?

Issue #2340, opened 2025-09-25
[tuner] Parallelize candidate generation


kuhar:
On some dispatches, I've seen candidate generation take close to 30 seconds. We should be able to drastically speed this up by parallelizing candidate generation.

We can do that by introducing 'meta-constraints' that divide the search space. For example. we can have a meta-constraints for the exact MFMA intrinsic or for the compilation pipeline. We should make sure that whatever constraints we pick they should try to divide the search space equally. A counterexample to that would be saying that one set of solver invocations handles tile sizes <= 64 and the other >64 -- we might have much more candidates in the latter.

This will give us `n` disjoint partitions over the search space that can be solved by `n` parallel solver invocations, one per thread.

PR #2756, opened 2026-01-07
[Tuner] Add multiprocessing to candidate_gen via divided constraint sets

RattataKing:
This PR tackles this issue: #2340 , where the previous Tuner could take ~30s during candidate generation. By splitting search space into multiple constraint sets and adding multiprocessing to it can speed up this phase.

In `generate_generic_contraction_solutions()`, constraints are now divided based on available mma intrinsic layouts. Each child process will work on the independent constraint set.

### DIFF
--- a/amdsharktuner/tests/constraint_generator_test.py
+++ b/amdsharktuner/tests/constraint_generator_test.py
@@ -681,7 +681,7 @@ def test_ContractionZ3Constants_to_meta() -> None:
 
 
 def test_ContractionZ3Constants_from_meta_dict() -> None:
-    meta = {
+    meta: dict[str, str | list[str]] = {
         "m_vals": ["m0", "m1"],
         "n_vals": ["n0"],
         "k_vals": ["k0"],


amdsharktuner/amdsharktuner/libtuner.py
```diff
@@ -393,6 +393,15 @@ def parse_arguments(
...
     )
+    candidate_gen_args.add_argument(
+        "--enable-parallel-generation",
+        type=bool,
+        default=True,
+        help=(
+            "Enable multiprocessing during candidate generation to speed up the process."
+            "Currently supported only for contraction and convolution."
+        ),
+    )
```
bangtianliu: I donot think this is correct. 

argparse with type=bool converts the string "False" using bool("False"), which returns True (any non-empty string is truthy in Python). 

You can follow this one
https://github.com/nod-ai/amd-shark-ai/blob/d47ff9c0bc833b7a39104da456254dee024760bf/amdsharktuner/amdsharktuner/libtuner.py#L391-L395

bangtianliu (changes requested):
Nice work! Left three comments from my initial review. I'll take a closer look tomorrow.

amdsharktuner/tests/constraint_generator_test.py
```diff
@@ -686,3 +706,65 @@ def test_get_z3_solutions() -> None:
     )
     results = list(constraint_generator.get_z3_solutions(z3_constraint_set))
     assert results == []
+
+
+def test_generate_solutions_with_multiprocessing(
+    tuner_ctx: common.TunerContext, gpu_target_info: iree_gpu.TargetInfo
+) -> None:
+    # Use "spawn" to avoid fork() warnings in pytest.
+    constraint_generator.process_utils.multiprocessing.set_start_method("spawn")
```
bangtianliu: CI errors here

amdsharktuner/amdsharktuner/constraint_generator.py
```diff
@@ -341,9 +424,36 @@ def generate_generic_contraction_solutions(
...
+        ]
+        tuner_ctx.logger.debug(
+            f"Will generate [{len(constraint_payload_list)}] constraint solvers."
+        )
+        print(
+            f"Running {len(constraint_payload_list)} worker processes (one per mega-constraint set)"
+        )
+        executor = process_utils.MultiprocessExecutor(
+            num_workers=len(constraint_payload_list),
+        )
```
bangtianliu: When there's only 1 MMA intrinsic, multiprocessing adds overhead with no benefit.
Adding a warning/fallback when len(constraint_payload_list) == 1

Consider (not belong to this current PR): splitting by additional dimensions to enable parallelism even with a single MMA intrinsic
- Tile size range/Subgroup count combination/ Workgroup Size

kuhar: If the overhead is negligible, I'd leave this as-is to keep the implementation simple

amdsharktuner/tests/constraint_generator_test.py
```diff
@@ -22,7 +22,10 @@
     dispatch_parser,
 )
 
-from amdsharktuner.test_utils import tuner_ctx
+from amdsharktuner.test_utils import tuner_ctx, configure_pytest_multiprocessing
+
+
+configure_pytest_multiprocessing()
```
kuhar: I think the standard way to do this is to use one of the pytest hooks like https://docs.pytest.org/en/latest/reference/reference.html#pytest.hookspec.pytest_sessionstart

amdsharktuner/tests/conftest.py
```diff
@@ -1,4 +1,10 @@
-from __future__ import annotations
+# Copyright 2025 Advanced Micro Devices, Inc.
```
bangtianliu: 2026 now, I made the same mistake in another PR lol

amdsharktuner/amdsharktuner/candidate_gen.py
```diff
@@ -247,6 +247,7 @@ def generate_solutions(
     allowed_waves_per_eu: list[int] = [2],
     pipeline_options_search_space: dispatch_constraints.PipelineOptionsSearchSpace = dispatch_constraints.PipelineOptionsSearchSpace(),
     codegen_pipeline: iree_codegen.DispatchLoweringPassPipeline = iree_codegen.DispatchLoweringPassPipeline.LLVMGPUVectorDistribute,
+    enable_multiprocessing: bool = False,
```
bangtianliu: > enable this unconditionally without extra CLI flags

Maybe we should change the default to `True`.  Or this parameter could be removed entirely if multiprocessing should always be on.

amdsharktuner/amdsharktuner/libtuner.py
```diff
@@ -917,8 +917,8 @@ def generate_candidate_specs(
         solutions = list(solutions_iter)
         solution_gen_end_time = time.perf_counter()
         elapsed_time = solution_gen_end_time - solution_gen_start_time
-        logging.debug(f"Completed candidate generation in {elapsed_time:.6f}s\n")
-        logging.debug(f"Max search space size: {len(solutions)}")
+        logging.critical(f"Completed candidate generation in {elapsed_time:.6f}s\n")
+        logging.critical(f"Max search space size: {len(solutions)}")
```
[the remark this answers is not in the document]
RattataKing: Changed to level `INFO` and `DEBUG`, and set logger to print level INFO and above to the console

amdsharktuner/amdsharktuner/candidate_gen.py
```diff
@@ -247,9 +247,12 @@ def generate_solutions(
     allowed_waves_per_eu: list[int] = [2],
     pipeline_options_search_space: dispatch_constraints.PipelineOptionsSearchSpace = dispatch_constraints.PipelineOptionsSearchSpace(),
     codegen_pipeline: iree_codegen.DispatchLoweringPassPipeline = iree_codegen.DispatchLoweringPassPipeline.LLVMGPUVectorDistribute,
+    enable_multiprocessing: bool = True,
```
kuhar: can we drop these too and always enable multiporocessing?

amdsharktuner/amdsharktuner/libtuner.py
```diff
@@ -665,7 +665,7 @@ def run_iree_benchmark_module_command(benchmark_pack: BenchmarkPack):
             **extra_flags,
         )
     except ireert.benchmark.BenchmarkTimeoutError as e:
-        logging.info(
+        logging.debug(
```
kuhar: Can you move all these logging changes to a separate PR?

amdsharktuner/tests/conftest.py
```diff
@@ -0,0 +1,14 @@
...
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+
+from typing import Any
+
+from amdsharktuner.test_utils import configure_pytest_multiprocessing
+
+
+def pytest_sessionstart(session: Any) -> None:
+    configure_pytest_multiprocessing()
```
kuhar: Do we still need this helper function? Maybe we can do it here? Also, doesn't this ensure that `set_start_method` is set exactly once?

amdsharktuner/amdsharktuner/candidate_gen.py
RattataKing: Do you want to entirely remove serialization processing part? i.e. 
```
if enable_multiprocessing:
  ...
else:
  z3_solutions_iter = get_z3_solutions(constraints_list[0])
```

PR #2434, opened 2025-10-06
[Tuner] Add `SolutionTrace` to track z3 feature vars in candidate generation

RattataKing:
common.py:
- Added `SolutionTrace`, to track variable values that generated from z3 constraints; the contraction type is `ContractionSolutionTrace`.
- Modified `TuningConfiguration`, added a new attr to store `SolutionTrace`.
- Added `CandidateProfile`, to pass needed candidate info from candidate_gen.py to libtuner.py.

constraints_generator.py:
- Modified `generate_generic_contraction_solutions()`, attach `SolutionTrace` to `TuningConfiguration` when return.
- Modified `DispatchTuner`, added new methods

candidate_gen.py:
- Modified `generate_configs_and_td_specs()`, which retrieves `SolutionTrace` from `TuningConfiguration` and attaches to `CandidateProfile`, and returns as a list to libtuner.py.

libtuner.py:
- Modified `TuningClient`, to track `dispatch_kind` during tuning for future flexibility.

PR #2760, opened 2026-01-09
[Tuner] Fix tuning_records bugs

RattataKing:
### Problem
Set flag `--num-candidates` with a value >= max number of candidates in entire search space, e.g. SMT solver has total 1265 solutions for `dispatch_sample.mlir`, invoke tuner with `--num-candidates=4096` will crash at `def compile()` phase with `IndexError: list index out of range`.

### Root Cause
Issue is caused by #2434 and #2744 :
In file `candidate_ordering.py`, func `build_tuning_records_from_order()` is designed to create the `tuning_records` without baseline entry.
```
len(tuning_client.tuning_records) = all candidates - baseline
```
PR #2434 unintentionally dropped the last candidates, masked the index issue.

### Solution
Add baseline entry to `tuning_records`, match list index to the candidate id, and skip the baseline entry when converting to csv.


kuhar (approved):
Thanks

bangtianliu (approved):

PR #2831, opened 2026-02-20
[tuner] Fix BOO tuner logging issues

bangtianliu:
This PR makes two changes for BOO Tuner: 
- Set TURBINE_LOG_LEVEL=WARNING to prevent torch modules/MLIR IR from cluttering console output
- Add finally block to prevent handler accumulation across benchmarks

Assisted-by:  [Claude Code](https://claude.ai/code)


rkayaith (approved):

kuhar (approved):
OK, testing this seems to require a lot of heroics -- let's drop the test and land the previous iteration.

PR #2838, opened 2026-02-26
Add support for tuning SDPA/attention operations with VectorDistribute

keshavvinayak01:
Attention dispatches would silently fail instead of selecting the Vector Distribute pipeline. This change fixes that behavior.


amdsharktuner/amdsharktuner/libtuner.py
```diff
@@ -776,20 +776,42 @@ def generate_candidate_specs(
...
+            get_iree_codegen_pipeline(codegen_pipeline),
         )
+        # Attention only supports VectorDistribute, so always include
+        # VectorDistribute tuners to allow attention ops to be matched.
+        if codegen_pipeline != CodegenPipelines.llvmgpu_vector_distribute:
+            vd_tuners = candidate_gen.get_supported_dispatch_tuners(
+                tuning_client.target_info.arch,
+                get_iree_codegen_pipeline(CodegenPipelines.llvmgpu_vector_distribute),
+            )
+            dispatch_tuners += [t for t in vd_tuners if t not in dispatch_tuners]
```
bangtianliu: Silently overriding the user's requested pipeline is surprising behavior. The user asked for TileAndFuse but the tuner quietly switches to VectorDistribute — that's a hidden    
  side effect that can cause confusion when debugging results.
                                                                                                                                                                                           
 Consider: if the dispatch is attention and the user didn't select VectorDistribute, warn them and skip (or error), rather than secretly changing the pipeline.

bangtianliu (changes requested):
No tests

PR #2839, opened 2026-02-26
Relax MMA intrinsic compatibility to allow flexible accumulator/result dtypes

keshavvinayak01:
Relaxing the MMA intrinsics compatibility check allows different accumulator/result dtypes, which is essential to making attention ops compile, otherwise failing with "unsupported MMA" errors.


bangtianliu (changes requested):
Good, we have tests for this PR.

PR #2840, opened 2026-02-26
Add denormal flush support as a tunable pipeline option

keshavvinayak01:
Adding support for denorm_flushing in the pipeline.
 _Might be_ conducive to better performance, as empirically tested for tuned attention benchmarks locally.

### DIFF
--- a/.github/workflows/ci-amdsharktank.yml
+++ b/.github/workflows/ci-amdsharktank.yml
@@ -74,6 +74,7 @@ jobs:
             -e amdsharktank/
           git clone https://github.com/iree-org/wave.git
           cd wave
+          git checkout 775987d38c7f7e63555fdecf000272edaf862a6a
           curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
           export PATH="$HOME/.cargo/bin:$PATH"
 
--- a/amdsharktank/build_tools/install_test_dependencies.sh
+++ b/amdsharktank/build_tools/install_test_dependencies.sh
@@ -56,6 +56,7 @@ if (($IREE_UNPINNED)); then
     pip uninstall -y wave-lang
     git clone https://github.com/iree-org/wave.git
     cd wave
+    git checkout 775987d38c7f7e63555fdecf000272edaf862a6a
     curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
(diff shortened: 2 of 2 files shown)


amdsharktuner/amdsharktuner/rocm/rocm_common.py
```diff
@@ -124,12 +127,20 @@ def get_translation_info_config(
...
-        }
-    )
+    config_dict_entries: dict[str, ir.Attribute] = {
+        common.GPU_PIPELINE_OPTIONS_KEY: pipeline_options,
+        common.LLVM_FUNC_ATTRS_KEY: waves_per_eu_dict,
+    }
+
+    # Add denormal_fp_math_f32 attribute if denorm_flushing is specified.
+    # When denorm_flushing is True, use "preserve-sign" to flush denormals to zero.
+    if denorm_flushing:
```
bangtianliu: Instead of mannual parse, we should expose python binding for this.  Not for blocking, at least add a TODO comment here.

amdsharktuner/amdsharktuner/rocm/rocm_dispatch_constraints.py
```diff
@@ -673,24 +673,35 @@ class PipelineOptionsSearchSpace:
...
 
 def generate_allowed_pipeline_options(
     pipeline_options_search_space: PipelineOptionsSearchSpace,
-) -> list[iree_gpu.PipelineOptionsAttr]:
-    pipeline_options_list = []
+) -> list[tuple[iree_gpu.PipelineOptionsAttr, bool]]:
+    """
+    Generate a list of (PipelineOptionsAttr, denorm_flushing) tuples.
+
+    denorm_flushing is passed separately since it's applied via llvm_func_attrs
```
bangtianliu: Docstring is wrong here
  Should say something like: "denorm_flushing is passed separately since it's a top-level translation info config attribute, not a PipelineOptionsAttr field."

bangtianliu:
In this PR, `allowed_denorm_flushing `is always `[False]` (the default value), so denorm flushing is never actually explored in the tuning space.

We already have similar support for `--prefetch-num-stages-options` and `--waves-per-eu-options`, so it would be good to follow the same pattern here, defaulting to False but allowing advanced users to provide all available options. 

Additionally, if denorm flushing is only applicable to attention ops, we should log a warning when it's specified for other op types.

bangtianliu (changes requested):

PR #2844, opened 2026-02-27
[tuner] add direct conv support along TileAndFuse pipeline

bangtianliu:
This PR adds support for direct convolution lowering in the TileAndFuse pipeline, complementing the existing IGEMM approach.
  
  Assisted-by: Claude Code


kuhar (commented):
I remember hearing/reading some update that direct conv through TAF doesn't bring improvements -- is this accurate or am I imagining things? @bangtianliu 

If it does not, do we still want to land it?

PR #2863, opened 2026-03-12
[tuner] Avoid textual IR for tuner tests

bangtianliu:
This PR consolidates MLIR builder functions into test_utils.py, replacing string templates with programmatic op construction.

Assisted-by: [Claude Code](https://claude.ai/code)
 
Issue: https://github.com/nod-ai/amd-shark-ai/issues/1291


kuhar (changes requested):
I'm not convinced this is the right direction -- textual IR is convenient in tests and probably easier to maintain that dedicated builders, since we can easily copy paste from IREE / MLIR lit tests. Whereas python builders and not always documented well and people who only work with the C++ implementation are generally not very familiar with those bindings.

The issues with textual IR is usually when the code is not exercised by any tests and may bitrot, but this is not a consideration here since these tests are always executed and we use nightly IREE builds -- if something breaks, we will immediately learn about it.

PR #2897, opened 2026-03-31
[tuner] Add col_major to MMA attrs and drop marker attributes for attention

bangtianliu:
To sync the changes from the IREE side https://github.com/iree-org/iree/pull/23633, which removes attention transpose intrinsic hacks.  Two PRs are required: 

- Update the Python binding (IREE PR): https://github.com/iree-org/iree/pull/23757 adds `col_major` parameter to `MMAAttr` and `VirtualMMAAttr` Python binding
- Corresponding changes in Tuner side (This Tuner PR):  sets `col_major=True` on MMA/VirtualMMA attrs when `QK` output layout matches `PV` input layout, and removes the now-unused `attention_qk_matmul` and `attention_pv_matmul` marker attributes from decomposition configs.

Assisted-by: [Claude Code](https://claude.ai/code)

### DIFF
--- a/amdsharktuner/amdsharktuner/rocm/rocm_dispatch_constraints.py
+++ b/amdsharktuner/amdsharktuner/rocm/rocm_dispatch_constraints.py
@@ -639,12 +639,13 @@ def getMMAAttr(
     lhs_type: ir.IntegerType | ir.FloatType,
     rhs_type: ir.IntegerType | ir.FloatType,
     mma_intrinsics: list[iree_gpu.MMAIntrinsic | iree_gpu.VirtualMMAIntrinsic],
+    col_major: bool = False,
 ) -> iree_gpu.MMAAttr | iree_gpu.VirtualMMAAttr:
     for mma_intrinsic in mma_intrinsics:
         if isinstance(mma_intrinsic, iree_gpu.MMAIntrinsic):
-            mma_attr = iree_gpu.MMAAttr.get(mma_intrinsic)
+            mma_attr = iree_gpu.MMAAttr.get(mma_intrinsic, col_major)
         else:
-            mma_attr = iree_gpu.VirtualMMAAttr.get(mma_intrinsic)
+            mma_attr = iree_gpu.VirtualMMAAttr.get(mma_intrinsic, col_major)
 
         a_type, b_type, c_type = mma_attr.abc_element_types
         mnk = mma_attr.mnk_shape
@@ -666,7 +667,7 @@ def getMMAAttr(
     raise ValueError(
         f"No matching MMA intrinsic found for "
         f"output_type={output_type}, lhs_type={lhs_type}, rhs_type={rhs_type}, "
-        f"m={m}, n={n}, k={k}."
+        f"m={m}, n={n}, k={k}, col_major={col_major}."
     )
 
 
--- a/amdsharktuner/amdsharktuner/rocm/rocm_dispatch_constraints.py
+++ b/amdsharktuner/amdsharktuner/rocm/rocm_dispatch_constraints.py
@@ -476,6 +476,7 @@ def generate_attention_vector_distribute_constraints(
     subgroup_m_count: z3.ArithRef,
     subgroup_n_count: z3.ArithRef,
     can_reuse_qk_output_for_pv_input: z3.BoolRef,
+    use_col_major: z3.BoolRef,
     gpu_target_info: iree_gpu.TargetInfo,
 ):
     m_tile, n_tile, k_tile = tile_sizes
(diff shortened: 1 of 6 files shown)


amdsharktuner/amdsharktuner/rocm/rocm_solutions.py
```diff
@@ -509,6 +509,11 @@ def generate_attention_solutions(
         def lookup(var):
             return model[var].as_long()
 
+        # Determine if layouts match to set col_major on MMA attrs.
+        # When QK output layout matches PV input layout, using col_major
+        # eliminates the conflict between chained MMA in attention.
+        layouts_match = bool(model[can_reuse_qk_output_for_pv_input])
```
RattataKing: I think this is incorrect, the `col_major` from upstream IREE is not a tunable parameter, it's a derived stamp checking whether `qk_output` layout can match to `pv_input` rhs layout.

I think it should be `layouts_match = bool(model[can_reuse_rhs])`

amdsharktuner/amdsharktuner/rocm/rocm_dispatch_constraints.py
RattataKing, line 556: Needs to update constraints a bit to match upstream, https://github.com/iree-org/iree/pull/23631/ drops `qk_acc == pv_lhs` entirely

RattataKing:
For the `col_major` config, if it's set incorrectly from tuner side, will it pass IREE and not fail at compilation or benchmarking phase, and end up generating wrong output secretly?

RattataKing (changes requested):

amdsharktuner/amdsharktuner/rocm/rocm_dispatch_constraints.py
bangtianliu: Yes, this IREE PR is not synced from the tuner side

bangtianliu:
> For the `col_major` config, if it's set incorrectly from tuner side, will it pass IREE and not fail at compilation or benchmarking phase, and end up generating wrong output secretly?

It will drop down the compilation ratio, refer to here:  https://github.com/nod-ai/amd-shark-ai/issues/2274 for more details. 

For the correctness part, I did not investigate too much. Probably yes, since reading data in a different layout. Adding a numerical correctness check should be another TODO for the tuner.

amdsharktuner/amdsharktuner/rocm/rocm_solutions.py
bangtianliu: Yes, you are right here.

amdsharktuner/amdsharktuner/rocm/rocm_solutions.py
```diff
@@ -591,7 +602,7 @@ def lookup(var):
 
         workgroup_size = lookup(sg_m_cnt) * lookup(sg_n_cnt) * lookup(subgroup_size)
 
-        # Set prefetch_num_stages based on whether layouts match.
+        # Set prefetch_num_stages based on whether QK output can be reused.
         # 0/1 = disable prefetching, 2 = two-stage pipeline (default),
         # 3 = three-stage pipeline (separate read, write, compute stages).
         layouts_match = bool(model[can_reuse_qk_output_for_pv_input])
```
RattataKing, line 608: NIT: Layout reuse can only happen when `can_reuse_rhs==True`, see all amd encoded mma: https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/Codegen/Dialect/GPU/IR/IREEGPUAttrs.cpp#L287-L730, `qk_acc == pv_lhs` will always be False.

To tighten constraints, I think tuner can just align with IREE, set `col_major=True` when `qk_acc == pv_rhs`, and avoid producing candidates where `col_major=False` even if `rhs_reuse==True`  (iree would resolve layout conflicts later but apparently result in slower performance).

Therefore, we can prune some z3 variables to simplify the current constraints, only keep one variable `can_reuse_rhs`, drop `can_reuse_qk_output_for_pv_input, use_col_major, ...`, set `use_col_major = bool(model[can_reuse_rhs])`, and I think `prefetch_num_stages==2` would only work when there are no layout conflicts, which means it only depends on `can_reuse_rhs` too.

RattataKing (approved):

amdsharktuner/amdsharktuner/rocm/rocm_solutions.py
RattataKing: but is qk_acc==pv_lhs always False?

bangtianliu: I think you made one good point, I will keep it on my TODO list

